PR #10, opened 2019-10-26
Issue #9, Implement Keyboard mgmt engine

JIoJIaJIu:
### Pull Request Description
Implementation of #9 ticket

### Important Notes
- Closures uses forget() and aren't bounded to instance lifetime because of the strange issue with browser (browser events aren't captured if forget isn't used, but works in test environment). Hence removing event listeners isn't happen on KeyboardEngine drop
- Tests aren't developed correctly cause requires [understanding Futures in wasm](https://github.com/rustwasm/wasm-bindgen/issues/1126), will do in next PR

### Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*.rs]
+indent_size = 4
+indent_style = space
+trim_trailing_whitespace = true
+max_line_length = 80
--- a/lib/core/src/lib.rs
+++ b/lib/core/src/lib.rs
@@ -279,6 +279,7 @@ mod example_02 {
     use crate::display::shape::text::TextComponentBuilder;
     use crate::display::shape::text::TextComponentProperties;
 
+
     #[wasm_bindgen]
     #[allow(dead_code)]
     pub fn run_example_text() {
@@ -354,6 +355,93 @@ mod example_02 {
             editor.scroll(Vector2::new(0.0, -0.01));
         }
         world.workspace_dirty.set(workspace_id);
+    }
+}
+
+// ==================
+// === Example 04 ===
+// ==================
+mod example_04 {
+    use wasm_bindgen::prelude::*;
+    use basegl_system_web::keyboard_engine::KeyboardEngine;
+    use basegl_system_web::{window};
+    use web_sys::KeyEvent;
+
+    #[wasm_bindgen(start)]
+    pub fn run_example_keyboard_management() {
+        init_keyboard_engine();
+    }
+
+    /// Init keyboard handlers on Window with the specific combinations
+    ///
+    /// Combinations:
+    ///  - a
+    ///  - Shift-a
+    ///  - Control-Shift-a
+    ///  - Control-Space
+    ///  - Alt-Space
+    fn init_keyboard_engine() {
+        let keyboard_engine = KeyboardEngine::new(&window().unwrap());
+        let handle1 = keyboard_engine.capture(
+            vec![
+                KeyEvent::DOM_VK_A
+            ],
+            Box::new(|| {
+                web_sys::console::log_1(&"Key pressed: a".into());
+            })
+        );
+        std::mem::forget(handle1);
+        let handle2 = keyboard_engine.capture(
+            vec![
+                KeyEvent::DOM_VK_A,
+                KeyEvent::DOM_VK_SHIFT
+            ],
+            Box::new(|| {
+                web_sys::console::log_1(&"Key pressed: Shift-a".into());
+            })
+        );
+        std::mem::forget(handle2);
+        let handle3 = keyboard_engine.capture(
+            vec![
+                KeyEvent::DOM_VK_A,
+                KeyEvent::DOM_VK_SHIFT,
+                KeyEvent::DOM_VK_CONTROL
+            ],
+            Box::new(|| {
+                web_sys::console::log_1(&"Key pressed: Control-Shift-a".into());
+            })
+        );
+        std::mem::forget(handle3);
+        let handle4 = keyboard_engine.capture(
+            vec![
+                KeyEvent::DOM_VK_SPACE,
+                KeyEvent::DOM_VK_CONTROL
+            ],
+            Box::new(|| {
+                web_sys::console::log_1(&"Key pressed: Control-Space".into());
+            })
+        );
+        std::mem::forget(handle4);
+        let handle5 = keyboard_engine.capture(
+            vec![
+                KeyEvent::DOM_VK_SPACE,
+                KeyEvent::DOM_VK_ALT
+            ],
+            Box::new(|| {
+                web_sys::console::log_1(&"Key pressed: Alt-Space".into());
+            })
+        );
+        std::mem::forget(handle5);
+        let handle6 = keyboard_engine.capture(
+            vec![
+                KeyEvent::DOM_VK_SPACE,
+                KeyEvent::DOM_VK_ALT
+            ],
+            Box::new(|| {
+                web_sys::console::log_1(&"Roger that! Key pressed: Alt-Space".into());
+            })
+        );
+        std::mem::forget(handle6);
     }
 }
 
--- a/lib/system/web/src/keyboard_engine/bindings.rs
+++ b/lib/system/web/src/keyboard_engine/bindings.rs
@@ -1,0 +1,46 @@
+use bigint::uint::U256;
+use std::cell::RefCell;
+use std::collections::HashMap;
+
+use crate::keyboard_engine::callback_registry::*;
+
+type BindingsMap = HashMap<U256, CallbackRegistry>;
+
+#[derive(Default)]
+pub struct Bindings {
+    pub data: RefCell<BindingsMap>,
+}
+
+impl Bindings {
+    pub fn new() -> Self{
+        Self {
+            data: RefCell::new(HashMap::new())
+        }
+    }
+
+    pub fn add<F: CallbackMut>
+    (&self, key: U256, callback: F) -> CallbackHandle {
+        let mut data = self.data.borrow_mut();
+        match data.get_mut(&key) {
+            Some(registry) => {
+                registry.add(callback)
+            }
+            None => {
+                let mut registry = CallbackRegistry::new();
+                let handle = registry.add(callback);
+                data.insert(key, registry);
+                handle
+            }
+        }
+    }
+
+    pub fn remove(&self, key: U256) {
+        self.data.borrow_mut().remove(&key);
+    }
+
+    pub fn call_by_key(&self, key: U256) {
+        if let Some(registry) = self.data.borrow_mut().get_mut(&key) {
+            registry.call();
+        }
+    }
+}
--- a/lib/system/web/src/keyboard_engine/mod.rs
+++ b/lib/system/web/src/keyboard_engine/mod.rs
@@ -1,0 +1,117 @@
+pub mod callback_registry;
+pub mod bindings;
+
+use std::cell::Cell;
+use bigint::uint::U256;
+use web_sys::{KeyboardEvent};
+use wasm_bindgen::JsCast;
+use wasm_bindgen::prelude::*;
+
+use crate::*;
+use self::bindings::*;
+use self::callback_registry::*;
+
+/// Keyboard Engine that handles user keyboard combinations
+///
+/// Keyboard Engine stores bindings as HashMap:
+/// {
+///     U256:          [Callback1, Callback2],
+///     U256:          [Callback1],
+/// }
+/// when the combination is pressed related callbacks are invoked
+/// Pressed combination is encoded into U256 bitstring
+pub struct KeyboardEngine {
+    pub bindings: Rc<Bindings>,
+    pub state:    Rc<Cell<U256>>,
+    pub target:   web_sys::EventTarget
+}
+
+impl KeyboardEngine {
+    pub fn new(target: &web_sys::EventTarget) -> Self {
+        let mut this = Self {
+            bindings: Rc::new(Bindings::new()),
+            state:    Rc::new(Cell::new(U256::from(0))),
+            target:   target.clone()
+        };
+        this.init();
+        this
+    }
+
+    pub fn from_tag_name(html_tag: &str) -> Result<Self> {
+        let element = document()?
+            .get_elements_by_tag_name(html_tag)
+            .item(0).ok_or_else(|| Error::missing("there is no such element"));
+
+        Ok(KeyboardEngine::new(&element.unwrap()))
+    }
+
+    fn init(&mut self) {
+        self.bind_keydown_event();
+        self.bind_keyup_event();
+    }
+
+    fn bind_keydown_event(&self) {
+        let state = Rc::clone(&self.state);
+        let bindings = Rc::clone(&self.bindings);
+        let callback = Box::new(move |event: web_sys::KeyboardEvent| {
+            if event.repeat() {
+                return;
+            }
+
+            let key_code = event.key_code();
+            let bit = U256::from(1) << (key_code as usize);
+            let key = state.get() | bit;
+            state.set(key);
+            bindings.call_by_key(key);
+        });
+        let callback = Closure::wrap(callback as Box<dyn FnMut(_)>);
+
+        let callback_ref = callback.as_ref().unchecked_ref();
+        self.target
+            .add_event_listener_with_callback("keydown", callback_ref).unwrap();
+        callback.forget();
+    }
+
+    fn bind_keyup_event(&self) {
+        let bindings = Rc::clone(&self.bindings);
+        let state =  Rc::clone(&self.state);
+        let callback = Box::new(move |event: KeyboardEvent| {
+            let key_code = event.key_code();
+            let bit = U256::from(1) << (key_code as usize);
+            let key = state.get() ^ bit;
+            state.set(key);
+            bindings.call_by_key(key);
+        });
+        let callback = Closure::wrap(callback as Box<dyn FnMut(_)>);
+
+        let callback_ref = callback.as_ref().unchecked_ref();
+        self.target
+            .add_event_listener_with_callback("keyup", callback_ref).unwrap();
+        callback.forget();
+    }
+
+    /// Returns U256 number that represents bitstring of pressed keys
+    fn get_key_bits(&self, combination: Vec<u32>) -> U256 {
+        combination.iter()
+            .fold(U256::from(0), |acc, x| {
+                let bit = U256::from(1) << (*x as usize);
+                acc | bit
+            })
+    }
+
+    /// Capture key combination
+    ///
+    /// Uses u32 as key code from WebIDL
+    /// https://github.com/rustwasm/wasm-bindgen/blob/913fdbc3daff65952b5678a34b98e07d4e6e4fbb/crates/web-sys/webidls/enabled/KeyEvent.webidl
+    pub fn capture<F: CallbackMut>
+    (&self, combination: Vec<u32>, callback: F) -> CallbackHandle {
+        let key = self.get_key_bits(combination);
+        self.bindings.add(key, callback)
+    }
+
+    /// Drops watching specific combination
+    pub fn drop_capture(&self, combination: Vec<u32>) {
+        let key = self.get_key_bits(combination);
+        self.bindings.remove(key);
+    }
+}
--- a/lib/system/web/tests/keyboard_engine.rs
+++ b/lib/system/web/tests/keyboard_engine.rs
@@ -1,0 +1,126 @@
+use js_sys::Promise;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::*;
+use wasm_bindgen_test::*;
+use web_sys::{ KeyEvent, KeyboardEvent, KeyboardEventInit};
+
+use basegl_system_web::*;
+use basegl_system_web::keyboard_engine::KeyboardEngine;
+
+wasm_bindgen_test_configure!(run_in_browser);
+
+#[wasm_bindgen_test]
+async fn capture_keypress() {
+    let elem = document().unwrap().create_element("canvas").unwrap();
+    let keyboard_engine = KeyboardEngine::new(&elem);
+    let key_a = KeyEvent::DOM_VK_A;
+
+    let promise = Promise::new(&mut |resolve, _| {
+        let handle = keyboard_engine.capture(vec![key_a], Box::new(move || {
+            web_sys::console::log_1(&"Key has been captured".into());
+            resolve.call0(&JsValue::NULL).ok();
+        }));
+        std::mem::forget(handle);
+    });
+
+    let mut event_init = KeyboardEventInit::new();
+    event_init.key_code(KeyEvent::DOM_VK_A);
+    let event = KeyboardEvent::new_with_keyboard_event_init_dict("keydown", &event_init).unwrap();
+    elem.dispatch_event(&event).unwrap();
+    JsFuture::from(promise).await.unwrap();
+}
+
+#[wasm_bindgen_test]
+async fn capture_keypress_on_tag_name() {
+    let document = document().unwrap();
+    let elem = document.create_element("canvas").unwrap();
+    document.first_child().unwrap().append_child(&elem).expect("ok");
+
+    let keyboard_engine = KeyboardEngine::from_tag_name("canvas").unwrap();
+    let key_b = KeyEvent::DOM_VK_B;
+    let promise = Promise::new(&mut |resolve, _| {
+        let handle = keyboard_engine.capture(vec![key_b], Box::new(move || {
+            web_sys::console::log_1(&"Key has been captured".into());
+            resolve.call0(&JsValue::NULL).ok();
+        }));
+        std::mem::forget(handle);
+    });
+    let mut event_init = KeyboardEventInit::new();
+    event_init.key_code(KeyEvent::DOM_VK_B);
+    let event = KeyboardEvent::new_with_keyboard_event_init_dict("keydown", &event_init).unwrap();
+    elem.dispatch_event(&event).unwrap();
+    JsFuture::from(promise).await.unwrap();
+}
+
+#[wasm_bindgen_test]
+async fn capture_keypress_with_two_callbacks() {
+    let elem = document().unwrap().create_element("canvas").unwrap();
+
+    let keyboard_engine = KeyboardEngine::new(&elem);
+    let key_a = KeyEvent::DOM_VK_A;
+
+    let promise = Promise::new(&mut |resolve, _| {
+        let handle = keyboard_engine.capture(vec![key_a], Box::new(move || {
+            web_sys::console::log_1(&"Key has been captured. callback #1".into());
+            resolve.call0(&JsValue::NULL).ok();
+        }));
+        std::mem::forget(handle);
+    });
+    let promise2 = Promise::new(&mut |resolve, _| {
+        let handle = keyboard_engine.capture(vec![key_a], Box::new(move || {
+            web_sys::console::log_1(&"Key has been captured. callback #2".into());
+            resolve.call0(&JsValue::NULL).ok();
+        }));
+        std::mem::forget(handle);
+    });
+    let mut event_init = KeyboardEventInit::new();
+    event_init.key_code(KeyEvent::DOM_VK_A);
+    let event = KeyboardEvent::new_with_keyboard_event_init_dict("keydown", &event_init).unwrap();
+    elem.dispatch_event(&event).unwrap();
+    JsFuture::from(promise).await.unwrap();
+    JsFuture::from(promise2).await.unwrap();
+}
+
+#[wasm_bindgen_test]
+fn uncapture_keypress() {
+    let elem = document().unwrap().create_element("canvas").unwrap();
+
+    let keyboard_engine = KeyboardEngine::new(&elem);
+    let key_a = KeyEvent::DOM_VK_A;
+    let callback1 = Box::new(|| {
+        panic!("Shouldn't be executed");
+    });
+    let callback2 = Box::new(|| {
+        web_sys::console::log_1(&"Key has been captured. callback #2".into());
+    });
+    let handle1 = keyboard_engine.capture(vec![key_a], callback1);
+    let handle2 = keyboard_engine.capture(vec![key_a], callback2);
+    std::mem::drop(handle1);
+
+    let mut event_init = KeyboardEventInit::new();
+    event_init.key_code(KeyEvent::DOM_VK_A);
+    let event = KeyboardEvent::new_with_keyboard_event_init_dict("keydown", &event_init).unwrap();
+    elem.dispatch_event(&event).unwrap();
+}
+
+#[wasm_bindgen_test]
+fn drop_capture_keypress() {
+    let elem = document().unwrap().create_element("canvas").unwrap();
+
+    let keyboard_engine = KeyboardEngine::new(&elem);
+    let key_a = KeyEvent::DOM_VK_A;
+    let callback1 = Box::new(|| {
+        panic!("Shouldn't be executed");
+    });
+    let callback2 = Box::new(|| {
+        panic!("Shouldn't be executed");
+    });
+    let handle1 = keyboard_engine.capture(vec![key_a], callback1);
+    let handle2 = keyboard_engine.capture(vec![key_a], callback2);
+    keyboard_engine.drop_capture(vec![key_a]);
+
+    let mut event_init = KeyboardEventInit::new();
+    event_init.key_code(KeyEvent::DOM_VK_A);
+    let event = KeyboardEvent::new_with_keyboard_event_init_dict("keydown", &event_init).unwrap();
+    elem.dispatch_event(&event).unwrap();
+}
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ wasm-pack.log
 
 # A thirdparty downloaded js
 /lib/core/msdf-sys/msdfgen_wasm.js
+
+# VIM
+*.sw?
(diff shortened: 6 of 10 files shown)


lib/core/src/lib.rs
```diff
@@ -27,10 +27,18 @@ pub mod system {
...
 use wasm_bindgen::prelude::*;
+use basegl_system_web::keyboard_engine::KeyboardEngine;
 
 #[wasm_bindgen(start)]
 pub fn start() {
     let world = World::new();
     world.add_workspace("canvas");
     world.start();
+
+    let mut ke = KeyboardEngine::from_tag_name("input", true);
```
wdanilo: Let's attach the KeyboardEngine to the whole screen, not only the input box. 
Plus, let's use more descriptive names. Arguably, we find `keyboard_engine` better name than `ke`.

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+use std::cell::RefCell;
+use std::rc::Rc;
+use wasm_bindgen::JsCast;
+use wasm_bindgen::prelude::*;
+use itertools::Itertools;
+
+enum Action {
+    KeyUp,
+    KeyDown
+}
```
wdanilo: The name `Action` is misleading. `KeyUp` or `KeyDown` are not actions - they are more like the current state of a key. Moreover, it's always preferred to use short names, what about some modules and names like `key::state::Up` instead?

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+type Bindings = HashMap<String, Box<dyn FnMut()>>;
+
+#[wasm_bindgen]
+pub struct KeyboardEngine {
+    bindings:        Rc<RefCell<Bindings>>,
+    _element:         web_sys::Element,
+    _prevent_default: bool,
+    //keydown_cb:    Closure<dyn FnMut(web_sys::KeyboardEvent)>,
+    //keyup_cb:      Closure<dyn FnMut(web_sys::KeyboardEvent)>,
+    _state:          Rc<RefCell<BTreeSet<String>>>
```
wdanilo: Don't use leading underscores in struct fields names until they are phantom values. Moreover, we tend to make every field public until it could break API or be not secure to use - is it the case with these fields?

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+impl KeyboardEngine {
+    pub fn new(element: &web_sys::Element, prevent_default: bool) -> Self {
```
wdanilo: Using `bool` in arguments is almost always a bad idea. When you see code `KeyboardEngine::new(foo, true)` you don't understand what this `true` means. It is always better either to create custom variants or use the builder pattern / modifier, like `KeyboardEngine::new("foo").prevent_default()` instead. 😃

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+        let state = Rc::new(RefCell::new(BTreeSet::new()));
+        let bindings = Rc::new(RefCell::new(HashMap::new()));
+        //let (keydown_cb, keyup_cb) = KeyboardEngine::init(element, bindings.clone(), state.clone());
+        KeyboardEngine::init(element, bindings.clone(), state.clone(), prevent_default);
```
wdanilo: `bindings.clone()` this is misleading as it doesnt show the true cost of the code. Its better to write `Rc::clone(bindings)` instead, or define a method which clearly tells that this is cheap like `bindings.clone_ref()`. We usually do the latter, as it is more general (the former works only with `Rc`)

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+            _state: state,
+
+            //keydown_cb: keydown_cb,
+            //keyup_cb: keyup_cb,
+        }
+    }
+
+    pub fn from_tag_name(html_tag: &str, prevent_default: bool) -> Self {
+        let window = web_sys::window().expect("no global window exists");
+        let document = window.document().expect("should have document on window");
```
wdanilo: Before landing on master all functions should handle errors in a nice way, without panics. Errors should never be written as strings - they need to be datatypes which can pretty print when needed :)

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn init (
+        element:         &web_sys::Element,
+        bindings:        Rc<RefCell<Bindings>>,
+        state:           Rc<RefCell<BTreeSet<String>>>,
+        prevent_default: bool
+    //) -> (
+    //    Closure<dyn FnMut(web_sys::KeyboardEvent)>,
+    //    Closure<dyn FnMut(web_sys::KeyboardEvent)>
+    ) {
+        let capture_key_cb = Rc::new(Box::new(move |mut code: String, action: Action| {
```
wdanilo: Let's use better names. I don't know what `cb` means here

wdanilo: Btw, instead of making such complex expression with so many chained things, lets distribute it to variables :)

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+            code.make_ascii_lowercase();
+            match action {
+                Action::KeyDown => {
+                    state.borrow_mut().insert(code);
+                    let key = state
+                        .borrow()
+                        .iter()
+                        .join("-");
+
+                    if let Some(f) = bindings.borrow_mut().get_mut(&key) { f(); }
```
wdanilo: I don't really like the idea to searching for keys using strings. Sure, `Ctrl-A` is a nice visual way to show users the meaning of the key, but this should only be pretty printing utility. Comparisons should be done using bitfields and masks.

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+enum Action {
+    KeyUp,
+    KeyDown
+}
+
+type Bindings = HashMap<String, Box<dyn FnMut()>>;
+
+#[wasm_bindgen]
+pub struct KeyboardEngine {
+    bindings:        Rc<RefCell<Bindings>>,
```
wdanilo: Try not using internal mutability as far as you can. Of course, in our final application we might want to keep `KeyboardEngine` as internal mutable struct, but then we will keep `Rc<RefCell<KeyboardEngine>>`. Try not to use `RefCell` in **any place** in your code now.

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Capture key press combination
+    /// combination has notation as <C-m> <M-Space> where
+    ///   C/Control - Contorl
+    ///   M/Meta - Meta
+    ///   A/Alt - Alt
+    ///   S/Shift - Shift
+    ///   Space
+    ///   a-z - alphabetic letters
+    pub fn capture(&mut self, combination: String, callback: Box<dyn FnMut()>) {
```
wdanilo: Using `combination` as String is not the best idea here as well. We need some custom parser, which will always be slow and error-prone. Why not `keyboard_engine.capture([key::Control, key::A], ...)` instead? :) 

It could be easily changed to bitfield mask inside of the function!

lib/system/web/tests/keyboard_engine.rs
```diff
@@ -0,0 +1,40 @@
...
+    let mut ke = KeyboardEngine::from_tag_name("canvas", false);
+    ke.capture(String::from("b"), Box::new(|| {
+        web_sys::console::log_1(&"Key has been captured".into());
+    }));
+    let mut event_init = KeyboardEventInit::new();
+    event_init.code("b");
+    let event = KeyboardEvent::new_with_keyboard_event_init_dict("keydown", &event_init).unwrap();
+    elem.dispatch_event(&event).unwrap();
+    assert_eq!(2 + 2, 4); // TODO
+}
```
wdanilo: Big applause for tests 👏 
Amazing that you are doing them in the right way (emitting events from code and testing results this way)

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    pub fn uncapture(&mut self, combination: String) {
+        let key = self.parse_combination(combination);
+        self.bindings.borrow_mut().remove(&key);
+    }
+
+    /// input combination as "C-p"
+    /// and returs string with repr "char(0x11)char(0x70)" in asceding key code order
+    fn parse_combination(&self, combination: String) -> String {
```
wdanilo: After having bitfields we dont need parser anymore! :)

lib/core/src/lib.rs
```diff
@@ -27,10 +27,18 @@ pub mod system {
...
+    ke.capture("a".to_string(), Box::new(|| { web_sys::console::log_1(&"Key pressed: a".into()); }) );
+    ke.capture("Shift-a".to_string(), Box::new(|| { web_sys::console::log_1(&"Key pressed: Shift-a".into()); }) );
+    ke.capture("C-Shift-a".to_string(), Box::new(|| { web_sys::console::log_1(&"Key pressed: Control-Shift-a".into()); }));
+    ke.capture("C-Space".to_string(), Box::new(|| { web_sys::console::log_1(&"Key pressed: Control-Space".into()); }));
+    ke.capture("A-Space".to_string(), Box::new(|| { web_sys::console::log_1(&"Key pressed: Alt-Space".into()); }));
```
wdanilo: Btw, I know this could sound hard and not natural, but lets try using the code style that we've got described in our docs. The earlier you start using it, the easier it will be for all of us read the same codebase. For example, a lot of your lines are over 80 chars length. Keeping the limit of 80 chars makes the code easier to read vertically, however, it often requires to think about better var names as well, so using it from the beginning makes sense, as it will impact the style (and naming) of your code.

JIoJIaJIu: Sure, I thought that you only use `rustfmt`, when I did commit I had seen style guide. I'm going to pass through it and edit the code

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: Do you have an idea how **to mutable** bindings field then?

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: Got it, in fact it's values that are not used, so I've masked them from linter

JIoJIaJIu:
Hello @wdanilo , thanks for such detailed review!
During the current week I will provide the changes

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: I decided not to create separated function, hence I've removed the seciont

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: I'm using `RefCell` not for internal mutability but cause I need to have few owners and mutate the field at the same time (because of the js async nature). If I use just `Rc` I got an `None` on `Rc::get_mut` when have more then 1 reference. Maybe you know how to do it better?

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: I've reworked the section, but would like to note that `keyCode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)
[KeyboardEventInit section](https://www.w3.org/TR/uievents/#idl-keyboardeventinit)

> In practice, keyCode and charCode are inconsistent across platforms and even the same implementation on different operating systems or using different localizations. This specification does not define values for either keyCode or charCode, or behavior for charCode. In conforming UI Events implementations, content authors can instead use key and code.

but `key` and `code` are strings

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
...
+        let state = Rc::new(RefCell::new(BTreeSet::new()));
+        let bindings = Rc::new(RefCell::new(HashMap::new()));
+
+        KeyboardEngine::init(target, Rc::clone(&bindings), Rc::clone(&state));
+
+        Ok(Self {
+            bindings,
+            state,
+            target: target.clone()
+        })
```
wdanilo: This function cannot fail, so why it returnrs Result?

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
...
+    ///
+    /// Uses u32 as key code from WebIDL
+    /// https://github.com/rustwasm/wasm-bindgen/blob/913fdbc3daff65952b5678a34b98e07d4e6e4fbb/crates/web-sys/webidls/enabled/KeyEvent.webidl
+    pub fn capture(
+        & self,
+        combination: Vec<u32>,
+        callback:    Box<dyn FnMut()>
+    ) {
+        let key = combination.iter().copied().collect::<BTreeSet<u32>>();
+        let mut bindings = self.bindings.borrow_mut();
```
wdanilo: I would create some kind of callback-abstraction, which hides all internal mutability and exposes simple API. Using `borrow_mut` in various functions is often considered a bad practice. Instead, we can create something like `BindingRegistry` (or other name), which handles all bindings for us and exposes a nice API :)

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            None => {
+                bindings.insert(key, vec![callback]);
+            }
+        }
+    }
+
+    /// Drops watching specific combination
+    // TODO support removing specific callback by ref
+    pub fn uncapture(& self, combination: Vec<u32>) {
```
wdanilo: I was just about to write the same thing that is in todo :D We want to register some gui actions, like - display full-screen visualization. And we want to be able to assign several actions to the same shortcut. Then we also want to be able to unregister some of them, not always all of them.

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
...
+
+                let key = state.borrow();
+                if let Some(vec) = bindings.borrow_mut().get_mut(&key) {
+                    vec.iter_mut().for_each(|func| func());
+                }
+            });
+            let keyup_callback = Closure::wrap(
+                keyup_callback as Box<dyn FnMut(_)>);
+            target.add_event_listener_with_callback(
+                "keyup", keyup_callback.as_ref().unchecked_ref())
```
wdanilo: This code is definitely too long and contains some strange sections inside. We try to write a code where each function has max of 10-15 lines. If it exceeds that limit, it should almost for sure be refactred to several functions. Basically reading the code line by line should be like reading a book - each line should contain a function call telling what is happening there. Here we got something much more complex, which is hard to read and hard to track.

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
+use std::collections::{HashMap, BTreeSet};
+
+use web_sys::{KeyboardEvent};
+use wasm_bindgen::JsCast;
+use wasm_bindgen::prelude::*;
+
+use crate::*;
+
+type Bindings = HashMap<BTreeSet<u32>, Vec<Box<dyn FnMut()>>>;
```
wdanilo: I dont understand one thing here. We dont have keyboards with more than 110 keys. In fact, every keyboard in the world contains less than 128 keys (am I right here, or am I missing some strange cases?). If Im right, then instead of `BTreeSet<u32>` we can just keep here `u128` as a bitfield, which will be much faster to compare and work on.

wdanilo: `Vec<Box<dyn FnMut()>>` - should be refactored to some kind of `CallbackRegistry` which should manage the callbacks there. Anyway, I'm not sure if `Vec` is the best thing here as we want to add and remove callbacks. Maybe something like `HashMap<*const (), Box<dyn FnMut()>>` instead (of course hidden in a nice type)?

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
...
+        KeyboardEngine::new(&element.unwrap())
+    }
+
+    fn init (
+        target:   &web_sys::EventTarget,
+        bindings: Rc<RefCell<Bindings>>,
+        state:    Rc<RefCell<BTreeSet<u32>>>
+    ) {
+        {
+            let state = Rc::clone(&state);
```
wdanilo: There is a lot of strange code here. Here for example you take `state` ownership in the function and the first thing you do is you clone the RC. It doesnt make sense because youve got the ownership, right?

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
...
+/// Keyboard Engine stores bindings as HashMap:
+/// {
+///     [keyA, keyB]:          [Callback1, Callback2],
+///     [keyA, keyB, keyCtrl]: [Callback1],
+/// }
+/// when the combination is pressed related callbacks are invoked
+pub struct KeyboardEngine {
+    pub bindings: Rc<RefCell<Bindings>>,
+    pub state:    Rc<RefCell<BTreeSet<u32>>>,
+    pub target:   web_sys::EventTarget
```
wdanilo: What if you don't keep `Rc<RefCell<...>>` on fields, but instead you keep ownership of all th fields and also you create another structure, like `KeyboardEngineShared (Rc<RefCell< KeyboardEngine>>)` instead? This way you can keep significant portion of the API as mutable and use interior mutability abstracted away.

lib/system/web/src/keyboard_engine.rs
```diff
@@ -0,0 +1,122 @@
...
+        let key = combination.iter().copied().collect::<BTreeSet<u32>>();
+        let mut bindings = self.bindings.borrow_mut();
+        match bindings.get_mut(&key) {
+            Some(callbacks) => {
+                callbacks.push(callback);
+            }
+            None => {
+                bindings.insert(key, vec![callback]);
+            }
+        }
```
wdanilo: This logic is a responsibility of some kind of callback manager, not capturing function!

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: I did it here cause I then move the state into the first callback, but I need it in the second one

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: Why not? I'm capturing here the key press, and callback management is a part of implementation

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: Do you suggest me to map every existed key into some bit position by my own? But the mapping table needs to support. Do you see any way to map keyCode/key into the bit position

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: The main problem that I have at this point that I need to
1. Have 2 closures with **static** lifetime which mutates the data
2. So I'm not able to borrow the data, only move i
3. And I cannot use `Rc` without `RefCell` or `Cell` cause I have multiple borrowing

If I got right, you suggest me to move bindins into `BindingRegistry`, but it doesn't change something, cause either I need to store internal data for `BindingRegistry` in `RefCell` or wrap it in `RefCell` itself as it's done right now. If I move everything in separated structures, I can make `KeyboardEngine` without internal mutability but I get new two that have it

How do you see I ca
1. Have 2 static closures with moved data that I should mutate
2. And don't use internal mutability at the same time
3. And what pattern at this point can do it nicer as it's right now?

lib/system/web/src/keyboard_engine.rs
JIoJIaJIu: Doesn't it increase complexity? I need `RefCell` as I described above for being able to handle keyboard events. Moving this logic to another struct will lead that all the code we have right now appears in the another struct what creates a mess - why I need the `KeyboardEngineInternal` then. 
KeyboardEngine should be bounded with logic of subscribing on keyboard events, cause it's the core functionality and without it it loses the meaning.  
I tried different approaches and haven't found better way because of `WasmClosure` 

I've checked Servo source code and found that using `Cell` is really common there even for simple structs
[KeyboardEvent.rs](https://github.com/servo/servo/blob/8cb614531183335ea320f60cb2236c7f4ba64480/components/script/dom/keyboardevent.rs#L25)  
[Event.rs](https://github.com/servo/servo/blob/8cb614531183335ea320f60cb2236c7f4ba64480/components/script/dom/event.rs#L50)

What I did at this point was change `RefCell` to `Cell` for state

lib/system/web/src/keyboard_engine/bindings.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl Bindings {
+    pub fn new() -> Self{
+        Self {
+            data: RefCell::new(HashMap::new())
+        }
+    }
+
+    pub fn add(&self, key: U256, callback: Box<dyn FnMut()>) {
+        let mut data = self.data.borrow_mut();
```
wdanilo: You should use Shrinkwrap here and deref coercion, so you could write `self.borrow_mut()` instead.

lib/system/web/src/keyboard_engine/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+            let key = state.get() | bit;
+            state.set(key);
+            bindings.call_by_key(key);
+        });
+        let callback = Closure::wrap(callback as Box<dyn FnMut(_)>);
+
+        let callback_ref = callback.as_ref().unchecked_ref();
+        self.target
+            .add_event_listener_with_callback("keydown", callback_ref).unwrap();
+        callback.forget();
```
wdanilo: isn't it leaking memory here?

.editorconfig
```diff
@@ -0,0 +1,5 @@
+[*.rs]
+indent_size = 4
+indent_style = space
+trim_trailing_whitespace = true
+max_line_length = 80
```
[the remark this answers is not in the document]
JIoJIaJIu: Are you sure? I wondered that this is RUST guidelines you've shared with me also

lib/system/web/src/keyboard_engine/bindings.rs
```diff
@@ -0,0 +1,55 @@
+use bigint::uint::U256;
+use std::cell::RefCell;
+use std::collections::HashMap;
+
+use crate::keyboard_engine::callback_registry::*;
+
+type BindingsMap = HashMap<U256, CallbackRegistry>;
```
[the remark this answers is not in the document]
JIoJIaJIu: I've got the list of the key_codes from [WebIDL bindings](https://github.com/rustwasm/wasm-bindgen/blob/913fdbc3daff65952b5678a34b98e07d4e6e4fbb/crates/web-sys/webidls/enabled/KeyEvent.webidl) that are used in web_sys. As you may see the biggest number is **0xFE**

lib/system/web/src/keyboard_engine/bindings.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(Default)]
+pub struct Bindings {
+    pub data: RefCell<BindingsMap>,
+}
+
+impl Bindings {
+    pub fn new() -> Self{
+        Self {
+            data: RefCell::new(HashMap::new())
+        }
```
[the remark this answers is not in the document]
JIoJIaJIu: Make sense, I am not experienced with `Default` trait, cause it's new feature for me, I'm going to read about it

lib/system/web/src/keyboard_engine/mod.rs
JIoJIaJIu: It's, the main idea, that if I would like to avoid to use `.forget()` I must to make my KeyboardEngine instance static for having ownership for closures. I did few attempts to implement it, but all of them were looked ugly, so I stayed with `forget()`

.editorconfig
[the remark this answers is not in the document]
wdanilo: Oh, I did not realize that editorconfig works across mutliple editors, interesting. In such a case, yes we can leave it here. I'm wondering if IntelliJ uses it, I need to check it out. Thanks for letting me know!

lib/core/src/lib.rs
```diff
@@ -27,10 +27,81 @@ pub mod system {
...
+///  - Alt-Space
+fn init_keyboard_engine() {
+    let keyboard_engine = KeyboardEngine::new(&window().unwrap());
+    keyboard_engine.capture(
+        vec![
+            KeyEvent::DOM_VK_A
+        ],
+        Box::new(|| {
+            web_sys::console::log_1(&"Key pressed: a".into());
+        })
```
wdanilo: A thing here - If you use `Box::new` explicitly it is leaking implementation detail. What if some day you came to an idea to use `Rc` instead of `Box`? All clients code will need to be changed, which is bad. The `capture` function should get a generic type and pack it to `Box` in smart constructor.

lib/system/web/src/keyboard_engine/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Uses u32 as key code from WebIDL
+    /// https://github.com/rustwasm/wasm-bindgen/blob/913fdbc3daff65952b5678a34b98e07d4e6e4fbb/crates/web-sys/webidls/enabled/KeyEvent.webidl
+    pub fn capture(
+        &self,
+        combination: Vec<u32>,
+        callback:    Box<dyn FnMut()>
+    ) {
+        let key = self.get_key_bits(combination);
+        self.bindings.add(key, callback);
+    }
```
wdanilo: A minor thing - we are not using rustfmt anymore cause it has a lot of errors, it isnt very flexivble and produces a long code. We are working on a specification of our style now, for now - see our codebase (on the buffer management brannch or the code submitted recently by @farmaazon ). In particular, this function should be formatted so:

```rust
    pub fn capture
    (&self, combination: Vec<u32>, callback: Box<dyn FnMut()>) {
        let key = self.get_key_bits(combination);
        self.bindings.add(key, callback);
    }
```

lib/system/web/src/keyboard_engine/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// Capture key combination
+    ///
+    /// Uses u32 as key code from WebIDL
+    /// https://github.com/rustwasm/wasm-bindgen/blob/913fdbc3daff65952b5678a34b98e07d4e6e4fbb/crates/web-sys/webidls/enabled/KeyEvent.webidl
+    pub fn capture(
+        &self,
+        combination: Vec<u32>,
+        callback:    Box<dyn FnMut()>
```
wdanilo: As I've written above, the `Box` is implementaion detail ,so the type here should be something like `F: FnMut()` or `F: Fn()` and should be boxed inside of `capture`. This way the interface is much more clean :)

lib/system/web/src/keyboard_engine/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+    ) {
+        let key = self.get_key_bits(combination);
+        self.bindings.add(key, callback);
+    }
+
+    pub fn uncapture(
+        &self,
+        combination:  Vec<u32>,
+        callback_ptr: *const dyn FnMut()
```
wdanilo: Here is a longer comment about why I think its sub-optimal and some ideas to improve it, but before we make it better, let's have a chat if these ideas are correct :)

So, in the future we want to display in GUI something like you know from other editors - a table telling - if you press `a + b` you get action `foo`, and if you press `g` then you get action `bar` and `action baz`. You know, something like:

| Binding  | Action |
| ------------- | ------------- |
| cmd + backspace  | editor:clear_line |
| cmd + arrow left  | editor:jump_to_line_start  |

So we will keep somewhere a bidirectional map of `String` <-> `combination` (string is the action). 

This mechanism is very gui-specific, sometimes you want to handle it in different way, so I'm not sure we would like to keep such bidirectional map as part of this PR / core design. So here are 2 ways - either we keep it as the core design or not. Let me first focus on the later case:

1. In such a case we can create another library around this library which defines bidirectional map keeping `String` <-> `CallbackID` information (whatever `CallbackID` is).
2. The client always needs to keep track of all callbacks that were registered (by keeping their IDs).
3. There should be an easy way to clean up the callback. I'm not sold to the idea to manually calling `uncapture`, as if you forgot to do it, you lose the information about the id and you have no way to get it easily back - you are leaking memory and the best way to revert from this stage is to reset all callbacks, which is kind of ugly.
4. So in such a case we could redefine the function to `fn captur(combination: ..., callback: F) -> CallbackHandle`. The `CallbackHandle` would be a special guard which when dropped (when it goes out of scope) will update the callback map. In such scenario you owuld never want to use uncapture manually, cause you just need to drop `CallbackHandle` and thats it. Then in the high-level API we can just keep bimap of `String` <-> `CallbackHandle` and if we want to uncapture something we can just thorw the `CallbackHandle` from the map and we are done - always safe and always without leaking memory.

------
The second approach would be to just embed the `String <->` Action` map here, so you could `uncapture` callbacks by their names. However, this limits the use cases of this library only to things which have names. In our GUI its always true, but designing libraries in less generic way without really benefits of it is not always the best method, so I feel the former approach is better.

What do u think about it ? :)

.editorconfig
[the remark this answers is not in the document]
JIoJIaJIu: Sure, it works through plugins so all famous editors support it, there about IDEA that supports it by default - https://www.jetbrains.com/help/idea/configuring-code-style.html#editorconfig

lib/system/web/src/keyboard_engine/mod.rs
JIoJIaJIu: Hm, what do you think of such design

1. `bindings` uses 
```
{
   U256: [ ActionTrait ]
}
```
2. So we need to implement **ActionTrait** for any type, for example for **Callback** and **Action**
3. For examle it could look like 
```
{
    U256: [ Callback1, Callback2, ActionOpenEditor, ActionCloseEditor]
}
```
4. **Action type** contains its `ID`
5. GUI setup will look as 2 actions:
    1. Create **Action**
     ```
      let action = Action::new(open::editor)
     ```
     2. Add **Action** callbacks
     ```
      keyboard_engine.capture(combination, action)
     ```
6. We can create global singleton **ActionEngine** with predefined actions and implement getting action by its `String` name / `ID`
   ```
     action_engine.bind(ID, combination)
   ```
   That does
    ```
      let action = getFromActionRegistry(ID);
      if !action {
        action = createDumpActionAndAddToRegistry(ID)
      }
      keyboard_engine.capture(combination, action)
    ```
    and provide
    ```
       action.get()
       action.list()
    ```

wdanilo: If I understand correctly, your solution assumes the existence of some kind of a global counter, right? So every time you create an action, you get a new id? If so, I don't like it because of the following reasons:
- Global counters are always bad. It's hard to use them in a multithreaded / multitenancy environment.
- The IDs are used only to differentiate actions in the action map right? If so, they are not part of the action per se. They are used only by their management layer, so logically they belong there.

Does it make sense? If so, I think I'd still go with the guard design here. :)

lib/core/src/lib.rs
JIoJIaJIu: I'm using `Box` here for allocating in the heap. I think that if we would like to use `Rc` for example then the value should be cloned from there and dereferenced.  Do you suggest to create something like?
```
fn capture<F: FnMut()> (..:, callback: F) { 
```

lib/system/web/src/keyboard_engine/mod.rs
JIoJIaJIu: I think you understood me not 100% clear
From your first post I like more the first approach - `create another library around this library which defines bidirectional map`
> the existence of some kind of a global counter, right

Some instance, is it global or not depends on the usage and implementation. The instance contains all registered actions and their ids. In fact we need to register actions somewhere

> So every time you create an action, you get a new id?

Why not? In this case ID can be the string as you mentioned in `String <-> CallbackID`

> The IDs are used only to differentiate actions in the action map right?

Yes, I mean that case

> They are used only by their management layer, so logically they belong there

Agreed. I'm just wondering about your described feature. IMHO it would be better not to merge `Actions` with `KeyboardEngine`

What do you think?

wdanilo: @JIoJIaJIu Oh damn, I haven't seen your reply, sorry for that. I was sure you will ping me on Discord when you progress with this issue and I just realized you replied here :(

So, I like the approach to `create another library around this library which defines bidirectional map` as well. Let's stay with it.

> Why not? In this case ID can be the string as you mentioned ...
Sure! I was rather asking if I understand that correctly. There is just one problem with ID: you need to unregister the callbacks manually. If you drop the ID you have a problem. 

Guro, take a look at a callback engine we just included in our system. It uses `CallbackHandle` as a result type and unregisters callbacks lazily (when firing the event). A very similar version could be created to unregister callbacks immediately when handle gets dropped. I'd love to know your thoughts about it vs your proposition above: https://github.com/luna/basegl/blob/wip/wd/buffer-management/lib/core/src/control/callback.rs

wdanilo:
@JIoJIaJIu we used a lot of ideas from your implementation in the frp engine we have developed. It has now keyboard bindings which base on the same assumptions that you have developed. Thank you for this contribution. I will close this PR as it will not be merged in this form, but I wanted to let you know that actually it was very helpful for us. To learn more, take a look here:
https://github.com/luna/ide/tree/master/lib/frp

JIoJIaJIu:
@wdanilo sure, thank you
make me know if you need any other contributions from my side, I wasn't able to contact you in the discord

PR #39, opened 2019-11-12
Add msdfgen-sys crate

farmaazon:
Added sys-crate wrapping msdfgen-wasm library.

### Pull Request Description
Added msdfgen-sys crate which wraps msdfgen-wasm library api in rust functions and structures.
The msdfgen-wasm library is automatically downloaded during build.

### Checklist

- [x] All code conforms to the [Rust](https://github.com/luna/enso/blob/master/doc/rust-style-guide.md) style guides as appropriate.


lib/core/msdf-sys/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+authors = ["Adam Obuchowicz <adam.obuchowicz@luna-lang.com>"]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+wasm-bindgen = "0.2.53"
+js-sys = "0.3.30"
```
wdanilo: Why do you need `js-sys` for this binding? Maybe its necessary, but itsn ot clear to me yet why.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
+extern crate wasm_bindgen;
+#[macro_use] extern crate slice_as_array;
```
wdanilo: This is old-rust. You dont need to use it anymore. Just use normal `use` syntax instead: `use slice_as_array`.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use wasm_bindgen::prelude::*;
+use js_sys::*;
```
wdanilo: We try not to import with `*` modules unless we use `prelude` or something similar. Maybe this is the correct usage here, but I would be more careful tbh.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+const MAX_MSDF_SIZE : usize = 32;
+pub const MSDF_CHANNELS_COUNT : usize = 3;
+const MSDF_DATA_SIZE : usize = MAX_MSDF_SIZE*MAX_MSDF_SIZE*MSDF_CHANNELS_COUNT;
```
wdanilo: According to style guide, we should not make things private unless it is a security threat.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
+extern crate wasm_bindgen;
+#[macro_use] extern crate slice_as_array;
+
+use wasm_bindgen::prelude::*;
+use js_sys::*;
+
+const MAX_MSDF_SIZE : usize = 32;
+pub const MSDF_CHANNELS_COUNT : usize = 3;
```
wdanilo: These constants are not described - why 3 and not 7? Any explanation should be provided. Moreover, what does `max msdf size` mean? `msdf` is a name of the library, so is this size of this library in megabytes or something else?

lib/core/msdf-sys/src/lib.rs
wdanilo: Minor thing: lets use spaces between operators

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub data   : [f32;MSDF_DATA_SIZE]
+}
+
+pub struct MSDFParameters {
+    width                         : usize,
+    height                        : usize,
+    edge_coloring_angle_threshold : f64,
+    range                         : f64,
+    scale                         : (f64, f64),
+    translate                     : (f64, f64),
```
wdanilo: Lets not use tuples. They are not descriptive and accessing elements by `.0` and `.1` is hacky.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    edge_threshold                : f64,
+    overlap_support               : bool
+}
+
+#[wasm_bindgen(module = "msdfgen_wasm.js")]
+extern {
+
+fn ccall(
```
wdanilo: this name is not descriptive. Lets either rename it or add docs here. In fact docs should be added to other methods as well (See the code guide for rust - docs section)

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    name        : &str,
+    return_type : &str,
+    types       : js_sys::Array,
+    values      : js_sys::Array
+) -> JsValue;
+
+fn getValue(address: usize, a_type: &str) -> JsValue;
+
+fn _msdfgen_maxMSDFSize() -> usize;
```
wdanilo: why these names start with underscore?

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    range                           : f64,
+    scale_x                         : f64,
+    scale_y                         : f64,
+    translate_x                     : f64,
+    translate_y                     : f64,
+    edge_threshold                  : f64,
+    overlap_support                 : bool
+) -> usize;
+
+fn _msdfgen_freeFont(font_handle: JsValue);
```
wdanilo: this name uses mixed-case style

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+}
+
+fn make_js_array_from_data_view(data: &[u8]) -> Uint8Array {
+    unsafe {
+        return Uint8Array::view(data)
+    }
+}
```
wdanilo: Wait, this is unsafe function, right? When I was talking about refactoring out such functions, I meant - refactor the smallest portion that is actually safe. Unless I'm missing something, this one is still very unsafe as it may gets invalidated by alloc, right?

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+pub fn load_font_memory(data: &[u8]) -> FontHandle {
+
+    let param_types = js_sys::Array::new_with_length(2);
+    param_types.set(0, JsValue::from_str("array"));
+    param_types.set(1, JsValue::from_str("number"));
```
wdanilo: why not `"number".into()` ?

wdanilo: There is a lot of hardcoded literals here - `2`, `0`, `1`. Lets make an abstraction for that. I mean, a function (or macro if we really need to compile compute the numbers), so we can use it like `let param_types = js_array!["array".into(), "number".into()];`

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    let params = js_sys::Array::new_with_length(2);
+    params.set(0, JsValue::from(make_js_array_from_data_view(data)));
+    params.set(1, JsValue::from_f64(data.len() as f64));
+    let handle = ccall(
+        "msdfgen_loadFontMemory",
+        "number",
```
wdanilo: There are literals everywhere - see the code guide, we should either remove literals or assign them to well-described vars.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        params);
+    FontHandle { handle }
+
+}
+
+impl Drop for FontHandle {
+    fn drop(&mut self) {
+        _msdfgen_freeFont(self.handle.clone())
+    }
+}
```
wdanilo: This part of the code should be defined close to `FontHandle` as it implements its crucial functionality. Basically, we should define structs, types, and associated impls next to each other. Such blocks should be separated with visual sections:
```
// ==================
// === FontHandle ===
// ==================

// struct def
// related type defs
// impls

// =====================
// === SomethingElse ===
// =====================

File: lib/core/msdf-sys/src/lib.rs [hunk 14]
```diff
@@ -0,0 +1,163 @@
...
+
+pub fn generate_msdf(
+    font    : &FontHandle,
+    unicode : u32,
+    params  : MSDFParameters
+) -> MSDF {
+    assert!(params.width <= MAX_MSDF_SIZE);
+    assert!(params.height <= MAX_MSDF_SIZE);
```
wdanilo: very, very minor thing - lets align the operators!

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        params.width,
+        params.height,
+        font.handle.clone(),
+        unicode,
+        params.edge_coloring_angle_threshold,
+        params.range,
+        params.scale.0,
+        params.scale.1,
+        params.translate.0,
+        params.translate.1,
```
wdanilo: this is ugly

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        params.edge_threshold,
+        params.overlap_support
+    );
+    let mut data : [f32;MSDF_DATA_SIZE] = [0.0;MSDF_DATA_SIZE];
+    for i in 0..params.width*params.height*MSDF_CHANNELS_COUNT {
```
wdanilo: this is magic. This multiplication and everything should be assigned to a meaningful variable.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        data[i] = getValue(output_address + i*4, "float")
```
wdanilo: `i*4` is magic

lib/core/msdf-sys/src/lib.rs
farmaazon: It's unavoidable when we use C api

lib/core/msdf-sys/src/lib.rs
farmaazon: MSDF is a Multi-channel signed distance field - a data structure like image. Library name is msdfgen (because it generates msdf). Shall I give here explanation what MSDF is?

lib/core/msdf-sys/Cargo.toml
farmaazon: I used Uint8Array for passing fonts in memory to js: https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html

lib/core/msdf-sys/src/lib.rs
farmaazon: I can't choose its name, because this is function provided by emscripten. But I'll add comment here

lib/core/msdf-sys/src/lib.rs
wdanilo: Yep, a short info would be desirable + link to more info.

lib/core/msdf-sys/build.rs
```diff
@@ -0,0 +1,39 @@
+extern crate download_lp;
+
+mod msdfgen_wasm {
+
+    pub const VERSION  : &str = "v1.0";
+    pub const FILENAME : &str = "msdfgen_wasm.js";
+
+    pub fn download() {
+        let url = format!(
+            "https://github.com/luna/msdf/releases/download/{version}/{filename}",
```
wdanilo: I would refactor the address of our github to a variaable as well. I believe that we may want to refer to it from different places, so refactoring it now could be super helpful. Until we need it in other places, lets leave it in this library / module, but as a variable.

lib/core/msdf-sys/build.rs
```diff
@@ -0,0 +1,39 @@
...
+            version = VERSION,
+            filename = FILENAME
+        );
+
+        if std::path::Path::new(FILENAME).exists() {
+            std::fs::remove_file(FILENAME).unwrap();
+        }
+
+        // Note [Downloading to src dir]
+        download_lp::download(url.as_str(), ".").unwrap();
```
wdanilo: I don't like the idea to keep 3rd "binaries" in src folder - maybe we create a third-party folder where such things are downlaoded? We can as well download them to `dist` (or whatever the destination directory is called). After thinking lomnger about it, I'm more and more convinced that we should reuse dist folder here as this is just some kind of building artifact.

lib/core/msdf-sys/build.rs
```diff
@@ -0,0 +1,39 @@
+extern crate download_lp;
+
+mod msdfgen_wasm {
+
+    pub const VERSION  : &str = "v1.0";
+    pub const FILENAME : &str = "msdfgen_wasm.js";
+
+    pub fn download() {
```
wdanilo: I know it may sound very verbose, but before merging to master each top-level public function has to has a documentation. This way we can generate nice docs to our libraries for new people joining the porject. See this guide: https://github.com/luna/enso/blob/master/doc/rust-style-guide.md#documentation-comments

lib/core/msdf-sys/src/internal.rs
```diff
@@ -0,0 +1,63 @@
...
+    #[wasm_bindgen(js_name="ccall")]
+    pub fn emscripten_call_function(
+        name        : &str,
+        return_type : &str,
+        types       : js_sys::Array,
+        values      : js_sys::Array
+    ) -> JsValue;
+
+    #[wasm_bindgen(js_name="getValue")]
+    pub fn emscripten_get_value_from_memory(address: usize, a_type: &str) -> JsValue;
```
wdanilo: its >80 chars

lib/core/msdf-sys/src/internal.rs
```diff
@@ -0,0 +1,63 @@
...
+        scale_y                         : f64,
+        translate_x                     : f64,
+        translate_y                     : f64,
+        edge_threshold                  : f64,
+        overlap_support                 : bool
+    ) -> usize;
+
+    #[wasm_bindgen(js_name="_msdfgen_freeFont")]
+    pub fn msdfgen_free_font(font_handle: JsValue);
+
```
wdanilo: Very minor thing - unnecessary enter here

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+
+pub struct Font {
+    handle: JsValue
+}
+
+impl Font {
+    // Loading font from memory
+    //
+    // Loads font from a any format which freetype library can handle.
+    // See https://www.freetype.org/freetype2/docs/index.html for reference
```
wdanilo: This is not a doc comment. Doc comments start with `///`. Moreover, a minor thing - some time ago I was not thinking about it this way, but I learned from @iamrecursion to always put pride in creating nice docs, and it really pays off. So let's always remember about finishing sentences with dots etc. I know this may sounds too restrictive, but it pays off, as we want to keep the codebase really high quality.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub fn load_from_memory(data: &[u8]) -> Self {
+        let param_types = js_sys::Array::of2(
+            &JsValue::from_str(emscripten_data_types::ARRAY),
+            &JsValue::from_str(emscripten_data_types::NUMBER)
+        );
+        unsafe { // Note [Usage of Uint8Array::view]
```
wdanilo: Please see https://github.com/luna/enso/blob/master/doc/rust-style-guide.md#source-notes
Especially: `Any unsafe usage of a function must be accompanied by a source note that explains what makes this particular usage safe.`

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+            params.range,
+            scale.x,
+            scale.y,
+            translate.x,
+            translate.y,
+            params.edge_threshold,
+            params.overlap_support
+        );
+        let mut data : [f32; Self::MAX_DATA_SIZE] = [0.0; Self::MAX_DATA_SIZE];
+        let data_size = params.width*params.height*Self::CHANNELS_COUNT;
```
wdanilo: Again very minor thing - spaces around operators. Damn, we need some day to fix rustfmt to support our style. Its so buggy now its hard to be used, but comments like these coulb be tiresome for you, sorry Adam.

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+}
+
+
+
+#[cfg(test)]
+mod tests {
+    extern crate wasm_bindgen_test;
+    extern crate slice_as_array;
```
wdanilo: Do we need to use `extern crate` in Rust 2018? I believe its not needed anymore

lib/core/msdf-sys/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        };
+        // when
+        let msdf = MutlichannelSignedDistanceField::generate(
+            &font,
+            'A' as u32,
+            params,
+            Vector2D { x: 1.0, y: 1.0 },
+            Vector2D { x: 0.0, y: 0.0 }
+        );
+        // then
```
wdanilo: What do these "when" and "get" comments mean?

lib/core/msdf-sys/build.rs
farmaazon: Technically it is not public because:
a) it's a build script - not designed to be imported anywhere
b) it's a private module of build script

lib/core/msdf-sys/src/lib.rs
farmaazon: An there is? // Note [Usage of Uint8Array::view]

lib/core/msdf-sys/src/lib.rs
farmaazon: These are sections inspired by Spock:
given - a test setup and parameters
when - the thing we're testing
then - checking assertions

lib/core/msdf-sys/build.rs
farmaazon: But the cargo "target" directory is like "dist" directory, and bound js files cannot be downloaded there, as I wrote in note. And the whole "msdf-sys" crate is by convention a wrapper for third party library, so do we really need additional "third-party" directory?

And this file is reused in future builds - the build.rs script is run only on first build or when the script itself changed

lib/core/msdf-sys/build.rs
farmaazon: done

lib/core/msdf-sys/src/internal.rs
farmaazon: fixed

lib/core/msdf-sys/src/internal.rs
farmaazon: fixed

lib/core/msdf-sys/src/lib.rs
farmaazon: fixed

lib/core/msdf-sys/src/lib.rs
farmaazon: No problem. fixed

lib/core/msdf-sys/src/lib.rs
farmaazon: It isn't. Removed

lib/core/msdf-sys/build.rs
wdanilo: Oh the note is in fact explaining it. Sorry I misread it earlier. Have you been asking about it on Rust Discord channel ?

lib/core/msdf-sys/src/lib.rs
wdanilo: Oh, hmm, we've been pasting the comments in the place of code, we never used the note-style. But I like it. We can add it to our code style.

PR #48, opened 2019-11-18
HTML Rendering System

notdanilo:
### Pull Request Description
This PR provides the implementation of the HTML Rendering System and its components:

* HTMLRenderer: To render the camera transforms in the HTMLScene div and objects transforms in each HTMLObject div.
* HTMLObject: Struct holding a HtmlElement and Transform
* HTMLScene: A collection of `HTMLObject`s
* Camera: Struct holding a Transform to position and orientate the camera, and a projection matrix.

### Important Notes
To improve productivity, I've created a TestContainer when running `wasm-pack test` in a browser, so we can see its visual results.

### Checklist


lib/core/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Wojciech Danilo <wojciech.danilo@gmail.com>"]
 edition = "2018"
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["rlib", "cdylib"]
```
wdanilo: Why we need `rlib`?

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
+use std::collections::HashSet;
+use std::rc::{Rc, Weak};
+use std::hash::{Hash, Hasher};
+pub trait Hashable = Hash + PartialEq;
```
wdanilo: There should be 2 empty lines between imports and code + please, use the code separators:

```
// ================
// === Hashable ===
// ================
```
to divide parts of the file related to different entities. If there is a single entitiy, we use just a single separator on top of the file.

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
+use std::collections::HashSet;
```
wdanilo: Wait, after reviewing this file I realized that are not using `WeakSet` anywhere in you codebase, so why this file is committed? :D

Please read and reply to the reviews though.

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct WeakRef<T : Hashable> {
+    elem : Rc<T>,
+    owner : Rc<WeakSet<T>>
+}
+
+impl<T : Hashable> Drop for WeakRef<T> {
+    fn drop(&mut self) {
+        unsafe {
+            // no worries because it will happen synchronously
+            Rc::get_mut_unchecked(&mut self.owner).data.remove(&WeakElem { elem : Rc::downgrade(&self.elem) });
```
wdanilo: Please fix this and other lines that have >80 chars. Be sure to read our code style guide.

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub struct WeakRef<T : Hashable> {
+    elem : Rc<T>,
+    owner : Rc<WeakSet<T>>
+}
+
+impl<T : Hashable> Drop for WeakRef<T> {
+    fn drop(&mut self) {
+        unsafe {
+            // no worries because it will happen synchronously
```
wdanilo: Please use style guide regarding comments in code. Plus, this comment is not specific enough - comments regarding `unsafe` blocks should be really detailed and explain why we need such blocks.

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl<T : Hashable> Hash for WeakElem<T> {
+    fn hash<H : Hasher>(&self, state : &mut H) {
+        match self.elem.upgrade() {
+            Some(rc) => rc.hash(state),
+            None => Option::<Rc<T>>::None.hash(state)
+        }
+    }
+}
```
wdanilo: Why not derive?

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl<T: Hashable> PartialEq for WeakElem<T> {
+    fn eq(&self, other : &Self) -> bool {
```
wdanilo: a very minor thing. We use `self` and `that` as a convention.

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
...
+impl<T : Hashable> WeakSet<T> {
+    pub fn empty() -> Rc<Self> {
+        Rc::new(Self { data : HashSet::new() })
+    }
+
+    pub fn insert(mut self : &mut Rc<Self>, elem : T) -> WeakRef<T> {
+        let elem = Rc::new(elem);
+        unsafe {
+            Rc::get_mut_unchecked(&mut self).data.insert(WeakElem { elem : Rc::downgrade(&elem) });
+        }
```
wdanilo: Uncommented unsafe

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
...
+                match rhs {
+                    Some(_) => false,
+                    None => true
+                }
+            }
+        })
+    }
+}
+
+impl<T : Hashable> Eq for WeakElem<T> {}
```
wdanilo: Minor formatting: `T: Hashable`

lib/core/src/data/weakset.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+        WeakRef { elem : elem, owner : self.clone() }
+    }
+
+    pub fn len(&self) -> usize {
+        self.data.len()
+    }
+
+    pub fn iter<'a>(&'a self) -> impl Iterator<Item = Rc<T>> + 'a {
+        self.data.iter().map(|x| x.elem.upgrade().unwrap()) // Unwrapping here is fine because it's contextually guaranteed to work. So there is no need to filter !is_none out
```
wdanilo: First, line too long, second: I dont understand the comment. What does it mean "contextually safe"? If it is safe, its need to be described why, I dont see it instantly when reading the code.

lib/core/src/display/scene/camera.rs
```diff
@@ -0,0 +1,20 @@
...
+use nalgebra::geometry::Perspective3;
+use std::f32::consts::PI;
+
+#[derive(Shrinkwrap)]
+#[shrinkwrap(mutable)]
+pub struct Camera {
+    #[shrinkwrap(main_field)]
+    pub object: Object,
+    pub projection: Matrix4<f32>,
+}
```
wdanilo: We align `:` in struct defs. I know this is stupid to write such comments, but until we find a way to fix / modify rustfmt we need to do it by hand.

lib/core/src/display/scene/camera.rs
```diff
@@ -0,0 +1,20 @@
...
+
+impl Camera {
+    pub fn perspective(fov: f32, aspect: f32, znear: f32, zfar: f32) -> Self {
+        let projection =
+            Perspective3::new(aspect, fov / 180.0 * PI, znear, zfar).as_matrix().clone();
```
wdanilo: We try not to create multiline-expressions. Please refactor it to variables instead.

wdanilo: What do you need `clone` for? You are creating the struct in this line, so what is the reason for cloning it?

lib/core/src/display/scene/htmlobject.rs
```diff
@@ -0,0 +1,49 @@
+use super::Object;
+use crate::system::web::{create_element_as, Result};
```
wdanilo: we're not using mutliple imports in a single line. This is because its much easier to manage one-import per line (disable, remove). When we will get rustfmt fixed some day it will automatically transform such imports to single-line ones.

lib/core/src/display/scene/htmlobject.rs
```diff
@@ -0,0 +1,49 @@
...
+use nalgebra::Vector3;
+use web_sys::HtmlElement;
+
+#[derive(Shrinkwrap)]
+#[shrinkwrap(mutable)]
+pub struct HTMLObject {
+    #[shrinkwrap(main_field)]
+    pub object: Object,
+    pub element: HtmlElement,
+    pub dimension: Vector3<f32>,
```
wdanilo: I believe these are `dimensions`, not `dimension`, right?

lib/core/src/display/scene/htmlobject.rs
```diff
@@ -0,0 +1,49 @@
...
+        let element = create_element_as::<HtmlElement>(name)?;
+        Ok(Self::from_element(element))
+    }
+
+    pub fn from_element(element: HtmlElement) -> Self {
+        let style = element.style();
+        style.set_property("transform-style", "preserve-3d").expect("transform-style: preserve-3d");
+        style.set_property("position", "absolute").expect("position: absolute");
+        style.set_property("width", "0px").expect("width: 0px");
+        style.set_property("height", "0px").expect("height: 0px");
```
wdanilo: What these `.expect(...)` do? I mean, they throw custom panic, right? If so, what is the exact message of panic thrown here for example?

lib/core/src/display/scene/htmlobject.rs
```diff
@@ -0,0 +1,49 @@
...
+        Self { object: Object::new(), element, dimension: Vector3::new(0.0, 0.0, 0.0) }
```
wdanilo: are `Object` and `Vector` instances of `Default`?

lib/core/src/display/scene/htmlobject.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    // We need to validate html. We can use Result<Node, Error> from
+    // element.children[0].
```
wdanilo: I don't understand this comment. Was it meant to be doc comment or TODO comment? Either way, it needs a better explanation.

lib/core/src/display/scene/htmlobject.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub fn set_dimension(&mut self, width: f32, height: f32) {
+        self.dimension = Vector3::new(width, height, 0.0);
+        let style = self.element.style();
+        style.set_property("width", &format!("{}px", width)).expect("set width");
+        style.set_property("height", &format!("{}px", height)).expect("set height");
+    }
+
+    pub fn get_dimension(&mut self) -> &Vector3<f32> {
+        &self.dimension
+    }
```
wdanilo: There is a convention in rust that getters do not have `get_` prefix.

lib/core/src/display/scene/htmlobject.rs
```diff
@@ -0,0 +1,49 @@
...
+        style.set_property("height", "0px").expect("height: 0px");
+        Self { object: Object::new(), element, dimension: Vector3::new(0.0, 0.0, 0.0) }
+    }
+
+    // We need to validate html. We can use Result<Node, Error> from
+    // element.children[0].
+    pub fn from_html_string(html: &str) -> Result<Self> {
+        let element = create_element_as::<HtmlElement>("div")?;
+        element.set_inner_html(html);
+
```
wdanilo: Minor thing - unnecessary empty line

lib/core/src/display/scene/htmlrenderer.rs
```diff
@@ -0,0 +1,104 @@
+use super::{Camera, HTMLScene};
+
+fn eps(value: f32) -> f32 {
+    if value.abs() < 1e-10 {
+        0.0
+    } else {
+        value
+    }
+}
```
wdanilo: Hehe, I love this function :D 
Make it shorter:
```rust
fn eps(value: f32) -> f32 {
    if value.abs() < 1e-10 { 0.0 } else { value }
}
```

lib/core/src/display/scene/htmlrenderer.rs
```diff
@@ -0,0 +1,104 @@
...
+
+pub struct HTMLRenderer {}
+
+impl HTMLRenderer {
+    pub fn new() -> Self {
+        Self {}
+    }
```
wdanilo: Each struct which implements `new` without args should have `Default` instance. Plus, this is a special situation when the function is so short, we definitely can make it one line long

lib/core/src/display/scene/htmlrenderer.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    pub fn render(&self, camera: &mut Camera, scene: &HTMLScene) {
+        let (view_width, view_height) = scene.get_dimension();
+        let fov = camera.projection[5] * view_height / 2.0;
```
wdanilo: `camera.projection[5]`

lib/core/src/display/scene/htmlrenderer.rs
```diff
@@ -0,0 +1,104 @@
...
+                    eps(transform[11]),
+                    eps(transform[12]),
+                    eps(-transform[13]),
+                    eps(transform[14]),
+                    eps(transform[15]),
+                    view_width / 2.0,
+                    view_height / 2.0
+                ),
+            )
+            .expect("set camera transform");
```
wdanilo: This is over 30 lines expression. Should be refactored out. Moreover, why are you negating some values? Are we keeping matrices that are different that those needed in CSS3D? If so, why ?

lib/core/src/display/scene/htmlrenderer.rs
```diff
@@ -0,0 +1,104 @@
...
+                                eps(transform[8]),
+                                eps(transform[9]),
+                                eps(transform[10]),
+                                eps(transform[11]),
+                                eps(transform[12]),
+                                eps(transform[13]),
+                                eps(transform[14]),
+                                eps(transform[15])
+                            ),
+                        )
```
wdanilo: This looks like copy-paste from previous code. Definitely to be refactored.

lib/core/src/display/scene/htmlscene.rs
```diff
@@ -0,0 +1,50 @@
+use super::HTMLObject;
```
wdanilo: Format this file according to previous comments, make lines max 80 chars, make more vars, so you dont have multi-line exprs after refactoring, and this will be a very nice file! <3

lib/core/src/display/scene/transform.rs
```diff
@@ -0,0 +1,84 @@
...
+        s1 * c2 * c3 + c1 * s2 * s3,
+        c1 * s2 * c3 - s1 * c2 * s3,
+        c1 * c2 * s3 + s1 * s2 * c3,
+    ))
+}
+
+pub struct Transform {
+    pub position:   Vector3<f32>,
+    pub quaternion: UnitQuaternion<f32>,
+    pub scale:      Vector3<f32>,
```
wdanilo: `position` and `scale` are the names of physical properties of objects. `quaternion` is a name of a math structure. I may be wrong, but I feel it's inconsistent here.

lib/core/src/display/scene/transform.rs
```diff
@@ -0,0 +1,84 @@
+use nalgebra::{Matrix4, Quaternion, UnitQuaternion, Vector3};
+
+// Note [Quaternion YXZ Euler angles]
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// To comply with Threejs impl, we generate a Quaternion applying rotation in
+// the order: pitch -> roll -> yaw, instead of roll -> pitch -> yaw based on
+// https://github.com/mrdoob/three.js/blob/master/src/math/Quaternion.js#L199
```
wdanilo: A question - why we comply with Theejs impl? Based on this comment I understand it is not a standard way of handling it, right? Is it more performant?

lib/core/src/display/scene/transform.rs
```diff
@@ -0,0 +1,84 @@
...
+    // based on https://github.com/mrdoob/three.js/blob/master/src/math/Matrix4.js#L732
+    pub fn to_homogeneous(&self) -> Matrix4<f32> {
+        let (x, y, z, w) = (
+            self.quaternion.coords.x,
+            self.quaternion.coords.y,
+            self.quaternion.coords.z,
+            self.quaternion.coords.w,
+        );
+        let (x2, y2, z2) = (x + x, y + y, z + z);
+        let (xx, xy, xz) = (x * x2, x * y2, x * z2);
```
wdanilo: very minor - alignment of `,`

lib/core/src/display/scene/transform.rs
```diff
@@ -0,0 +1,84 @@
...
+        self.scale = Vector3::new(x, y, z);
+    }
+
+    pub fn set_rotation(&mut self, roll: f32, pitch: f32, yaw: f32) {
+        self.quaternion = from_euler_angles_pry(roll, pitch, yaw);
+    }
+
+    // Note [Transform to Matrix4 composition]
+    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // based on https://github.com/mrdoob/three.js/blob/master/src/math/Matrix4.js#L732
```
wdanilo: This should be a doc comment

lib/core/src/data/weakset.rs
notdanilo: My mistake! I considered using it, but then I relied on the existing opt_vec implementation. Should we keep it?

lib/core/Cargo.toml
notdanilo: We need it to make the crate available in the tests.

See https://github.com/rust-lang/cargo/issues/6659

lib/core/src/data/weakset.rs
notdanilo: weakset.rs was removed as suggested

lib/core/src/data/weakset.rs
notdanilo: weakset.rs was removed as suggested

lib/core/src/display/scene/htmlscene.rs
notdanilo: A newer file formatted with `cargo fmt` should be available. Would you check it?

lib/core/src/display/scene/transform.rs
notdanilo: You are right. I renamed `quaternion` to `rotation`. I also renamed `position` to `translation`.

lib/core/src/display/scene/transform.rs
notdanilo: I left this as a source comment and provided a detailed explanation on how it works

lib/core/src/display/scene/transform.rs
notdanilo: There is no performance difference. It's just a convention and we can change it whenever we want.

As I was developing this functionality based on Three.js, I used it as a visual reference and I saw something was wrong. That's when I realized the rotation order was different.

lib/core/src/display/scene/transform.rs
notdanilo: I think it was fixed by `cargo fmt`. Please check

lib/core/src/display/scene/camera.rs
notdanilo: A E S T H E T I C S is not stupid. Fixed!

lib/core/src/display/scene/camera.rs
notdanilo: Fixed by cargo fmt

notdanilo: as_matrix() returns a ref to Matrix4.

But it's a Copy type, so I updated it to use a deref as suggested by clippy.

lib/core/src/display/scene/htmlobject.rs
notdanilo: I am not sure I understood what you meant. Is the suggestion here to change it to

```
use crate::system::web::create_element_as;
use crate::system::web::Result;
```

?

lib/core/src/display/scene/htmlobject.rs
notdanilo: Yes, fixing it.

lib/core/src/display/scene/htmlobject.rs
notdanilo: Object yes, but not Vector3.
I have updated it to `object : Default::new()`

lib/core/src/display/scene/htmlobject.rs
notdanilo: I just implemented it. This function needs to return an Err if the html text is invalid.

lib/core/src/display/scene/htmlrenderer.rs
notdanilo: Damn you, rustfmt!

lib/core/src/display/scene/htmlrenderer.rs
notdanilo: I just derived it from Default :+1:

lib/core/src/display/scene/htmlrenderer.rs
notdanilo: Is this comment incomplete?

If this isn't clear I commented it with an explanation:
```
        // Note [fov from projection matrix]
        // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        // https://github.com/mrdoob/three.js/blob/22ed6755399fa180ede84bf18ff6cea0ad66f6c0/examples/js/renderers/CSS3DRenderer.js#L275
```

lib/core/src/display/scene/htmlrenderer.rs
notdanilo: Omg, this is horrible. I will beautify this code.

About the css matrices: I had my original implementation prior to the Euler rotation order being different the one in Three.js. I didn't want to lose much time with it, so I ended up taking this https://github.com/mrdoob/three.js/blob/22ed6755399fa180ede84bf18ff6cea0ad66f6c0/examples/js/renderers/CSS3DRenderer.js#L100 as an implementation reference.

This negation looks like a transform needed by CSS3D. And we might don't need that for WebGL, for instance.

It would make sense for us to keep the representation the same between WebGL Rendering objects and HTML Rendering objects.

lib/core/src/display/scene/htmlrenderer.rs
notdanilo: The negation is different. This formatted code is definitely hard to read.

lib/core/src/display/rendering/object.rs
```diff
@@ -0,0 +1,47 @@
...
+    ///
+    /// * roll - rotates around x-axis (in radians)
+    /// * pitch - rotates around y-axis (in radians)
+    /// * yaw - rotates around z-axis (in radians)
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// // You can have rust code between fences inside the comments
+    /// // If you pass --test to Rustdoc, it will even test it for you!
```
wdanilo: I dont think this is part of our docs :)

lib/core/src/display/rendering/transform.rs
```diff
@@ -0,0 +1,100 @@
...
+
+// To comply with Threejs impl, we generate a Quaternion applying rotation in
+// the order: pitch -> roll -> yaw, instead of roll -> pitch -> yaw based on
+// https://github.com/mrdoob/three.js/blob/master/src/math/Quaternion.js#L199
+fn from_euler_angles_pry(
+        roll: f32,
+        pitch: f32,
+        yaw: f32
+    ) -> UnitQuaternion<f32>
+{
```
wdanilo: Format it like this instead please:
```rust
fn from_euler_angles_pry
(roll: f32, pitch: f32, yaw: f32) -> UnitQuaternion<f32> {
...
}
```

lib/core/src/display/rendering/transform.rs
```diff
@@ -0,0 +1,100 @@
...
+    let (s1, c1): (f32, f32) = (roll * 0.5 as f32).sin_cos();
+    let (s2, c2): (f32, f32) = (pitch * 0.5 as f32).sin_cos();
+    let (s3, c3): (f32, f32) = (yaw * 0.5 as f32).sin_cos();
```
wdanilo: very minor, but this code would look nicer with `*` aligned

lib/core/src/display/rendering/transform.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl Transform {
+    /// Creates an identity transform
+    pub fn identity() -> Self {
+        Self {
+            translation : Vector3::new(0.0, 0.0, 0.0),
+            rotation :    UnitQuaternion::identity(),
+            scale :       Vector3::new(1.0, 1.0, 1.0),
+        }
```
wdanilo: You can use 1-line exprs here to make it more clean and even shorter:
```rust
let translation = Vector3::new(0.0, 0.0, 0.0);
let rotation    = UnitQuaternion::identity();
let scale       = Vector3::new(1.0, 1.0, 1.0);
Self { translation, rotation, scale }
```

lib/core/src/lib.rs
```diff
@@ -10,6 +10,9 @@
 // === Module Structure Reexport ===
 // =================================
 
+#[macro_use]
+extern crate shrinkwraprs;
```
wdanilo: this is old rust style. use `use` instead. Or just import our `Prelude` instead :D

lib/system/web/src/lib.rs
```diff
@@ -130,6 +131,16 @@ macro_rules! fmt {
 // ===================
 // === DOM Helpers ===
 // ===================
+
+pub fn dyn_into<T, U>(obj : T) -> Result<U>
+where T : wasm_bindgen::JsCast + Debug,
+      U : wasm_bindgen::JsCast
+{
```
wdanilo: open brace goes to previous line

lib/system/web/src/lib.rs
```diff
@@ -163,3 +179,42 @@ pub fn request_animation_frame(f: &Closure<dyn FnMut()>) -> Result<i32> {
...
+impl AttributeSetter for web_sys::HtmlElement {
+    fn set_attribute_or_panic(&self, name : &str, value : &str) {
+        self.set_attribute(name, value)
+            .unwrap_or_else(|_|
+                panic!("Failed to set attribute \"{}\" = \"{}\" on \"{:?}\"",
+                        name,
+                        value,
+                        self
+                )
+            );
```
wdanilo: Ouch :O 
Single line exprs to the rescue!

lib/core/src/display/rendering/htmlobject.rs
```diff
@@ -0,0 +1,76 @@
...
+        element.set_property_or_panic("transform-style", "preserve-3d");
+        element.set_property_or_panic("position", "absolute");
+        element.set_property_or_panic("width", "0px");
+        element.set_property_or_panic("height", "0px");
+
+        Self {
+            object: Default::default(),
+            element,
+            dimensions: Vector2::new(0.0, 0.0)
+        }
```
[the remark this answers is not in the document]
farmaazon: For me, the @notdanilo version is better - i looks nearly exactly the same, and I immediately see that this is filling fields of return value. When I see `let` I never know where it can be used

lib/core/src/display/rendering/camera.rs
```diff
@@ -0,0 +1,45 @@
+use super::Object;
+use nalgebra::base::Matrix4;
+use nalgebra::geometry::Perspective3;
+use std::f32::consts::PI;
+
+/// A 3D camera representation with its own 3D
+// `Transform` and projection matrix
```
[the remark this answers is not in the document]
notdanilo: Fixed!

lib/core/src/data/opt_vec.rs
```diff
@@ -47,6 +47,15 @@ impl<T> OptVec<T> {
         item.iter().for_each(|_| self.free_ixs.push(ix));
         item
     }
+
+    /// Gets the amount of items as Some(element)
+    pub fn len(&self) -> usize {
+        self.items.len() - self.free_ixs.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
```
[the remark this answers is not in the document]
notdanilo: /// Returns true if vector contains no element.
Like in Vec too

lib/core/src/data/opt_vec.rs
```diff
@@ -47,6 +47,15 @@ impl<T> OptVec<T> {
         item.iter().for_each(|_| self.free_ixs.push(ix));
         item
     }
+
+    /// Gets the amount of items as Some(element)
```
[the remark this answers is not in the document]
notdanilo: /// Returns the number of elements in the vector, also referred to as its 'length'.
Just like in Vec

lib/core/src/display/rendering/camera.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Shrinkwrap)]
+#[shrinkwrap(mutable)]
+pub struct Camera {
+    #[shrinkwrap(main_field)]
+    pub object : Object,
+    pub projection : Matrix4<f32>,
```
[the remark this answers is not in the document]
notdanilo: What's the expected alignment here?

lib/core/src/display/rendering/camera.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl Camera {
+    /// Creates a Camera with perspective projection
+    ///
+    /// # Arguments
+    /// fov - Field of View in degrees
+    /// aspect - Aspect ratio of the screen
+    /// znear - Near distance clipping
+    /// zfar - Far distance clipping
```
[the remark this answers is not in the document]
notdanilo: I agree that obvious naming is usually better than documentation. I will rename them to
```
field_of_view_degrees: f32,
aspect_ratio: f32,
near_clipping: f32,
far_clipping: f32
```

lib/core/src/display/rendering/camera.rs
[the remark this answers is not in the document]
notdanilo: You mean to align ":"? If that's so, it's fixed.

lib/core/src/display/rendering/camera.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// # Example
+    /// ```
+    /// use basegl::display::rendering::Camera;
+    /// let dimension = (1920.0, 1080.0);
+    /// let camera = Camera::perspective(
+    ///     45.0,
+    ///     dimension.0 / dimension.1,
+    ///     1.0,
+    ///     1000.0
+    /// );
```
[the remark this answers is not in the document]
notdanilo: I think it's interesting to provide examples on things that might not be obvious for people with zero math or 3D graphics background. A Camera should be a high level concept usable by every programmer.

I made the example smaller although:

    /// ```
    /// use basegl::display::rendering::Camera;
    /// let dimension = (1920.0, 1080.0);
    /// let aspect_ratio = dimension.0 / dimension.1;
    /// let camera = Camera::perspective(45.0, aspect_ratio, 1.0, 1000.0);
    /// ```

If you think this is dispensable, let me know.

lib/core/src/display/rendering/camera.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// ```
+    pub fn perspective(fov: f32, aspect: f32, znear: f32, zfar: f32) -> Self {
+        let projection = *Perspective3::new(
+                aspect,
+                fov / 180.0 * PI,
+                znear,
+                zfar
+            ).as_matrix();
```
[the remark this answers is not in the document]
notdanilo: Now that I have renamed the args names, this expression won't fit in 80 chars. This is how it looks now:
```
        let projection = *Perspective3::new(
                aspect_ratio,
                field_of_view_degrees / 180.0 * PI,
                near_clipping,
                far_clipping
            ).as_matrix();
```

lib/core/src/display/rendering/camera.rs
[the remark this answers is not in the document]
wdanilo: Much better!

lib/core/src/display/rendering/camera.rs
```diff
@@ -0,0 +1,45 @@
...
+        Self { object: Default::default(), projection }
```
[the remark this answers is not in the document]
notdanilo: Done

lib/core/src/display/rendering/htmlobject.rs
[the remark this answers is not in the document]
notdanilo: Default isn't implemented for Vectors. We need to use Vector2::new(x,y).

I rather use 1-liners. So @wdanilo version looks cleaner for me.
If we use `let` right before `Self`, it it's immediately clear where it's being used.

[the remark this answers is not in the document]
notdanilo: We can write these expressions like this so it be used to keep 1-liners smaller than 80 chars.

lib/core/src/display/rendering/htmlobject.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /// Creates a HTMLObject from a HTML string
+    /// # Example
+    /// ```rust,no_run
+    /// use basegl::display::rendering::HTMLObject;
+    /// let html_string = "<b>hello</b>";
+    /// let object = HTMLObject::from_html_string(html_string)
+    ///                         .expect("valid object");
+    /// assert_eq!(object.element.inner_html(), html_string);
```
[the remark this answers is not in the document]
notdanilo: Working on it

lib/core/src/display/rendering/htmlobject.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// ```
+    pub fn from_html_string(html: &str) -> Result<Self> {
+        let element = create_element("div")?;
+        element.set_inner_html(html);
+        match element.children().item(0) {
```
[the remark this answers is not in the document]
notdanilo: It gets the first item :man_shrugging: But I got a substitute for it:

`match element.first_element_child() {`

lib/core/src/display/rendering/htmlobject.rs
```diff
@@ -0,0 +1,76 @@
...
+            Some(element) => Ok(Self::from_element(dyn_into(element)?)),
+            None => Err(Error::missing("valid HTML")),
+        }
+    }
+
+    /// Sets the underlying HtmlElement dimension
+    pub fn set_dimensions(&mut self, width: f32, height: f32) {
+        self.dimensions = Vector2::new(width, height);
+        self.element.set_property_or_panic("width", &format!("{}px", width));
+        self.element.set_property_or_panic("height", &format!("{}px", height));
```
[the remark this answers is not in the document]
notdanilo: AsRef is super useful here. I updated the code

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -0,0 +1,88 @@
+use super::{Camera, HTMLScene};
+use crate::system::web::StyleSetter;
+
+fn eps(value: f32) -> f32 {
+    if value.abs() < 1e-10 { 0.0 } else { value }
+}
+
+/// A renderer for `HTMLObject`s
```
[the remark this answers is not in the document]
notdanilo: dots everywhere

lib/core/src/display/rendering/scene.rs
```diff
@@ -0,0 +1,24 @@
...
+    ///
+    /// # Arguments
+    /// * id - the HtmlElement container's id
+    pub fn new(id: &str) -> Result<Self> {
+        let container = dyn_into(get_element_by_id(id)?)?;
+        Ok(Self { container })
+    }
+
+    /// Gets the HtmlElement container's dimensions
+    pub fn get_dimension(&self) -> (f32, f32) {
```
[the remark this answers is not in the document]
notdanilo: Update it to Vector2<f32> and renamed the function to `get_dimensions`

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -0,0 +1,88 @@
...
+        scene.div
+            .element
+            .set_property_or_panic("perspective", &format!("{}px", fov));
+
+        // Note [CSS Matrix3D from Camera]
+        let t = camera
+                    .transform
+                    .to_homogeneous()
+                    .try_inverse()
+                    .expect("inverse");
```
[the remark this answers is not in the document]
notdanilo: This is the best I could do:

```
        let t = camera.transform.to_homogeneous().try_inverse()
                      .expect("Render couldn't get camera's matrix inverse");
```

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl HTMLRenderer {
+    /// Creates a HTMLRenderer
+    pub fn new() -> Self { Default::default() }
+
+    /// Renders the `Scene` from `Camera`'s point of view
+    pub fn render(&self, camera: &mut Camera, scene: &HTMLScene) {
+        let (view_width, view_height) = scene.get_dimension();
+        // Note [fov from projection matrix]
+        let fov = camera.projection[5] * view_height / 2.0;
```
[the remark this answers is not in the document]
notdanilo: I clarified it. It's actually not related to fov. It's the near plane distance.

```
        // Note [znear from projection matrix]
        let expr = camera.projection[(1, 1)]; // expr = 2.0 * near / (height)
        let near = expr * view_dim.y / 2.0;
```

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        let matrix3d = format!(
+            "matrix3d({}, {}, {}, {},
+                      {}, {}, {}, {},
+                      {}, {}, {}, {},
+                      {}, {}, {}, {})",
+                      eps(t[ 0]), eps(-t[ 1]), eps(t[ 2]), eps(t[ 3]),
+                      eps(t[ 4]), eps(-t[ 5]), eps(t[ 6]), eps(t[ 7]),
+                      eps(t[ 8]), eps(-t[ 9]), eps(t[10]), eps(t[11]),
+                      eps(t[12]), eps(-t[13]), eps(t[14]), eps(t[15]));
```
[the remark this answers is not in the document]
notdanilo: I moved this to

```
pub trait IntoCSSMatrix {
    fn into_css_matrix(&self) -> String;
}
```

And I also remove the need to transform the Object matrix. We only need to invert the Y coordinates in the Camera matrix now.

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -0,0 +1,63 @@
+use super::Camera;
+use super::HTMLScene;
+use super::IntoCSSMatrix;
+use crate::system::web::StyleSetter;
+use crate::prelude::*;
+
+fn eps(value: f32) -> f32 {
```
wdanilo: This definitely needs docs. I assume nobody who has not coded GL engines would understand why its needed.

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -0,0 +1,63 @@
...
+
+impl HTMLRenderer {
+    /// Creates a HTMLRenderer.
+    pub fn new() -> Self { default() }
+
+    /// Renders the `Scene` from `Camera`'s point of view.
+    pub fn render(&self, camera: &mut Camera, scene: &HTMLScene) {
+        let half_d = scene.get_dimensions() / 2.0;
+        // Note [znear from projection matrix]
+        let expr = camera.projection[(1, 1)]; // expr = 2.0 * near / (height)
```
wdanilo: This hardcoded `(1,1)` should have a dedicated method or something. It's not descriptive enough what it really does.

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -0,0 +1,63 @@
...
+        scene.div.element.set_property_or_panic("perspective", &near);
+
+        let mut transform = camera
+                      .transform.to_homogeneous().try_inverse()
+                      .expect("Render couldn't get camera's matrix inverse")
+                      .map(|a| eps(a));
+
+        // Negating the second column to invert Y.
+        // Equivalent to scaling by (1.0, -1.0, 1.0).
+        transform.row_part_mut(1, 4).iter_mut().for_each(|a| *a = -*a);
```
wdanilo: instead of writing comment + code which is hard to understand, what about refactoring it to a function with a nice name like `invert_y`? :)

lib/core/src/display/rendering/transform.rs
```diff
@@ -0,0 +1,172 @@
...
+    fn into_css_matrix(&self) -> String;
+}
+
+impl IntoCSSMatrix for Matrix4<f32> {
+    fn into_css_matrix(&self) -> String {
+        let mut iter = self.iter();
+        let item = iter.next().expect("Matrix4 should have the first item");
+        let acc = format!("matrix3d({}", item);
+        let mut ret = iter.fold(acc, |acc, item| format!("{}, {}", acc, item));
+        ret.push(')');
```
wdanilo: This is ugly. `push(')')` is hacky. First concatenate children, then do `format!("matrix3d({})", ...)` instead.

lib/core/src/display/rendering/object.rs
```diff
@@ -0,0 +1,47 @@
+use super::Transform;
+
+/// Base structure for representing a 3D object in a `Scene`
+pub struct Object {
+    pub transform : Transform,
+}
+
+impl Default for Object {
+    fn default() -> Self {
+        Self { transform: Transform::identity() }
```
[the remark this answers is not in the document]
notdanilo: We want the default transform to be identity so it doesn't transform anything: T * I = T

lib/core/src/display/rendering/object.rs
```diff
@@ -0,0 +1,47 @@
+use super::Transform;
+
+/// Base structure for representing a 3D object in a `Scene`
+pub struct Object {
```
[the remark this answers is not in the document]
notdanilo: I like to simply call it Object because it's going to represent an object (like the real world thing we all are used to) in a virtual space.

Do you think calling it Object might be confusing?

lib/core/src/display/rendering/renderer.rs
```diff
@@ -0,0 +1,11 @@
...
+
+/// Base structure for our Renderers
+#[derive(Default)]
+pub struct Renderer {}
+
+impl Renderer {
+    pub fn new() -> Self {
+        Default::default()
+    }
+}
```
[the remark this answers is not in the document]
notdanilo: It isn't used. I made it when I was modelling Renderer -> HTMLRenderer inheritance relationship. But it ended up empty.

lib/core/src/display/rendering/htmlrenderer.rs
notdanilo: I really don't know how to simply describe it without explaining how a projection matrix is composed.

PR #49, opened 2019-11-18
Add wasm-pack test to github workflows

farmaazon:
### Pull Request Description
To the test workflow added a step where wasm-pack test is run for each crate in workspace.

### Important Notes
- A script was added, which runs wasm-pack test for each workspace members, because The wasm-pack itself cannot read workspaces. The script was written in rust, therefore a special crate for build-and-utility-scripts in rust was created.
- The wasm32 target does not support -Zprofile flag, therefore wasm_bindgen_tests cannot work with the used coverage reports generators. For now I left it as it is, in future maybe someone find another tool supporting wasm32.

### Checklist


script/rust/src/bin/test-all.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+/// Call wasm-pack test for each workspace member
+///
+/// This function reads workspace members list from `Cargo.toml` in current
+/// directory, and call `wasm-pack test` each member. All script arguments
+/// are passed to wasm-pack test process.
+fn main()
+{
```
wdanilo: formatting

lib/system/web/js/resize_observer.js
```diff
@@ -54,12 +54,12 @@ class Pool {
...
 let resizeObserverPool = new Pool((...args) => new ResizeObserver(...args))
 
-export function resize_observe(target, f) {
+function resize_observe(target, f) {
     let id = resizeObserverPool.reserve(resize_observer_update(f))
     resizeObserverPool[id].observe(target)
 }
 
-export function resize_unobserve(id) {
+function resize_unobserve(id) {
```
wdanilo: Why those changes?

script/rust/src/bin/test-all.rs
```diff
@@ -0,0 +1,43 @@
+
+fn get_workspace_members(
+    cargo_toml_root : toml::Value
+) -> Vec<String> {
```
wdanilo:
```rust
fn get_workspace_members
(cargo_toml_root : toml::Value) -> Vec<String> {
```

lib/system/web/js/resize_observer.js
farmaazon: Once I changed test in core from #[test] to #[wasm_bindgen_test] and it turns out that wasm-pack test does not work with "export" word (even with --node option...). But I don't know why, I don't have a deep knowledge of Js.

farmaazon: The exact error is ```Running /home/adam-praca/dev/basegl/target/wasm32-unknown-unknown/debug/deps/basegl-c02b39c2bb2e07d5.wasm
/home/adam-praca/dev/basegl/target/wasm32-unknown-unknown/wbg-tmp/snippets/basegl-system-web-056d2c0780aef067/js/resize_observer.js:57
export function resize_observe(target, f) {
^^^^^^

SyntaxError: Unexpected token 'export'
```
But without exports the example 01-scene does not work. Now I'm putting version which works on both browser and wasm-pack test

script/rust/src/bin/test-all.rs
farmaazon: fixed

script/rust/src/bin/test-all.rs
farmaazon: fixed

lib/system/web/js/resize_observer.js
wdanilo: strange :/

PR #58, opened 2019-11-26
HTMLRenderer Benchmarks

notdanilo:
### Pull Request Description
This PR adds:
1. benchmark tests for our HTMLRenderer;
2. New benchmark and test macros: `#[web_test]` and `#[web_bench]`

### Important Notes
* We now have a web benchmark capability with API similar to Rust's Bencher.

### Checklist


wdanilo:
Lets merge that after we finish benchmarks vs threejs and and improve our perf.

lib/core/src/display/rendering/scene.rs
```diff
@@ -1,24 +1,99 @@
...
+}
+
+// =============
+// === Scene ===
+// =============
 
 /// A collection for holding 3D `Object`s.
-#[derive(Debug)]
+//#[derive(Debug)]
 pub struct Scene {
```
farmaazon: This struct keeps scene of html objects only, or of any kind of objects?

lib/core/src/display/rendering/transform.rs
```diff
@@ -52,16 +52,31 @@ impl Transform {
         self.translation = Vector3::new(x, y, z);
     }
 
-    /// Set Transform's scale.
+    /// Gets Transform's translation.
```
farmaazon: Maybe it is not part of this review, but... there is nalgebra::Projective2/3 structure, which I think does exactly the same thing as our Transform, doesn't it?

lib/core/src/math/utils.rs
```diff
@@ -56,3 +71,8 @@ pub fn invert_y(mut m: Matrix4<f32>) -> Matrix4<f32> {
     m.row_part_mut(1, 4).iter_mut().for_each(|a| *a = -*a);
     m
 }
+
+// Note [2D array to 1D array]
+// ===========================
+// We need to interpret [[f32; 4]; 4] as [f32; 16] to view as Float32Array.
+// The memory layout is the same, so this operation is safe.
```
farmaazon: Not quite. The `Float32Array` from `view` may be invalidated if there is _any_ allocation done after its creation. So if someone not suspecting anything (because function is safe, right?) creates Float32Array from matrix and makes a simple Box before reading array, he may be surprised

farmaazon: And you assumes that Matrix4 from nalgebra keeps data in `[[f32; 4]; 4]` what might be not the case, in fact i don't see in nalgebra docs any guarantees about this. So the code is very unstable, a upgrade of nalgebra may cause unexpected behaviours (or new optimization in rustc)

lib/core/tests/bench_test.js
```diff
@@ -0,0 +1,6 @@
+export function set_gradient_bg(dom, r, g, b) {
+  let components = r + "," + g + "," + b;
```
farmaazon: Or `[r,g,b].join()` if you like it

lib/web-test-proc-macro/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "web-test-proc-macro"
+version = "0.1.0"
+authors = ["Danilo Guanabara <danguafer@gmail.com>"]
```
farmaazon: recently we start to put a global enso team email

lib/web-test-proc-macro/src/lib.rs
```diff
@@ -0,0 +1,55 @@
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use syn::*;
+use quote::quote;
+
+#[proc_macro_attribute]
+pub fn web_test(_args: TokenStream, input: TokenStream) -> TokenStream {
```
farmaazon: What this macro is doing? I'm not good in reading macros definitions, a short comment would help here

lib/web-test/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "web-test"
+version = "0.1.0"
+authors = ["Danilo Guanabara <danguafer@gmail.com>"]
```
farmaazon: Enso team

lib/web-test/src/benchcontainer.rs
```diff
@@ -0,0 +1,53 @@
+use crate::prelude::*;
+use crate::system::web::create_element;
+use crate::system::web::dyn_into;
+use crate::system::web::NodeInserter;
+use crate::system::web::StyleSetter;
+use web_sys::HtmlElement;
+use super::Container;
+
+/// Html benchmark container.
```
farmaazon: "Html benchmarks" sounds like there were html program we want to measure with benchmarks :) Maybe "Html container displaying benchmark results"?

lib/system/web/src/animationframeloop.rs
```diff
@@ -0,0 +1,60 @@
...
+        let nop_closure = Closure::wrap(nop_func);
+        let callback    = Rc::new(RefCell::new(nop_closure));
+
+        let run  = true;
+        let data = Rc::new(RefCell::new(AnimationFrameData { run }));
+
+        let callback_clone = callback.clone();
+        let data_clone     = data.clone();
+
+        *callback.borrow_mut() = Closure::wrap(Box::new(move || {
```
farmaazon: We expects that function will be called only once from js, so we can use [once_for_js](https://docs.rs/wasm-bindgen/0.2.55/wasm_bindgen/closure/struct.Closure.html#method.once_into_js)

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -2,11 +2,23 @@ use crate::prelude::*;
...
+#[wasm_bindgen(module = "/js/htmlrenderer.js")]
+extern "C" {
+    fn set_object_transform(dom: &JsValue, matrix_array: &JsValue);
+    fn setup_perspective(dom: &JsValue, znear : &JsValue);
+    fn setup_camera_transform(
+        dom          : &JsValue,
+        znear        : &JsValue,
+        half_width   : &JsValue,
+        half_height  : &JsValue,
+        matrix_array : &JsValue);
```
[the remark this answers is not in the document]
notdanilo: It won't fit in 80chars even if I do that :(

lib/core/src/display/rendering/htmlscene.rs
```diff
@@ -18,45 +45,53 @@ use crate::data::types::Index;
...
+    /// Sets the HTMLScene DOM's dimensions.
+    pub fn set_dimensions(&mut self, dimensions : Vector2<f32>) {
+        self.html_data.set_dimensions(dimensions);
+        self.scene.set_dimensions(dimensions);
     }
 
     /// Moves a HTMLObject to the Scene and returns an index to it.
     pub fn add(&mut self, object: HTMLObject) -> Index {
-        self.camera.element.append_child_or_panic(&object.element);
+        self.html_data.camera.element.append_child_or_panic(&object.element);
```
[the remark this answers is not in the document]
notdanilo: It's private now and we can access it through its iterators

lib/core/src/display/rendering/scene.rs
notdanilo: It's supposed to hold other objects too. I am extending it now that I am implementing a PhysicsRenderer

lib/core/src/display/rendering/scene.rs
```diff
@@ -1,24 +1,99 @@
...
-    pub container : HtmlElement,
+    pub dom : HtmlElement,
+    _resize : ResizeObserver,
```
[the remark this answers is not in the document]
notdanilo: Yes, that's the case here. I need to hold the resource as long `Scene` lives, but I am not accessing `_resize` anywhere in the code.

lib/core/src/display/rendering/scene.rs
```diff
@@ -1,24 +1,99 @@
...
+
+        let data_clone = data.clone();
+        let resize = Closure::new(move |width, height| {
+            let mut data = data_clone.borrow_mut();
+            data.dimensions = Vector2::new(width as f32, height as f32);
+            for callback in &data.callbacks {
+                callback(&data.dimensions);
+            }
+        });
+        let _resize = ResizeObserver::new(&dom, resize);
```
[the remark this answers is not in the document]
notdanilo: I renamed them to `resize_closure` and `_resize_observer`

lib/core/src/display/rendering/transform.rs
notdanilo: Our Transform struct is used to keep track of Position, Rotation and Scale separately. The CPU operates faster over the Transform this way. Once we send it to the GPU, we can convert it to homogeneous matrix.

lib/core/src/math/utils.rs
notdanilo: @farmaazon I am not sure I understood the scenario. What kind of allocation can break it? Can you provide more detail, maybe with an example?

About the memory layout, here is the documentation for AsRef:

https://docs.rs/nalgebra/0.19.0/nalgebra/base/struct.Matrix.html#impl-AsRef%3C%5B%5BN%3B%204%5D%3B%204%5D%3E

This kind of memory layout for matrices is common in math libraries. If the Matrix memory layout (including `majorness`) breaks in future releases, we can catch it with our units tests.

I have added this test to make sure we have the expected memory layout:

```rust
    #[test]
    fn matrix4_memory_layout() {
        use super::Matrix4;
        let matrix = Matrix4::<f32>::new
            ( 1.0, 5.0,  9.0, 13.0
            , 2.0, 6.0, 10.0, 14.0
            , 3.0, 7.0, 11.0, 15.0
            , 4.0, 8.0, 12.0, 16.0 );
        let matrix = matrix.as_ref();
        unsafe {
            let layout = &*(matrix as *const [[f32; 4]; 4] as *const [f32; 16]);
            for (i, n) in layout.iter().enumerate() {
                assert_eq!((i + 1) as f32, *n);
            }
        }
    }
```

lib/core/tests/bench_test.js
notdanilo: Wouldn't the explicit string concatenation be faster? I am calling this function 400 times per frame.

lib/system/web/src/animationframeloop.rs
notdanilo: `Closure::once_into_js` returns a `JsValue` which cannot be used with `request_animation_frame`

lib/web-test-proc-macro/src/lib.rs
notdanilo: I added this doc
```rust
/// #[web_test] creates a [320.0, 240.0] div with id = fn_name and appends it
/// into the document.
```

and this note

```rust
            // We insert Container::new("Tests", fn_name, 320.0, 240.0)
            // at the beginning of the function block.
            parsed.block.stmts.insert(0, stmt);
```

lib/web-test/src/benchcontainer.rs
notdanilo: Fixed!

lib/core/src/math/utils.rs
notdanilo: Also, this can make this function safe:
```rust
#[derive(Shrinkwrap)]
pub struct Float32ArrayView<'a> {
    #[shrinkwrap(main_field)]
    array   : Float32Array,
    phantom : PhantomData<&'a Float32Array>
}

pub trait IntoFloat32Array {
    fn into_float32array<'a>(&'a self) -> Float32ArrayView<'a>;
}

impl IntoFloat32Array for Matrix4<f32> {
    fn into_float32array<'a>(&'a self) -> Float32ArrayView<'a> {
        // Note [2D array to 1D array]
        unsafe {
            let matrix = self.as_ref();
            let matrix = matrix as *const [[f32; 4]; 4];
            let array = Float32Array::view(&*(matrix as *const [f32; 16]));
            let phantom = PhantomData;
            Float32ArrayView { array, phantom }
        }
    }
}
```

wdanilo: @farmaazon is right. This is **VERY** unsafe and we should not use it. See the docs of `js_sys::Float32Array::view`: 
```
Views into WebAssembly memory are only valid so long as the backing buffer isn't resized in JS. Once this function is called any future calls to Box::new (or malloc of any form) may cause the returned value here to be invalidated. Use with caution!

Additionally the returned object can be safely mutated but the input slice isn't guaranteed to be mutable.

Finally, the returned object is disconnected from the input slice's lifetime, so there's no guarantee that the data is read at the right time.
```

Keeping this `View` in a structure is a very, very, very bad idea. Danilo, I've got a request - when writing docs about safety, please address **every** safety note from **every** unsafe function you use there, including for example the `view` function. 

Btw, I think we should introduce the same thing the Firefox team did:

wdanilo: tall_unsafe_1

wdanilo: tall_unsafe_2

lib/core/tests/htmlrenderer.rs
```diff
@@ -1,30 +1,37 @@
...
+use wasm_bindgen::JsValue;
 
-pub mod common;
+#[wasm_bindgen(module = "/tests/bench_test.js")]
+extern "C" {
+    fn set_gradient_bg(
+        dom           : &JsValue,
+        red           : &JsValue,
+        green         : &JsValue,
+        blue          : &JsValue);
```
wdanilo: indentation

lib/core/tests/htmlrenderer.rs
```diff
@@ -1,30 +1,37 @@
...
         let scene = HTMLScene::new("nonexistent_id");
         assert!(scene.is_err(), "nonexistent_id should not exist");
     }
 
-    #[wasm_bindgen_test]
+    #[web_test]
     fn object_behind_camera() {
-        TestContainer::new("object_behind_camera", 320.0, 240.0);
         let mut scene = HTMLScene::new("object_behind_camera")
                                   .expect("Failed to create HTMLScene");
```
wdanilo: As I've written above, let's jst always keep indentation in newline to be 4 spaces, not depending on previous line.

lib/system/web/src/animationframeloop.rs
```diff
@@ -0,0 +1,60 @@
...
+// ============================
+
+struct AnimationFrameData {
+    run : bool
+}
+
+pub struct AnimationFrameLoop {
+    forget : bool,
+    data   : Rc<RefCell<AnimationFrameData>>
+}
```
wdanilo: You've mentioned that you want to merge this animation frame loop with or other one. Is this merge going to happen in another PR?

lib/system/web/src/lib.rs
```diff
@@ -244,6 +259,26 @@ impl NodeAppender for Node {
                 )
             );
     }
+
+    fn prepend_or_panic(&self, node : &Node) {
```
wdanilo: The naming is inconsistent. We've got `append_CHILD_or_panic` and `prepend_or_panic`.

lib/web-test-proc-macro/src/lib.rs
wdanilo: The docs shold mention how to use it. It's actially not obvious to me reading the current docs and signature.

lib/core/src/math/utils.rs
farmaazon: With AsRef you're right, that's legitimate use of casting.
As for `Flotat32Array::view` the example is:
```rust
let arr = matrix.into_float32array();
let some_data : Box<Whatever> = Box::new(); // here is new allocation, 
// so created view is invalidated (see a docs fragment mentioned by @wdanilo )
let_pass_this_array_to_js(arr); // however an user does not expect it 
// and pass invalid view somewhere else

File: lib/core/js/htmlrenderer.js [hunk 18]
```diff
@@ -0,0 +1,22 @@
...
+
+export function setup_camera_transform(
+                        dom,
+                        znear,
+                        half_width,
+                        half_height,
+                        matrix_array) {
+  let transform = "translateZ(" + znear + "px)"
+                + arr_to_css_matrix3d(matrix_array)
+                + "translate(" + half_width + "px" + ", " + half_height + "px)"
```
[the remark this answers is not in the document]
farmaazon: "In no lang" is too much said. In c++ is a common custom to make longer indent inside a one instruction, and shorter in blocks. And I think the default Intellij settings for Java make longer indents as well. 
However, I also opts for using 4 spaces everywhere

lib/core/src/data/opt_vec.rs
```diff
@@ -91,6 +91,32 @@ impl<'a, T> IntoIterator for &'a OptVec<T> {
...
+
+impl<'a, T> Iterator for IterMut<'a, T> {
+    type Item = &'a mut T;
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(opt_item) = self.iter.next() {
+            // If opt_item has some item, we return it, else we try the next one
+            if let Some(item) = opt_item { Some(item) } else { self.next() }
+        } else { None }
+    }
+}
```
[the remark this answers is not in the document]
farmaazon: The sense of this code is to skip `None` and unwrap `Some` - that cannot be done with aliases... perhaps...

lib/core/js/htmlrenderer.js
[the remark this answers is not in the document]
notdanilo: I am not sure I got it. Shouldn't we prioritize align `+` (concatenation) with `=` (assignment)?

lib/core/src/data/opt_vec.rs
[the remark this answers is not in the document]
notdanilo: What @farmaazon said. This Iterator was implemented so we don't need to handle the Option in  Vec<Option<T>> for every iteration.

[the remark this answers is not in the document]
notdanilo: @wdanilo I just saw your stylish solution with `filter_map`. We will keep it.

lib/core/src/math/utils.rs
notdanilo: Thanks for bringing that to my attention. I missed the Unsafety note of `Float32Array::view`.

I qualified the function as unsafe and I also provided documentation pointing out why it's unsafe.

lib/web-test-proc-macro/src/lib.rs
notdanilo: We now have an usage example for `#[web_test]` and `#[web_test]`

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -22,32 +34,40 @@ impl HTMLRenderer {
...
 
-        scene.div   .element.set_property_or_panic("perspective", near);
-        scene.camera.element.set_property_or_panic("transform"  , css);
+        // Note [znear from projection matrix]
+        let half_dim     = scene.get_dimensions() / 2.0;
+        let expr         = camera.projection[(1, 1)];
+        let near         = (expr * half_dim.y).into();
+
+        let half_width   = half_dim.x.into();
+        let half_height  = half_dim.y.into();
```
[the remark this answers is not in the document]
notdanilo: It makes sense!

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -22,32 +34,40 @@ impl HTMLRenderer {
...
-        let trans_z   = format!("translateZ({})", near);
-        let matrix3d  = trans_cam.into_css_matrix();
-        let trans     = format!("translate({}px,{}px)", half_dim.x, half_dim.y);
-        let css       = format!("{} {} {}", trans_z, matrix3d, trans);
 
-        scene.div   .element.set_property_or_panic("perspective", near);
-        scene.camera.element.set_property_or_panic("transform"  , css);
+        // Note [znear from projection matrix]
+        let half_dim     = scene.get_dimensions() / 2.0;
+        let expr         = camera.projection[(1, 1)];
```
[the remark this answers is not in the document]
notdanilo: This is accessing the m(1, 1) element of the homogeneous matrix, which means the y-scale. I created a method with this signature: `Camera::get_y_scale() -> f32`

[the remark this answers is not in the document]
notdanilo: @wdanilo it's cleaner than getting the y-scale through projection[(1, 1)]. But this y-scale can be understood as an internal of homogeneous matrices for keeping the aspect ratio right which I think shouldn't be exposed in Camera.

The Renderer is supposed to know how a homogeneous matrix works, that's why I directly got the y-scale from the matrix [1, 1] coordinate.

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -22,32 +34,40 @@ impl HTMLRenderer {
...
+
+        let half_width   = half_dim.x.into();
+        let half_height  = half_dim.y.into();
 
-        for object in &scene.objects {
+        let matrix_array = trans_cam.into_float32array();
+
+        setup_perspective(&scene.html_data.div.element, &near);
+        setup_camera_transform(
+            &scene.html_data.camera.element,
```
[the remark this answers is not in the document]
notdanilo: Updated it to
```rust
        let scene_dom    = &scene.html_data.div.element;
        let camera_dom   = &scene.html_data.camera.element;
```

and

```
        setup_perspective(&scene_dom, &near);
        setup_camera_transform(
            &camera_dom,
```

lib/core/src/display/rendering/htmlscene.rs
```diff
@@ -2,12 +2,39 @@ use crate::prelude::*;
 
 use super::HTMLObject;
 use super::Scene;
-use crate::data::opt_vec::OptVec;
+use crate::data::opt_vec::*;
```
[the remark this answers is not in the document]
notdanilo: Updated it to
```
use crate::data::opt_vec::OptVec;
use crate::data::opt_vec::Iter;
use crate::data::opt_vec::IterMut;
```

Just a heads up. I started to use wildcards on the camera-system branch like this:

In `rendering/html/mod.rs`:
```rust
mod htmlscene;
pub use htmlscene::*;
```

This will make all the public API in `htmlscene` available in `basegl::rendering::html`, so you can use them as:

```rust
basegl::rendering::html::HTMLScene;
```

instead of

```rust
basegl::rendering::html::htmlscene::HTMLScene;
```

lib/core/src/display/rendering/htmlscene.rs
```diff
@@ -18,45 +45,53 @@ use crate::data::types::Index;
...
+        let objects   = default();
+
+        scene .dom    .append_child_or_panic(&div.element);
+        div   .element.append_child_or_panic(&camera.element);
+        camera.element.set_property_or_panic("transform-style", "preserve-3d");
+
+        let html_data = Rc::new(HTMLSceneData::new(div, camera));
+
+        let html_data_clone = html_data.clone();
+        scene.add_resize_callback(Box::new(move |dimensions| {
```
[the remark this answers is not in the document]
notdanilo: You are right! I abstracted this function like this:

```rust
    pub fn add_resize_callback<T>(&mut self, callback : T)
        where T : Fn(&Vector2<f32>) + 'static {
        self.data.borrow_mut().resize_callbacks.push(Box::new(callback));
    }
```

lib/core/src/data/opt_vec.rs
```diff
@@ -145,4 +171,27 @@ mod tests {
...
         }
     }
+
+    #[test]
+    fn test_iter_mut() {
+        let mut v = OptVec::new();
+
+        let  ix1 = v.insert(|_| 0);
+        let _ix2 = v.insert(|_| 1);
+        let _ix3 = v.insert(|_| 2);
```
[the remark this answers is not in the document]
notdanilo: I fixed the conflicts

lib/core/src/math/utils.rs
wdanilo: I don't see it. The code is still:

```rust
impl IntoFloat32Array for Matrix4<f32> {
    fn into_float32array(&self) -> Float32Array {
        // Note [2D array to 1D array]
        unsafe {
            let matrix = self.as_ref();
            Float32Array::view(&*(matrix as *const [[f32; 4]; 4] as *const [f32; 16]))
        }
    }
}
```

wdanilo: Btw, let's move the note to the doc comment of the whole function

notdanilo: @wdanilo I pushed it few minutes ago. Please check it now.

lib/system/web/src/animationframeloop.rs
farmaazon: Right. Then there is another function you may consider: `Closure::once`

lib/core/js/htmlrenderer.js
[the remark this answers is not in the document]
wdanilo: You can align `=` and `+`, but then move `=` to newline as well. Even better would be to distribute it to vars :)

lib/core/src/display/rendering/htmlrenderer.rs
```diff
@@ -22,32 +34,48 @@ impl HTMLRenderer {
...
+        for object in scene {
             let mut transform = object.transform.to_homogeneous();
             transform.iter_mut().for_each(|a| *a = eps(*a));
-            let matrix3d = transform.into_css_matrix();
-            let css      = format!("translate(-50%, -50%) {}", matrix3d);
-            object.element.set_property_or_panic("transform", css);
+
+            // Note [Unsafe Float32ArrayView]
+            let matrix_array = unsafe { transform.into_float32_array_view() };
+            set_object_transform(&object.element, &matrix_array);
```
wdanilo: This code is error prone. It's too easy to insert a line between these lines and make it break. Please put inside `unsafe` block the whole unsafe code and refactor it to a safe function. Document the function extensively. Documentation of each function constining unsafe block should be documented using the special Safety / Unsafety sections. We should not use `unsafe` call as a part of safe code. The idea is to refactor unsafe fragments to self-contained safe functions which are guaranteed to work correctly.

lib/core/src/display/rendering/htmlscene.rs
```diff
@@ -18,45 +45,53 @@ use crate::data::types::Index;
 #[shrinkwrap(mutable)]
 pub struct HTMLScene {
```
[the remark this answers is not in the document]
notdanilo: We no longer have HTMLScene. There is only Scene<T> :)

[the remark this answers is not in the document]
notdanilo: And DOM was moved to GraphicsRenderer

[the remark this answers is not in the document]
notdanilo: @wdanilo but now `HTMLRenderer` derefs to `GraphicsRenderer` and both of them has `set_dimension`. (I just moved the function from `Scene` to `Renderer` along with the DOM).

What would be a good alternative solution for it?

lib/core/src/display/rendering/htmlscene.rs
```diff
@@ -18,45 +45,53 @@ use crate::data::types::Index;
...
+        let camera    = HTMLObject::new("div")?;
+        let objects   = default();
+
+        scene .dom    .append_child_or_panic(&div.element);
+        div   .element.append_child_or_panic(&camera.element);
+        camera.element.set_property_or_panic("transform-style", "preserve-3d");
+
+        let html_data = Rc::new(HTMLSceneData::new(div, camera));
+
+        let html_data_clone = html_data.clone();
```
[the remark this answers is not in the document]
notdanilo: I usually declare the variables just before its use. In this case `html_data` takes ownership of `camera` and we have to mutate it before moving.

lib/core/src/display/rendering/htmlscene.rs
```diff
@@ -18,45 +45,53 @@ use crate::data::types::Index;
...
+        self.html_data.camera.element.append_child_or_panic(&object.element);
         self.objects.insert(|_| object)
     }
 
     /// Removes and retrieves a HTMLObject based on the index provided by
     pub fn remove(&mut self, index: usize) -> Option<HTMLObject> {
         let result = self.objects.remove(index);
         result.iter().for_each(|object| {
-            self.camera.element.remove_child_or_panic(&object.element);
+           self.html_data.camera.element.remove_child_or_panic(&object.element);
```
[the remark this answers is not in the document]
notdanilo: DOM management and rendering was moved to `HTMLRenderer` and this function no longer exist

lib/core/src/display/rendering/htmlscene.rs
```diff
@@ -72,3 +107,19 @@ impl HTMLScene {
...
+    }
+}
+
+impl<'a> IntoIterator for &'a mut HTMLScene {
+    type Item = &'a mut HTMLObject;
+    type IntoIter = IterMut<'a, HTMLObject>;
+    fn into_iter(self) -> Self::IntoIter {
+        (&mut self.objects).into_iter()
+    }
+}
```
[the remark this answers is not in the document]
notdanilo: I agree. This functionality is too simple for that much writing :( But afaik there is another way to express it differently.

lib/core/src/display/rendering/scene.rs
```diff
@@ -1,24 +1,99 @@
...
+// === Scene ===
+// =============
 
 /// A collection for holding 3D `Object`s.
-#[derive(Debug)]
+//#[derive(Debug)]
 pub struct Scene {
-    pub container : HtmlElement,
+    pub dom          : HtmlElement,
+    _resize_observer : ResizeObserver,
```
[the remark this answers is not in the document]
notdanilo: Because `_resize_observer` will not be used, but we need to keep track of it until `Scene` is dropped. This functionality was moved to `GraphicsRenderer`.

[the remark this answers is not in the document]
notdanilo: @wdanilo actually, I have been mentioning that DOM resizing was moved to `GraphicsRenderer`, but I made a separation of concerns and created `DomContainer` and `GraphicsRenderer` owns it. This is the right place to look it.

lib/core/src/display/rendering/scene.rs
```diff
@@ -1,24 +1,99 @@
...
+// =================
+// === SceneData ===
+// =================
+
+#[derive(Derivative)]
+#[derivative(Debug)]
+struct SceneData {
+    dimensions : Vector2<f32>,
+    #[derivative(Debug="ignore")]
+    resize_callbacks : Vec<ResizeCallback>
```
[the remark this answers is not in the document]
notdanilo: I will leave that for the future then

lib/core/src/display/rendering/scene.rs
```diff
@@ -1,24 +1,99 @@
...
+        self.dom.set_property_or_panic("height", format!("{}px", dimensions.y));
+        self.data.borrow_mut().dimensions = dimensions;
+    }
+
+    /// Gets the Scene DOM's dimensions.
     pub fn get_dimensions(&self) -> Vector2<f32> {
-        let width  = self.container.client_width()  as f32;
-        let height = self.container.client_height() as f32;
-        Vector2::new(width, height)
+        self.data.borrow().dimensions
```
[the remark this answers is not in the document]
notdanilo: Welp, clippy told me that :P

lib/system/web/src/animationframeloop.rs
notdanilo: This will happen when we get rid of our duplicates

lib/system/web/src/lib.rs
notdanilo: Renamed to:

```rust
    fn append_or_panic(&self, node : &Node);
    fn prepend_or_panic(&self, node : &Node);
    fn insert_before_or_panic(&self, node : &Node, reference_node : &Node);
```

lib/core/src/display/rendering/htmlrenderer.rs
notdanilo: I created a safe function wrapping the unsafe code :+1:

lib/core/src/display/rendering/dom_container.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Gets the Scene DOM's dimensions.
+    pub fn dimensions(&self) -> Vector2<f32> {
+        self.data.borrow().dimensions
+    }
+
+    /// Adds a ResizeCallback.
+    pub fn add_resize_callback<T>(&mut self, callback:T)
+        where T : Fn(&Vector2<f32>) + 'static {
+        self.data.borrow_mut().resize_callbacks.push(Box::new(callback));
+    }
```
wdanilo: Why we are using `'static` here? I'm trying to always remove every `'static` from codebase - thats why Im asking about it. 

Ok, I see why. You are converting that to `dyn` trait. In usch a case, I would create a type alias for that and encode the `'static` in this type alias in order to explicitly tell why its needed. Somethingl ike:

```
trait ResizeCallbackFn = where Self:  Fn(&Vector2<f32>) + 'static
```
and change the line to
```
where T: ResizeCallbackFn
```

lib/core/src/display/rendering/html/html_renderer.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+fn set_object_transform(dom: &JsValue, matrix: &Matrix4<f32>) {
+    // Note [Unsafe Float32ArrayView]
+    unsafe {
+        let matrix_array =  matrix.into_float32_array_view();
+        js::set_object_transform(&dom, &matrix_array);
+    }
+}
```
wdanilo: In such case, after refactoring it ot separate function, its IMO better to put the documentation + Safety section directly on the function `set_object_transform`. The whole point of refactoring it this way is that we are just encapsulating unsafe things in a safe interface, so it should be documented here. There is no more need to do some redirection - and thats super cool, cause we dont need to search for the docs in the file anymore! :)

lib/core/src/display/symbol/buffer.rs
```diff
@@ -306,7 +306,7 @@ impl<T> Builder<T> {
 // ========================
 
 // TODO The following code should be refactored to use the new macro `eval-tt`
-// TODO engine. Some utils, like `cartesian` macro should also be refactored
+// TODO rendering. Some utils, like `cartesian` macro should also be refactored
```
wdanilo: It doesnt make sense now 😂 What is `eval-tt` rendering? :O Btw, Danilo, I suggest using IntelliJ refactoring, it works really nice and saves ton of time.

lib/core/src/math/utils.rs
```diff
@@ -12,18 +17,52 @@ pub trait IntoCSSMatrix {
...
+    phantom : PhantomData<&'a Float32Array>
+}
+
+pub trait IntoFloat32ArrayView {
+    /// # Safety
+    /// Views into WebAssembly memory are only valid so long as the backing buffer isn't resized in JS. Once this function is called any future calls to Box::new (or malloc of any form) may cause the returned value here to be invalidated. Use with caution!
+    ///
+    /// Additionally the returned object can be safely mutated but the input slice isn't guaranteed to be mutable.
+    ///
+    /// Finally, the returned object is disconnected from the input slice's lifetime, so there's no guarantee that the data is read at the right time.
```
wdanilo: Way over 100 chars :P

lib/core/src/math/utils.rs
```diff
@@ -12,18 +17,52 @@ pub trait IntoCSSMatrix {
...
+impl IntoFloat32ArrayView for Matrix4<f32> {
+    unsafe fn into_float32_array_view(&self) -> Float32ArrayView<'_> {
+        // Note [2D array to 1D array]
+        let matrix = self.as_ref();
+        let matrix = matrix as *const [[f32; 4]; 4];
+        let array = Float32Array::view(&*(matrix as *const [f32; 16]));
+        let phantom = PhantomData;
+        Float32ArrayView { array, phantom }
+    }
+}
```
wdanilo: This is the only implementation of this trait and unsafe function - in such case, why do we want to keep it as a trait, not just as a single unsafe function? Maybe it's a good design, but Id like to confirm it.

lib/core/src/math/utils.rs
```diff
@@ -56,3 +112,8 @@ pub fn invert_y(mut m: Matrix4<f32>) -> Matrix4<f32> {
     m.row_part_mut(1, 4).iter_mut().for_each(|a| *a = -*a);
     m
 }
+
+// Note [2D array to 1D array]
+// ===========================
+// We need to interpret [[f32; 4]; 4] as [f32; 16] to view as Float32Array.
+// The memory layout is the same, so this operation is safe.
```
wdanilo: Imo this comment on unsafe function is still misleading. It tells it is safe, while the function is marked unsafe.

lib/core/src/math/utils.rs
```diff
@@ -38,11 +77,28 @@ mod tests {
...
+            , 2.0, 6.0, 10.0, 14.0
+            , 3.0, 7.0, 11.0, 15.0
+            , 4.0, 8.0, 12.0, 16.0 );
+        let matrix = matrix.as_ref();
+        unsafe {
+            let layout = &*(matrix as *const [[f32; 4]; 4] as *const [f32; 16]);
+            for (i, n) in layout.iter().enumerate() {
+                assert_eq!((i + 1) as f32, *n);
+            }
+        }
```
wdanilo: Undocumented usage of unsafe. Moreover, I assume that this was meant to check if our "flattenning method" works. But instead of using the method it copy-pastes the logic code, so in fact it doesnt check if our utils are correct. We can refactor one of the codes, and this tests will be not synchronised anymore.

lib/core/src/display/rendering/dom_container.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl fmt::Debug for DOMContainer {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self.dom)?;
+        write!(f, "{:?}", self.resize_observer)?;
+        write!(f, "{:?}", self.data.borrow())
+    }
+}
```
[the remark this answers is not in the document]
notdanilo: I may have missed it. I am using `#[derive(Debug)]` instead.

lib/core/src/display/rendering/dom_container.rs
notdanilo: Done. I also fixed `GraphicsRenderer::add_resize_callback`

lib/core/src/display/symbol/buffer.rs
notdanilo: Fun fact: I used IntelliJ refactoring everywhere lol I don't know what happened here.

I renamed it back to `engine`.

lib/core/src/math/utils.rs
notdanilo: Oopsy! It's fixed

lib/core/src/display/rendering/html/html_renderer.rs
notdanilo: Ah, yes! But the only reason I am redirecting is because I am using the same artifice in `setup_camera_transform`, so I didn't want to have a duplicate comment. Makes sense?

lib/core/src/math/utils.rs
notdanilo: The reasoning here is that we might want to use this view on other objects as well and because I think it's prettier to use this function as a method-like fashion like `matrix.into_float32_array_view()`.

lib/core/src/math/utils.rs
notdanilo: In this function there are two different unsafe operations: raw pointer casting and the use of `Float32Array::view`. What if I group here? Will it be more evident?

```rust
impl IntoFloat32ArrayView for Matrix4<f32> {
    unsafe fn into_float32_array_view(&self) -> Float32ArrayView<'_> {
        // Note [2D array to 1D array]
        let matrix = {
            let matrix = self.as_ref();
            let matrix = matrix as *const [[f32; 4]; 4];
            &*(matrix as *const [f32; 16]);
        };
        let array = Float32Array::view(matrix);
        let phantom = PhantomData;
        Float32ArrayView { array, phantom }
    }
}
```

lib/core/src/math/utils.rs
notdanilo: Thanks for pointing it out. I refactored this code to:

```rust
pub fn matrix4_to_array(matrix:&Matrix4<f32>) -> &[f32; 16] {
    // We need to interpret [[f32; 4]; 4] as [f32; 16] to view as Float32Array.
    // The memory layout is the same, so this operation is safe.
    unsafe {
        &*(matrix.as_ref() as *const [[f32; 4]; 4] as *const [f32; 16])
    }
}
```

lib/core/src/math/utils.rs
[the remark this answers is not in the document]
notdanilo: Ah, yes! This comment was copied from `Float32Array::view` and I didn't review it after adapting it.

The doc was fixed.

lib/core/src/math/utils.rs
wdanilo: I was refering to the sentence `We need to interpret [[f32; 4]; 4] as [f32; 16] to view as Float32Array. The memory layout is the same, so this operation is safe.` - which is not really true, as the whole operation is marked unsafe. If the whole operation is marked unsafe, we should not have a comment telling "hey, this is safe".

lib/core/src/display/rendering/html/html_renderer.rs
wdanilo: I would copy-paste it. Unsafe blocks are rare and are extremaly dangereous. Its error prone to refaactoring. Lets keep safety notes close to the source.

lib/core/src/display/symbol/buffer.rs
wdanilo: Oh, it was done by intellij? Thats funny :D

lib/core/src/math/utils.rs
wdanilo: This does not solve my issue. The code which you are testing for correctness is this one `&*(matrix.as_ref() as *const [[f32; 4]; 4] as *const [f32; 16])`, right? It is copy-pasted here from its original definition point in `into_float32_array_view`. Thus if I change the definition of `into_float32_array_view` then this test will not test the correctness anymore. If you want to test if an unsafe function works correctly, write tests for `into_float32_array_view` instead, not for copy-pasted partial implementation of it. We see now that these are connected because they are in a single PR, but soon, after merge they will become 2 codes with similar implementations, and after someone refactors one of them, they can not be synchronised anymore.

notdanilo: @wdanilo it was refactored into a `matrix4_to_array` function and I am using the same function in `into_float32_array_view`:

```rust
impl IntoFloat32ArrayView for Matrix4<f32> {
    unsafe fn into_float32_array_view(&self) -> Float32ArrayView<'_> {
        let matrix = matrix4_to_array(&self);
        let array = Float32Array::view(matrix);
        let phantom = PhantomData;
        Float32ArrayView { array, phantom }
    }
}
```

and in our tests

```rust
    #[test]
    fn matrix4_memory_layout() {
        use super::Matrix4;
        use super::matrix4_to_array;

        let matrix = Matrix4::<f32>::new
            ( 1.0, 5.0,  9.0, 13.0
            , 2.0, 6.0, 10.0, 14.0
            , 3.0, 7.0, 11.0, 15.0
            , 4.0, 8.0, 12.0, 16.0 );

        let layout = matrix4_to_array(&matrix);
        for (i, n) in layout.iter().enumerate() {
            assert_eq!((i + 1) as f32, *n);
        }
    }
```

So they aren't copy-pasted.

lib/core/src/math/utils.rs
notdanilo: As described in the comment bellow, this section of code was wrapped into a safe function called `matrix4_to_array`.

lib/core/src/math/utils.rs
wdanilo: This is definitely step in the right direction, but now we've got an unsafe function which is not marked unsafe!  The whole "safety" topic is very, very important for us and we should never accept it this way. Every unsafe function should be marked as unsafe and have safety notes. 

```rust
pub fn matrix4_to_array(matrix:&Matrix4<f32>) -> &[f32; 16] {
    // To interpret [[f32; 4]; 4] as [f32; 16].
    // The memory layout is the same, so this operation is safe.
    unsafe {
        &*(matrix.as_ref() as *const [[f32; 4]; 4] as *const [f32; 16])
    }
}
```

notdanilo: @wdanilo my understanding is that the compiler isn't smart enough to know that `[[f32; 4]; 4]` is the same of `[f32; 16]`. It's just a different way to access this data.

The only way I found to achieve it right now is to use unsafe methods. But why would it be unsafe if the memory layout is the same?

@farmaazon do you have any consideration to make on this topic?

wdanilo: @notdanilo Oh I'm so sorry, you are right. I was somehow sure that this function includes a unsafe view on JS memory and it is still unsafe. All good now!

PR #61, opened 2019-12-02
Rust code style guide

wdanilo:
Rust code style guide

### DIFF
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -221,7 +221,7 @@ impl Printer for GlobalVarStorage {
 - Commas between comples expressions (including arg list) are spaced.
 - Commas between simple elements are not spaced: `Result<Self,Error>`.
 - Arguments to functions are not spaced: `build(builder,"out",qual)`.
-
+- Operators are always spaced: `let foo = a + b * c;`.
 
 ### Function definitions
 The following examples show proper function styles:
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -96,6 +96,7 @@ impl Default for TransformOrder {
     fn default() -> Self { Self::ScaleRotateTranslate }
 }
 
+
 // =============================
 // === HierarchicalTransform ===
 // =============================
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -26,6 +26,141 @@ to be handled manually.
 Each line in a source file should have max of 80 chars of text (including 
 comments).
 
+### Imports 
+Imports should be divided into 4 groups separated by blank lines. Items in the
+groups should be sorted alphabetically.
+```rust
+// Group 1: sub-module definitions.
+// Group 2: prelude-like imports.
+// Group 3: local-crate imports.
+// Group 4: external imports.
+```
+
+For example:
+```rust
+pub mod display_object;
+
+use crate::prelude::*;
+
+use crate::closure;
+use crate::data::opt_vec::OptVec;
+use crate::dirty;
+use crate::system::web::group;
+
+use nalgebra::Matrix4;
+use nalgebra::Vector3;
+```
+
+
+### Sections
+
+Source files should be divided into sections. Section should be placed before 
+each new "concept" defined in a file. By "concept" we normally mean a structure
+with related implementations. In case the related implementations use some 
+helper structs with a very small implementations, these helper structs may be 
+defined in the same section. Moreover, the code in each section should be divided
+into sub-sections, grouping related definitions into groups. At least one section
+should be defined in a file (if there is at least one struct definition as well).
+For example:
+
+```rust
+// =================
+// === AxisOrder ===
+// =================
+
+/// Defines the order in which particular axis coordinates are processed. Used
+/// for example to define the rotation order in `DisplayObject`.
+pub enum AxisOrder { XYZ, XZY, YXZ, YZX, ZXY, ZYX }
+
+impl Default for AxisOrder {
+    fn default() -> Self {Self::XYZ}
+}
+
+
+// =================
+// === Transform ===
+// =================
+
+/// Defines the order in which transformations (scale, rotate, translate) are
+/// applied to a particular object.
+pub enum TransformOrder {
+    ScaleRotateTranslate,
+    ScaleTranslateRotate,
+    RotateScaleTranslate,
+    RotateTranslateScale,
+    TranslateRotateScale,
+    TranslateScaleRotate
+}
+
+impl Default for TransformOrder {
+    fn default() -> Self { Self::ScaleRotateTranslate }
+}
+
+// =============================
+// === HierarchicalTransform ===
+// =============================
+
+pub struct HierarchicalTransform<OnChange> {
+    transform        : Transform,
+    transform_matrix : Matrix4<f32>,
+    origin           : Matrix4<f32>,
+    matrix           : Matrix4<f32>,
+    pub dirty        : dirty::SharedBool<OnChange>,
+    pub logger       : Logger,
+}
+
+impl<OnChange> HierarchicalTransform<OnChange> {
+    pub fn new(logger:Logger, on_change:OnChange) -> Self {
+        let logger_dirty     = logger.sub("dirty");
+        let transform        = default();
+        let transform_matrix = Matrix4::identity();
+        let origin           = Matrix4::identity();
+        let matrix           = Matrix4::identity();
+        let dirty            = dirty::SharedBool::new(logger_dirty,on_change);
+        Self {transform,transform_matrix,origin,matrix,dirty,logger}
+    }
+}
+
+// === Getters ===
+
+impl<OnChange> HierarchicalTransform<OnChange> {
+    pub fn position(&self) -> &Vector3<f32> {
+        &self.transform.position
+    }
+
+    pub fn rotation(&self) -> &Vector3<f32> {
+        &self.transform.rotation
+    }
+
+    ...
+}
+
+// === Setters ===
+
+impl<OnChange:Callback0> HierarchicalTransform<OnChange> {
+    pub fn position_mut(&mut self) -> &mut Vector3<f32> {
+        self.dirty.set();
+        &mut self.transform.position
+    }
+
+    pub fn rotation_mut(&mut self) -> &mut Vector3<f32> {
+        self.dirty.set();
+        &mut self.transform.rotation
+    }
+
+    ...
+}
+```
+
+
+### Vertical spacing
+We use the following amount of vertical spacing:
+- 2 blank lines after imports
+- 2 blank lines before each section.
+- 1 blank line before and after sub-section.
+- 1 blank line before functions / structures / impls.
+- 1 blank line at the end of the file.
+
 
 ### Multiline Expressions
 Most (preferably all) expressions should be single line. Multiline expression is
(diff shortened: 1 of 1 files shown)


docs/style-guide.md
```diff
@@ -0,0 +1,291 @@
...
+
+```rust
+pub fn new() -> Self {
+    let sub_logger  = logger.sub("shape_dirty");
+    let shape_dirty = ShapeDirty::new(sub_logger,on_dirty.clone());
+    let sub_logger  = logger.sub("mesh_registry_dirty");
+    let dirty_flag  = MeshRegistryDirty::new(sub_logger,on_dirty);
+    Self {shape_dirty,dirty_flag}
+}
+```
```
farmaazon: I'd like to put here a case we discussed on call (before Michał and Danilo joined): a chain of processing functions, characteristic for iterators. Compare:
```rust
impl<'a> CharVerticesBuilder<'a> {
    // ........
    pub fn build(&mut self, character : char) -> OneCharVertices {
        // a bunch if lets computing ultimate transformation
        let pair_iter            = GLYPH_VERTICES_BASE_LAYOUT.iter();
        let point_iter           = pair_iter.map(|(x, y)| Point2::new(*x, *y));
        let transformed          = point_iter.map(|p| ultimate_vertex_trans * p);
        let transformed_iterable = transformed.map(|p| point_to_iterable(p));
        transformed_iterable.flatten().collect()
    }
```
with
```rust
impl<'a> CharVerticesBuilder<'a> {
    // ........
    pub fn build(&mut self, character : char) -> OneCharVertices {
        // a bunch if lets computing ultimate transformation
        GLYPH_VERTICES_BASE_LAYOUT.iter()
            .map(|(x, y)| Point2::new(*x, *y))
            .map(|p| ultimate_vertex_trans * p)
            .map(|p| point_to_iterable(p))
            .flatten().collect()
    }
```
For me the second case is much less noisy. It's hard to make a meaningful label here.

And, unfortunately I still think same about optional example we discussed:
```rust
    fn get_kerning_trans(&mut self, ch : char) -> Translation2<f32> {
        let prev_char = self.previous_char;
        let opt_value = prev_char.map(|lc| self.font.get_kerning(lc, ch));
        let value     = opt_value.unwrap_or(0.0);
        Translation2::new(value, 0.0)
    }
```
What is the value of introduced label `opt_value`? I usually see that this is _optional_, from its type
```rust
    fn get_kerning_trans(&mut self, ch : char) -> Translation2<f32> {
        let prev_char = self.previous_char;
        let value     = prev_char.map(|lc| self.font.get_kerning(lc, ch))
                                 .unwrap_or(0.0);
        Translation2::new(value, 0.0)
    }
```
We have the same number of lines, and avoid making zero-value labels. I propose using this heavy-alignment of new function - the processing function should be a really simple call (or it actually deserve a label). 

As a rule, there should be only one processing call per line, and from the call we should easily read what the result is.

wdanilo:
Adam, abbreviations are arguably worse. For me, `vertex_transformation` is as good as `vtx_trans`. In fact, majority of Rust libs use abbreviations, and the whole lang was designed around those (e.g. `impl`, `fn`, etc). Answering the comments:

- Increasing the line length from 80 to 100 chars - after this change we are not longer able to work in 2 column layout nicely, however I'm tending to agree here that in such a verbose lang like rust this may be a good choice. We can do it.

- `Result<Self, Error>` instead of `Result<Self,Error>`, or `fn test(foo: String, bar: Int)` instead o `fn test(foo:String, bar:Int)` - pleas no. It is much faster to parse with eyes where arguments end if you've got them separated more than type args. With proper sytnax highlighting the type operator (`:`) should do its job well even when glued together like this.

- Regarding the examples you provided – sure, naming variables syntactically does not make much sense. But naming them **semantically** makes much more sense actually. For example:

 ```rust
impl<'a> CharVerticesBuilder<'a> {
    // ........
    pub fn build(&mut self, character : char) -> OneCharVertices {
        // a bunch if lets computing ultimate transformation
        let glyphs           = GLYPH_VERTICES_BASE_LAYOUT.iter();
        let points           = pair_iter.map(|(x, y)| Point2::new(*x, *y));
        let char_points      = point_iter.map(|p| ultimate_vertex_trans * p);
        let char_points_iter = char_points.map(|p| point_to_iterable(p));
        char_points_iter.flatten().collect()
    }
```

And the naming is still bad. I don't really know what happens after the step `ultimate_vertex_trans`. What does it do? What is the result of that? I'm asking **semantically**, not **syntactically**. The answer is clearly not "transformed" - as it does not tell semantically what this var is about. The same applies to the second example - long chains of transformations are hard to read and in order to read it more like a book - I prefer each step to be described by var which has semantically descriptive name.

Regarding the second example, I personaly prefer much more
```rust
    fn get_kerning_trans(&mut self, ch : char) -> Translation2<f32> {
        let prev_char = self.previous_char;
        let opt_value = prev_char.map(|lc| self.font.get_kerning(lc, ch));
        let value     = opt_value.unwrap_or(0.0);
        Translation2::new(value, 0.0)
    }
``` 
over

```rust
    fn get_kerning_trans(&mut self, ch : char) -> Translation2<f32> {
        let prev_char = self.previous_char;
        let value     = prev_char.map(|lc| self.font.get_kerning(lc, ch))
                                 .unwrap_or(0.0);
        Translation2::new(value, 0.0)
    }
```

especially when this line is in the middle of other lines and does not break the alignment rules. If we could, please lets stay with the former one.

docs/style-guide.md
farmaazon: So, after discussion with Wojtek we decided to leave _single line expressions_ rule as it is. It's hard to define accurately when the function calls chain are ok and when not, and the many let's in code are just a nuisance I can manage.

docs/style-guide.md
```diff
@@ -0,0 +1,291 @@
...
+            Self::OutStorage (qual) => build!(builder,"out",qual),
+        }
+    }
+}
+```
+
+
+### Spaces 
+- Type operator is not spaced: `fn test(foo:String, bar:Int) { ... }`.
+- Commas between comples expressions (including arg list) are spaced.
```
notdanilo: Example?
&& "Comples" typo.

docs/style-guide.md
```diff
@@ -0,0 +1,291 @@
...
+- Commas between simple elements are not spaced: `Result<Self,Error>`.
+- Arguments to functions are not spaced: `build(builder,"out",qual)`.
```
notdanilo: I personally think spaced args are more readable: `build(builder, "out", qual)`.

docs/style-guide.md
```diff
@@ -0,0 +1,291 @@
+# Rust Style Guide
+
+## Motivation - why not to completely rely a formatting tool?
```
mwu-tow: This question conflates formatting with a style. Style is so much more, as next paragraph answers. 
But it is all pretty obvious. I'd state the question — "why not rely on a formatting tool for formatting". And that second paragraph answers.

docs/style-guide.md
```diff
@@ -0,0 +1,291 @@
...
+
+The docs of rustfmt state that "formatting code is a mostly mechanical task 
+which takes both time and mental effort. By using an automatic formatting tool, 
+a programmer is relieved of this task and can concentrate on more important 
+things.". While in many cases it is true, if the uthor of code does not take
```
mwu-tow: Basically the thought of this paragraph is "style is more then just formatting" but we introduce it as if we were to say "formatters don't deliver what they promise". But the quoted statement actually stands true for this case.

docs/style-guide.md
```diff
@@ -0,0 +1,291 @@
...
+comments).
+
+### Imports 
+Imports should be divided into 4 groups separated by blank lines. Items in the
+groups should be sorted alphabetically.
+```rust
+// Group 1: sub-module definitions.
+// Group 2: prelude-like imports.
+// Group 3: local-crate imports.
+// Group 4: external imports.
```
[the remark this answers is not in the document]
wdanilo: Very interesting idea. @farmaazon @notdanilo thoughts?

docs/style-guide.md
notdanilo: I think it's easier to find groups of `id:type` if there is no space between them. So, for me, it's a preferable style when defining the function parameters in a single line.

wdanilo: Guys, I would stay with the no-spaces rules here tbh. In Rust, we're using `,` to separate things. We use extra spacing when defining functions only to divide `something:Type` groups apart, but when calling a function there is no need for such operation. Using both `,` and a space is separating args like with two spaces. If you use color highlighting then its obvious where the `,` is. 

I recognize this (as the majority of this doc) as not very objective preference, so I'm just asking to stay with the current description here.

docs/style-guide.md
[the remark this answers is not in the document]
wdanilo: @mwu-tow After thinking about it, I'm against this rule. All deps developed by us in the lib folder are meant to be refactored aslibraries in the future. After such refactoring a LOT of files needs to be changed, so this rule is very impractical imo.

wdanilo:
@mwu-tow I cannot somehow answer your comment `I don't like the rule, examples shown by Adam are much easier to read without all that ...`, so I'll do it here:

I understand these rationales, but dividing long chains into well-described lets is a "generic" rule that works. Let me explain here a few points:
- Dividing should always involve semantically named variables. We've been already chatting about the examples, and for example, I don't understand looking at the code what each line does there. I would understand if the steps would be named semantically correct.
- I'm sure that developers with years of experience of writing high-quality code, like you guys, would take care of writing well structured chained operations, but it is a very rare case. And TBH, even in your code sometime the steps are not obvious and require a very strong context-knowledge to understand what this 10-lines long-expression does. 
- If we want to work with the community, we need to take extra care of code transparency. Thus both this rule as well as writing comments on all public interfaces may seem sometimes like overkill, but it allows us to both keep the code widely understandable, as well as guarantees that the code we get from the community is understandable as well.

docs/style-guide.md
```diff
@@ -0,0 +1,291 @@
...
+            Self::InStorage  (qual) => build!(builder,"in" ,qual),
+            Self::OutStorage (qual) => build!(builder,"out",qual),
+        }
+    }
+}
+```
+
+
+### Spaces 
+- Type operator is not spaced: `fn test(foo:String, bar:Int) { ... }`.
```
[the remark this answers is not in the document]
wdanilo: As @notdanilo mentioned above:
> I think it's easier to find groups of id:type if there is no space between them. So, for me, it's a preferable style when defining the function parameters in a single line.

wdanilo:
Guys, I'm merging this style guide. Let's use it strictly from now on. In case we would like to introduce any changes, everyone is welcome to create PR and start a discussion about the change. I am aware that not all the guides are the favorite guides for everyone here, and I'm also aware it is not possible to write a style guide which will be OK for everyone. Many, many changes were applied after your comments, and I'm really thankful for the help in making it into better shape, I highly appreciate it ❤️. I also hope, that we will be happy when using these rules and I deeply hope that no one would be mad about not including some of the requested changes. I think that the correct way to progress now is to start using it and in case we will see real code usage which will drastically benefit from another style, let's definitely show this code and let's discuss it in a separate thread. I believe that this guide is missing a lot in order to be a "perfect guide". On the other hand, making a "perfect guide" may not be possible, as including more and more corner cases makes keeping the rules in our heads more and more impossible. Anyway, I strongly encourage to improve it in the future, and maybe find a way to further simplify the rules.

PR #67, opened 2019-12-05
Implemented newlines

farmaazon:
### Pull Request Description
TextComponent will properly display text containing newline characters.

### Important Notes
* Code of the TextComponent was somewhat prepared for optimal clipping text and scrolling; each line of text have an own fixed-size buffer fragment, so the lines should be easily extended and fragments reused.
* The first commit (98643721a94d93ef9ec67613ed6658437b04713f) is only moving and renaming glyph_square.rs file. For best review experience you may skip this commit in diff 

### Checklist


lib/core/src/lib.rs
```diff
@@ -131,21 +128,23 @@ mod example_03 {
...
 
-            let all_cases     = iproduct!(0..fonts.len(), SIZES.iter());
+            let all_cases     = iproduct!(0..fonts.len(), 0..SIZES.len());
 
-            for (i, (font, size)) in all_cases.enumerate() {
+            for (font, size) in all_cases {
 
-                let line_position = nalgebra::Vector2::new(-0.95, 0.9 - 0.064*(i as f64));
+                let x = -0.95 + 0.6 * (size as f64);
+                let y = 0.90 - 0.45 * (font as f64);
```
wdanilo: I did not notice it earlier - why do we use such hardcoded literals here? If we have to, please refactor them to vars.

lib/core/src/text.rs
```diff
@@ -1,65 +1,150 @@
...
+    /// Render text
+    pub fn display(&self) {
+        let gl_context = &self.gl_context;
+
+        gl_context.use_program(Some(&self.gl_program));
+        self.bind_buffer_to_attribute("position",&self.buffers.vertex_position);
+        self.bind_buffer_to_attribute("texCoord",&self.buffers.texture_coordinates);
+        self.setup_blending();
+        gl_context.bind_texture(Context::TEXTURE_2D, Some(&self.gl_msdf_texture));
+        gl_context.draw_arrays(WebGlRenderingContext::TRIANGLES,0,self.buffers.vertices_count() as i32);
```
wdanilo: Over 100 chars

lib/core/src/text.rs
```diff
@@ -1,65 +1,150 @@
...
+        let lines_text = self.text.as_ref().split('\n');
+        let lines_iter = lines_text.map(Self::initialize_line);
+        lines_iter.collect()
+    }
+
+    fn initialize_line(text:&str) -> Line {
+        Line {
+            content : text.to_string(),
+        }
+    }
```
wdanilo: Very minor thing, but I think we should add it to the style guide:
```rust
    fn initialize_line(text:&str) -> Line {
        let content = text.to_string();
        Line {content}
    }
```

lib/core/src/text.rs
```diff
@@ -158,14 +194,14 @@ impl<'a> TextComponentBuilder<'a> {
...
-            , tex_type
-            , data
+                , tex_level
+                , internal_fmt
+                , width
+                , height
+                , border
+                , format
+                , tex_type
+                , data
```
wdanilo: Again minor thing but I assume this change (and the one above) was accidental

lib/core/src/text/buffer.rs
```diff
@@ -0,0 +1,153 @@
+pub mod glyph_square;
+pub mod line;
+
+use crate::prelude::*;
+
+use crate::text::buffer::glyph_square::{GlyphVertexPositionBuilder, GlyphTextureCoordinatesBuilder, GlyphAttributeBuilder, BASE_LAYOUT_SIZE};
```
wdanilo: Over 100 chars

lib/core/src/text/buffer.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    fn build_fragments(displayed_lines:usize) -> Vec<BufferFragment> {
+        let indexes   = 0..displayed_lines;
+        let fragments = indexes.map(|_| BufferFragment { assigned_line : None });
+        fragments.collect()
+    }
+
+    /// Refresh the whole buffers making them display the given lines.
+    pub fn refresh_all<'a>(&mut self, gl_context:&Context, lines:&[Line], font:&'a mut FontRenderInfo, to_window:&Transform2<f64>) {
```
wdanilo: Way over 100 chars. We need to automate checking it 😂

lib/core/src/text/buffer.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    fn refresh_vertex_position_buffer<'a, Iter>
+    ( &self
+    , gl_context:&Context
+    , content_with_index:Iter
+    , font:&mut FontRenderInfo
+    , to_window:&Transform2<f64>
+    ) where Iter : ExactSizeIterator<Item=(&'a str,usize)>
+    {
```
wdanilo: Just styling note but imo important - lets always use open curly bracese in previous line please!

lib/core/src/text/buffer/line.rs
```diff
@@ -0,0 +1,156 @@
...
+    max_line_size    : usize,
+    squares_produced : usize,
+    glyph_builder    : GlyphBuilder,
+    chars_iterator   : Option<std::str::Chars<'a>>,
+}
+
+pub type
+LineVerticesBuilder<'a,'b>           = LineAttributeBuilder<'a,GlyphVertexPositionBuilder<'b>>;
+pub type
+LineTextureCoordinatesBuilder<'a,'b> = LineAttributeBuilder<'a,GlyphTextureCoordinatesBuilder<'b>>;
```
wdanilo: If we cannot change the name to be shorter (andplease do if we can somehow), I would definitely write it this way:
```rust
pub type LineVerticesBuilder<'a,'b> = 
    LineAttributeBuilder<'a,GlyphVertexPositionBuilder<'b>>;
pub type LineTextureCoordinatesBuilder<'a,'b> = 
    LineAttributeBuilder<'a,GlyphTextureCoordinatesBuilder<'b>>;
```

lib/core/src/lib.rs
farmaazon: Well, this is code for an example, and it is well-written anyway. Soon I think I remove this example and just create a test with html like Danilo did

lib/core/src/text.rs
farmaazon: Yeah, that's Intellij work. If you know how to tell him "be less eager in autoformatting", give me a clue.

lib/core/src/text.rs
farmaazon: Well, this is the one I don't agree with. I don't see the point of creating let here, and in more complex case it is nice to see clearly what is assigned to a field (without quicksearchng for label name).

I see the point in doing this for function arguments, where reader does not see what the arguments are without jumping to definition. But here the field names are mentioned, so this is not the case.

In other words, making labels for fields arguments is like doing this:
```rust
fn abs(x:usize) -> usize {
     let result = if x >= 0 {x} else {-x};
    result
}
```

farmaazon: I'd like to make this rule: don't male labels for structure fields if it is constructed as the return of the function (if you construct structure in the middle of function body you probably need to split this function anyway)

lib/core/src/text/buffer/line.rs
farmaazon: Usually I'm against shorting names just because they don't fit line. And I'm usually against abbreviations. But then I discover that `GlyphTextureCoordsBuilder` is better than `GlyphTextureCoordinatesBuilder` :D 

So, each rule has exceptions :)

PR #88, opened 2019-12-17
Text editing

farmaazon:
### Pull Request Description
Added operation for editing text in text component, which does the minimal required buffer refresh.

### Checklist

### DIFF
--- a/lib/core/src/text.rs
+++ b/lib/core/src/text.rs
@@ -38,14 +38,15 @@ use web_sys::WebGlTexture;
 /// commits
 #[derive(Debug)]
 pub struct TextComponent {
-    pub content     : TextComponentContent,
-    pub position    : Point2<f64>,
-    pub size        : Vector2<f64>,
-    pub text_size   : f64,
-    gl_context      : WebGl2RenderingContext,
-    gl_program      : Program,
-    gl_msdf_texture : WebGlTexture,
-    buffers         : TextComponentBuffers,
+    pub content       : TextComponentContent,
+    pub position      : Point2<f64>,
+    pub size          : Vector2<f64>,
+    pub text_size     : f64,
+    gl_context        : WebGl2RenderingContext,
+    gl_program        : Program,
+    gl_msdf_texture   : WebGlTexture,
+    msdf_texture_rows : usize,
+    buffers           : TextComponentBuffers,
 }
 
 impl TextComponent {
@@ -75,12 +76,17 @@ impl TextComponent {
 
     /// Render text
     pub fn display(&mut self, fonts:&mut Fonts) {
+        self.buffers.refresh(&self.gl_context,self.content.refresh_info(fonts));
+
+        if self.msdf_texture_rows != fonts.get_render_info(self.content.font).msdf_texture.rows() {
+            self.update_msdf_texture(fonts);
+        }
+
         let gl_context     = &self.gl_context;
         let vertices_count = self.buffers.vertices_count() as i32;
 
-        self.buffers.refresh(gl_context,self.content.refresh_info(fonts));
         gl_context.use_program(Some(&self.gl_program));
-        self.update_uniforms();
+        self.update_uniforms(fonts);
         self.bind_buffer_to_attribute("position",&self.buffers.vertex_position);
         self.bind_buffer_to_attribute("tex_coord",&self.buffers.texture_coords);
         self.setup_blending();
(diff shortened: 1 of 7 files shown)


lib/core/src/text.rs
```diff
@@ -75,29 +76,66 @@ impl TextComponent {
...
-    fn update_uniforms(&self) {
+    fn update_uniforms(&self, fonts:&mut Fonts) {
         let gl_context          = &self.gl_context;
         let to_scene            = self.to_scene_matrix();
         let to_scene_ref        = to_scene.as_ref();
+        let msdf_width          = MsdfTexture::WIDTH as f32;
+        let msdf_height         = fonts.get_render_info(self.content.font).msdf_texture.rows() as f32;
         let to_scene_loc        = gl_context.get_uniform_location(&self.gl_program,"to_scene");
+        let msdf_size_loc       = gl_context.get_uniform_location(&self.gl_program,"msdf_size");
         let transpose           = false;
```
wdanilo: A very minor thing but the spacing here is too big.

lib/core/src/text.rs
```diff
@@ -75,29 +76,66 @@ impl TextComponent {
...
+            ( target
+            , tex_level
+            , internal_fmt
+            , width
+            , height
+            , border
+            , format
+            , tex_type
+            , data
+            );
```
wdanilo: I'm just wondering if we really want such code to occupy 11 lines. I mean, is it more readable this way?

lib/core/src/text/buffer.rs
```diff
@@ -144,27 +141,27 @@ impl TextComponentBuffers {
...
-        let displayed_lines      = self.displayed_lines(content.lines.len());
+    fn setup_buffers(&mut self, gl_context:&Context, refresh:RefreshInfo) {
+        let displayed_lines      = self.displayed_lines(refresh.lines.len());
         let all_fragments        = 0..self.fragments.fragments.len();
-        let mut builder          = self.create_fragments_data_builder(content.font);
+        let mut builder          = self.create_fragments_data_builder(refresh.font);
 
         self.fragments.reassign_fragments(displayed_lines);
-        self.fragments.build_buffer_data_for_fragments(all_fragments,&mut builder,content.lines);
+        self.fragments.build_buffer_data_for_fragments(all_fragments,&mut builder,refresh.lines.as_ref());
```
wdanilo: Ok, so we need to add to CI checking the line length :P

lib/core/src/text/buffer/fragment.rs
```diff
@@ -292,6 +296,14 @@ impl BufferFragments {
         }
     }
 
+    /// Mark as dirty all fragments with dirty assigned line.
```
wdanilo: TBH, I completely do not understand this doc comment :P

lib/core/src/text/buffer/glyph_square.rs
```diff
@@ -225,11 +227,12 @@ impl<'a> GlyphAttributeBuilder for GlyphTextureCoordsBuilder<'a> {
 
     /// Compute texture coordinates for `ch`
```
wdanilo: This doc comment is not a sentence + it does not describe all the things it should describe. By inspecting the code I understand that it also takes into consideration previous glphs to get proper spacing right?

lib/core/src/text.rs
```diff
@@ -175,22 +213,15 @@ impl<'a,'b,Str:AsRef<str>> TextComponentBuilder<'a,'b,Str> {
         let gl_context        = self.workspace.context.clone();
         let gl_program        = self.create_program(&gl_context);
         let gl_msdf_texture   = self.create_msdf_texture(&gl_context);
-        let lines             = self.split_lines();
-        let font              = self.fonts.get_render_info(self.font_id);
         let display_size      = self.size / self.text_size;
-        let content_ref       = ContentRef{lines:lines.as_ref(),font};
-        let buffers           = TextComponentBuffers::new(&gl_context,display_size,content_ref);
+        let mut content       = TextComponentContent::new(self.font_id,self.text.as_ref());
+        let buffers           = TextComponentBuffers::new(&gl_context,display_size,content.refresh_info(self.fonts));
```
wdanilo: I don't like the name of the variable. What's about `i_am_a_way_too_long_line` instead? :D

lib/core/src/text/content.rs
```diff
@@ -0,0 +1,370 @@
...
+        TextChange {replaced,
+            lines : TextComponentContent::split_to_lines(text)
+        }
+    }
+
+    /// A type of this change. See `ChangeType` doc for details.
+    pub fn change_type(&self) -> ChangeType {
+        if self.lines.is_empty() {
+            panic!("Invalid change");
+        } else if self.replaced.start.line == self.replaced.end.line && self.lines.len() == 1 {
```
wdanilo: Its hard to undersstand if the condition is correct and why its working the way its working. I would consider refactoring it somehow.

lib/core/src/text/content.rs
```diff
@@ -0,0 +1,370 @@
...
+        }
+    }
+
+    fn split_to_lines(text:&str) -> Vec<String> {
+        let split      = text.split('\n');
+        let without_cr = split.map(Self::cut_cr);
+        without_cr.map(|s| s.to_string()).collect()
+    }
+
+    fn cut_cr(from:&str) -> &str {
```
wdanilo: Either doc or better name :D

lib/core/src/text/content.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    /// Apply change to content.
+    pub fn make_change(&mut self, change:TextChange) {
+        match change.change_type() {
+            ChangeType::Simple    => self.make_simple_change(change),
+            ChangeType::Multiline => self.make_multiline_change(change),
+        }
+    }
+    
+    fn make_simple_change(&mut self, change:TextChange) {
```
wdanilo: what is a "simple" change? Should be documented or named better.

lib/core/src/text/content.rs
```diff
@@ -0,0 +1,370 @@
...
+            self.dirty_lines.add_lines_range(start_line..=end_line);
+        }
+    }
+    
+    fn mix_content_into_change(&mut self, change:&mut TextChange) {
+        self.mix_first_edited_line_into_change(change);
+        self.mix_last_edited_line_into_change(change);
+    }
+
+    fn mix_first_edited_line_into_change(&self, change:&mut TextChange) {
```
wdanilo: Lacks doucmentation. I know its private, still I dont understand what it really does by the name (what does "mix" mean?)

lib/core/src/text.rs
farmaazon: For me It's better to be one-two lines, the current shape came from https://github.com/luna/basegl/pull/62#discussion_r353527495

lib/core/src/text/buffer/glyph_square.rs
farmaazon: No, not this function. It only compute texture coordinates, and that depends on single char only

lib/core/src/text/buffer/fragment.rs
farmaazon: This is a method of BufferFragments. Fragments may be dirty (that's comes from their definition described in `BufferFragment` doc). Lines also may be dirty. Fragments may have line assigned. 

I assume that the reader will have knowledge of above facts. And I hope I don't have to explain what `dirty` means?

lib/core/src/text.rs
farmaazon: Ok, I see, that in times of that comment we had 80 chars limit.

lib/core/src/text/content.rs
farmaazon: Right, it's not obvious even in context of its usage. I'll add a doc comment.

PR #91, opened 2019-12-20
60 steps per second Physics Simulator (Kinematics, Spring, Air Dragging)

notdanilo:
### Pull Request Description
This PR delivers a fixed time step Physics Simulator with kinematics, spring and air dragging physics.

### Important Notes
This PR deliver two important elements:
* Animator, which targets a fixed time step execution.
* PhysicsSimulator and their properties.

The simulator steps per second default is 60 steps per second. And it isn't configurable yet due the nature of the currently used formulas, which doesn't take time into consideration, varying results when we run at different time step rates.

### Checklist


lib/core/src/animation/animator.rs
```diff
@@ -0,0 +1,51 @@
...
+
+struct AnimatorData {
+    closure          : Box<dyn FnMut(f32)>,
+    previous_time    : Option<f32>,
+    step_duration    : f32,
+    accumulated_time : f32
+}
+
+impl AnimatorData {
+    pub fn new<F:FnMut(f32) + 'static>(steps_per_second:f32, f:F) -> Self {
```
wdanilo: Let's not keep `'static` naked in types. There is a risk it populates to different functions and poisons a lot of code. Sometimes after refactoring it would not be needed anymore, but no one would notice that. Please create a type alias for such complex types always.

lib/core/src/animation/animator.rs
```diff
@@ -0,0 +1,51 @@
...
+        let closure          = Box::new(f);
+        let previous_time    = None;
+        let step_duration    = 1.0 / steps_per_second;
+        let accumulated_time = zero();
+        Self { closure, previous_time, step_duration, accumulated_time }
```
wdanilo: Very, very minor one - spacing. When evaluating a function, there are no spaces after commas.

lib/core/src/animation/animator.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+// ================
+// === Animator ===
+// ================
+
+/// This structure is aimed to run a closure at a fixed time rate.
+pub struct Animator {
+    _animation_loop: AnimationFrameLoop
```
wdanilo: Why the variable starts with `_` ?

lib/core/src/animation/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod physics;
+mod animator;
+
+pub use animator::*;
```
wdanilo: Please add enters to end of files :)

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,248 @@
...
+
+// ======================
+// === DragProperties ===
+// ======================
+
+/// This structure contains air dragging properties.
+#[derive(Default, Clone)]
+pub struct DragProperties {
+    amount : Rc<RefCell<f32>>
+}
```
wdanilo: Why `DragProperties` is defined as `RefCell`? Why don't we just keep all properites as plain data. I havent found in code any place you need to modify it from different places at once. The same applies to SpringProperties.

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,248 @@
...
+
+// ========================
+// === PhysicsSimulator ===
+// ========================
+
+/// A 60 steps per second physics simulator used to simulate `Properties`.
+pub struct PhysicsSimulator {
+    _animator : Animator
+}
```
wdanilo: I don't really understand why `PhysicsSumulator` contains only `Animator`. It should also own all the physics properties, right? I mean, if it only contains `Animator`, then it is more like `Animator` then the whole "physics engine"

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Simulates `Properties` on `object`.
+    pub fn new<T>(mut object:T, properties:Properties) -> Self
+    where T: HasPosition + 'static {
+        let steps_per_second = 60.0;
+        let _animator = Animator::new(steps_per_second, move |delta_time| {
+            properties.kinematics().set_position(object.position());
+            simulate_spring(&mut properties.kinematics(), &properties.spring());
+            simulate_dragging(&mut properties.kinematics(), &properties.drag());
+            simulate_kinematics(&mut properties.kinematics(), delta_time);
+            object.set_position(properties.kinematics().position());
```
wdanilo: Hmm, everything here seems really nicely decoupled (that is super cool)m, but I reall dont understand why you keep so many objects behind `RefCell`. If you could design that as keeping the ownership, this is **always** better way to do. I believe that spring properties and drag properties could be part of physics engine - then it will own that and ou will not use RefCells at all. Basically lets agree that using RefCells is bad and we should do it only if we need and ponly if we want to get nicer API. I believe that wihtout RefCells this user-facing API will be the same, am I right?

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,248 @@
...
+fn simulate_kinematics(properties:&mut KinematicProperties, dt:f32) {
+    properties.set_velocity(properties.velocity() + properties.acceleration() * dt);
+    properties.set_position(properties.position() + properties.velocity()     * dt);
+}
+
+/// Simulate dragging on `KinematicProperties`.
+fn simulate_dragging(kinematics:&mut KinematicProperties, drag:&DragProperties) {
+    let velocity = kinematics.velocity();
+    let speed    = velocity.norm();
+    kinematics.set_velocity(velocity / (1.0 + drag.amount() * speed));
```
wdanilo: I believe that this equation for velocity change is wrong. Where are you using mass property of the object? Moreover, the whole equation is not physicaslly correct.

lib/core/src/traits.rs
```diff
@@ -0,0 +1,8 @@
+use nalgebra::Vector3;
```
wdanilo: The name of this file is very bad. It does not tell semantically what is here. I would never think that it contains `Position` getter and setter. Moreover, this file will become a big garbage soon if we allow for it becuase everyone will put there things.

lib/core/tests/physics_simulator.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        let aspect_ratio = view_dim.x / view_dim.y;
+        let mut camera = Camera::perspective(45.0, aspect_ratio, 1.0, 1000.0);
+        camera.set_position(Vector3::new(0.0, 0.0, 29.0));
+
+        let kinematics     = default();
+        let mass           = 1.0;
+        let coefficient    = 10.0;
+        let fixed_point    = zero();
+        let spring         = SpringProperties::new(mass, coefficient, fixed_point);
```
wdanilo: mass is not part of spring properties. It is a property of the object. It should be inside of physics simulator or bundled with the object, but definitely this is not mass of a spring.

lib/core/tests/physics_simulator.rs
```diff
@@ -0,0 +1,74 @@
...
+            let x = 32.0 * (random() - 0.5) as f32;
+            let y = 24.0 * (random() - 0.5) as f32;
+            let z = 0.0;
+            let position = Vector3::new(x, y, z);
+            properties.spring().set_fixed_point(position);
+            target.set_position(position);
+        });
+
+        std::mem::forget(simulator);
+        std::mem::forget(animator);
```
wdanilo: Please dont use it even in tests. It is too error prone.

lib/core/tests/physics_simulator.rs
```diff
@@ -0,0 +1,74 @@
...
+        let every = 2.0;
+        let animator  = Animator::new(1.0 / every, move |_| {
+            let x = 32.0 * (random() - 0.5) as f32;
+            let y = 24.0 * (random() - 0.5) as f32;
+            let z = 0.0;
+            let position = Vector3::new(x, y, z);
+            properties.spring().set_fixed_point(position);
+            target.set_position(position);
+        });
+
```
wdanilo: How the API would be used with a camera? Could you show the code / pseudo code of that?

lib/core/src/animation/animator.rs
notdanilo: Created this:
```rust
pub trait FnAnimation = FnMut(f32) + 'static;
```

lib/core/src/animation/animator.rs
notdanilo: Fixed

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub fn new(position:Vector3<f32>, velocity:Vector3<f32>, acceleration:Vector3<f32>) -> Self {
+        let data = Rc::new(RefCell::new(KinematicData { position, velocity, acceleration }));
+        Self { data }
+    }
+
+    /// Gets velocity.
+    pub fn velocity(&self) -> Vector3<f32> {
+        self.data.borrow().velocity
+    }
+
```
wdanilo: We agreed some time ago that we dont need to put docs on getters and setters cause it is just stupid and doesnt bring any more information and jsut takes your time. Just fyi. However, please then refactor getters and setters so:

```rust
// === Getters ===

impl KinematicProperties {
    ...
}


// === Setters ===

impl KinematicProperties {
    ...
}
```

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,248 @@
...
+        let data = Rc::new(RefCell::new(PropertiesData::new(kinematics, spring, drag)));
+        Self { data }
+    }
+
+    pub fn kinematics(&self) -> KinematicProperties { self.data.borrow().kinematics.clone() }
+
+    pub fn set_kinematics(&mut self, kinematics:KinematicProperties) {
+        self.data.borrow_mut().kinematics = kinematics
+    }
+    pub fn spring(&self) -> SpringProperties { self.data.borrow().spring.clone() }
```
wdanilo: Let's not use single line functions until you crreate al ot of similar functions and you want to align them to show the similarity. It is arguably hard to read them. I think we got it somewhere in runtime docs.

lib/core/src/traits.rs
notdanilo: @wdanilo I agree. Any suggestion where I should move `HasPosition` to?

lib/core/src/animation/physics.rs
notdanilo: @wdanilo I am lost now. We had a discussion about the new equation I was using which was taking time into consideration and you asked me to use the older one because you said it was physically correct. This is the older one. What should I do?

lib/core/src/animation/physics.rs
notdanilo: Yeah, the bunch of RefCells were refactored out of the code as we discussed on Discord. The API is nicer too.

lib/core/src/animation/physics.rs
notdanilo: Haha the physics properties are moved to `Animator`'s closure. There is no need to keep a reference of it inside `PhysicsSimulator` too.

lib/core/tests/physics_simulator.rs
notdanilo: I need to keep them alive, because I don't want the closures inside of them to die.

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,249 @@
...
+        self.mass = mass
+    }
+}
+
+impl HasPosition for KinematicsProperties {
+    fn position    (&self) -> Vector3<f32>            { self.position }
+    fn set_position(&mut self, position:Vector3<f32>) { self.position = position }
+}
+
+// =============================
```
wdanilo: This is very very minor, but please remember about it - before these big headers we put 3 empty lines

lib/core/tests/physics_simulator.rs
wdanilo: You need to keep them alive till when? When they are destroyed, what really happens if you dont do it?

notdanilo: @wdanilo I need to keep them alive while the bencher.iter's closure is alive.

lib/core/src/animation/physics.rs
notdanilo: https://github.com/luna/basegl/blob/master/docs/style-guide.md#vertical-spacing defines 2 empty lines between sections.

lib/core/src/animation/physics.rs
wdanilo: I understand that now the problem is solved and we know how to code it? :)

lib/core/src/traits.rs
wdanilo: @notdanilo as agreed, leave it somewhere and Ill fix its placement when refactoring files

lib/core/tests/physics_simulator.rs
wdanilo: So why don't you just refer them in the iter body instead?

lib/core/src/animation/physics.rs
wdanilo: Thanks for cathcing it. I need to fix the style guide :(

lib/core/src/animation/physics.rs
notdanilo: @wdanilo yes, we came to an agreement! :+1:

notdanilo:
@wdanilo a consideration about my last commit:

We now have three different kind of animators which can be useful in many different situations.

`FixedStepAnimator` was renamed from `Animator`. It runs a `FnAnimation` closure at fixed delta time.
`DifferentialTimeAnimator` runs a `FnAnimation` closure at every frame, which can result variable delta times.
`ContinuousTimeAnimator` runs a `FnAnimation` closure at every frame, but its input is a continuous time. It can be used for playback and we can set its playback time.

`FixedStepAnimator` is implemented with `DifferentialTimeAnimator` which is implemented with `ContinuousTimeAnimator`.

But I am making direct use of `FixedStepAnimator` and `ContinuousTimeAnimator` to implement fixed step simulation with interpolated values.

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,308 @@
...
+
+/// Runs a simulation step.
+fn simulate(properties:&mut PhysicsProperties, delta_time:f32) -> Vector3<f32> {
+    let spring        = properties.spring();
+    let drag          = properties.drag();
+    let mut net_force = zero();
+    properties.mod_kinematics(|mut kinematics| {
+        net_force += spring.force(&kinematics);
+        net_force += drag.force(&kinematics);
+        simulate_kinematics(&mut kinematics, &net_force, delta_time);
```
notdanilo: @wdanilo see this section. This code could be far more extensible and it could be refactored to

```rust
struct Properties {
    forces : Vec<Box<dyn PhysicsForce>>,
    kinematics : KinematicsProperties
}

for f in properties.forces {
    net_force += f.force();
}
```

But I can't do it because only `PhysicsProperties` implements the shared mutability pattern we wanted to get rid of to avoid RefCell borrowing overhead.

If each property implemented the same design pattern, I would be able to remove `PhysicsKinematics` parameter from

```rust
fn force(&self, kinematics:&PhysicsKinematics)
```

And share it in `SpringProperties`, because it needs to know the body position.

And I would also be able to mutate `SpringProperties` without getting its reference from

```rust
struct PhysicsProperties {
    pub fn mod_spring<F:FnOnce(&mut SpringProperties)>(&mut self, f:F) {
        f(&mut self.data.borrow_mut().spring)
    }
}
```

In fact, this interface would be deprecated.

I didn't want to make this move without consulting for better time management. I am moving back to the `Camera Animation System` PR now.

lib/core/src/animation/fixed_step_animator.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+
+
+// ================
+// === Animator ===
+// ================
+
+/// This structure is aimed to run a closure at a fixed time rate.
```
wdanilo: But this sentence is not true. Frames can take longer than 1/step_duration. So you cannot guarantee that this structure runs a closure at fixed time rate. This sentence is misleading.

lib/core/src/animation/fixed_step_animator.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl FixedStepAnimator {
+    pub fn new<F:FnAnimation>(steps_per_second:f32, f:F) -> Self {
+        let mut data               = FixedStepAnimatorData::new(steps_per_second, f);
+        let _differential_animator = DifferentialTimeAnimator::new(move |delta_time| {
+            data.accumulated_time += delta_time;
+            while data.accumulated_time > data.step_duration {
+                data.accumulated_time -= data.step_duration;
+                (data.closure)(data.step_duration);
```
wdanilo: I still think this is bad that you pass to the inner closure always `1 / step_duration`. It would be much more sensible to pass there a **real time that passed between frames**. Now you are repeating information. In order to create `FixedStepAnimator` you need to provide `steps_per_second` and a function. Why then the first argument to this function is always `1 / steps_per_second`? It is redundant information and you should never pass reduntant information around.

lib/core/src/animation/differential_time_animator.rs
```diff
@@ -0,0 +1,47 @@
...
+
+
+
+// ================================
+// === DifferentialTimeAnimator ===
+// ================================
+
+/// This structure runs an animation every frame with the time difference from the last frame as
+/// its input.
+pub struct DifferentialTimeAnimator {
```
wdanilo: Are you sure you wanted to use the "differential" word here? You are not computing any differentials there, as far as I see.
https://en.wikipedia.org/wiki/Differential_(mathematics)
https://en.wikipedia.org/wiki/Differential_of_a_function

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,308 @@
...
+        Self { coefficient: amount }
+    }
+}
+
+impl PhysicsForce for DragProperties {
+    fn force(&self, kinematics:&KinematicsProperties) -> Vector3<f32> {
+        let velocity  = kinematics.velocity;
+        let speed     = velocity.norm();
+        if speed > 0.0 {
+            velocity.normalize() * speed.powf(2.0) * -0.5 * self.coefficient
```
wdanilo: What is this `* -0.5` ?  I believer it should not be in this equation. Moreover, I may be wrong, but `speed.powf(2.0)` is not the fastest way of computing a power of 2.

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,308 @@
...
+        } else {
+            Vector3::new(0.0, 0.0, 0.0)
+        }
```
wdanilo: Why not default ?

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,308 @@
...
+            let transition       = time / step_time / 1000.0;
+            object.set_position(linear_interpolation(current_position, next_position, transition));
+        });
+
+        let _fixed_step_animator   = FixedStepAnimator::new(steps_per_second, move |_| {
+            continuous_animator.set_time(0.0);
+            let mut next_position          = next_position.borrow_mut();
+            *current_position.borrow_mut() = *next_position;
+            *next_position                 = simulate(&mut properties, step_time);
+        });
```
wdanilo: I havent seen it before, but why are we using so many naked RefCells here and why do we need to use both `FixedStepAnimator` and `ContinousTimeAnimator` in Physics simulator?

lib/core/src/animation/physics.rs
wdanilo: https://github.com/rust-lang/rust/issues/47234

notdanilo: @wdanilo the simplified formula is `F = 1/2 * coefficient * velocity ^ 2`. This is the intensity of the drag force. But the actual drag force is opposite to the velocity, thus I am negating it.

I will replace `speed.powf(2.0)` by `speed * speed`.

lib/core/src/animation/physics.rs
notdanilo: Replaced by `zero()`.

lib/core/src/animation/fixed_step_animator.rs
notdanilo: Welp, you can aim a target, but it can miss it :D

```rust
/// This structure attempts to run a closure at a fixed time rate.
```

Sounds better?

lib/core/src/animation/fixed_step_animator.rs
notdanilo: Keeping the input of the closure as `step_duration` even if it's taking more than `step_duration` is an useful assumption. But let me also discuss about the design:

If `step_duration = 1s` and the frame actually takes `2.5s`, it will call the closure twice with `step_duration = 1s` and leave the remainder `0.5s` to be accumulated.

We could run it a third time with `0.5s`, but the remainder value would be variable and it isn't useful for fixed step physics.

Analyzing the above situation further, we can see that this could result in 4 iterations in `3s`, instead of 3 iterations in `3s`, if the first frame takes `2.5s` (two `1s` calls and one `0.5s` call) and the second one takes `0.5s` (just one `0.5s` call).

lib/core/src/animation/differential_time_animator.rs
notdanilo: We agreed renaming `DifferentialTimeAnimator` to `Animator`.

lib/core/src/animation/physics.rs
notdanilo: @wdanilo

1. `FixedStepAnimator` runs the simulation step.
2. `ContinuousTimeAnimator` runs the simulator every frame, interpolating the current simulation step and next simulation step results.

This can be refactored to only one `ContinuousTimeAnimator`, but it requires the fixed step execution logic to be reimplemented inside `PhysicsSimulator`.

lib/core/src/animation/fixed_step_animator.rs
wdanilo: Much better. But the docs should explain more. What happens if it fails - that it runs the closure couple of times etc. I know it, you know it, but the reader of the docs should know it as well :D

lib/core/src/animation/physics.rs
wdanilo: We need 1 loop and no re-implementation of logic. I'm 99% sure it would be possible with the right design.

lib/core/src/animation/fixed_step_animator.rs
wdanilo: If step durration is `1s` and frame took `2.5s` you should finally run 3 frames and use the interpolation to display the correct value between `2s` and `3s`. 

I completely dont feel this point: `Keeping the input of the closure as step_duration even if it's taking more than step_duration is an useful assumption.` - why? You are providing this information when creating the `FixedStepAnimator`. Why you need to provide to the closure the same information that you provided when defining the closure?

lib/core/src/animation/fixed_step_animator.rs
notdanilo: @wdanilo this?
```rust
/// This structure attempts to run a closure at a fixed time rate.
///
/// # Internals
/// If, for instance, we want to run FnAnimation once per second, it's step duration
/// (FnAnimation(delta_time)) will be 1 second. But keep in mind that if the actual frame takes
/// longer, say 2 seconds, FnAnimation will be called twice in the same moment, but its delta_time
/// parameter will always be fixed to 1 second.
```

lib/core/src/animation/continuous_time_animator.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+
+// === Setters ===
+
+impl ContinuousTimeAnimator {
+    pub fn set_time(&mut self, time:f32) {
+        // TODO: get_performance is slow. Replace it with better approach.
+        let current_time = get_performance().expect("Couldn't get performance").now() as f32;
```
wdanilo: Ok, thinking more about it, using explicit `get_performance` is bad idea. The js function `requestAnimationFrame` provides its callback with a well compted time, so we should use it instead. Its important, lets fix it: https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

lib/core/src/animation/physics.rs
```diff
@@ -0,0 +1,320 @@
...
+pub struct SpringProperties {
+    coefficient : f32,
+    fixed_point : Vector3<f32>
+}
+
+impl Default for SpringProperties {
+    fn default() -> Self {
+        Self::new(zero(),zero())
+    }
+}
```
wdanilo: Why not deriving it? Here and other places

notdanilo: It has been said a few times, Vector3 doesn't implement default.

PR #93, opened 2019-12-23
Cursors in text

farmaazon:
### Pull Request Description
This implements displaying of cursors in text component.
But in fact, the main part of this PR is refactoring of TextComponentContent structure, so the position of cursors may be quick and easily obtained. As a side effect, the horizontal scrolling is faster, especially in the middle of very long lines.

### Checklist


lib/backend/webgl/src/lib.rs
```diff
@@ -109,3 +111,45 @@ fn handle_error<T: CompilationTarget>(ctx:&Context, target:ErrorTarget, t:T) ->
...
+    }
+}
+
+/* Note [unsafe buffer_data]
+ *
+ * The Float32Array::view is safe as long there are no allocations done
+ * until it is destroyed. This way of creating buffers were taken from
+ * wasm-bindgen examples
+ * (https://rustwasm.github.io/wasm-bindgen/examples/webgl.html)
+ */
```
wdanilo: I would propose to always keep safety doc on the function which is a wrapper for some unsafe functionality. Jumping to a definition by labels makes the code harder to navigate. Do you agree?

wdanilo: Btw, you can reuse here our existing functionality. Please see `buffer/item.rs` - it implements the same unsafe buffer view for a variety of types.

lib/core/src/lib.rs
```diff
@@ -233,77 +234,96 @@ mod example_01 {
...
+            let change = TextChange::insert(cursor.position, string.as_str());
+            editor.content.make_change(change);
+            let new_cursor_position = CharPosition {
+                line: editor.content.lines.len()-1,
+                column : editor.content.lines.last().unwrap().len(),
+            };
+            cursor.position = new_cursor_position;
+        }
+        if start_scrolling <= js_sys::Date::now() {
+            editor.scroll(Vector2::new(0.0, -0.01));
```
wdanilo: This `y` scroll value seems very small. I remember looking at the demo when the text was "jumping up" several times per second. I understand that its `y` movement is encoded here. My question is if the `y` and `x` position are aligned with `y` and `x` pixels respectivelly? Is moving `2` to right equal to moving 2 pixels rights? I would like to talk about it also in context of our camera - probably this should be part of integration task the next sprint.

lib/core/src/text.rs
```diff
@@ -1,49 +1,58 @@
...
 // === TextComponent ===
 // =====================
 
+#[derive(Debug)]
+pub struct TextComponentProperties {
+    pub position        : Point2<f64>,
+    pub size            : Vector2<f64>,
+    pub text_size       : f64,
+    pub color           : Color<f32>,
+}
```
wdanilo: Alignment

lib/core/src/lib.rs
farmaazon: So, some answers are in docs for `scroll` function used here. So: No, they aren't aligned with pixels. the 1.0 is equal to line's height (for both dimensions).

wdanilo: Ok, we need to talk about it on Monday when we will be planning merging it with our camera system.

lib/core/src/display/shape/text.rs
```diff
@@ -1,49 +1,58 @@
...
-use web_sys::WebGlBuffer;
 use web_sys::WebGlTexture;
 
 
 // =====================
 // === TextComponent ===
 // =====================
 
+#[derive(Debug)]
+pub struct TextComponentProperties {
```
wdanilo: Missing docs

lib/core/src/display/shape/text.rs
```diff
@@ -1,49 +1,58 @@
...
+    pub position  : Point2<f64>,
+    pub size      : Vector2<f64>,
+    pub text_size : f64,
+    pub color     : Color<f32>,
+}
+
 /// Component rendering text
 ///
 /// This component is under heavy construction, so the api may easily changed in few future
 /// commits
```
wdanilo: These docs commits are not senteces, as they do not end with a dot :D

lib/core/src/display/shape/text/content.rs
```diff
@@ -173,6 +175,13 @@ impl TextComponentContent {
         }
     }
 
+    pub fn line(& mut self, index:usize) -> LineRef {
```
wdanilo: Docs

PR #95, opened 2019-12-27
Performance fine-tuning & monitoring

wdanilo:
### Pull Request Description
- Implementation of https://github.com/luna/basegl/issues/83
- Partial implementation of (see comments on the ticket) https://github.com/luna/basegl/issues/77
- A lot has changed regarding how we can run and test the application. The `README` was updated to reflect the changes: https://github.com/luna/basegl/blob/120484928a6757804b013336c19bdbf0a8d8373c/README.md

### Important notes
- I'm committing back the ugly `examples/package-lock.json`. Without it the `npm install` fails now on MacOS Catalina, and it seems that having this ugly thing in the repo is the only way we can prevent a catastrophe from happening.

### Checklist


README.md
```diff
@@ -26,59 +27,59 @@ rustup component add rustfmt                # Install the code auto-formatter.
...
+cd examples
 npm install
-npm run start
+npm run start 
 ```
 
-You can now open the following address in your browser: http://localhost:8080.
+You can now navigate to http://localhost:8080 and play with the demo scenes!
 
-There are also _web test_ which are run in browser and produce some output. To
```
farmaazon: Why this section was deleted? I think this information should be available somewhere

examples/package-lock.json
```diff
@@ -0,0 +1,5646 @@
+{
```
farmaazon: AFAIR we don't want to have so big package-lock.json files in our repo, do we?

lib/core/src/debug/monitor.rs
```diff
@@ -0,0 +1,603 @@
...
+
+fn document() -> web_sys::Document {
+    window().document().unwrap_or_else(|| panic!("Cannot access window.document."))
+}
+
+fn body() -> web_sys::HtmlElement {
+    document().body().unwrap_or_else(|| panic!("Cannot access window.document.body."))
+}
+
+fn performance() -> Performance {
```
farmaazon: These are rather laconic names for functions in module. Perhaps you want to follow "getters without get" rule, but I think these function are not plain getters, so could be namet "get_performance_info" or something like that

lib/core/src/debug/monitor.rs
```diff
@@ -0,0 +1,603 @@
...
+        dom.append_child(&canvas).unwrap();
+
+        let mut out = Self{user_config,config,width,height,dom,panels,canvas,context,is_first_draw};
+        out.update_config();
+        out
+    }
+}
+
+impl Monitor {
+    /// Creates a new, empty monitor instance. Ise the `add` method to fill it with samplers.
```
farmaazon: Ise -> Use

lib/core/src/debug/monitor.rs
```diff
@@ -0,0 +1,603 @@
...
+    panels        : Vec<Panel>,
+    canvas        : HtmlCanvasElement,
+    context       : CanvasRenderingContext2d,
+    is_first_draw : bool,
+}
+
+
+// === Public API ===
+
+// TODO: All the `unwraps` below should be handled nicer when this lib will be finished.
```
farmaazon: I don't understand this todo. What lib, the whole basegl? Will basegl have the nicer handling of web_sys/js_sys errors?

Or if it's this "monitor" library, why we're trying to merge it if it's not finished?

lib/system/web/src/lib.rs
```diff
@@ -328,3 +329,8 @@ impl NodeRemover for Node {
         self.remove_child(node).unwrap_or_else(panic_msg);
     }
 }
+
+#[wasm_bindgen(inline_js = "export function request_animation_frame2(f) { requestAnimationFrame(f) }")]
+extern "C" {
+    pub fn request_animation_frame2(closure: &Closure<dyn FnMut()>) -> i32;
```
farmaazon: What is this for? I don't see any usages

README.md
wdanilo: 3 reasons:
1. All scripts should be short commands, if you want to put something to readme, create a script for it in scripts.
1. It sounds not very professional to me. "There are also _web test_ which are run in browser and produce some output". Its like - "yeah there are some tests and some output, yo". 
2. If I understand correctly, the `./scripts/test` contain those, right? 

Anyway, I'd be happy to add more info about it, ut I need help with it.

examples/package-lock.json
wdanilo: Please see readme of this PR.

lib/core/src/debug/monitor.rs
wdanilo: Hmm, but this is `window.performance` getter. It is not "performance info", it is "performance object". I'm happy to change the name, but I dont like "get_performance_info" too much. What do you think?

lib/core/src/debug/monitor.rs
wdanilo: This is about "monitor" lib. Monitoring was not my task - it is task of @notdanilo , but I needed it to performance fine tune things, so I created initial implementation here.

lib/system/web/src/lib.rs
wdanilo: Right! That was a test, thanks for catching!

examples/package-lock.json
farmaazon: Oh, silly me

README.md
farmaazon: Ad 3. no they don't. The script run all tests in headless mode. And if you need for short script for that, it should be a part of this PR, or a task should be created for this.

lib/core/src/debug/monitor.rs
farmaazon: Ok, I see that "performance object" is something javasscript specific? If you think it will be clear in usage places, then leave it as it is.

README.md
wdanilo: As we agreed on phone call, ill add the script in next pr.

PR #106, opened 2020-01-08
Cursor navigation

farmaazon:
### Pull Request Description
Created function for navigation cursors by single steps.

### Checklist


lib/core/src/display/shape/text.rs
```diff
@@ -12,7 +12,7 @@ use crate::display::world::Workspace;
 use crate::display::render::webgl::Context;
 use crate::display::shape::text::buffer::TextComponentBuffers;
 use crate::display::shape::text::content::TextComponentContent;
-use crate::display::shape::text::cursor::Cursors;
+use crate::display::shape::text::cursor::{Cursors, Step, CursorNavigation};
```
wdanilo: We're not using this import style.

lib/core/src/display/shape/text.rs
```diff
@@ -94,6 +94,12 @@ impl TextComponent {
         }
     }
 
+    pub fn navigate_cursors(&mut self, step:Step, selecting:bool, fonts:&mut Fonts) {
```
wdanilo: I relly don't like the idea that a font-map is involved in a cursor navigation operation. It seems to me like two concepts which should not mix. What is the reason that `fonts` have to be used here? Btw, I would find a better name for `Fonts`, like a `FontRegistry`.

lib/core/src/display/shape/text.rs
```diff
@@ -125,8 +131,7 @@ impl TextComponent {
     }
 
     fn refresh_cursors(&mut self, fonts:&mut Fonts) {
-        let cursors_changed = !self.cursors.dirty_cursors.is_empty();
-        if cursors_changed {
+        if self.cursors.dirty {
             let gl_context = &self.gl_context;
             let content    = &mut self.content;
```
wdanilo: I'm happy this could could be obsolete after our integration. It's tricky to handle it manually like here.

lib/core/src/display/shape/text/content.rs
```diff
@@ -124,10 +124,19 @@ impl TextChange {
...
 
+
+
 // ============================
 // === TextComponentContent ===
 // ============================
 
+/// A position of character in multiline text.
+#[derive(Copy,Clone,Debug,PartialEq,Eq,PartialOrd,Ord)]
+pub struct CharPosition {
```
wdanilo: I think that a name `TextLocation` reflects better the intention of this struct.

lib/core/src/display/shape/text/content.rs
```diff
@@ -124,10 +124,19 @@ impl TextChange {
     }
 }
 
+
+
 // ============================
 // === TextComponentContent ===
 // ============================
 
+/// A position of character in multiline text.
```
wdanilo: A very minor thing, missing "a" - "A position of character in a multiline text."

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -45,6 +43,17 @@ impl Cursor {
...
 
+    /// Extend the selection to cover the given range. Cursor itself may be moved, and will be
+    /// on the same side of selection as before.
+    pub fn extend_selection(&mut self, range:&Range<CharPosition>) {
+        let new_start = range.start.min(self.position).min(self.selected_to);
+        let new_end   = range.end.max(self.position).max(self.selected_to);
+        match self.position.cmp(&self.selected_to) {
+            Ordering::Less => { *self = Cursor{position:new_start, selected_to:new_end  } },
+            _              => { *self = Cursor{position:new_end  , selected_to:new_start} },
+        }
```
wdanilo: This could be refactored so, that `*self = ` would be before the match.

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -63,15 +72,165 @@ impl Cursor {
...
+        self.move_cursor_to_position(cursor,new_position);
+    }
+
+    /// Get cursor position at begin of given line.
+    pub fn line_begin(&self, line_index:usize) -> CharPosition {
+        CharPosition {
+            column : 0,
+            line   : line_index,
+        }
+    }
```
wdanilo: I believe that this should be a smart constructor of `CharPosition` used as `CharPosition::line_begin(...)`.

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -63,15 +72,165 @@ impl Cursor {
...
+    pub fn content_end(&self) -> CharPosition {
+        CharPosition {
+            column : self.content.lines.last().unwrap().len(),
+            line   : self.content.lines.len() - 1,
+        }
+    }
+
+    /// Get cursor position for the next char from given position. Returns none if at end of
+    /// whole document.
+    pub fn next_char(&self, position:&CharPosition) -> Option<CharPosition> {
```
wdanilo: I would name it `next_char_position`

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -63,15 +72,165 @@ impl Cursor {
...
+    }
+
+    /// Get cursor position one line behind the given position, such the new x coordinate of
+    /// displayed cursor on the screen will be nearest the current value.
+    pub fn line_down(&mut self, position:&CharPosition) -> Option<CharPosition> {
+        let next_line = Some(position.line + 1).filter(|l| *l < self.content.lines.len());
+        next_line.map(|line| self.near_same_x_in_another_line(position,line))
+    }
+
+    fn new_position(&mut self, position:CharPosition, step:&Step) -> CharPosition {
```
wdanilo: I know this is private function, but a comment of what it does and what it is used for would be really helpful here.

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -63,15 +72,165 @@ impl Cursor {
...
+    fn near_same_x_in_another_line(&mut self, position:&CharPosition, line:usize)
+    -> CharPosition {
+        let x_position = Cursor::x_position_of_cursor_at(position,self.content,self.fonts);
+        let column     = self.column_near_x(line,x_position);
+        CharPosition{line,column}
+    }
+
+    /// Get the column number in given line, so the cursor will be as near as possible the
+    /// `x_position` in _text space_. See `display::shape::text::content::line::Line`
+    /// documentation for details about _text space_.
```
wdanilo: Love these comments, super helpful, good job! :)

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -100,43 +259,93 @@ lazy_static! {
 /// a WebGL buffer with vertex positions of all cursors.
 #[derive(Debug)]
 pub struct Cursors {
-    pub cursors       : Vec<Cursor>,
-    pub dirty_cursors : HashSet<usize>,
-    pub buffer        : WebGlBuffer,
+    pub cursors : Vec<Cursor>,
+    pub dirty   : bool,
+    pub buffer  : Option<WebGlBuffer>,
```
wdanilo: I don't like the hierarchy here - mixing display logic with data logic. However, I believe no change is required here, as after moving to sprite system this will be cleaned up. This is just a note for you Adam, so you see how I think about it.

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -100,43 +259,93 @@ lazy_static! {
...
+        self.merge_overlapping_selections();
+        self.dirty = true;
+    }
+
+    /// Number of vertices in cursors' buffer.
+    pub fn vertices_count(&self) -> usize {
+        self.cursors.len() * CURSOR_BASE_LAYOUT_SIZE
+    }
+
+    fn merge_overlapping_selections(&mut self) {
```
wdanilo: This definitely needs a doc.

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -100,43 +259,93 @@ lazy_static! {
...
+                match merged {
+                    Some(merged_range) => {
+                        self.cursors[i-1].extend_selection(&merged_range);
+                        self.cursors.remove(i);
+                        last_range = merged_range;
+                    },
+                    None               => {
+                        last_range = current_range;
+                        i += 1
+                    }
```
wdanilo: In a case where multiple match branches are not line-by-line, I would not align `=>`. It looks strange. Btw, all toolsthat I know which align `=>` (like scalafmt) do it only when matches are line by line.

lib/core/src/display/shape/text.rs
farmaazon: It's used in navigating line up/down. To find what column in another line have x position near current cursor we must know the width of each character. And that is in font info

PR #108, opened 2020-01-09
Animation Manager (aka EasingAnimator)

notdanilo:
### Pull Request Description
This PR provides a new type of Animator called EasingAnimator, which can animate any object that implements HasPosition, from `origin_position` to `target_position` using an arbitrary function from our list of easing functions.

### Important Notes
- This PR provides all the easing functions from https://easings.net/
- EasingAnimator follows the same design of `Camera Animation System`'s `Animators`

### Checklist


lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,88 @@
...
+use std::rc::Rc;
+use std::cell::RefCell;
+
+
+// ==========================
+// === EasingAnimatorData ===
+// ==========================
+
+struct EasingAnimatorData {
+    object              : Box<dyn HasPosition>,
```
wdanilo: I don't like this. EasingAnimatorData should work on anything, not necessary something which has position. I want for example to connect it to a Div opacity value. It should accept a lambda which it should call with a new value instead.

lib/core/src/animation.rs
```diff
@@ -2,4 +2,5 @@
 
 pub mod physics;
 pub mod animator;
-pub mod position;
\ No newline at end of file
+pub mod position;
+pub mod easing;
```
farmaazon: According to our guidelines, there should be a newline at end of file

lib/core/src/animation/animator/continuous.rs
```diff
@@ -84,24 +86,28 @@ impl ContinuousAnimatorData {
...
+                let absolute_start_ms = data.absolute_start_ms();
+                let absolute_start_ms = absolute_start_ms.unwrap_or_else(|| {
+                    data.set_absolute_start_ms(Some(current_time));
+                    current_time
+
+                });
+                let relative_start_ms = data.relative_start_ms();
+                let relative_time_ms  = current_time - absolute_start_ms + relative_start_ms;
+                data.on_animation_frame(relative_time_ms);
+            }
```
farmaazon: I'd really like to have this block in a separate function. It's not intuitive to search for "code run for each frame" in constructor.

So, the `if lest Some(data)` should remain here, but the main logic should be a separate function - maybe as member of ContinuousAnimatorData?

lib/core/src/animation/animator/continuous.rs
```diff
@@ -84,24 +86,28 @@ impl ContinuousAnimatorData {
...
-            });
-            let relative_start_ms = data_clone.relative_start_ms();
-            let relative_time_ms  = current_time - absolute_start_ms + relative_start_ms;
-            data_clone.on_animation_frame(relative_time_ms);
+            if let Some(data) = weak_data.upgrade() {
+                let absolute_start_ms = data.absolute_start_ms();
+                let absolute_start_ms = absolute_start_ms.unwrap_or_else(|| {
+                    data.set_absolute_start_ms(Some(current_time));
+                    current_time
+
```
farmaazon: remove this empty line

lib/core/src/animation/animator/continuous.rs
```diff
@@ -84,24 +86,28 @@ impl ContinuousAnimatorData {
 // === ContinuousAnimator ===
 // ==========================
 
-/// This structure runs an animation with continuous time as its input.
+/// This structure runs an animation with continuous time in milliseconds.
```
farmaazon: I think this doc could explain more. I don't know what is the meaning "the structure runs an animation". This docs should explain that animation is a registered event_loop callback calling repeatedly some animation_callback.

lib/core/src/animation/animator/continuous.rs
```diff
@@ -14,7 +16,7 @@ struct ContinuousTimeAnimatorProperties {
     callback          : Box<dyn AnimationCallback>,
     relative_start_ms : f32,
     absolute_start_ms : Option<f32>,
-    callback_guard    : Option<AnimationFrameCallbackGuard>
+    callback_guard    : Option<CallbackHandle>
```
farmaazon: Maybe "event_loop_callback_guard" to distinct this from the animation callback

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct EasingAnimator {
+    data : Rc<RefCell<EasingAnimatorData>>
+}
+
+impl EasingAnimator {
+    /// Creates an EasingAnimator using a `easing_function` to move `object` from
+    /// `origin_position` to `target_position` in `duration_seconds`.
+    pub fn new<T,F>
+    (mut event_loop   : &mut EventLoop
+     , easing_function  : F
```
farmaazon: the dots should be aligned with ()
And `where` should be intended one level less.

lib/core/src/animation/animator/easing.rs
farmaazon: And why the object is boxed instead of being just generic parameter?

lib/core/src/animation/easing.rs
```diff
@@ -0,0 +1,119 @@
...
+        pub fn [<$name _out>](t:f64) -> f64 { 1.0 - [<$name _in>](1.0 - t) }
+        /// A $name in out transition.
+        pub fn [<$name _in_out>](t:f64) -> f64 {
+            let t = t * 2.0;
+            if t < 1.0 {
+                [<$name _in>](t) / 2.0
+            } else {
+                ([<$name _out>](t - 1.0) + 1.0) / 2.0
+            }
+}
```
farmaazon: this } is wrongly aligned.

lib/core/src/control/callback.rs
```diff
@@ -66,27 +57,32 @@ impl Guard {
 /// handle lifetimes are strictly connected. As soon a handle is dropped, the callback is removed
 /// as well.
 #[derive(Derivative)]
-#[derivative(Debug, Default)]
-pub struct CallbackRegistry {
+#[derivative(Debug)]
+pub struct CallbackRegistry<F> {
     #[derivative(Debug="ignore")]
-    callback_list: Vec<(Guard, Box<dyn FnMut()>)>
+    callback_list: Vec<(Guard, F)>
```
farmaazon: Something weird is done here. The callback registry is no longer a callback registry, because it does not collect callbacks anymore, but some arbitrary type F. And `run_all` is in fact a function _running one callback for all Fs_, not _run all callbacks_

lib/core/src/control/event_loop.rs
```diff
@@ -40,18 +47,18 @@ impl EventLoop {
     /// Init the event loop.
     fn init(self) -> Self {
         let data = Rc::downgrade(&self.rc);
-        let main = move |_| { data.upgrade().map(|t| t.borrow_mut().run()); };
+        let main = move |time_ms| { data.upgrade().map(|t| t.borrow_mut().run(time_ms)); };
         with(self.rc.borrow_mut(), |mut data| {
             data.main = Some(Closure::new(main));
-            data.run();
+            data.run(0.0);
```
farmaazon: This 0.0 may be really surprising. Either pass here "performance.now" value, or just register main and let wait for the first frame.

lib/core/src/control/callback.rs
farmaazon: Ok, now I see why you did it that way, but I think it's much better to add argument type as a parameter. If someone will need to pass more argument, he can use tuple or create struct with named parameters.

lib/core/src/examples.rs
```diff
@@ -3,4 +3,5 @@
 pub mod sprite_system;
 pub mod text;
 pub mod shapes;
-pub mod camera_navigation;
\ No newline at end of file
+pub mod camera_navigation;
+pub mod easing_animator;
```
farmaazon: newline

lib/core/src/examples/camera_navigation.rs
```diff
@@ -13,7 +14,6 @@ use crate::animation::animator::continuous::ContinuousAnimator;
 use crate::animation::physics::inertia::*;
 
 use nalgebra::{Vector2, Vector3};
```
farmaazon: Let take the chance and fix this imports to be in a separate lines

lib/core/src/examples/easing_animator.rs
```diff
@@ -0,0 +1,282 @@
...
+use crate::system::web::create_element;
+use crate::system::web::NodeInserter;
+use crate::system::web::AttributeSetter;
+use crate::system::web::StyleSetter;
+use crate::display::render::css3d::Object;
+use crate::control::event_loop::EventLoop;
+
+use nalgebra::{Vector2, Vector3};
+
+use web_sys::{HtmlElement, HtmlCanvasElement, CanvasRenderingContext2d};
```
farmaazon: Imports in separate lines

lib/web-test/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2018"
 [dependencies]
 wasm-bindgen-test = "0.3.3"
 web-test-proc-macro  = { version = "0.1.0" , path = "../web-test-proc-macro" }
+basegl               = { version = "0.1.0" , path = "../core"       }
```
farmaazon: Please align all }

lib/core/src/control/callback.rs
wdanilo: I agree, this is no longer callback registry. As I've written earlier, if you have problem making it generic over the callback types, please make copy of the original implementation, name it `CallbackRegistry1`, which will be parametrized with 1 parameter (the type of argument to every callback) and I will merge these implementations later. But we should not change the API of the original callback registry, as the API does not behave right now like a callback firing utility.

lib/core/src/animation/animator/easing.rs
notdanilo: @farmaazon it's just for the case we want to reuse EasingAnimator and change it with other object that also implement `HasPosition`. We can't do that when after monomorphization. But I gotta ask: What's the trade-off between using boxed dyn and using generic parameters?

About the flexibility, I will replace `HasPosition` with a new trait called `Interpolatable` and extend our example to show it animating different properties.

lib/core/src/control/callback.rs
notdanilo: This is the way I found to be able to send multiple arguments to our callbacks. Unfortunately it removes some constraints and makes it possible to store any kind of type in the registry.

`#![fn_traits]` looks promising for our APIs, but I found in my tests it's unstable.

I will leave `CallbackRegistry` with the feel and looks of a callback firing utility and I will provide create `CallbackRegistry1<T:Copy>`. Please note that `T` has a `Copy` constraint, because when firing it with `run_all`, it tries to move the value and, well, I need a copy of the same argument for every callback.
```rust
    /// Fires all registered callbacks.
    pub fn run_all(&mut self, t:T) {
        self.clear_unused_callbacks();
        self.callback_list.iter_mut().for_each(move |(_,callback)| callback(t));
    }
```

Is there any suggestion you would like to make about this API?

lib/core/src/animation/animator/continuous.rs
notdanilo: @farmaazon thanks for pointing it out! The name certainly needs to be fixed.

lib/core/src/animation/animator/continuous.rs
notdanilo: I think this should be clear enough:
```rust
/// `ContinuousAnimator` registers itself in `EventLoop`, repeatedly calling an
/// `AnimationCallback` with the playback time in millisecond as its input.
```

lib/core/src/animation/animator/continuous.rs
notdanilo: It makes sense. I moved the whole block to `ContinuousAnimatorData`'s `on_animation_frame`.

lib/core/src/control/callback.rs
```diff
@@ -57,36 +67,63 @@ impl Guard {
...
-        self.callback_list.iter_mut().for_each(|(_,callback)| c(callback));
+        self.callback_list.iter_mut().for_each(move |(_,callback)| callback(t));
     }
 
     /// Checks all registered callbacks and removes the ones which got dropped.
     fn clear_unused_callbacks(&mut self) {
         self.callback_list.retain(|(guard,_)| guard.exists());
     }
-}
+}
```
farmaazon: It looks better, but still I think that instead of copying code for CallBackRegistry1 from CallbackRegistry we should have only one CallbackRegistry with one argument parameter. If someone does not want to pass arguments to their callbacks, he could just make () as an argument (as it is usually done for `Result` if function does not returns value, but might have an error).

@wdanilo I heard that you have a plan for callback registry. Shall we leave two versions or can we deduplicate as I described above?

lib/core/src/control/event_loop.rs
```diff
@@ -50,7 +49,9 @@ impl EventLoop {
         let main = move |time_ms| { data.upgrade().map(|t| t.borrow_mut().run(time_ms)); };
         with(self.rc.borrow_mut(), |mut data| {
             data.main = Some(Closure::new(main));
-            data.run(0.0);
+            if let Some(closure) = &data.main {
+                web::request_animation_frame(&closure).unwrap();
+            }
```
farmaazon: The if will be always true. What about 
```rust
let closure = Closure::new(main)
web::request_animation_frame(&closure).unwrap();
data.main = Some(closure)
```

lib/core/src/examples/easing_animator.rs
```diff
@@ -125,45 +170,49 @@ struct SubExample {
 
 impl SubExample {
     fn new<F>
-    ( mut event_loop   : &mut EventLoop
-    , graph_canvas     : Canvas
-    , animation_canvas : Canvas
-    , f                : &'static F
-    , origin_position  : Vector3<f32>
-    , target_position  : Vector3<f32>) -> Self
```
farmaazon: As I know, this is properly formatted signature, not like in the new version. Or maybe the guidelines changed?

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Creates an EasingAnimator using a `easing_function` to move `object` from
+    /// `initial_value` to `final_value` in `duration_seconds`.
+    pub fn new<F,C>
+    ( mut event_loop                : &mut EventLoop
+    , mut easing_animation_callback : C
+    , easing_function               : F
+    , initial_value                 : T
+    , final_value                   : T
+    , duration_seconds              : f64) -> Self
+    where F:FnEasing, C:EasingAnimationCallback<T> {
```
notdanilo: @wdanilo @farmaazon please note the API change.

I am no longer attaching the `Interpolable` object to `EasingAnimator`. Now we just need to set the initial and final values and pass a closure to `EasingAnimator` to get the interpolated values :)

lib/core/src/examples/easing_animator.rs
notdanilo: @farmaazon I am not sure I got it. Isn't it complying to this rule?
```rust
pub fn new<Dom:Str>
( dom        : Dom
, logger     : Logger
, on_dirty   : OnDirty
, on_remove  : OnRemove
, on_replace : OnReplace
) -> Result<Self,Error> {
    ...
}
```

lib/core/src/control/event_loop.rs
notdanilo: I updated it to
```rust
            data.main = Some(Closure::new(main));
            web::request_animation_frame(&data.main.as_ref().unwrap()).unwrap();
```

I thought about this
```rust
let closure = Closure::new(main)
web::request_animation_frame(&closure).unwrap();
data.main = Some(closure)
```

But `closure` execute these lines in `EventLoopData`'s `run`:

```rust
        let callback_id = self.main.as_ref().map_or(default(), |main| {
            callbacks.run_all(time_ms);
            web::request_animation_frame(main).unwrap()
        });
```

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
...
+
+pub trait EasingAnimationCallback<T> = FnMut(T) + 'static;
+
+
+
+// ==========================
+// === EasingAnimatorData ===
+// ==========================
+
+struct EasingAnimatorData<T:Interpolable<T>+'static> {
```
wdanilo: Because you used `+'static` here - it propagated all the way down to `EasingAnimator` and also to `impls` below. I mentioned some time ago that almost every usage of `'static` in other contexts than `F:Fn() + 'static` when `F` gets Boxed is an orange lamp. Sure, compiler tells you that this is a possible solution, but in 90% of cases this is ugly solution. Like here. Please remove it and do not use `'static` if not really, really needed.

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub fn new<F,C>
+    ( mut event_loop                : &mut EventLoop
+    , mut easing_animation_callback : C
+    , easing_function               : F
+    , initial_value                 : T
+    , final_value                   : T
+    , duration_seconds              : f64) -> Self
+    where F:FnEasing, C:EasingAnimationCallback<T> {
+        let continuous_animator = None;
+        let data                 = EasingAnimatorData{
```
wdanilo: Alignment

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
...
+        };
+        let data = Rc::new(RefCell::new(data));
+        let weak = Rc::downgrade(&data);
+        let continuous_animator = ContinuousAnimator::new(&mut event_loop, move |time_ms| {
+            if let Some(data) = weak.upgrade() {
+                let data             = data.borrow();
+                let duration_seconds = data.duration_seconds;
+                let initial_value    = data.initial_value;
+                let final_value      = data.final_value;
+                let time_seconds     = time_ms as f64 / 1000.0 / duration_seconds;
```
wdanilo: 2 minor performance conerns:
- Computing `1000.0 / duration_seconds` every tick doesnt make sense
- `time_ms as f64` I don't like casting values every tick

lib/core/src/animation/animator/easing.rs
wdanilo: I highly appreciate how you make the code layout look nice, good job! ❤️

I was thinking about this example, and I believe that we could get a more clear layout this way:

```rust
    pub fn new<F:FnEasing, C:EasingAnimationCallback<T>>
    ( mut event_loop                : &mut EventLoop
    , mut easing_animation_callback : C
    , easing_function               : F
    , initial_value                 : T
    , final_value                   : T
    , duration_seconds              : f64) -> Self {
```

The reason for that is that the first line has already the variables and you left it empty already. So there is a place to define what `F` and what `C` is. I would not have so big problem if you used more descriptive names than `F` and `C`. It is not bad to use short names if they are described, and when I see `F:FnEasing` it is perfectly clear what happens here :)

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
...
+                let time_seconds     = clamp(time_seconds, 0.0, 1.0);
+                let time_seconds     = easing_function(time_seconds) as f32;
+                let value = linear_interpolation(initial_value, final_value, time_seconds);
+                easing_animation_callback(value);
+            }
+        });
+        data.borrow_mut().continuous_animator = Some(continuous_animator);
+        Self { data }
```
wdanilo: I hate this pattern. It is the best we can get here, still, I hate it :P 
I had to write it to express my hate to ugly rust patterns that cannot be expressed in a more secure way :D

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /// Animates the attached object from `initial_value` to `final_value` in
```
wdanilo: What is "the attached object"?

lib/core/src/animation/animator/fixed_step.rs
```diff
@@ -15,11 +15,14 @@ use crate::system::web::animation_frame_loop::AnimationFrameLoop;
 /// This struct counts the intervals in a time period.
 #[derive(Debug)]
 pub struct IntervalCounter {
+    /// Interval duration.
     pub interval_duration : f32,
+    /// The amount of time accumulated by `IntervalCounter`'s `add_time`.
     pub accumulated_time  : f32
 }
```
wdanilo: We have not been talking about it earlier, but I believe that in case we document pub fields, they should not be aligned and they should be separated vertically. What do you think about this instead? CC @farmaazon 

```rust
pub struct IntervalCounter {
    /// Interval duration.
    pub interval_duration: f32,

    /// The amount of time accumulated by `IntervalCounter`'s `add_time`.
    pub accumulated_time: f32
}
```

lib/core/src/animation/animator/fixed_step.rs
```diff
@@ -15,11 +15,14 @@ use crate::system::web::animation_frame_loop::AnimationFrameLoop;
...
 
 impl IntervalCounter {
+    /// Creates `IntervalCounter` which counts how many intervals with an `interval_duration`.
```
wdanilo: I'm ok with it. Just FYI - the function `new` I always just document as `/// Constructor.` because it is abvious what it does. However, I put bigger emphasis on documenting the structure itself. The information about counting intervals etc - describes what the structure does and how it operates, it does not strictly documents the `new` function itself.

lib/core/src/animation/easing.rs
```diff
@@ -0,0 +1,118 @@
+//! This file provides several easing functions used for transition animation.
```
wdanilo: Everything here looks nice and cool! One thing raised my attention. Everything here is defined on `f64`. In many cases, I will be animating `f32` values. Casting them back and forth is rather a bad idea. Can we make these implementations polymorphic? At least for both `f64` and `f32`? If not (If that would be hard), then I would prefer to choose `f32` over `f64`, as all WebGL values are `f32`.

lib/core/src/animation/easing.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// Easing function signature.
+pub trait FnEasing = Fn(f64) -> f64 + 'static;
+
+
+
+// ========================
+// === Easing functions ===
+// ========================
+// Reference: http://easings.net/en
```
wdanilo: This should be in the description of the module.

lib/core/src/animation/easing.rs
```diff
@@ -0,0 +1,118 @@
...
+    if t < 1.0 {
+        back_in_params(t, overshoot) / 2.0
+    } else {
+        (back_out_params(t - 1.0, overshoot) + 1.0) / 2.0
+    }
+}
+
+easing_fn!(pub fn back(t:f64) -> f64 { back_in_params(t, 1.70158) });
+
+/// A elastic in transition with params.
```
wdanilo: A very minor comment. "elastic in" is a one description, should be spelled "An elastic-in transition with parameters." instead.

lib/core/src/animation/physics/inertia.rs
```diff
@@ -119,23 +124,37 @@ impl KinematicsProperties {
...
+    pub fn velocity(&self) -> Vector3<f32> {
+        self.velocity
+    }
+
+    /// `Acceleration` getter.
+    pub fn acceleration(&self) -> Vector3<f32> {
+        self.acceleration
+    }
+
+    /// `Mass` getter.
```
wdanilo: I'm sad we cannot disable the requirement of writing docs for getters :P It's sad.

lib/core/src/control/callback.rs
wdanilo: Let's leave 2 versions now. I will merge them.

lib/core/src/control/callback.rs
```diff
@@ -94,3 +95,35 @@ impl CallbackRegistry {
...
+    pub fn run_all(&mut self, t:T) {
+        self.clear_unused_callbacks();
+        self.callback_list.iter_mut().for_each(move |(_,callback)| callback(t));
+    }
+
+    /// Checks all registered callbacks and removes the ones which got dropped.
+    fn clear_unused_callbacks(&mut self) {
+        self.callback_list.retain(|(guard,_)| guard.exists());
+    }
+}
```
wdanilo: Newline :D

lib/core/src/control/event_loop.rs
```diff
@@ -40,24 +46,26 @@ impl EventLoop {
     /// Init the event loop.
     fn init(self) -> Self {
         let data = Rc::downgrade(&self.rc);
-        let main = move |_| { data.upgrade().map(|t| t.borrow_mut().run()); };
+        let main = move |time_ms| { data.upgrade().map(|t| t.borrow_mut().run(time_ms)); };
         with(self.rc.borrow_mut(), |mut data| {
             data.main = Some(Closure::new(main));
-            data.run();
+            web::request_animation_frame(&data.main.as_ref().unwrap()).unwrap();
```
wdanilo: What is the purpose of this change? I believe the previous version was working in the same way, but I might be missing something here :)

lib/core/src/control/event_loop.rs
```diff
@@ -95,21 +103,23 @@ impl Default for EventLoopData {
 
 impl EventLoopData {
     /// Create new instance.
-    pub fn run(&mut self) {
+    pub fn run(&mut self, time_ms:f32) {
         (self.on_loop_started)();
         let callbacks   = &mut self.callbacks;
         let callback_id = self.main.as_ref().map_or(default(), |main| {
-            callbacks.run_all();
+            callbacks.run_all(time_ms);
```
wdanilo: I have to admit that the naming of variables in this and previous PR is so nice. It is **MUCH** better than it was in the first PRs. Really good job Danilo! ❤️

lib/core/src/control/event_loop.rs
```diff
@@ -95,21 +103,23 @@ impl Default for EventLoopData {
...
             web::request_animation_frame(main).unwrap()
         });
         self.main_id = callback_id;
         (self.on_loop_finished)();
     }
 
+    /// Sets a callback which is called when the loop started.
     pub fn set_on_loop_started<F:FnMut()+'static>(&mut self, f:F) {
```
wdanilo: I believe we have a type alias for `FnMut() + 'static`

lib/core/src/control/mouse_manager.rs
```diff
@@ -276,17 +294,13 @@ impl MouseManager {
         Ok(EventListener::new(self.data.target(), "contextmenu", listener))
     }
 
-    /// Adds mouse down event callback and returns its listener object.
-    add_callback!(add_mouse_down_callback, MouseClick, "mousedown");
+    add_callback!(mouse_down, MouseClick, "mousedown");
```
wdanilo: Good change! 👍

lib/core/src/display/navigation/navigator.rs
```diff
@@ -54,7 +54,7 @@ impl Navigator {
     }
 
     fn start_simulator
-    ( mut event_loop:&mut AnimationFrameLoop
+    ( mut event_loop:&mut EventLoop
     , camera:Camera) -> (PhysicsSimulator, PhysicsProperties) {
```
wdanilo: Can we do it in single line now?

lib/core/src/examples/easing_animator.rs
```diff
@@ -0,0 +1,331 @@
...
+            easing_function,
+            graph_canvas,
+            animation_canvas,
+            easing_animator
+        };
+        let data = Rc::new(RefCell::new(data));
+        let weak = Rc::downgrade(&data);
+        let easing_animator = EasingAnimator::new(
+            &mut event_loop,
+            move |value| { weak.upgrade().map(|data| data.borrow_mut().properties = value); },
```
wdanilo: This are examples so I'm ok with it, but in normal code, I would really dislike naked refcells like here. If you see a simple way to just wrap those in a newtype, please do. If not, lets leave them.

lib/core/src/math/utils.rs
```diff
@@ -1,11 +1,11 @@
-#![allow(missing_docs)]
```
wdanilo: Wait, I removed this file in the past. I don't think we should have `math` top-level package, as we do not have many utils there. Please keep them in sub-modules until we get significant amount of those.

lib/core/src/examples/easing_animator.rs
farmaazon: Yes - the old version. I saw change adding the space before each comma.

lib/core/src/animation/animator.rs
```diff
@@ -1,8 +1,11 @@
+//! This file implements Animator.
```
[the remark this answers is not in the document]
notdanilo: I made it:
```rust
//! This file implements the Animation struct which runs a callback once per frame with a time
//! difference from the last frame as its input.
```

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
+//! This file provides the implementation of EasingAnimator.
```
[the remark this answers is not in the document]
notdanilo: I used your suggestion linking to https://easings.net/en

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use super::ContinuousAnimator;
+use crate::animation::easing::FnEasing;
+use crate::control::event_loop::EventLoop;
+use crate::math::utils::linear_interpolation;
+use crate::math::utils::Interpolable;
+
+use nalgebra::clamp;
+use std::rc::Rc;
+use std::cell::RefCell;
```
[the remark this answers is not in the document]
notdanilo: I wasn't aware it re-exported `rc` and `cell`. :) I am using it now.

lib/core/src/animation/animator/easing.rs
notdanilo: Docs of the old implementation I missed to update.

lib/core/src/animation/animator/fixed_step.rs
notdanilo: I thought about doing it, it feels more readable to me. I will change it accordingly

lib/core/src/animation/animator/fixed_step.rs
notdanilo: @wdanilo I think it's interesting to document `new` as it takes some parameters which only makes sense during its construction.

lib/core/src/animation/animator/easing.rs
notdanilo: I removed most of them, but L46 `impl<T:Interpolable<T>+'static> EasingAnimator<T>` still seemed necessary. I really don't understand why it needs an explicit lifetime because T is moved as a copy and not as a reference.

lib/core/src/animation/easing.rs
notdanilo: I made it f32. I tried to make polymorphic using nalgebra::RealField, but it wasn't trivial. I can work on it later.

I added this comment

```rust
// TODO: It would be useful to make the easing functions polymorphic. Maybe we can make use of
// nalgebra::RealField?
```

lib/core/src/animation/easing.rs
notdanilo: Fixd

lib/core/src/control/event_loop.rs
notdanilo: @wdanilo, `data.run` now takes a time input and since we don't have this information at this moment, we call instead `request_animation_frame`.

lib/core/src/control/event_loop.rs
notdanilo: Thx :D

lib/core/src/animation/animator/fixed_step.rs
```diff
@@ -42,16 +47,16 @@ impl IntervalCounter {
...
 
 impl FixedStepAnimatorData {
-    pub fn new<F:AnimationCallback>(steps_per_second:f32, f:F) -> Self {
-        let closure          = Box::new(f);
+    pub fn new<F:AnimationCallback>(steps_per_second:f64, f:F) -> Self {
+        let callback         = Box::new(f);
         let step_duration    = 1000.0 / steps_per_second;
         let counter          = IntervalCounter::new(step_duration);
-        Self { closure,counter }
+        Self { callback,counter }
```
wdanilo: This is a very very minor thing, but lets use a single style - `{` is like `(` and `}` is like `)` , so instead of `{ callback,counter }` wer just write `{callback,counter}`.

lib/core/src/animation/easing.rs
```diff
@@ -0,0 +1,122 @@
...
+
+
+
+// ========================
+// === Easing functions ===
+// ========================
+// Reference: http://easings.net/en
+
+macro_rules! easing_fn {
+    (pub fn $name:ident(t:f32) -> f32 $block:block) => { paste::item! {
```
wdanilo: Why don't you write a macro `easing_fns` and just put all functions in it instead of using this macro so often below? :)

lib/core/src/animation/animator/easing.rs
```diff
@@ -0,0 +1,93 @@
...
+// ======================
+// === EasingAnimator ===
+// ======================
+
+/// This struct animates from `origin_position` to `target_position` using easing functions.
+pub struct EasingAnimator<T:Interpolable<T>> {
+    data : Rc<RefCell<EasingAnimatorData<T>>>
+}
+
+impl<T:Interpolable<T>+'static> EasingAnimator<T> {
```
wdanilo: Let's rename this `'static` to `CallbackArgument`.

lib/core/src/animation/easing.rs
```diff
@@ -0,0 +1,122 @@
+//! This module implements easing formulas and related utilities. To learn more about easing
+//! functions, please refer to the following link: https://easings.net/en
+
+// TODO: It would be useful to make the easing functions polymorphic. Maybe we can make use of
+// nalgebra::RealField?
+
+use core::f32::consts::PI;
+
+/// Easing function signature.
+pub trait FnEasing = Fn(f32) -> f32 + 'static;
```
wdanilo: I don't like this style of writing this code. It looks like the `'static` was attached to `f32`. Write `pub trait FnEasing = (Fn(f32) -> f32) + 'static;` or `pub trait FnEasing = 'static + Fn(f32) -> f32;` instead :)

lib/core/src/animation/position.rs
```diff
@@ -1,11 +1,12 @@
-#![allow(missing_docs)]
+//! This module defines the `HasPosition` trait.
```
wdanilo: As agreed on Discord, lets remove this trait and use lambdas instead.

lib/core/src/control.rs
```diff
@@ -2,3 +2,4 @@
 
 pub mod callback;
 pub mod io;
+pub mod event_loop;
```
wdanilo: newline

lib/core/src/control/callback.rs
wdanilo: This comment was not applied ^^^ :)

lib/core/src/control/event_loop.rs
wdanilo: What is this time? Would it be ok if we pass `0` instead? Or the time doesnt start with 0?

lib/core/src/examples/easing_animator.rs
```diff
@@ -0,0 +1,331 @@
...
+        };
+        let data = Rc::new(RefCell::new(data));
+        let weak = Rc::downgrade(&data);
+        let easing_animator = EasingAnimator::new(
+            event_loop,
+            move |value| { weak.upgrade().map(|data| data.borrow_mut().properties = value); },
+            f,
+            initial_value,
+            final_value,
+            2.0
```
wdanilo: Please, refactor out this value to a variable - when reading the code I dont remember what it is and jumping to other file is bad here.

lib/core/src/animation/animator.rs
```diff
@@ -1,16 +1,20 @@
+//! This module implements the Animation struct which runs a callback once per frame with a time
+//! difference from the last frame as its input.
+
 pub mod continuous;
 pub mod fixed_step;
+pub mod easing;
 
 use continuous::ContinuousAnimator;
-use crate::system::web::animation_frame_loop::AnimationFrameLoop;
+use crate::control::event_loop::EventLoop;
```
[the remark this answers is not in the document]
notdanilo: What about this?

```rust
mod event_loop;
pub use event_loop::*;
```

lib/core/src/control/event_loop.rs
notdanilo: We can't pass 0 because it's fed with `performance.now()`. It returns the elapsed time, in milliseconds, from the beginning of the execution.

PR #112, opened 2020-01-15
Texture rendering

farmaazon:
### Pull Request Description
This PR contains code created for rendering Textures. Also it allows for create texture from memory, not only from url location.

I want to merge it despite not being well tested, because it alters somehow the structures used in texture management, what may have an influence over next tasks people are doing.

### Checklist


lib/core/src/display/symbol.rs
```diff
@@ -216,22 +238,26 @@ impl Symbol {
     /// Creates a new VertexArrayObject, discovers all variable bindings from shader to geometry,
     /// and initializes the VAO with the bindings.
     fn init_variable_bindings(&mut self, var_bindings:&[shader::VarBinding]) {
-        self.vao      = Some(VertexArrayObject::new(&self.context));
-        self.uniforms = default();
-        self.textures = default();
+        const MAX_TEXTURE_UNITS:u32 = 16;
```
wdanilo: Dont hardcode 16 slots - check it!
Sample API usage: https://webglreport.com/?v=2
Stats: https://webglstats.com/webgl/parameter/MAX_TEXTURE_IMAGE_UNITS
As you can see, many devices nowadays do not support 16 textures. 4% of devices support 32 textures.

lib/core/src/display/symbol.rs
```diff
@@ -258,8 +291,14 @@ impl Symbol {
             match uniform {
                 AnyUniform::Prim(uniform) =>
                     self.uniforms.push(UniformBinding::new(name,location,uniform)),
-                AnyUniform::Texture(uniform) =>
-                    self.textures.push(TextureBinding::new(name,location,uniform)),
+                AnyUniform::Texture(uniform) => {
+                    let texture_unit = texture_unit_iter.next().unwrap_or_else(|| {
+                        panic!("Texture unit limit exceeded.");
```
wdanilo: If you can, do not panic! Emit error message and drop that texture. It is better to have a working program with glitched graphics than a crashed application.

lib/core/src/display/symbol.rs
```diff
@@ -258,8 +291,14 @@ impl Symbol {
...
+                    });
+                    let binding      = TextureBinding::new(name,location,uniform,texture_unit);
```
wdanilo: Spacing

lib/core/src/display/symbol/material.rs
```diff
@@ -17,15 +17,17 @@ use crate::system::gpu::types::*;
 #[derive(Clone,Debug)]
 pub struct VarDecl {
     /// The GLSL type of the variable.
-    pub tp: glsl::PrimType,
+    pub tp : glsl::PrimType,
```
wdanilo: Minor - we do not put spaces before `:` if it is not aligned with other lines.

lib/core/src/display/symbol/material.rs
```diff
@@ -58,6 +60,7 @@ pub struct Material {
 pub trait Input = Into<VarDecl> + GpuDefault;
 
 impl Material {
+
```
wdanilo: There should not be this newline.

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -326,49 +331,72 @@ with_texture_format_relations!(generate_internal_format_instances []);
 
 
 
-// =====================
-// === TextureSource ===
-// =====================
+// =======================
+// === Texture sources ===
```
wdanilo: This is a heading. In English headings have all words capitalized.

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -326,49 +331,72 @@ with_texture_format_relations!(generate_internal_format_instances []);
...
+// =======================
+
+// === Empty Texture ===
+
+/// An empty texture, created as 1x1 texture in Rgba format.
+pub struct EmptyTexture;
```
wdanilo: This is not empty texture. It is an UnitTexture. Empty texture IMO means that you create a texture but does not initialize it with data (e.g. for the needs of framebuffers). Let's talk about it via phone tomorrow.

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -326,49 +331,72 @@ with_texture_format_relations!(generate_internal_format_instances []);
...
-    Url(String)
+
+// === Texture from url ===
+
+/// Texture loaded from given url. It is loaded to gpu asynchronously, putting a mocked texture
+/// before
+pub struct TextureFromUrl {
+    /// An url from where the texture is loaded.
+    pub url       : String,
 }
```
wdanilo: This should be named `Url` instead. This type is very generic, it does not have anything in common with textures.

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -326,49 +331,72 @@ with_texture_format_relations!(generate_internal_format_instances []);
...
     }
 }
 
 
+// === Texture from memory ===
+
+/// A texture created from slice.
+pub struct TextureFromMemory<'a,T> {
+    /// A slice from which the texture data are taken.
+    pub view   : &'a[T],
```
wdanilo: I don't like the idea that the texture does not own this data. IMO data should be owned by the texture and everyone who has access to the uniform should be able to modify this data. Let's talk about it via phone.

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -326,49 +331,72 @@ with_texture_format_relations!(generate_internal_format_instances []);
...
 
-/// Texture representation.
-#[derive(Derivative)]
-#[derivative(Clone(bound=""))]
-#[derivative(Debug(bound=""))]
-pub struct Texture<InternalFormat,ElemType> {
-    source  : TextureSource,
-    phantom : PhantomData2<InternalFormat,ElemType>,
+/// A Texture.
+pub struct Texture<Source,InternalFormat,ElemType> {
```
wdanilo: I'm not sure if the texture should be parametrized with `Source`. The longer I think about it the more I dislike this idea. Let's have a call about it.

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -326,49 +331,72 @@ with_texture_format_relations!(generate_internal_format_instances []);
...
-impl<I:InternalFormat,T:TextureItemType> Texture<I,T> {
-    /// Constructor.
-    pub fn new<S:Into<TextureSource>>(source:S) -> Self {
-        let source  = source.into();
+impl<S,I,T> Texture<S,I,T> {
+    /// Create a new texture from given source.
+    pub fn new(source:S) -> Self {
         let phantom = PhantomData;
-        Self {source,phantom}
+        Texture{source,phantom}
```
wdanilo: spacing

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -427,31 +509,32 @@ pub struct BoundTexture<I,T> {
 /// Texture bound to GL context.
 #[derive(Debug)]
 pub struct BoundTextureData<I,T> {
-    texture    : Texture<I,T>,
     gl_texture : WebGlTexture,
     context    : Context,
+    format     : PhantomData<I>,
+    elem_type  : PhantomData<T>,
 }
```
wdanilo: First of all the name `format` is misleading. This is `internal_format`, not `format. Second of all, I would keep the names more visible. Why not this instead?

```rust
pub struct BoundTextureData<InternalFormat,ElemType> {
    gl_texture : WebGlTexture,
    context    : Context,
    phantom    : PhantomData2<InternalFormat,ElemType>,
}
```

lib/core/src/system/gpu/data/texture.rs
```diff
@@ -427,31 +509,32 @@ pub struct BoundTexture<I,T> {
...
     pub fn init_mock(&self) {
-        let texture         = &self.texture;
         let target          = Context::TEXTURE_2D;
         let level           = 0;
-        let internal_format = texture.gl_internal_format();
-        let format          = texture.gl_format().into();
-        let elem_type       = texture.gl_elem_type();
+        let internal_format = Context::RGBA as i32;
+        let format          = Context::RGBA;
+        let elem_type       = Context::UNSIGNED_BYTE;
```
wdanilo: I don't like it. These parameters should never be provided manually. They should always be infered from our magic table like they were before.

lib/core/src/display/symbol.rs
farmaazon: Ok, I read the documentation of OpenGl, where 16 is a minimum. And I can read MAX_TEXTURE_IMAGE_UNITS, but the question is if we want to make library which may behave differently on user machines?

lib/core/src/display/symbol/material.rs
farmaazon: The coding style guide put the space in its example

lib/core/src/system/gpu/data/texture.rs
farmaazon: Ok, but that is actually a fix for this function. Below you created a RGBA data without checking what the actual format is.

And for me the final solution should be rather, that you don't specify format and elem_type while create UnitTexture.

lib/core/src/system/gpu/data/texture.rs
farmaazon: There was `I` and `T` before, here and in many other places, so I though that is your code style :)

lib/core/src/system/gpu/data/texture.rs
farmaazon: But I dislike idea that the FontRenderInfo know anything about uniforms etc. - and it should be able to extend msdf data, and we don't want to copy them each time the texture is created.

lib/core/src/system/gpu/data/texture.rs
farmaazon: Especially if you want to texture data being owned by texture, the parameter should remain. When I want to extend texture data, I need to specify what the source is (extending texture data is valid only for Memory source), or making ugly pattern match with one valid variant and panics elsewhere.

lib/core/src/system/gpu/data/texture.rs
farmaazon: Ok, the `TextureFromMemory` (and therefore `Texture` itself) could own the data, but not BoundTexture - but that is how it works now.

lib/core/src/display/symbol.rs
```diff
@@ -25,6 +25,7 @@ use crate::system::gpu::shader::Context;
 use crate::system::gpu::data::buffer::IsBuffer;
 use crate::system::gpu::data::uniform::AnyUniform;
 use crate::system::gpu::data::uniform::AnyTextureUniform;
+use crate::system::gpu::data::uniform::AnyTextureUniformOps;
 use crate::system::gpu::data::uniform::AnyPrimUniform;
 use crate::system::gpu::data::uniform::AnyPrimUniformOps;
```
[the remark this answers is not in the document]
farmaazon: I'll fix it, but I admit that for me it not convenient form: 
a) as a code writer: before I could use some structs not knowing where they are, because IDE find them for me. Now I must find, in which _types_ crate they are
b) as a code reader: before I could see in file without IDE support from where the struct/trait was used. Now I don't know from which `types::*` the object comes.

And usually if I write code, I do it in IDE, but often read code somewhere else (e.g. doing a review).

PR #119, opened 2020-01-21
Integration of WebGL and CSS scenes and renderers

notdanilo:
### Pull Request Description
This PR removes duplicated concepts such as hierarchical transforms and cameras of WebGL and CSS3D renderers. Now both features are based on DisplayObject concept.

### Important Notes
- CSS3D renderer was refactored with DisplayObject
- The implementation was reduced to three objects: HTMLScene, HTMLObject and HTMLRenderer.

### Checklist


lib/core/src/display/navigation/navigator.rs
```diff
@@ -80,11 +79,12 @@ impl Navigator {
         let camera_clone         = camera.clone();
         let mut properties_clone = properties.clone();
         let panning_callback     = move |pan: PanEvent| {
+            let y_scale = camera_clone.projection_matrix().m11;
```
wdanilo: This is magic. It needs to be exposed as function from camera. With a good, good documentation.

lib/core/src/display/navigation/navigator.rs
```diff
@@ -95,25 +95,23 @@ impl Navigator {
...
                 let point      = zoom.focus;
                 let normalized = normalize_point2(point, dom_clone.dimensions());
                 let normalized = normalized_to_range2(normalized, -1.0, 1.0);
 
                 // Scale X and Y to compensate aspect and fov.
-                let x                     = -normalized.x * persp.aspect;
+                let x                     = -normalized.x * camera.screen().aspect();
                 let y                     =  normalized.y;
-                let fov_slope             = camera.get_y_scale();
+                let fov_slope             = camera.projection_matrix().m11;
```
wdanilo: magic again :)

lib/core/src/display/camera/camera2d.rs
```diff
@@ -144,12 +149,16 @@ impl Camera2dData {
         let transform_dirty        = TransformDirty2::new(logger.sub("transform_dirty"),());
         let transform_dirty_copy   = transform_dirty.clone();
         let transform              = DisplayObjectData::new(logger);
-        let zoom_uniform           = globals.add_or_panic("zoom",1.0);
+        let zoom_update_callback   = Box::new(|_| ());
```
farmaazon: Consider usage of Option<Box<ZoomUpdateFn>> - that type actually gives a information that the callback is optional, so it's set only in some cases. That's the way @mwu-tow prefer.

lib/core/src/display/navigation/navigator.rs
farmaazon: But the variable name is a documentation here, isn't it?

lib/core/src/display/navigation/navigator.rs
farmaazon: ok, here I don't know whats going on too :)

lib/core/src/system/web/dom/html/html_renderer.rs
```diff
@@ -169,18 +169,20 @@ impl HTMLRenderer {
...
         let half_dim     = self.renderer.container.dimensions() / 2.0;
-        let y_scale      = camera.get_y_scale();
+        let projection   = camera.projection_matrix();
+        let y_scale      = projection.m11;
         let y_scale      = y_scale * half_dim.y;
 
-        match camera.camera_type() {
-            CameraType::Perspective(_) => {
+        match camera.projection() {
+            Projection::Perspective{ .. } => {
```
farmaazon: I don't think there should be spaces in `{}`

lib/core/src/display/navigation/navigator.rs
wdanilo: Nope, the name doesn't tell me anything what it is about. What does `y_scale` mean for the camera?

notdanilo: @wdanilo @farmaazon so... When we get perspective's matrix [1][1], we are getting y field of view slope. I have exposed this method in `Camera2d`:

```rust
    /// Gets y field of view slope.
    pub fn fovy_slope(&self) -> f32 {
        self.projection_matrix().m11
    }
```

lib/core/src/display/navigation/navigator.rs
notdanilo: mathemagics
mathemagics

notdanilo:
@wdanilo @farmaazon the PR is ready for review. The code was fully refactored to use `DisplayObject`s and both `WebGL` and `CSS3D` renderers are ready to be integrated in a single scene.

Before merging I have two considerations left I would like to discuss with you:
1. Should it belong to `basegl::system::web::dom::html`? I think it should be moved to `basegl::display`, but I would like to hear from you where it should be placed in.
2. What do you think about renaming `HTMLScene`, `HTMLObject` and `HTMLRenderer` to `CSS3DScene`, `CSS3DObject` and `CSS3DRenderer`?

lib/core/src/display/world/scene.rs
```diff
@@ -23,9 +23,11 @@ pub struct Scene {
...
 
 impl Scene {
-    pub fn new(logger:Logger, globals:&UniformScope) -> Self {
-        let root   = DisplayObjectData::new(logger.sub("root"));
-        let camera = Camera2d::new(logger.sub("camera"),globals);
+    pub fn new(logger:Logger, width:f32, height:f32, globals:&UniformScope) -> Self {
+        let root    = DisplayObjectData::new(logger.sub("root"));
+        let camera  = Camera2d::new(logger.sub("camera"),width,height);
+        let uniform = globals.add_or_panic("zoom", 1.0);
+        camera.add_zoom_update_callback(move |zoom| uniform.set(zoom));
```
wdanilo: This uniform is not stored, it is destoryed when going out of this scope. The fact that it is working is a bug. You should stroe unifoirm and assume that when it gous out of scope it will be removed.

lib/core/src/system/web/dom.rs
```diff
@@ -1,24 +1,6 @@
...
-
-pub use camera::*;
-pub use transform::*;
-pub use graphics_renderer::*;
-pub use dom_container::*;
-
 #[warn(missing_docs)]
 pub mod html;
+
+pub use basegl_system_web::dom::*;
```
wdanilo: nwline

lib/core/src/system/web/dom/html/html_renderer.rs
```diff
@@ -125,63 +125,62 @@ impl HTMLRendererData {
...
-                js::setup_perspective(&self.data.div, &y_scale.into());
+    pub fn render(&self, camera: &mut Camera2d, scene: &HTMLScene) {
+        camera.update();
+        let trans_cam  = camera.transform().matrix().try_inverse();
+        let trans_cam  = trans_cam.expect("Camera's matrix is not invertible.");
+        let trans_cam  = trans_cam.map(eps);
+        let trans_cam  = invert_y(trans_cam);
+        let half_dim   = self.container.dimensions() / 2.0;
+        let fov_slope  = camera.fovy_slope();
+        let y_scale    = fov_slope * half_dim.y;
```
wdanilo: We need to describe it. It is named here `y_scale` it is passed to a function which argument is named `znear`. So `znear` is the same as `Y-axis` scale of something. I do not understand the logic. What is exactly this value responsible for? I'd be really thankful for an explanation.

lib/core/src/system/web/dom/html/html_scene.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub fn add(&mut self, object: HTMLObject) -> Ix {
+        self.display_object.add_child(&object.display_object);
+        self.objects.insert(object)
+    }
+
+    /// Removes and retrieves a HTMLObject based on the index provided by
+    pub fn remove(&mut self, index: Ix) -> Option<HTMLObject> {
+        let object = self.objects.remove(index);
+        if let Some(object) = &object {
+            self.display_object.remove_child(&object.display_object);
```
wdanilo: If `HtmlScene` is display object, then you should do `self.remove_child`. There should not be need to redirect it to a field. Display objects add thsi method automatically to things which implement some trait - see display object definition for more info,

lib/core/tests/html_renderer.rs
```diff
@@ -187,24 +191,24 @@ mod tests {
...
+        renderer.container().dom.set_property_or_panic("background-color", "black");
 
         for _ in 0..1000 {
-            let mut object = HTMLObject::new("div")
+            let mut object = HTMLObject::new(logger.clone(), "div")
                                     .expect("Failed to create object");
             object.set_dimensions(1.0, 1.0);
-            object.set_scale(0.5, 0.5, 0.5);
+            object.set_scale(Vector3::new(0.5, 0.5, 0.5));
             scene.add(object);
```
wdanilo: The API should be consistent. `Scene` is display object, so adding children should be done by using `add_child`.

lib/system/web/src/dom.rs
```diff
@@ -0,0 +1,3 @@
+mod dom_container;
+
+pub use dom_container::*;
```
wdanilo: newline

lib/core/src/display/world/scene.rs
notdanilo: @wdanilo is it wrong to assume that it will live as long as the closure lives?

lib/core/src/system/web/dom/html/html_renderer.rs
notdanilo: @wdanilo sorry to mix up names. `y_scale` is indeed meaningless. When we multiply `fov_slope` by half the screen height, we get the distance from which the camera can cover all the UI plane. The variable names were fixed.

lib/core/src/display/navigation/navigator.rs
wdanilo: Ok, I can be dumb and just don't understand simple things (if so, I am sorry), but I still do not understand :( Few questions:
1. What is "y field of view slope"? I know what is field of view, but I dont know what is the "slope". Google isnt helpful here.
2. We obtain this "y slope" (whatever it is), why in the code belowe we do `let z = fov_slope;`? How the Y slope is related to the `z` component in the callback?

lib/core/src/display/world/scene.rs
wdanilo: Oh, good point. Hmm, the only problem is that if we will ever replace camera (render from 2 or more cameras), which we will, then we will to handle this in special case and this uniform will cause problems. It will be better to just store it in the scene.

lib/core/tests/html_renderer.rs
notdanilo: As decided on Discord, this will be part of the World integration task.

lib/core/src/system/web/dom/html/html_scene.rs
notdanilo: As decided on Discord, this will be part of the World integration task.

lib/core/src/system/web/dom/html/html_renderer.rs
notdanilo: The trigonometric formula was fixed.

lib/core/src/display/navigation/navigator.rs
notdanilo: I fixed the trigonometric formula.
```
    /|
   / | h
s /__|
   z
```
s = h / z

s = half fov y slope
z = the distance to cover all the screen we want to discover
h = half height

z = h / s

We can use `camera.half_fovy_slope()` which is equivalent to `(camera.fovy() / 2.0).tan()`, to get the slope value.

notdanilo:
@wdanilo I also brought back `ResizeObserver` to `DomContainer` in this PR.

lib/core/src/display/navigation/navigator.rs
```diff
@@ -80,11 +79,12 @@ impl Navigator {
         let camera_clone         = camera.clone();
         let mut properties_clone = properties.clone();
         let panning_callback     = move |pan: PanEvent| {
+            let fovy_slope = camera_clone.half_fovy_slope();
             // base_distance is a distance where the camera covers all the UI.
-            let base_distance = dom_clone.dimensions().y / 2.0 * camera_clone.get_y_scale();
+            let base_distance = dom_clone.dimensions().y / 2.0 / fovy_slope;
```
wdanilo: It looks correct now! :)

lib/core/src/display/navigation/navigator.rs
```diff
@@ -95,25 +95,23 @@ impl Navigator {
...
-                let direction             = Vector3::new(x, y, z).normalize();
-                let mut position          = properties.spring().fixed_point;
-                let zoom_amount           = zoom.amount * position.z;
-                position                 += direction   * zoom_amount;
-                let min_zoom              = persp.near  + min_zoom;
-                position.z                = clamp(position.z, min_zoom, max_zoom);
+                let x            = -normalized.x * camera.screen().aspect();
+                let y            =  normalized.y;
+                let fovy_slope   = camera.half_fovy_slope();
+                let z            = 1.0 / fovy_slope;
```
wdanilo: It makes perfect sense. I would change here few things. Variable name `fovy_slope` -> `half_fovy_slope`. And `let y_unit = 1.0` :)

lib/core/src/display/scene.rs
```diff
@@ -130,7 +125,7 @@ pub struct SceneData {
     pipeline        : RenderPipeline,
     composer        : RenderComposer,
     stats           : Stats,
-
+    uniform         : Uniform<f32>
```
wdanilo: This is a terrible variable name ! It is like a variable named "int" :P

lib/core/src/display/scene.rs
```diff
@@ -150,14 +145,22 @@ impl {
         let sub_logger      = logger.sub("symbols");
         let variables       = UniformScope::new(logger.sub("global_variables"),&context);
         let symbols         = SymbolRegistry::new(&variables,&stats,&context,sub_logger,on_change);
-        let shape           = Shape::default();
+        let window = crate::system::web::window();
+        let inner_width = window.inner_width().unwrap().as_f64().unwrap() as f32;
+        let inner_height = window.inner_height().unwrap().as_f64().unwrap() as f32;
```
wdanilo: spacing

lib/core/tests/html_renderer.rs
```diff
@@ -18,57 +18,59 @@ extern "C" {
...
         // We move the Camera behind the object so we don't see it.
         camera.set_position(Vector3::new(0.0, 0.0, -100.0));
 
-        renderer.render(&mut camera, &scene);
+        renderer.render(&mut camera,&scene);
     }
 
-    fn create_scene(renderer : &HTMLRenderer) -> Scene<HTMLObject> {
-        let mut scene : Scene<HTMLObject> = Scene::new();
+    fn create_scene(logger:Logger, renderer:&HtmlRenderer) -> HtmlScene {
```
wdanilo: You should never require clone on call side. Use here `logger:&Logger` instead.

lib/core/tests/html_renderer.rs
```diff
@@ -18,57 +18,59 @@ extern "C" {
...
+        let mut scene: HtmlScene = HtmlScene::new(logger.clone());
```
wdanilo: this is also clone on call side. The api should take reference to it instead (here and in other objects)

lib/system/web/src/dom/dom_container.rs
```diff
@@ -1,16 +1,17 @@
...
+use basegl_prelude::*;
 
-use crate::system::web::get_element_by_id;
-use crate::system::web::dyn_into;
-use crate::system::web::Result;
-use crate::system::web::StyleSetter;
-use crate::system::web::intersection_observer::IntersectionObserver;
+use crate::get_element_by_id;
+use crate::dyn_into;
+use crate::Result;
```
wdanilo: What it is? Where it is defined?

lib/core/src/display/world.rs
```diff
@@ -267,6 +266,14 @@ impl World {
         self.rc.borrow_mut().run();
     }
 
+    pub fn event_loop(&self) -> EventLoop {
+        self.rc.borrow().event_loop.clone()
+    }
+
+    pub fn scene<F:FnMut(&Scene)>(&self, mut f:F) {
+        f(&self.rc.borrow().scene)
+    }
```
wdanilo: This is very unsafe function, as it is easy to mutable borrow mutliple times. `Scene` is a shared reference object, create instead function `scene(&self) -> Scene`.

lib/core/src/examples/shapes.rs
```diff
@@ -24,6 +26,24 @@ pub fn run_example_shapes() {
 }
 
 fn init(world: &World) {
+    let container      = DomContainer::from_id("app").expect("Couldn't get container");
+    let mut event_loop = world.event_loop();
+    let mut camera     = None;
+    world.scene(|scene| camera = Some(scene.camera()));
```
wdanilo: This can be nicer - after applying comments above, it will be.

lib/core/src/examples/shapes.rs
```diff
@@ -24,6 +26,24 @@ pub fn run_example_shapes() {
 }
 
 fn init(world: &World) {
+    let container      = DomContainer::from_id("app").expect("Couldn't get container");
```
wdanilo: I don't like that we create this manually. It should be hidden. Scene has a reference to dom object. Maybe it should use `DomContainer` internally. Anyway, user-facing code of the lib should not require this line.

lib/core/src/examples/shapes.rs
```diff
@@ -24,6 +26,24 @@ pub fn run_example_shapes() {
...
+    world.scene(|scene| camera = Some(scene.camera()));
+    let camera     = camera.unwrap();
+    camera.update();
+
+    let screen = camera.screen();
+    let fovy_slope = camera.half_fovy_slope();
+    let x = 0.0;
+    let y = 0.0;
+    let z = screen.height / 2.0 / fovy_slope;
+    camera.set_position(Vector3::new(x, y, z));
```
wdanilo: This file should show how users of basegl are meant to use the library. Requiring the user to write such boileprlate code is a bad design. This code should not be here. The code should be as simple as the threejs API. Moreover, attaching a navigator should NOT move the camera. Attaching navigator should keep camera where it was. The whole code should look like this here:

```
fn init(world: &World) {
    let camera = world.scene().camera();
    let navigator = Navigator::new(&camera, &scene).ok();

    let s1 = Circle("25.0 + 20.0*sin(input_time/1000.0)");
    ...
```

lib/core/src/examples/shapes.rs
```diff
@@ -24,6 +26,24 @@ pub fn run_example_shapes() {
...
+
+    let navigator = Navigator::new(&mut event_loop, &container, camera);
+    let navigator = navigator.expect("Couldn't create navigator");
+    std::mem::forget(navigator);
```
wdanilo: Please, do not use `mem::forget` in the codebase unless you really need to. Here we can do fine without it.

lib/system/web/src/dom/dom_container.rs
notdanilo: @wdanilo please note that I moved `dom_container` to `system/web` crate.

lib/core/tests/html_renderer.rs
notdanilo: Fixing

lib/core/tests/html_renderer.rs
notdanilo: Fixing

lib/core/src/examples/shapes.rs
notdanilo: @wdanilo wait, but this is the initialization of the camera position for the example. We have a different initialization for `camera_navigation.rs`.

We can of course define the initial position of the camera. It can be initialized to `vec3(width/2,height/2,height/2/fovy_slope)` in `Scene::new`

wdanilo: Camera was always initialized automatically in these examples, why we need to initialize it by hand now?

lib/core/src/display/camera/camera2d.rs
```diff
@@ -144,27 +184,57 @@ impl Camera2dData {
...
+            clipping,
+            alignment,
+            zoom,
+            zoom_update_callback,
+            native_z,
+            view_matrix,
+            projection_matrix,
+            view_projection_matrix,
+            projection_dirty,
+            transform_dirty
```
wdanilo: Why this change? It takes sooo much space now. Just keep it in 2 lines please :)

lib/core/src/display/scene.rs
```diff
@@ -150,14 +145,22 @@ impl {
         let sub_logger      = logger.sub("symbols");
         let variables       = UniformScope::new(logger.sub("global_variables"),&context);
         let symbols         = SymbolRegistry::new(&variables,&stats,&context,sub_logger,on_change);
-        let shape           = Shape::default();
+        let window          = crate::system::web::window();
+        let inner_width     = window.inner_width().unwrap().as_f64().unwrap() as f32;
+        let inner_height    = window.inner_height().unwrap().as_f64().unwrap() as f32;
+        let shape           = Shape::new(inner_width, inner_height);
```
wdanilo: This is shape initializer. It should be refactored to something like `Shape::from_window()` or something. This logic should not be in scene constructor.

lib/core/src/display/scene.rs
```diff
@@ -150,14 +145,22 @@ impl {
...
         let listeners       = Self::init_listeners(&logger,&canvas,&shape,&shape_dirty);
         let symbols_dirty   = dirty_flag;
-        let camera          = Camera2d::new(logger.sub("camera"),&variables);
+        let camera          = Camera2d::new(logger.sub("camera"),width,height);
+        let zoom_uniform    = variables.add_or_panic("zoom", 1.0);
         let text_components = default();
         let on_resize       = default();
         let stats           = stats.clone();
-
+        let uniform         = zoom_uniform.clone();
```
wdanilo: I added here a comment the previous time. This variable name is terrible.

lib/core/src/display/scene.rs
```diff
@@ -176,7 +179,11 @@ impl {
         let composer = RenderComposer::new(&pipeline,&context,&variables,width,height);
 
         Self {pipeline,composer,root,canvas,context,symbols,camera,symbols_dirty,shape,shape_dirty,logger
-             ,listeners,variables,on_resize,text_components,stats}
+             ,listeners,variables,on_resize,text_components,stats,zoom_uniform}
+    }
+
+    pub fn canvas(&self) -> web_sys::HtmlCanvasElement {
+        self.canvas.clone()
```
wdanilo: use `clone_ref` to indicate that it is a cheap clone.

lib/core/src/display/scene.rs
```diff
@@ -232,6 +239,10 @@ impl {
         }
     }
 
+    pub fn camera(&self) -> Camera2d {
+        self.camera.clone()
```
wdanilo: use `clone_ref` to indicate that it is a cheap clone.

lib/core/src/system/web/dom/html/html_scene.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use super::HtmlObject;
+use crate::display::object::DisplayObjectData;
+use data::opt_vec::*;
+
+
+
+// =============
+// === HtmlScene ===
+// =============
```
wdanilo: Minor fixes needed :)

lib/core/src/display/camera/camera2d.rs
notdanilo: It's 3 lines now :P But yup

lib/core/src/display/scene.rs
notdanilo: Fixed

lib/core/src/display/scene.rs
notdanilo: There is no `clone_ref` in `HtmlCanvasElement`, but I am changing it everywhere else when possible

lib/core/src/system/web/dom/html/html_scene.rs
notdanilo: @wdanilo what needs to be fixed here?

PR #129, opened 2020-01-30
Removing global state from world, allowing creating multiple world in…

wdanilo:
### Pull Request Description
- Removes a bad design we did in the past.

### Important Notes
- Now creation of all symbols requires a `&World` argument, which is ugly but necessary.

### Checklist

### DIFF
--- a/lib/core/src/display/world.rs
+++ b/lib/core/src/display/world.rs
@@ -33,39 +33,6 @@ pub use stats::*;
 
 
 
-// ===========================
-// === UNSAFE GLOBAL STATE ===
-// ===========================
-
-/// The following code is very unsafe and should disappear in the future. We are using it in order
-/// to keep the API simple and future-proof. Each `Stage` is associated with HTML Canvas element,
-/// and we currently support a single stage only. Although this is very low-priority task, as we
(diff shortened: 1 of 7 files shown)


lib/core/src/display/world.rs
```diff
@@ -226,7 +193,7 @@ impl World {
     pub fn new(world_data: WorldData) -> Self {
         let rc = Rc::new(RefCell::new(world_data));
         let out = Self {rc};
-        init_global_variables(&out);
+//        init_global_variables(&out);
```
farmaazon: So, if we resign from global variables, I would remove this line instead of commenting it

lib/core/src/examples/shapes.rs
```diff
@@ -39,7 +39,7 @@ fn init(world: &World) {
     let s2 = s1.translate(25.0,0.0);
     let s3 = &s1 + &s2;
 
-    let shape_system = ShapeSystem::new(&s3);
+    let shape_system = ShapeSystem::new(world, &s3);
```
farmaazon: Usually we don't put space after comma in this case, do we?

PR #132, opened 2020-01-30
New Text Component

farmaazon:
### Pull Request Description
Refactored TextComponent to use the sprites and glyph system, and renamed it to TextField. Also do much refactoring stuff and add missing documentation.

### Checklist


lib/core/src/display/shape/glyph/system.rs
```diff
@@ -92,8 +92,9 @@ impl Line {
     /// The replacing strings will reuse glyphs which increases performance of rendering text.
     pub fn replace_text<Chars>(&mut self, chars:Chars, fonts:&mut FontRegistry)
     where Chars : Iterator<Item=char> + Clone {
-        let font = fonts.get_render_info(self.font_id);
-        let pen  = PenIterator::new(self.baseline_start,self.height,chars.clone(),font);
+        let font        = fonts.get_render_info(self.font_id);
+        let chars_count = chars.clone().count();
```
wdanilo: What is chars and how costly is the clone here?

lib/core/src/display/shape/glyph/system.rs
```diff
@@ -108,10 +109,36 @@ impl Line {
...
+impl Line {
+    /// Baseline starting point.
+    pub fn baseline_start(&self) -> &Vector2<f32> { &self.baseline_start }
+    /// Line's height.
+    pub fn height        (&self) -> f32           { self.height          }
+    /// Number of glyphs, giving the maximum length of displayed line.
+    pub fn length        (&self) -> usize         { self.glyphs.len()    }
+    /// Font id.
+    pub fn font_id       (&self) -> FontId        { self.font_id         }
+}
```
wdanilo: These docs do not introduce any information. `Line.height` is obviously a line's height. In suhc case I would use `allow(missing_docs)`.

lib/core/src/display/shape/glyph/system.rs
```diff
@@ -219,6 +251,7 @@ impl GlyphSystem {
         material.add_input("zoom"       , 1.0);
         material.add_input("msdf_range" , FontRenderInfo::MSDF_PARAMS.range as f32);
         material.add_input("color"      , Vector4::new(0.0,0.0,0.0,1.0));
+        material.add_output("id", Vector4::<u32>::new(0,0,0,0));
```
wdanilo: Add comment to remove it in the future pls.

lib/core/src/display/shape/text/content.rs
```diff
@@ -1,28 +1,33 @@
...
 // ==================
 
 /// Set of dirty lines' indices
 #[derive(Debug)]
 pub struct DirtyLines {
-    pub single_lines : HashSet<usize>,
-    pub range        : Option<RangeFrom<usize>>
+    /// A set of single edited lines.
+    pub single_lines: HashSet<usize>,
+    /// An open range of dirty lines. When the line is added or removed, all the lines below.
```
wdanilo: I dont understand this sentence:
```When the line is added or removed, all the lines below.```

File: lib/core/src/display/shape/text/content.rs [hunk 5]
```diff
@@ -68,17 +73,19 @@ impl DirtyLines {
...
 /// A change type
-///
-/// A change is simple if it's replace a fragment of one line with text without new lines. Otherwise
-/// its a multiline change.
-#[derive(Clone,Copy,Debug)]
+#[derive(Copy,Clone,Debug)]
 pub enum ChangeType {
-    Simple, Multiline
+    /// A simple change fragment of one line with text without new lines.
+    Simple,
```
wdanilo: In such case I'd prefer `SingleLine` name.

lib/core/src/display/shape/text/content/line.rs
```diff
@@ -62,33 +68,61 @@ impl Line {
         self.char_x_positions.clear();
         &mut self.chars
     }
+}
+
+
+
+// ======================
+// === Line full info ===
```
wdanilo: Very minor one. In English titles all words are capitalized.

lib/core/src/display/shape/text/cursor.rs
```diff
@@ -230,62 +238,44 @@ impl<'a,'b> CursorNavigation<'a,'b> {
...
-        self.dirty   = true;
+    }
+
+    /// Remove all cursors except the active one.
+    pub fn remove_additional_cursors(&mut self) {
+        self.cursors.drain(0..self.cursors.len()-1);
+    }
+
+    /// Return the active (last added) cursor.
+    pub fn active_cursor(&self) -> &Cursor {
```
wdanilo: If I understand correctly, all cursors are "active", not only the last one, right? If so, the naming is off.

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
+//! Rendering TextField.
```
wdanilo: If this is textifield impl, then the module should reflect that name.

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::display::shape::primitive::system::ShapeSystem;
+use crate::display::symbol::geometry::compound::sprite::Sprite;
+
+use nalgebra::Vector2;
+use nalgebra::Vector3;
+use crate::display::shape::text::TextFieldProperties;
+use crate::display::world::World;
+
+
+// =======================
```
wdanilo: spacing

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::display::shape::text::render::assignment::LineFragment;
+use crate::display::shape::text::render::selection::SelectionSpritesGenerator;
+use crate::display::shape::primitive::def::*;
+use crate::display::shape::primitive::system::ShapeSystem;
+use crate::display::symbol::geometry::compound::sprite::Sprite;
+
+use nalgebra::Vector2;
+use nalgebra::Vector3;
+use crate::display::shape::text::TextFieldProperties;
+use crate::display::world::World;
```
wdanilo: this should belong to previous group of imports

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+
+
+// =======================
+// === RenderedContent ===
+// =======================
+
+type GlyphLine = crate::display::shape::glyph::system::Line;
```
wdanilo: This is a line of TextField, right? If so, why its name is not just `Line` ? Being user of the lib, I would like to have `text_field::Line` type instead of `GlyphLine`, unless there are other lines than glyph lines.

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// Structure containing sprites bound to one cursor with its selection.
+#[derive(Debug)]
+pub struct CursorSprites {
```
wdanilo: The name CursorSprites do not mention selection. It is misleading imo. I know this is not very helpful, I cannot find better name now, but this is not the best one.

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// Cursor sprite.
+    pub cursor: Sprite,
+    /// Selection sprites.
+    pub selection: Vec<Sprite>,
+}
+
+/// Structure with all data and sprites required for rendering specific TextField.
+#[derive(Debug)]
+pub struct RenderedContent {
```
wdanilo: The docs tell that this is struct with data required FOR RENDERING (in the future), while the name of the struct tells it was already RENDERED (in the past).

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// System used for rendering glyphs.
+    pub glyph_system: GlyphSystem,
+    /// System used for rendering cursors.
+    pub cursor_system: ShapeSystem,
+    /// System used for rendering selections.
+    pub selection_system: ShapeSystem,
+    /// All drawn glyph lines.
+    pub glyph_lines: Vec<GlyphLine>,
```
wdanilo: Ok, after reading this line I do not know what is GlyphLine anymore. IT should have docs (even as private type).

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+                let line          = &content.lines[fragment.line_index];
+                let chars         = &line.chars()[fragment.chars_range.clone()];
+                glyph_line.set_baseline_start(bsl_start);
+                glyph_line.replace_text(chars.iter().cloned(),fonts);
+            }
+        }
+        self.display_object.update();
+    }
+
+    /// Update all displayed cursors with their selections.
```
wdanilo: What does "update" mean? This docs needs to be more explicite.

lib/core/src/display/shape/text/render.rs
```diff
@@ -0,0 +1,202 @@
...
+            let line_height   = self.line_height;
+            let system        = &self.selection_system;
+            let mut generator = SelectionSpritesGenerator {content,line_height,system};
+            sprites.selection.clear();
+            sprites.selection = generator.generate(&selection);
+        }
+        self.display_object.update();
+    }
+
+    fn baseline_start_for_fragment(fragment:&LineFragment, content:&mut TextFieldContentFullInfo)
```
wdanilo: Needs a doc. What does it mean? Especially what does "LineFragment" mean in this context? I know I can jump here . aand there to leanr more, but small docs with explanation would be better here :)

lib/core/src/display/shape/text/content.rs
farmaazon: certainly, I've eaten some words from it

lib/core/src/display/shape/text/cursor.rs
farmaazon: That depends. E.g when you have multiline cursor in Intellij, you edit using all the cursors, but only one is active (this one which highlights the line). Only the active cursor scrolls text on moving, and only this cursor's selection is extended on shift+alt+mouse click

lib/core/src/display/shape/glyph/system.rs
farmaazon: This is iterator over &str or Vec<char>. Cloning should be just a copying reference and length

lib/core/src/display/shape/text/render.rs
farmaazon: But this structure is not for users. The users see the `Line` struct which is part of `TextFieldContent`

lib/core/src/display/shape/text/render.rs
farmaazon: I assumend everywhere, that every cursor has possibly selection bound to it. See `Cursor` struct for instance

lib/core/src/display/shape/text/render.rs
farmaazon: Added docs for this alias anyway.

lib/core/src/display/shape/text/cursor.rs
farmaazon: Added docs for explaining that

lib/core/src/display/shape/glyph/system.rs
farmaazon: I remove some, and some I extended to provide more info.

lib/core/src/display/shape/text/content.rs
wdanilo: omnomnom :D

PR #148, opened 2020-02-03
File size checking 📏

notdanilo:
### Pull Request Description
This PR adds file size checking to our Github Actions workflow.

### Important Notes
- The current maximum file size is set to 2.1mb.

### Checklist


.github/workflows/build.yml
```diff
@@ -14,16 +14,25 @@ jobs:
...
       - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
       - name: Install Rust Nightly Toolchain
         uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly-2019-11-04
           override: true
-      - name: Run cargo check
+      - name: Install binaryen
+        run: npm install binaryen -g
```
wdanilo: I would like to move this responsibility to the `minimize_wasm.sh` script. It should install binaryen if it was not installed (of course not globally) and then run it. `minimize_wasm` is meant to be useful not only on CI, it sohuld work locally, and requiring binaryen to be installed locally or on CI is a bad assumption. So I believe this step should not be a CI step - it should be done in the script.

script/assert_max_wasm_size.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+echoerr() { echo "$@" 1>&2; }
+
+current_size=$(cat file_size)
+current_size="${current_size:0:$((${#current_size} - 1))}"
+
+max_size=2.1 # MB
+echo "Current size: ${current_size}M. expected maximum size: ${max_size}M"
```
wdanilo: Does this line print `"Current size: 3M. expected maximum size: 2M" ? If so then:

1. These are not valid sentences. Please, take care of the quality of messages.
2. "M" is not a proper abbreviation for Megabyte.

script/assert_max_wasm_size.sh
```diff
@@ -0,0 +1,16 @@
...
+if (( $(echo "$current_size <= $max_size" |bc -l) ));
+then
+  echo OK
+else
+  echoerr FAIL
+  exit 1
+fi
```
wdanilo: nl

script/assert_max_wasm_size.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+echoerr() { echo "$@" 1>&2; }
+
+current_size=$(cat file_size)
+current_size="${current_size:0:$((${#current_size} - 1))}"
```
wdanilo: What does this line do? Bash is magic, could you please refactor it to more variables please? By the way, I find bash super cool for 1-line scripts, but even with 10-lines scripts it starts to be unreadable. Maybe we should create bigger (bigger means more than 3 lines :P) scripts in python or Rust instead? https://github.com/DanielKeep/cargo-script

wdanilo:
Oh by the way, you havent removed the size checking from webpack in this PR. It should be removed as this PR introduces a much much better way to do it :)

script/assert_max_wasm_size.sh
notdanilo: This is the format returned by `du -h`. I will rename it to `MiB`.

.github/workflows/build.yml
notdanilo: I just moved the responsibility for installing binaryen to `minimize_wasm.sh`

script/assert_max_wasm_size.sh
notdanilo: It slices out the trailing M in 123.4M. I think this should be clear enough:

```
current_size=$(cat file_size)
current_size_len=$((${#current_size} - 1))
current_size="${current_size:0:current_size_len}"
```

But I will keep in mind cargo-script for future cases.

.github/workflows/build.yml
```diff
@@ -23,7 +23,6 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
-
```
wdanilo: Sometimes we leave newline between different actions sometimes not. Let's make it everywhere (newline between each action) please :)

.github/workflows/build.yml
```diff
@@ -56,22 +55,42 @@ jobs:
...
+  #      - uses: actions/checkout@v1
+  #      - uses: actions-rs/toolchain@v1
+  #        with:
+  #          toolchain: nightly-2019-11-04
+  #          override: true
+  #      - run: rustup component add rustfmt
+  #      - uses: actions-rs/cargo@v1
+  #        with:
+  #          command: fmt
+  #          args: --all -- --check
```
wdanilo: Lets remove it. We know now we will not be using it.

.github/workflows/build.yml
```diff
@@ -56,22 +55,42 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2019-11-04
+          override: true
+      - name: Install wasm-pack
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: wasm-pack --version 0.8.1
+      - name: Compile WASM
```
wdanilo: Strictly speaking it is not "Compile WASM" , this is "Compile to WASM", or "Compile Rust to WASM". I know the difference is subtle, but lets make it correct.

PR #156, opened 2020-02-06
Utilities for safe usage of handles with RefCell values.

mwu-tow:
### Pull Request Description
This PR provides `utilities` crate with a new `handles` module. 
It provides the general-purpose code that enables safer usage patterns of RefCell-based shared state, leveraging the asynchronous execution environment.
Please refer to the documentation for the details.

### Important Notes
Kudos to @farmaazon for coming up with this pattern while discussing #151.

### Checklist


lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
+//! This module implements utilities that allow for a safer usage pattern of
```
wdanilo: This file path does not tell what it is about. It should be `lib/ide/utils/src/cell/handle.rs`

lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
...
+
+
+// ================
+// == WeakHandle ==
+// ================
+
+/// A type that can provide `Weak` handle to a value under `RefCell`.
+pub trait IsWeakHandle: Clone {
+    /// Type of the data stored.
+    type Data;
```
wdanilo: This is a bad pattern. Basically the most composable way to do it is to create a separate trait only for type dependencies. An example of what it leads to is here. Right now I can have a polymorphic function `foo<T>(t:T)` and I want it to return its "Data" no matter if it was weak or strong handle. I cannot do it simply. Instead you should do:

```rust
pub trait HasContent {
    type Content;
}

pub trait IsWeakHandle : Clone + HasContent { ... }
```

And FYI, HasContent is already defined in Prelude :)

Moreover, the implementations would so simplify using this pattern . -we can then use default impls in an easy way.

lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
...
+
+// ==============
+// == WeakWith ==
+// ==============
+
+/// Future that shall call `cb` with the data under `handle`.
+#[derive(Debug)]
+pub struct WeakWith<Handle,Cb> {
+    handle : Handle,
+    cb     : Option<Cb>,
```
wdanilo: This sohuld be named `callback`. Its not an internal variable inside of a function. It is even mentioned in documentation.

lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+impl<Handle,Cb> StrongWith<Handle,Cb> {
+    /// Create new `StrongWith` value.
+    pub fn new<Data,R>
+    ( handle : Handle
+    , cb     : Cb)
+    -> StrongWith<Handle,Cb>
```
wdanilo: this should go to single line

lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
...
+impl<Handle,Cb> WeakWith<Handle,Cb> {
+    /// Creates the value.
+    pub fn new<Data,R>
+    ( handle : Handle
+    , cb     : Cb)
+    -> WeakWith<Handle,Cb>
+    where Handle : IsWeakHandle<Data = Data>,
+          Cb     : FnOnce(&mut Data) -> R, {
+        WeakWith {handle,cb:Some(cb)}
+    }
```
wdanilo: This definition makes the type inference more complex than it sohuld be. Moreover, it is hard to read `Data = Data`. Why not (assuming using `HasContent` as proposed above):
```rust
pub fn new<T> (handle:Handle, callback:Callback) -> WeakWith<Handle,Callback>
where Handle   : IsWeakHandle,
      Callback : FnOnce(&mut Content<Handle>) -> R {
    WeakWith {handle,callback:Some(callback)}
}
```

lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
...
+
+/// Packs given value into a Rc<RefCell<T>>.
+pub fn strong<T>(t:T) -> StrongHandle<T> { Rc::new(RefCell::new(t)) }
+
+/// A strong (shared ownership) handle to some internally mutable data.
+pub type StrongHandle<T> = Rc<RefCell<T>>;
+
+/// A weak handle to some internally mutable data.
+pub type WeakHandle<T> = Weak<RefCell<T>>;
+
```
wdanilo: If this is meant to be a safe interface I dont like the idea to still allow me to `borrow_mut` things here. Why not newtypes?

farmaazon: +1

lib/ide/utils/src/handle.rs
farmaazon: Could you provide better example than `foo<T>(t:T)`
I cannot figure out what can be an example of function which operate on trait with near-to-none semantic value.

lib/ide/utils/src/handle.rs
farmaazon: For me, the @mwu-tow signature is better, because it's pattern common in Rust (like all the operations on Iterators are defined)
Yeah, `Data = Data` is a little verbose. It is a drawback? For me it's not.

lib/ide/utils/src/handle.rs
mwu-tow: i went with just `cell`, so we have cell::WeakHandle (no handle::Handle)

lib/ide/utils/src/handle.rs
mwu-tow: done

lib/ide/utils/src/handle.rs
mwu-tow: Superseded by other suggested changes, there are no traits anymore.

lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Obtain strong (`Rc`) handle to the data.
+    fn strong_handle(&self) -> StrongHandle<Self::Data>;
+
+    /// Obtain weak handle to the data.
+    fn downgrade(&self) -> WeakHandle<Self::Data> {
+        Rc::downgrade(&self.strong_handle())
+    }
+
+    /// Returns `Future` that shall call `Cb` with the access to mutably
+    /// borrowed data.
```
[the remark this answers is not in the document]
mwu-tow: Done

lib/ide/utils/src/handle.rs
mwu-tow: Done

lib/ide/utils/src/handle.rs
mwu-tow: simplified, as the `new` does not really care about Callback type, only the `Future` impl

mwu-tow: `Data = Data` looks funny but imho it's clear once you saw it once — I thought about renaming one of them… but then why should we came up with different names for the very same entity.

(but code is gone anyway)

lib/ide/utils/src/handle.rs
mwu-tow: done

lib/ide/utils/src/handle.rs
```diff
@@ -0,0 +1,268 @@
...
+            let unboxed = try_load(&data2).await.unwrap();
+            assert_eq!(unboxed, INITIAL_VALUE);
+            const NEW_VALUE: i32 = INITIAL_VALUE + 1;
+            try_store(&data2, NEW_VALUE).await.unwrap();
+            let unboxed = try_load(&data2).await.unwrap();
+            assert_eq!(unboxed, NEW_VALUE);
+        };
+        block_on(fut);
+    }
+}
```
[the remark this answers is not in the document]
mwu-tow: done

PR #157, opened 2020-02-06
[wip] Controller Stubs

mwu-tow:
### Pull Request Description
This PR implements sample project, module and text controller stubs.

Please note that this code purpose is display a possible organization of controllers and their possible communication patterns, particularly including the safe refcell handles. While it comes with many methods, actually implementing all the controllers is covered by other tasks.

That's why many implementations contains `todo!` blocks and comments.

Also, a WIP implementation for JS-event loop-based executor is provided. It will be completed later during this milestone, but now it might prove useful for testing, so I upload it as well under `todo` module.

### Important Notes
The relevant changes are `controller.rs`.
Rest are helpers, also contents of #156 are provisionally included to keep the code compiling.

The naming patterns and organization shall be subjected to the future bikeshedding. 
Currently each controller is in its own module in `controllers.rs` and they follow a scheme where each controller has a struct named `Data` that stores its state and `WeakHandle` and `StrongHandle` to keep a reference to it. Names might be changed.

As the "real" implementations of the controllers appear, I expect they should be moved to the files of their own.

Similarly, I expect the project position in the repository to be changed as well, as `ide/ide` looks bad, as @notdanilo pointed out. However I did not want to delay this PR just to figure out the repo reorganization.

### Checklist


.gitignore
```diff
@@ -15,3 +15,6 @@ wasm-pack.log
 
 # A thirdparty downloaded js
 /lib/core/msdf-sys/msdfgen_wasm.js
+
+# npm
+node_modules
```
farmaazon: newline

lib/ide/ide/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+  'CloseEvent',
+  'Document',
+  'Element',
+  "ErrorEvent",
+  "MessageEvent",
+  'HtmlElement',
+  'Node',
+  'WebSocket',
+  'Window',
+]
```
farmaazon: I wonder if these changes does not go to conflict with @notdanilo work. Please make sure you are working on the same ide lib.

lib/ide/ide/src/controller.rs
```diff
@@ -0,0 +1,377 @@
...
+use std::collections::HashMap;
+use futures::task::LocalSpawnExt;
+
+use utils::handle::*;
+use utils::handle;
+
+pub type FallibleResult<T> = std::result::Result<T,failure::Error>;
+
+/// Helper that generates wrapper tuple struct around Weak<RefCell<Data>>.
+macro_rules! make_weak_handle {
```
farmaazon: Why we cannot just use "shrinkwrap" in these structures?

lib/ide/ide/src/controller.rs
```diff
@@ -0,0 +1,377 @@
...
+            }
+        }
+        /// Stores given module controller handle and returns it.
+        ///
+        /// Note: handle stored in the project controller is weak, so the caller
+        /// remains responsible for managing the module's controller lifetime.
+        pub fn insert_module(&mut self, data:module::Data) -> module::StrongHandle {
+            let path   = data.path.clone();
+            let module = strong(data);
+            // TODO what if already exists?
```
farmaazon: Then return the existing module. 

This is why we are playing with Rc handles here.

lib/ide/ide/src/controller.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    #[derive(Clone,Debug)]
+    pub struct Data {
+        pub path     : Location,
+        pub contents : String,
+        pub parent   : project::WeakHandle,
+    }
+
+    impl Data {
+        pub fn fetch_text(&self) -> impl Future<Output = FallibleResult<String>> {
```
farmaazon: Why this is not just an async function?

lib/ide/ide/src/controller.rs
```diff
@@ -0,0 +1,377 @@
...
+        pub async fn fetch_text(&self) -> FallibleResult<String> {
+            self.with(|data| data.fetch_text()).await?.await
+        }
+    }
+}
+
+mod text {
+    use super::*;
+
+    pub type StrongHandle = handle::StrongHandle<Data>;
```
farmaazon: I'd like to name it "TextControllerHandle"

lib/ide/ide/src/controller.rs
```diff
@@ -0,0 +1,377 @@
...
+                Event::Notification(n) => match n {
+                    FilesystemEvent(e) => Some(module::Location(e.path.0.clone()))
+                }
+            }
+        }
+    }
+
+    make_weak_handle!(WeakHandle,Data);
+
+    impl WeakHandle {
```
farmaazon: I think operating on Strong handles is more common, so I would implement all the logic on strong handles and having the weak handle we can just upgrade it.

lib/ide/ide/src/controller.rs
```diff
@@ -0,0 +1,377 @@
...
+        pub contents : String,
+        pub parent   : project::WeakHandle,
+    }
+
+    impl Data {
+        pub fn fetch_text(&self) -> impl Future<Output = FallibleResult<String>> {
+            let loc    = self.path.clone();
+            let parent = self.parent.clone();
+            async move {
+                parent.read_module(loc).await
```
farmaazon: As you know, I don't like refering to the parent, mainly because now everyone who calls `fetch_text` should be prepared for an error "Parent does not exists" which never occurs.

I think the module controller should have an access to FileManagerClient and ask him for file content directly.

notdanilo:
Just a small comment as I haven't reviewed it file by file:

Can we organize the file structure in `lib/ide` like in `lib/core`? Using `lib/ide/ide` is somewhat weird.

lib/ide/ide/src/controller.rs
```diff
@@ -0,0 +1,426 @@
...
+///
+/// This allows treating handles as separate types and fitting them with impl
+/// methods of their own. While not necessary, such implementation may allow
+/// hiding from user gritty details of `with` usage behind nice, easy API.
+macro_rules! make_handles {
+    ($data_type:ty) => {
+        /// newtype wrapper over StrongHandle.
+        #[derive(Shrinkwrap)]
+        #[derive(Clone,Debug)]
+        pub struct StrongHandle(pub utils::cell::StrongHandle<$data_type>);
```
farmaazon: I propose rename it for just `Handle` - the strong version should be the mainly used, so I don't need to see everywhere that handle is indeed strong.

lib/ide/ide/Cargo.toml
notdanilo: It doesn't conflict. I am actually merging this branch into my working branch :)

mwu-tow:
Closing. The stubs will be merged as aprt of #159, however the approach tested here will not be widely adopted.
All stubs will be gradually replaced with a proper implementations, as covered by their relevant tasks,

PR #165, opened 2020-02-10
Fix renderer & navigator resizing

notdanilo:
### Pull Request Description
This PR fixes two issues:

1. Css3dRenderer wasn't resizing like WebGL renderer.
2. Navigator wasn't updating physics properties on resize.

### Important Notes
This had to be fixed before delivering the initial project view because the Css3dRenderer resizing logic was incomplete.

### Checklist


lib/core/src/display/navigation/navigator.rs
```diff
@@ -62,9 +65,21 @@ impl Navigator {
...
+            let fixed_point       = properties.spring().fixed_point;
+            let position          = properties.kinematics().position();
+            let distance          = (position - fixed_point).magnitude();
+            const THRESHOLD : f32 = 1.0;
+            let position = if distance < THRESHOLD {
+                fixed_point
+            } else {
+                interpolated_position
+            };
+            if camera.transform().position() != position {
```
wdanilo: Why moving camera has some internal `if` ? What this code is about? And what is this threshold? Basically:

I believe that moving a camera should not need branching in logic - if I'm wrong, I'd love to talk about why and then the code needs to be described better because now it's not clear.

lib/core/src/display/navigation/navigator.rs
```diff
@@ -90,10 +104,22 @@ impl Navigator {
...
-
-        let dom_clone = dom.clone();
+            properties.mod_spring(|spring| { spring.fixed_point += Vector3::new(x, y, z); });
+        });
+
+        let mut dom_clone = dom.clone();
+        dom_clone.add_resize_callback(
+            enclose!((camera,mut properties) move |dimensions:&Vector2<f32>| {
+                camera.set_screen(dimensions.x, dimensions.y);
+                camera.update();
```
wdanilo: Camera `set_screen` and update is already handled by the scene. Scene is the owner of the attached camera. This is not resposnibility of Navigator!

lib/core/src/display/navigation/navigator.rs
```diff
@@ -90,10 +104,22 @@ impl Navigator {
...
+                let position = camera.transform().position();
+                properties.mod_kinematics(|kinematics| {
+                    kinematics.set_position(position);
+                    kinematics.set_velocity(Vector3::new(0.0, 0.0, 0.0));
+                });
+                properties.mod_spring(|spring| spring.fixed_point = position);
+            })
+        );
```
wdanilo: Do I understand correctly that after resize the movement of the navigator is stopped? If so, its not bad, but is this the only solution here?

lib/core/src/display/navigation/navigator.rs
notdanilo: @wdanilo yes, the logic is duplicated. But when this callback is ran, camera screen isn't updated, we need to be sure to update it.

lib/core/src/display/navigation/navigator.rs
notdanilo: @wdanilo there is a good reason for that: we want to avoid the camera to get updated on every frame if its position hasn't changed. We use the threshold to move the camera directly to its final destination (the spring's fixed point), because the position coming from the simulator is unstable, so we can't simply use this comparison `camera.transform().position() != position`.

This is important because we want to stop the simulation by setting the kinematics position and the spring fixed point from the camera position after `set_screen`.

lib/core/src/display/navigation/navigator.rs
notdanilo: To sum up my previous comments: yes, that's the solution. You can see there is a small amount of code change in this PR, but I actually spend lots of hours testing it. This is not a workaround, because it considers how the current architecture works. So this is the best solution I got without making a major architectural change.

lib/core/src/display/navigation/navigator.rs
wdanilo: Then there is something wrong with responsibility. We should never duplicate such logic, because tracking what is happening is really hard. Camera is updated by scene. Scene tracks resize events of the dom. So maybe scene should emit onresize events instead and you should listen to them instead fo directly on dom ?

lib/core/src/display/navigation/navigator.rs
wdanilo: @notdanilo how the position coming from simulator could be unstable? 2 follow up questions:
1. If the position coming from simulator is unstable, who's responsibility is it to fix it? Camera?
2. Why it was earlier working, and now it is not? What made the simulator unstable now?

lib/core/src/display/navigation/navigator.rs
wdanilo: I know, its always hard. I'm ok with this solution. Its not the most beautiful to stop the camera movement, but the resize will happen rarely so its ok :)

lib/core/src/animation/physics/inertia.rs
```diff
@@ -296,8 +296,18 @@ impl PhysicsSimulator {
...
-            callback(position);
+            let t          = transition as f32;
+            let interpolated_position = linear_interpolation(current_position,next_position, t);
+
+            let fixed_point = properties.spring().fixed_point;
+            let position    = properties.kinematics().position();
+            let distance    = (position - fixed_point).magnitude();
+            const FIXED_POINT_DISTANCE_THRESHOLD: f32 = 0.1;
+            if distance < FIXED_POINT_DISTANCE_THRESHOLD {
+                callback(fixed_point)
```
wdanilo: shouldnt we stop the simulator here? Stop its velocity etc?

lib/core/src/display/camera/camera2d.rs
```diff
@@ -95,6 +95,12 @@ impl Default for Clipping {
 // === Camera2dData ===
 // ====================
 
+/// Function used to return the updated screen width, height and `Camera2d`'s `DisplayObjectData`.
+pub trait ScreenUpdateFn = FnMut(f32,f32,DisplayObjectData) + 'static;
```
wdanilo: I don't expect function which updates screen dimensions to pass a `DisplayObjectData`. It is some internal object, it is a very strange design. If we tell camera to have a callback after screen update, it sohuld just call the callback. It can pass new scren dimension, but nothing more.

lib/core/src/display/camera/camera2d.rs
```diff
@@ -334,6 +349,11 @@ impl Camera2d {
     pub fn add_zoom_update_callback<F:ZoomUpdateFn>(&self, f:F) {
         self.rc.borrow_mut().add_zoom_update_callback(f);
     }
+
+    /// Adds a callback to notify when `screen` is updated.
+    pub fn add_screen_update_callback<F:ScreenUpdateFn>(&self, f:F) {
+        self.rc.borrow_mut().add_screen_update_callback(f);
+    }
```
wdanilo: Here is one thing I really dislike about this solution and ideas how to fix it. `add_screen_update_callback` adds a callbkack and **removes existing callback if it was there**. It was acceptible for me in the case of `add_zoom_update_callback`, because the idea was that the zoom update callback is registered always only by the owner of the camera (the scene), however, this could be improved as well. The problem with `add_screen_update_callback` is bigger, because the callback is registered by some arbitrary code (navigator). Do we have any guarantees that we will not create something like minimap view which will register its own calllback and accidently unregister this one? Nope. So either we need to panic if the callback was already registered (ugly solution) or allow registering multiple callbacks (nice solution) - by using our callback manager here.

lib/core/src/display/navigation/navigator.rs
```diff
@@ -62,9 +66,8 @@ impl Navigator {
         let drag               = DragProperties::new(1500.0);
         let properties         = PhysicsProperties::new(kinematics, spring, drag);
         let steps_per_second   = 60.0;
-        let properties_clone   = properties.clone();
         let callback           = move |position| camera.set_position(position);
-        let sim = PhysicsSimulator::new(steps_per_second,properties_clone,callback);
+        let sim = PhysicsSimulator::new(steps_per_second,properties.clone(),callback);
```
wdanilo: I know this is old code, so no need to fix that, ubut this constructor should accept reference like `PhysicsSimulator::new(steps_per_second,&properties),callback);` and do the clone inside if needed. In fadct, it shouldaccept anythign `<S:Into<Properties>>` (so it sohuld accept both reference as well as value). Im wiriting aobut it because when designing APIs lets do interfaces this way :)

lib/core/src/animation/physics/inertia.rs
```diff
@@ -296,8 +296,18 @@ impl PhysicsSimulator {
             }
 
             let transition = interval_counter.accumulated_time / interval_counter.interval_duration;
-            let position   = linear_interpolation(current_position,next_position,transition as f32);
-            callback(position);
+            let t          = transition as f32;
```
wdanilo: I know the next line would be split, but split is better than single char name :P Lets name it `transition` instead :)

lib/core/src/animation/physics/inertia.rs
```diff
@@ -296,8 +296,18 @@ impl PhysicsSimulator {
...
+            let interpolated_position = linear_interpolation(current_position,next_position, t);
+
+            let fixed_point = properties.spring().fixed_point;
+            let position    = properties.kinematics().position();
+            let distance    = (position - fixed_point).magnitude();
+            const FIXED_POINT_DISTANCE_THRESHOLD: f32 = 0.1;
+            if distance < FIXED_POINT_DISTANCE_THRESHOLD {
```
wdanilo: Why don't we keep it as a field in `PhysicsSimulator` so if one needs it could be modified per simulator ?

lib/core/src/animation/physics/inertia.rs
```diff
@@ -263,6 +263,40 @@ impl PhysicsProperties {
...
+// === SimulationThresholds ===
+// ============================
+
+/// A struct holding simulation thresholds used for computing optimizations.
+#[derive(Clone,Copy,Debug)]
+pub struct SimulationThresholds {
+    /// Used to snap object to fixed point if its distance is less than this threshold.
+    pub fixed_point_distance : f32,
+    /// Used to stop simulation computing if the object speed is less than this threshold.
+    pub speed                : f32
```
wdanilo: Very, Very minor thing - but lets just do it this way :) In case things are separated by docs, dont align them, it looks strange :)

lib/core/src/animation/physics/inertia.rs
```diff
@@ -280,10 +314,12 @@ pub struct PhysicsSimulator {
...
 impl PhysicsSimulator {
     /// Simulates `Properties` and inputs `Kinematics`' position in `PhysicsCallback`.
-    pub fn new<F:PhysicsCallback>
-    ( steps_per_second:f64
-    , mut properties:PhysicsProperties
-    , mut callback:F) -> Self {
+    pub fn new<F:PhysicsCallback,Properties:Into<PhysicsProperties>>
+    ( steps_per_second : f64
+    , properties       : Properties
+    , thresholds       : SimulationThresholds
```
wdanilo: Why don't we default the thresholds to some sensible, default values? Constructors should be minimal. If the user wants to change the threshold, they can do it after creating the `PhysicsSimulator`. Basically, a user-friendly API allows you to create things in the simplest possible manner, preferably defaulting settings to sensible values if such sensible values exist. In case of thresholds, we should definitely default them to make the constructor easier to user :)

lib/core/src/display/navigation/navigator.rs
```diff
@@ -90,10 +94,23 @@ impl Navigator {
...
 
-            properties_clone.mod_spring(|spring| { spring.fixed_point += Vector3::new(x, y, z); });
-        };
+            properties.mod_spring(|spring| { spring.fixed_point += Vector3::new(x, y, z); });
+        });
+
+        let transform       = camera.transform();
+        let resize_callback = camera.add_screen_update_callback(
+            enclose!((mut properties,transform) move |dimensions:&Vector2<f32>| {
+            println!("{}", dimensions);
```
wdanilo: this should not be comitted.

lib/core/src/display/scene.rs
```diff
@@ -375,7 +378,7 @@ impl {
             }
             self.logger.info("Rendering meshes.");
             self.symbols.render(&self.camera);
-//            self.css3d_renderer.render(&self.camera);
+            self.css3d_renderer.render(&self.camera);
```
wdanilo: Oh God, I believe ciommenting this was my accidental change. I am so sorry about it. I hope it did not cause any problems and you did not lose too much time tracking it down? :(

notdanilo: lol I spent 30m trying to figure that out :P but it's fine.

lib/core/src/animation/physics/inertia.rs
```diff
@@ -296,8 +332,27 @@ impl PhysicsSimulator {
...
+            let interpolated_position = linear_interpolation(
+                current_position,
+                next_position,
+                transition as f32
+            );
+
+            let fixed_point = properties.spring().fixed_point;
+            properties.mod_kinematics(|kinematics| {
+                let speed    = kinematics.velocity.magnitude();
+                if speed >= thresholds.speed {
```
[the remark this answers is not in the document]
notdanilo: It just don't call the callback function. I think it's fine, isn't it?

lib/core/src/display/scene.rs
wdanilo: @notdanilo I'm really, really sorry :(

lib/core/src/animation/physics/inertia.rs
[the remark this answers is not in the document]
wdanilo: But it should. It should snap the position. If I've got a camera which moves like `0.000001 px/s` I sohuld jsut stop the movement and snap it to the final position :)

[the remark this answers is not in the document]
notdanilo: The rule was changed. Callback only occurs when velocity != zero.

PR #172, opened 2020-02-12
Make JSON-RPC handler more async-compatible

mwu-tow:
### Pull Request Description
The JSON-RPC Handler type (and dependent ones) will no longer require owner to manually call `process_events` (aka "tick") every frame to ensure progress.

Instead they are able to provide caller with a `Future` that when spawned as a task in an executor, will ensure progress.

It is better in a few ways:
* as JSON-RPC APIs were asynchronous, the executor setup was required anyway already to use the `Handler` — requiring both executor and ticks was just unnecessarily complicated for the user code;
* this decouples JSON-RPC handling from execution model;
* executor will drive handlers smarter: it will know when they need to be woken (instead of 
bothering them on every frame) and will be able to make as much progress as possible without stalling even when multiple asynchronous components depend on each other and is independent on "ticking" order.

This is a step towards proper implementation of #143.

### Important Notes
Handler now uses explicit shared state (Rc<RefCell<Data>>). It is wrapped by the module and inaccessible outside. All borrow muts are defined by the handler module and are safe.

### Checklist


lib/ide/json-rpc/src/handler.rs
```diff
@@ -150,55 +128,135 @@ pub type OngoingCalls = HashMap<Id,oneshot::Sender<ReplyMessage>>;
...
-    /// Transports text messages between this handler and the peer.
-    pub transport       : Box<dyn Transport>,
-    /// Allows receiving events from the `Transport`.
-    pub incoming_events : std::sync::mpsc::Receiver<TransportEvent>,
-    /// Handle to send outgoing events.
-    pub outgoing_events : Option<UnboundedSender<Event<Notification>>>,
+    rc: Rc<RefCell<HandlerData<Notification>>>,
+}
+
+impl<Notification> Handler<Notification> {
```
farmaazon: We've got two impls block - please merge them or add subsections `// ===` between them

lib/ide/json-rpc/src/handler.rs
```diff
@@ -150,55 +134,141 @@ pub type OngoingCalls = HashMap<Id,oneshot::Sender<ReplyMessage>>;
...
+    }
+
+    /// Removes all the ongoing requests. This will be recognized by the `Future`s
+    /// as losing connection error.
+    pub fn clear_ongoing_requests(&self) {
+        with(self.rc.borrow_mut(), |mut data| {
+            data.ongoing_calls.clear()
+        })
+    }
+
```
wdanilo: Instead of doing all of these things by hand,  could you pelase use the `shared!` macro?

lib/ide/json-rpc/src/handler.rs
```diff
@@ -150,55 +134,141 @@ pub type OngoingCalls = HashMap<Id,oneshot::Sender<ReplyMessage>>;
...
+        });
+        event_rx
+    }
+
+    /// Creates a new stream with events from this handler.
+    ///
+    /// If such stream was already existing, it will be finished (and
+    /// continuations should be able to process any remaining events).
+    pub fn handler_event_stream(&self) -> impl Stream<Item = Event<Notification>> {
+        let (tx,rx)          = unbounded();
```
wdanilo: alignemnt

lib/ide/json-rpc/src/handler.rs
```diff
@@ -150,55 +134,141 @@ pub type OngoingCalls = HashMap<Id,oneshot::Sender<ReplyMessage>>;
...
+                    Ok(()) => {},
+                    Err(e) =>
+                        if e.is_full() {
+                            // Impossible, as per `futures` library docs.
+                            panic!("unbounded channel should never be full")
+                        } else if e.is_disconnected() {
+                            // It is ok for receiver to disconnect and ignore events.
+                        } else {
+                            // Never happens unless `futures` library changes API.
+                            panic!("unknown unexpected error")
```
wdanilo: `unknown unexpected error` sounds strange

lib/ide/json-rpc/src/handler.rs
```diff
@@ -150,55 +134,141 @@ pub type OngoingCalls = HashMap<Id,oneshot::Sender<ReplyMessage>>;
...
+    /// Sends a handler event to the event stream.
+    pub fn emit_event(&self, event:Event<Notification>) {
+        with(self.rc.borrow_mut(), |data| {
+            if let Some(event_tx) = data.outgoing_events.as_ref() {
+                match event_tx.unbounded_send(event) {
+                    Ok(()) => {},
+                    Err(e) =>
+                        if e.is_full() {
+                            // Impossible, as per `futures` library docs.
+                            panic!("unbounded channel should never be full")
```
wdanilo: We agreed that the errors will be proper sentences. We need to keep one style everywhere. We can talk about it once again, but if we are still on the smae page, lets make them proper sentecnes here :)

PR #200, opened 2020-02-18
Pass id map to parser calls. 

Josef-Vonasek:
### Checklist

### DIFF
--- a/lib/ide/parser/README.md
+++ b/lib/ide/parser/README.md
@@ -1,8 +1,0 @@
-You can run the tests by calling:
-```
-sbt syntaxJS/fullOptJS
-wasm-pack test --headless --firefox --chrome --safari
-```
-
-Note that --safari wasn't tested and there is also an external 
-[issue](https://github.com/rustwasm/wasm-pack/issues/611) with --chrome. 
--- a/lib/ide/parser/README.md
+++ b/lib/ide/parser/README.md
@@ -1,7 +1,7 @@
 You can run the tests by calling:
 ```
 sbt syntaxJS/fullOptJS
-wasm-pack test --headless --firefox --chrome --safari common/rust/parser
+wasm-pack test --headless --firefox --chrome --safari
 ```
 
 Note that --safari wasn't tested and there is also an external 
--- a/lib/ide/parser/src/api.rs
+++ b/lib/ide/parser/src/api.rs
@@ -1,21 +1,54 @@
 use crate::prelude::*;
 use ast::ID;
+use std::ops::Sub;
+
+use serde::Serialize;
 
 pub type Ast = ast::Ast;
-
 
 
 // ============
 // == Parser ==
 // ============
 
-pub type IDMap = Vec<((isize, isize), ID)>;
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+pub struct Index { pub value:usize }
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+pub struct Size { pub value:usize }
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+pub struct Span { pub index: Index, pub size: Size }
+
+#[derive(Clone, Debug, Default, Serialize)]
+pub struct IDMap(pub Vec<(Span, ID)>);
+
+impl Add for Size {
+    type Output = Size;
+    fn add(self, rhs:Size) -> Size {
+        Size { value: self.value + rhs.value }
+    }
+}
+
+impl Add<Size> for Index {
+    type Output = Index;
+    fn add(self, rhs:Size) -> Index {
+        Index { value: self.value + rhs.value }
+    }
+}
+
+impl Sub<Size> for Index {
+    type Output = Index;
+    fn sub(self, rhs:Size) -> Index {
+        Index { value: self.value - rhs.value }
+    }
+}
+
 
 /// Entity being able to parse Luna programs into Luna's AST.
 pub trait IsParser {
     fn parse(&mut self, program: String, ids: IDMap) -> Result<Ast>;
 }
-
 
 
 // ===========
--- a/lib/ide/parser/src/api.rs
+++ b/lib/ide/parser/src/api.rs
@@ -3,6 +3,7 @@ use ast::ID;
 use std::ops::Sub;
 
 use serde::Serialize;
+use serde::Deserialize;
 
 pub type Ast = ast::Ast;
 
@@ -11,16 +12,16 @@ pub type Ast = ast::Ast;
 // == Parser ==
 // ============
 
-#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct Index { pub value:usize }
 
-#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct Size { pub value:usize }
 
-#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct Span { pub index: Index, pub size: Size }
 
-#[derive(Clone, Debug, Default, Serialize)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct IDMap(pub Vec<(Span, ID)>);
 
 impl Add for Size {
--- a/lib/ide/parser/src/api.rs
+++ b/lib/ide/parser/src/api.rs
@@ -1,4 +1,5 @@
 use crate::prelude::*;
+use ast::ID;
 
 pub type Ast = ast::Ast;
 
@@ -8,9 +9,11 @@ pub type Ast = ast::Ast;
 // == Parser ==
 // ============
 
+pub type IDMap = Vec<((isize, isize), ID)>;
+
 /// Entity being able to parse Luna programs into Luna's AST.
 pub trait IsParser {
-    fn parse(&mut self, program: String) -> Result<Ast>;
+    fn parse(&mut self, program: String, ids: IDMap) -> Result<Ast>;
 }
 
 
--- a/lib/ide/parser/src/jsclient.rs
+++ b/lib/ide/parser/src/jsclient.rs
@@ -1,10 +1,11 @@
 #![cfg(target_arch = "wasm32")]
 
 use crate::prelude::*;
-
 use crate::api;
-use crate::api::{IsParser, IDMap};
-use crate::api::Error::ParsingError;
+
+use api::IsParser;
+use api::IDMap;
+use api::Error::ParsingError;
 
 use wasm_bindgen::prelude::*;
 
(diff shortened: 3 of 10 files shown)


lib/ide/parser/src/api.rs
```diff
@@ -8,9 +9,11 @@ pub type Ast = ast::Ast;
 // == Parser ==
 // ============
 
+pub type IDMap = Vec<((isize, isize), ID)>;
+
 /// Entity being able to parse Luna programs into Luna's AST.
 pub trait IsParser {
-    fn parse(&mut self, program: String) -> Result<Ast>;
+    fn parse(&mut self, program: String, ids: IDMap) -> Result<Ast>;
```
mwu-tow: The original code wasn't complaint in the first place (it was moved to this repo without refactoring) but when you touch it, please follow the coding guidelines.

There should be no space between the parameter name and its type (so `ids:IDMap` and so on).

lib/ide/parser/src/jsclient.rs
```diff
@@ -3,7 +3,7 @@
 use crate::prelude::*;
 
 use crate::api;
-use crate::api::IsParser;
+use crate::api::{IsParser, IDMap};
```
mwu-tow: Coding guidelines forbid this syntax, please use two separate "use" directives.

lib/ide/parser/src/jsclient.rs
```diff
@@ -53,10 +50,13 @@ impl Client {
...
+   fn parse(&mut self, program: String, ids: IDMap) -> api::Result<api::Ast> {
+       let json_ids = serde_json::to_string(&ids).expect(
+         "ID Map serialization should not fail."
+       );
+       match parse(program, json_ids) {
+           Ok(json_ast) => Err(ParsingError(json_ast)),
+           Err(_)       => Err(api::interop_error(Error::ScalaException())),
+       }
+   }
+}
```
mwu-tow: Files should have a trailing newline.

lib/ide/parser/src/jsclient.rs
```diff
@@ -53,10 +50,13 @@ impl Client {
...
-    fn parse(&mut self, _program: String) -> api::Result<api::Ast> {
-        match parse(_program) {
-            Ok(json_ast) => Err(ParsingError(json_ast)),
-            Err(_)       => Err(api::interop_error(Error::ScalaException())),
-        }
-    }
-}
+   fn parse(&mut self, program: String, ids: IDMap) -> api::Result<api::Ast> {
+       let json_ids = serde_json::to_string(&ids).expect(
+         "ID Map serialization should not fail."
```
mwu-tow: Truly should not happen... but if we already return Resutl, we can include this failure as a case in Error type and pass it to the caller.

lib/ide/parser/src/lib.rs
```diff
@@ -45,7 +47,7 @@ impl Parser {
...
 
 impl api::IsParser for Parser {
-    fn parse(&mut self, program: String) -> api::Result<api::Ast> {
-        self.deref_mut().parse(program)
-    }
-}
+   fn parse(&mut self, program: String, ids: IDMap) -> api::Result<api::Ast> {
+        self.deref_mut().parse(program, ids)
+   }
+}
```
mwu-tow: missing newline

lib/ide/parser/src/lib.rs
```diff
@@ -9,6 +9,8 @@ use crate::prelude::*;
 
 use std::panic;
 use std::ops::DerefMut;
+use crate::api::IDMap;
```
mwu-tow: "use crate::" declarations should be places between prelude and other use declarations.

lib/ide/parser/src/main.rs
```diff
@@ -8,7 +8,7 @@ fn main() {
     println!("Will parse: {}", program);
 
     let mut parser = parser::Parser::new_or_panic();
-    let     output = parser.parse(program);
+    let     output = parser.parse(program, vec![]);
```
mwu-tow: in place of `vec![]` I'd rather use `default()` -- this code should not need to commit to any specific container storage used by idmap.

lib/ide/parser/README.md
```diff
@@ -1,7 +1,7 @@
 You can run the tests by calling:
 ```
 sbt syntaxJS/fullOptJS
-wasm-pack test --headless --firefox --chrome --safari common/rust/parser
+wasm-pack test --headless --firefox --chrome --safari
```
mwu-tow: I think that this file can be removed altogether, as we already have a script that runs all the tests for all the packages (and otherwise this package is being tested like any else).
sbt command is not needed (or possible anymore, as the repos were split), as the JS build artifacts will be automatically downlaoded by the build.rs

lib/ide/parser/src/api.rs
```diff
@@ -8,9 +9,11 @@ pub type Ast = ast::Ast;
 // == Parser ==
 // ============
 
+pub type IDMap = Vec<((isize, isize), ID)>;
```
mwu-tow: I'd wrap this in a strong type, current usage of IDMap should not rely on it being `Vec`.
(actually, no one really uses it in other way then serializing and passing to JS — when we will use it, then we will actually know what underlying container works best)

lib/ide/parser/src/wsclient.rs
```diff
@@ -9,6 +9,7 @@ use websocket::{
 use crate::api;
 use api::Error::*;
 use Error::*;
+use crate::api::IDMap;
```
mwu-tow: use declarations should be ordered (crate usage together)

lib/ide/parser/tests/parsing.rs
```diff
@@ -133,7 +133,7 @@ impl Fixture {
     /// Runs parser on given input, panics on any error.
     fn parse(&mut self, program:&str) -> Ast {
         println!("parsing {}", program);
-        let ast = self.0.parse(program.into()).unwrap();
+        let ast = self.0.parse(program.into(), vec![]).unwrap();
```
mwu-tow: default

lib/ide/parser/src/wsclient.rs
```diff
@@ -84,7 +85,7 @@ impl From<serde_json::error::Error> for Error {
 /// All request supported by the Parser Service.
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub enum Request {
-    ParseRequest { program: String },
+    ParseRequest {program: String, ids: IDMap },
```
mwu-tow: Have I missed something or is the Scala parser-service still needing the same change?

lib/ide/parser/src/api.rs
Josef-Vonasek: I don't see how wrapping it removes the usage of `Vec`. 

I assume the wrapper would be struct, so we would just replace `vec![..]` with more verbose `IDMap { vec![] }`.

mwu-tow: Huh. To correctly hide the vec, one would need top provide construction methods, like `new` or  FromIterator trait.

But I've been thinking about this.

Actually what I don't like is weak typing (by weak typing I mean exposing raw integers and, most relevantly, tuples, instead of named types/fields).  Basically the shape of this IDMap is incidental, created to match the serialziation format for Scala interop, not to express data semantics.

The "proper" ID Map should not really expose the underlying storage, rather hide it behind some API. Also, there should be no tuples like (isize, isize) but we should solve it like we did in Scala — have proper struct for Position/Offset/Span. Having strong types also allows us providing methods that'll make usage much nicer.

But now I think we actually need two ID Map types:
1. the "proper", strongly typed one (like the one I'm describing)
2. the JS serialziation compatible one (like the one you have here).

The first one should be the one exposed in parser API. 
The second one should be created only incidentally, for serialization, when passing idmap to scala. It should not be part of user-targeted interface of this crate.

As part of this PR, I'm happy with having 1. as total stub (e.g. being only able to use default() to construct empty map). Its development will be likely driven by other tasks, like generating idmap from ast or persisting it in module file. And, most importantly, text API.

2. in its current shape is actually fine but its usage should be limited to rust/scala interop.

lib/ide/parser/src/api.rs
```diff
@@ -1,20 +1,57 @@
...
+use serde::Deserialize;
 
+pub type Ast = ast::Ast;
 
 
 // ============
 // == Parser ==
 // ============
 
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
```
mwu-tow: Please remove spaces between arguments. (same for other instances)

lib/ide/parser/src/api.rs
```diff
@@ -1,20 +1,57 @@
...
+pub struct Index { pub value:usize }
```
mwu-tow: I'd move `Index`, `Size` and `Span` to `utilities` crate (e.g. to some path like `src/data/text.rs`.
These are useful dictionary type that I'd encourage various part of codebase to use (and add more utilities to them). I don't consider them specific to parser. (particularly since we plan to use them in ast)

lib/ide/parser/src/api.rs
```diff
@@ -1,20 +1,57 @@
...
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Size { pub value:usize }
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Span { pub index: Index, pub size: Size }
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct IDMap(pub Vec<(Span, ID)>);
```
mwu-tow: This one I think should live next to AST structures. While currently it lives in parser, we will soon want a method that generates this from AST and I don't really think it is specific to parser.

lib/ide/parser/src/main.rs
```diff
@@ -1,5 +1,7 @@
 use parser::api::IsParser;
 
+use enso_prelude::default;
```
mwu-tow: If symbol is in prelude, we should import it whole (prelude::*) and place before other use declarations.
Also, we should prefer crate (parser's) prelude.

lib/ide/parser/src/wsclient.rs
```diff
@@ -84,7 +85,7 @@ impl From<serde_json::error::Error> for Error {
 /// All request supported by the Parser Service.
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub enum Request {
-    ParseRequest { program: String },
+    ParseRequest { program: String, ids: api::IDMap },
```
mwu-tow: redundant spaces after `:`

lib/ide/parser/src/wsclient.rs
```diff
@@ -196,14 +197,15 @@ impl Client {
 }
 
 impl api::IsParser for Client {
-    fn parse(&mut self, program: String) -> api::Result<api::Ast> {
-        let request  = Request::ParseRequest { program };
+
+   fn parse(&mut self, program:String, ids:api::IDMap) -> api::Result<api::Ast> {
+        let request  = Request::ParseRequest { program, ids };
```
mwu-tow: We don't place spaces in inner sides of {…} braces.
We don't place space between simple arguments. (no one knows what exactly is "simple" but single identifier is simple)

lib/ide/parser/tests/web.rs
```diff
@@ -1,23 +1,46 @@
 use parser::Parser;
+use uuid::Uuid;
 use wasm_bindgen_test::{wasm_bindgen_test_configure, wasm_bindgen_test};
+
 use parser::api::Error::ParsingError;
+use parser::api::IDMap;
+use parser::api::Span;
+use parser::api::Index;
+use parser::api::Size;
+use enso_prelude::default;
```
mwu-tow: bring in crate's prelude::*

lib/ide/parser/tests/web.rs
```diff
@@ -1,23 +1,46 @@
...
+use std::rc::Rc;
```
mwu-tow: this should come in as part of prelude

lib/ide/parser/src/api.rs
mwu-tow: Not yet resolved.

lib/ide/parser/src/api.rs
```diff
@@ -1,20 +1,57 @@
...
+    type Output = Index;
+    fn add(self, rhs:Size) -> Index {
+        Index { value: self.value + rhs.value }
+    }
+}
+
+impl Sub<Size> for Index {
+    type Output = Index;
+    fn sub(self, rhs:Size) -> Index {
+        Index { value: self.value - rhs.value }
```
mwu-tow: theres should be no spaces inside {…}  (here and same applies to instances above)

lib/ide/parser/src/api.rs
mwu-tow: BTW, if you add new types, there should be new sections for them. `Index` is not a "Parser".

PR #201, opened 2020-02-19
Navigator refactoring and other improvements on the new MouseManager

notdanilo:
### Pull Request Description
This PR removes the duplicated implementation of MouseManager, keeping the newest one.

### Important Notes
This task improves the MouseManager adding two events which weren't exposed: `wheel event` and `mouse leave event`.

### Checklist


lib/core/src/control/io/mouse.rs
```diff
@@ -1,407 +1,144 @@
-//! This module contains the `MouseManager` implementation, its associated structs such as
-//! `MousePositionEvent`, `MouseClickEvent` and `MouseWheelEvent`.
+//! This module contains implementation of a oldmouse manager and related utilities.
```
wdanilo: Is this correct?

lib/core/src/control/io/mouse.rs
```diff
@@ -1,407 +1,144 @@
...
-    pub position : Vector2<f32>,
-
-    /// The button which triggered the event.
-    pub button : MouseButton
+/// Shared event dispatcher.
+#[derive(Debug,Derivative)]
+#[derivative(Clone(bound=""))]
+#[derivative(Default(bound=""))]
+pub struct EventDispatcher<T> {
+    rc: Rc<RefCell<XCallbackRegistry1<T>>>
```
wdanilo: When creating `mouse2.rs` I created the `XCallbackRegistry1` because `CallbackRegistry1` had too many cosntraints. Can we fix this in this pr as well pls ?

lib/core/src/control/io/mouse/button.rs
```diff
@@ -1,4 +1,4 @@
-//! This module defines bindings to mouse buttons.
+//! This module defines bindings to oldmouse buttons.
```
wdanilo: Please fix all the docs - there is a lot of docs liek these.

lib/core/src/display/navigation/navigator/events.rs
```diff
@@ -236,90 +228,96 @@ impl NavigatorEvents {
...
+        self.disable_context_menu = Some(ignore_context_menu(target)?);
         Ok(())
     }
 
     fn initialize_mouse_end_event(&mut self) -> Result<()> {
-        let data         = Rc::downgrade(&self.data);
-        let closure      = move |_| {
+        let data              = Rc::downgrade(&self.data);
+        let listener          = self.mouse_manager.on_up.add(move |_:&OnUp| {
             if let Some(data) = data.upgrade() {
```
wdanilo: a minor thing - do not align `=` if they are in different blocks . So here, the last `=` is in a new block `{...}` , so the first 2 `=` sohuld be aligned with themselves but not with the last one.

lib/core/src/display/scene.rs
```diff
@@ -181,13 +182,13 @@ impl Mouse {
...
-                mouse2::Button1 => button1_pressed_ref.set(true),
-                mouse2::Button2 => button2_pressed_ref.set(true),
-                mouse2::Button3 => button3_pressed_ref.set(true),
-                mouse2::Button4 => button4_pressed_ref.set(true),
+                Button0 => button0_pressed_ref.set(true),
+                Button1 => button1_pressed_ref.set(true),
+                Button2 => button2_pressed_ref.set(true),
+                Button3 => button3_pressed_ref.set(true),
+                Button4 => button4_pressed_ref.set(true),
             }
```
wdanilo: I would really like to keep here `mouse::ButtonX` instead. I do not think that importing all button names to scope is a good idea as it pollutes the scope and also makes the code harder to read. Seeing "mouse::Button" in code makes it much more obvious.

lib/core/src/control/io/mouse.rs
notdanilo: I will fix the `oldmouse` word introduced in the comments.

PR #203, opened 2020-02-19
File Manager Integration

mwu-tow:
### Pull Request Description
This PR brings code necessary to fully integrate file manager and properly use it.
Two most important parts are:
* executor that can be integrated with basegl's event loop;
* implementation of the `Transport` trait based on `web_sys::WebSocket`.

I've also added some glue code to put the file manager connection into the project view.

Implements #143.

Has been already reviewed as #199, hopefully all feedback has been applied (or agreed that we won't change this at this point).

### Checklist


lib/ide/src/controller/project.rs
```diff
@@ -62,7 +61,25 @@ shared! { Handle
     }
 }
 
+impl State {
+    /// Creates a new project controller. Schedules all necessary execution with
+    /// the global executor.
+    pub fn new_running(file_manager_transport:impl Transport + 'static) -> Self {
```
farmaazon: I'm not sure if this should be a public function. Or just merge with these function in `Handle` The docs say that this function "creates a new project controller", but in fact this create _State_ and state alone is not project controller, is it?

lib/ide/src/executor/global.rs
```diff
@@ -0,0 +1,62 @@
...
+// This is safe, because the global mutable state is only accessible through the
+// two functions provided in this module, the code will be used only in a web
+// single-threaded environment (so no race conditions are possible), and the
+// functions do not allow leaking reference to the global executor.
+
+/// Obtains the reference to the current spawner.
+///
+/// Panics, if the global spawner hasn't been set previously with `set_spawner`.
+#[allow(unsafe_code)]
+pub fn with_spawner<R>(f:impl FnOnce(&dyn LocalSpawn) -> R) -> R {
```
farmaazon: In fact this is not much safer than before.

What if someone call "set_spawner" inside the passed function?

I wonder if we should not create RefCell, so we at least avoid ub.

lib/ide/src/executor/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Code dealing with executors, i.e. entities that are used to execute asynchronous
```
farmaazon: This is mod file, and we use another file layout for modules.

lib/ide/src/transport/web.rs
```diff
@@ -0,0 +1,250 @@
...
+            })
+        }
+    }
+
+    fn set_event_tx(&mut self, tx:mpsc::UnboundedSender<TransportEvent>) {
+        let tx1 = tx.clone();
+        self.set_on_message(move |e| {
+            let data = e.data();
+            if let Some(text) = data.as_string() {
+                let _ = tx1.unbounded_send(TransportEvent::TextMessage(text));
```
farmaazon: I still see the `let _` and you said that you will make a function for "send without caring about no recipients"

lib/shapely/impl/src/shared.rs
```diff
@@ -255,6 +255,11 @@ macro_rules! shared_struct {
                     operation(&mut self.rc.borrow_mut())
                 }
 
+                /// Wraps given data object into a shared handle.
+                pub fn new_from_data(data:$name_mut<$($params)*>) -> Self {
```
farmaazon: I think this one should be private (as with_borrowed is)

lib/ide/src/executor/global.rs
```diff
@@ -0,0 +1,56 @@
+//! This module defines a shared, global state that stores a task spawner.
+//!
+//! It is expected that as a part of one-time initialization routine, IDE shall
+//! call `set_spawner` to define a global spawner. Once that is done, the whole
```
wdanilo: Wouldn't `init_spawner` be better name here?

lib/ide/src/executor/global.rs
```diff
@@ -0,0 +1,56 @@
...
+// Note [Global Executor Safety]
+// =============================
+// This is safe, because the global mutable state is only accessed through the
+// two functions provided in this module, the code will be used only in a web
+// single-threaded environment (so no race conditions are possible), and the
+// functions do not leak reference to the global spawner.
+
+/// Spawns a task using the global spawner.
+/// Panics, if called when there is no global spawner set or if it fails to
+/// spawn task (e.g. because the connected executor was prematurely dropped).
```
wdanilo: Please move this comment to the docs of `set_spawner` under section **safety**.

lib/ide/src/executor/global.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::prelude::*;
+
+use futures::task::LocalSpawnExt;
+use futures::task::LocalSpawn;
+
+/// Global spawner handle.
+///
+/// It should be set up once, as part of the IDE initialization routine and
+/// remain accessible indefinitely.
+static mut SPAWNER: Option<Box<dyn LocalSpawn>> = None;
```
wdanilo: Could we use here [lazy-static](http://rust-lang-nursery.github.io/lazy-static.rs/lazy_static/index.html) instead?

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Underlying executor. Shared internally with the event loop callback.
+    executor    : Rc<RefCell<LocalPool>>,
+    /// Executor's spawner handle.
+    pub spawner : LocalSpawner,
+    /// Event loop that calls us on each frame.
+    event_loop  : Option<EventLoop>,
+    /// Handle to the callback - if dropped, loop would have stopped calling us.
+    /// Also owns a shared handle to the `executor`.
+    cb_handle   : Option<CallbackHandle>,
+}
```
wdanilo: Imo we should not align things with docs. I know we've been chatting about it, but the more I see it, the more I find it strange.

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+use futures::task::SpawnError;
+use futures::executor::LocalPool;
+use futures::executor::LocalSpawner;
+
+/// Executor. Uses a single-threaded `LocalPool` underneath, relying on basegl's
+/// `EventLoop` to do as much progress as possible on every animation frame.
+#[derive(Debug)]
+pub struct JSExecutor {
+    /// Underlying executor. Shared internally with the event loop callback.
+    executor    : Rc<RefCell<LocalPool>>,
```
wdanilo: Could we make a newtype for `Rc<RefCell<LocalPool>>`? We should make RefCells accessible in the smallest possible amount of places, so making a small, self-contained shared pool is better than this.

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+use futures::task::LocalSpawn;
+use futures::task::LocalFutureObj;
+use futures::task::SpawnError;
+use futures::executor::LocalPool;
+use futures::executor::LocalSpawner;
+
+/// Executor. Uses a single-threaded `LocalPool` underneath, relying on basegl's
+/// `EventLoop` to do as much progress as possible on every animation frame.
+#[derive(Debug)]
+pub struct JSExecutor {
```
wdanilo: It is named `JSExecutor` and the docs do not tell anything about JS connection here. They tell it is just executor. What is the difference between JSExecutor and Executor then ?

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+    cb_handle   : Option<CallbackHandle>,
+}
+
+impl JSExecutor {
+    /// Creates a new JS Executor. It is not yet running, use `schedule_running`
+    /// method to schedule it in an event loop.
+    pub fn new() -> JSExecutor {
+        let executor  = LocalPool::default();
+        let spawner   = executor.spawner();
+        let executor  = Rc::new(RefCell::new(executor));
```
wdanilo: alignment

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+        JSExecutor {
+            executor,
+            spawner,
+            event_loop : None,
+            cb_handle  : None,
+        }
```
wdanilo: if you create `let event_loop = None;` and `let cb_handle = None` then you can write this in a single line, which will make the whole code 4-lines shorter.

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// Creates a new JS executor with an event loop of its own. The event loop
+    /// will live as long as this executor.
+    pub fn new_running() -> JSExecutor {
+        let mut executor   = JSExecutor::new();
```
wdanilo: alignment

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+        executor.schedule_running(EventLoop::new());
+        executor
+    }
+
+    /// Returns a callback compatible with `EventLoop` that once called shall
+    /// attempt achieving as much progress on this executor's tasks as possible
+    /// without stalling.
```
wdanilo: what does mean "callback compatible with EventLoop" ?

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub fn runner_callback(&self) -> impl EventLoopCallback {
+        let executor = self.executor.clone();
```
wdanilo: This is a cheap clone, use `clone_ref` to indicate it instead.

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+        move |_| {
+            // Safe, because this is the only place borrowing executor and loop
+            // callback shall never be re-entrant.
+            let mut executor = executor.borrow_mut();
+            executor.run_until_stalled();
```
wdanilo: This would be obvious if the `executor` was a newtype. Now we need such comments to prove it safety, which is much worse because these comments can not survive refactoring of other people working on this code.

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// Registers this executor to the given event's loop. From now on, event
+    /// loop shall trigger this executor on each animation frame. To stop call
+    /// `stop_running`.
+    ///
+    /// The executor will keep copy of this loop handle, so caller is not
+    /// required to keep it alive.
+    pub fn schedule_running(&mut self, event_loop:EventLoop) {
+        let cb = self.runner_callback();
```
wdanilo: the name `runner_callback` sounds like there was an entity named `runner` and this is its callback. When reading the code I was sure till now thisi s the case here. The function `runner_callback` should be just named `run` instead.

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    /// Registers this executor to the given event's loop. From now on, event
+    /// loop shall trigger this executor on each animation frame. To stop call
+    /// `stop_running`.
+    ///
+    /// The executor will keep copy of this loop handle, so caller is not
+    /// required to keep it alive.
+    pub fn schedule_running(&mut self, event_loop:EventLoop) {
```
wdanilo: One question - why the executor does not own `EventLoop`? Why it does not create its own instance in constructor?

lib/ide/src/executor/global.rs
wdanilo: Moreover, I think it would be beneficial to tell in docs here what is a spawner. I read almost the whole file and still I don ot understand what is a spawner and what is its relation to executor. I know it "spawns tasks", but what it really means? What is an example of such task, what is the difference between spawning a task and calling a function, etc.

lib/ide/src/executor/web.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Stops event loop (previously assigned by `run` method) from calling this
+    /// executor anymore. Does nothing if no loop was assigned. To resume call
+    /// `schedule_running`.
+    ///
+    /// Drops the stored handle to the loop.
+    pub fn stop_running(&mut self) {
+        self.cb_handle  = None;
+        self.event_loop = None;
+    }
```
wdanilo: if this is named `stop_running` then the opposite method should be `start_running`

lib/ide/src/lib.rs
```diff
@@ -37,9 +52,78 @@ pub mod constants {
...
-    ProjectView::new().forget();
-}
\ No newline at end of file
+    let global_executor = setup_global_executor();
+    // We want global executor to live indefinitely.
+    std::mem::forget(global_executor);
+
+    let config = SetupConfig::new_mock();
+    executor::global::spawn(async move {
+        let error_msg    = "Failed to setup initial project view.";
```
wdanilo: alignment

lib/ide/src/transport/web.rs
```diff
@@ -0,0 +1,250 @@
...
+// =================
+// === Utilities ===
+// =================
+
+#[wasm_bindgen]
+extern "C" {
+    /// Converts given `JsValue` into a `String`.
+    #[allow(unsafe_code)]
+    #[wasm_bindgen(js_name="String")]
+    fn js_to_string(s: JsValue) -> String;
```
wdanilo: Does `dyn_into` or `unsafe_dyn_into` not allow for that? I like the name `unsafe_dyn_into` more as this is very unsafe. If you provide here nything which is not a string it will blow up.

lib/ide/src/transport/web.rs
```diff
@@ -0,0 +1,250 @@
...
+#[derive(Clone,Debug,Fail)]
+pub enum ConnectingError {
+    /// Failed to construct websocket. Usually this happens due to bad URL.
+    #[fail(display = "Invalid websocket specification: {}.", _0)]
+    ConstructionError(String),
+    /// Failed to establish connection. Usually due to connectivity issues,
+    /// wrong URL or server being down. Unfortunately, while the real error
+    /// cause is usually logged down in js console, we have no reliable means of
+    /// obtaining it programmatically. Reported error codes are utterly
+    /// unreliable.
```
wdanilo: Is there absolutely no way to even parse the error and get this information ? :(

lib/ide/src/transport/web.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// Wraps given WebSocket object.
+    pub fn new(ws:web_sys::WebSocket) -> WebSocket {
+        WebSocket {
+            ws,
+            on_message : default(),
+            on_close   : default(),
+            on_open    : default(),
+            on_error   : default(),
+        }
+    }
```
wdanilo: I would put all the "storages" in a new struct. This way a lot of the code would be nicer, like you can then derive Default for that and change the `new` function to:

```rust
    pub fn new(ws:web_sys::WebSocket) -> WebSocket {
        let storage = default();
        WebSocket {ws,storage}
    }
```

lib/ide/src/transport/web.rs
```diff
@@ -0,0 +1,250 @@
...
+        wst.wait_until_open().await?;
+        Ok(wst)
+    }
+
+    /// Awaits until `open` signal has been emitted. Clears any callbacks on
+    /// this `WebSocket`, if any has been set.
+    async fn wait_until_open(&mut self) -> Result<(),ConnectingError> {
+        // Connecting attempt shall either emit on_open or on_close.
+        // We shall wait for whatever comes first.
+        let (tx, mut rx) = mpsc::unbounded::<Result<(),()>>();
```
wdanilo: Please use better names then `tx` and `rx` - I was seriously trying to figure them out, but I cant.

lib/ide/src/transport/web.rs
```diff
@@ -0,0 +1,250 @@
...
+        let tx_clone = tx.clone();
```
wdanilo: if this is a cheap clone, please use `clone_ref`

lib/ide/src/transport/web.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /// Sets callback for the `open` event.
+    pub fn set_on_open(&mut self, f:impl FnMut(Event) + 'static) {
+        self.on_open.wrap(f);
+        self.ws.set_onopen(self.on_open.js_ref());
+    }
+
+    /// Clears all the available callbacks.
+    pub fn clear_callbacka(&mut self) {
```
wdanilo: I believe there is a type in this name, unless it is a polish word for "callback" - "ustaw callbacka" :D

lib/ide/utils/src/channel.rs
```diff
@@ -0,0 +1,21 @@
...
+                panic!("Unbounded channel should never be full.")
+            } else if e.is_disconnected() {
+                // It is ok for receiver to disconnect and ignore events. We
+                // want to just "emit", we do not care if someone is listening.
+            } else {
+                // Never happens unless `futures` library changes API.
+                panic!("Unrecognized error when sending event.")
+            }
+    }
+}
```
wdanilo: nl

lib/system/web/src/closure.rs
```diff
@@ -0,0 +1,60 @@
...
+pub trait ClosureFn<Arg> = FnMut(Arg) + 'static where Arg: ClosureArg;
+
+/// Stores an optional closure.
+/// The purpose it reduce boilerplate repeating when setting JS callbacks.
+#[derive(Debug,Derivative)]
+#[derivative(Default(bound=""))]
+pub struct ClosureStorage<Arg> {
+    /// The stored closure.
+    pub closure : Option<Closure<dyn FnMut(Arg)>>,
+}
```
wdanilo: If this is only for storing optional closure, I would name it `OptionalClosure`. The name `ClosureStorage` does not indicate its real purpose. Moreover, this has a hardcoded type to be `FnMut`, whcile `Closure` is generic over the type. I would make this generic as well or name it `OptionalFnMutClosure`.

lib/system/web/src/closure.rs
```diff
@@ -0,0 +1,60 @@
+use crate::prelude::*;
```
wdanilo: this module does not implement `closure` - it implements some utils for it. In fact it implements some kind of storage. I would move it to `src/closure/storage` instead. Right now, after browisng the filenames I was sure thisis an altenrative to wasm-bindgen implementation for a closure

mwu-tow: done

lib/system/web/src/closure.rs
mwu-tow: done

lib/ide/src/transport/web.rs
mwu-tow: done

lib/ide/src/transport/web.rs
mwu-tow: What is "cheap clone" ? 
Compiler says there's no such method. Am I missing some import?

lib/ide/src/executor/web.rs
mwu-tow: It means that we can pass it as an argument to `EventLoop.add_callback`.

lib/ide/src/executor/web.rs
mwu-tow: There is no difference. There is no such thing as the "Executor". Executor is a concept. JSExecutor is an executor.

But I feel sentiment that actually this executor is not directly about JS. The event loop is about JS, because it implements on top of `Window.requestAnimationFrame` but that's implementation detail of a different module.
Will rename.

lib/ide/src/executor/web.rs
mwu-tow: done

lib/ide/src/executor/web.rs
mwu-tow: The code is longer and more symbols are repeated.
But if you really want… 
done

lib/ide/src/executor/web.rs
mwu-tow: done

lib/ide/src/executor/web.rs
mwu-tow: What is "cheap clone" ? 
COmpiler says there's no such method. Am I missing somthing?

lib/ide/src/executor/web.rs
mwu-tow: See the answer to the comment about borrow_mut usage below.

lib/ide/src/executor/web.rs
mwu-tow: Not really. Safety comes not from whatever that newtype would wrap but from the EventLoop property and the whole setup.
To be safe it must be scheduled within at most one event loop and previous callback must be dropped when setting a new one.  This executor is just ensuring all these.  I think you cannot split this and "isolate" safety in just newtype wrapper.

lib/ide/src/executor/web.rs
mwu-tow: To give users of this module control over execution.

For example, IDE might want to reuse event loop with rest of basegl and manage exeuction times together (e.g. async execution time should get limited if rendering takes longer than usual).

lib/ide/src/executor/web.rs
mwu-tow: it was named like this before, it was changed per @farmaazon request and I do agree with him: "run" might suggest executing runner before returning. This only sets up runner to be called later by the event loop.

lib/ide/src/executor/web.rs
mwu-tow: done

lib/ide/src/lib.rs
mwu-tow: done

lib/ide/src/transport/web.rs
mwu-tow: None that I know of. Nor the people on Stack Overflow. 
Reportedly they make it impossible for security reasons.

lib/ide/src/transport/web.rs
mwu-tow: We would lose more on having write `.storages` than save here on `defaults`. 
Also, creating incidental data structures just to store things that can be defaulted does not feel nice to me.

lib/ide/src/transport/web.rs
mwu-tow: They mean "transmitter" and "receiver" respectively, Rust stdlib documentation and future-related libraries use them commonly for naming two ends of pipe-like entity. The terms came from electronics, I believe.

You can find these names already in many places in our codebase, it wouldn't make sense to change them only here.
If we want to get rid of them, let's make a separate task for it.

lib/ide/src/transport/web.rs
mwu-tow: Why is this unsafe? https://www.w3schools.com/jsref/jsref_string.asp 
This show quite numerous examples of calling with not a strings.

I believe `dyn_into`  is about low level casting, also no instance for String. (only JsString that then would need to be processed further to deal with unpaired utf16 surrogates)

lib/ide/src/controller/project.rs
```diff
@@ -62,7 +61,25 @@ shared! { Handle
...
+        crate::executor::global::spawn(ret.file_manager.runner());
+        ret
+    }
+}
+
 
+impl Handle {
+    /// Creates a new project controller. Schedules all necessary execution with
+    /// the global executor.
+    pub fn new_running(file_manager_transport:impl Transport + 'static) -> Self {
```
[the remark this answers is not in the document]
wdanilo: `Handle is a common name for pointer-like entities` - that's right. But it does not have anything in common with `State` so when browsing code I see `Handle` I do not understand it is a `StateHandle`. 

`Also, this PR is not about naming scheme of controllers.` - if a naming in code is strange we should chat about it. There are no PRs about naming only. 

Regarding shared macro, the names in shared macro sghould always follow convention `X` and `XData` (as soon as IntelliJ would understand it the `XData` name would be autogenerated. So this should be named `StateData` and `State` respectively.

lib/ide/src/executor/global.rs
wdanilo: After reading docs I was sure this should be set once. What are the use cases of setting it multiple times?

lib/ide/src/executor/web.rs
wdanilo: `CloneRef` is defined in prelude - it is meant to be implemented for strucutres which clone is "cheap", i.e. cloning a pointer (like `Rc`).

lib/ide/src/executor/web.rs
wdanilo: Let's have a call about it! :)

lib/ide/src/executor/web.rs
wdanilo: I don't fully understand it, again, lets have a short call about it :)

lib/ide/src/executor/web.rs
wdanilo: Good point!

lib/ide/src/transport/web.rs
wdanilo: Oh, I totally missed this part `js_name="String"`. All good! In such a case, this should be refacored to lib/system/web

lib/ide/src/transport/web.rs
wdanilo: `We would lose more on having write .storages than save here on defaults` - I do not understand it. Actually writing `.storages` feels nice to me as it sorts the logic in code nicely. If you do not want to write it, you can always write deref for it. Anyway, I feel like grouping all storages into a common struct is actually a nice design.

lib/ide/src/transport/web.rs
wdanilo: Yeah, I think it would be cool to change them. I do not agree that it doesnt make sense to change them only here. Just lets change them and slowly refactor the old code (or lets make task for it), but if we agree something should be named differently, we should not create wrong naming anymore I think.

PR #204, opened 2020-02-20
View x Controller Integration (into wip/mwu/fm-integration2)

notdanilo:
### Pull Request Description
This PR delivers view and controller integration: more specifically Project and Text views x controllers.

### Important Notes
* Press ctrl+s to save the current content to Path("initial file")

### Checklist


lib/ide/src/view/text_editor.rs
```diff
@@ -45,54 +88,65 @@ impl TextEditor {
         let text_size    = 16.0;
         let properties   = TextFieldProperties {font,text_size,base_color,size};
         let text_field   = TextField::new(&world,properties);
+        let key_listener = None;
+        let controller_clone = controller.clone_ref();
+        let text_field_clone = text_field.clone();
```
mwu-tow: Why line above we use `clone_ref` and here `clone`?

lib/ide/src/view/text_editor.rs
```diff
@@ -45,54 +88,65 @@ impl TextEditor {
...
+        executor::global::spawn(async move {
+            if let Ok(content) = controller_clone.read_content().await {
```
mwu-tow: I don't think that this async operation should keep the text controller alive. If the user closes all the text editors, the controller should also finish as soon as possible, and any ongoing operations should be abandoned.

Suggestion is:
1) get the future from `controller.read_content()` outside the async block, then you can await on it without requiring controller
2) pass weak handle to the controller and try upgrading it and writing, after future is done

lib/ide/src/view/text_editor.rs
```diff
@@ -45,54 +88,65 @@ impl TextEditor {
...
+                text_field_clone.write(&content);
```
mwu-tow: Also, I have mixed feelings about design where we open text editor first with no contents, then allow user writing contents, then we overwrite the contents with whatever comes from file manager (remember that e.g. due to some connectivity hiccup the read file answer may come with significant delay).

I'd consider having having the whole text editor creation being async operation depending on reading file contents (so we can editor only after file has been read and we know we can safely edit it).
This is not request for change now (unless you think it is a good idea and easy to implement), just a thought for future work.

lib/ide/src/view/text_editor.rs
```diff
@@ -13,25 +16,65 @@ use basegl::display::world::*;
...
+            let strings : Vec<String> = content.lines.iter().map(|l| l.chars().iter().collect()).collect();
+            for string in &strings {
+                text += string;
+                text += "\n"
+            }
+            text.pop(); // removes last \n
+        });
+
+        executor::global::spawn(async move {
+            if controller.store_content(text).await.is_err() {
```
mwu-tow: Don't pass the controller to the async close, we don't need to keep it alive while waiting for the store result.
You can have 
```rust
        let store_fut = controller.store_content(text);
        executor::global::spawn(async move {
            if store_fut.await.is_err() {
                println!("Failed to write file")
            }
        });
```

lib/ide/src/view/text_editor.rs
```diff
@@ -45,54 +88,65 @@ impl TextEditor {
...
+            if let Ok(content) = controller_clone.read_content().await {
+                text_field_clone.write(&content);
+            }
+        });
         world.add_child(&text_field);
 
-        Self {text_field,padding,position,size}.initialize()
+        let data = TextEditorData {controller,text_field,padding,position,size,key_listener};
+        let rc = Rc::new(RefCell::new(data));
+        Self{rc}.initialize()
```
mwu-tow: You can simplify this with `Self::new_from_data(data)`

lib/ide/src/view/text_editor.rs
mwu-tow: I don't think that `TextField`'s `write` is what we need here, it says that it appends text under the cursor. We want to set the whole TextField contents to `content`.

lib/ide/src/view/text_editor.rs
```diff
@@ -45,54 +88,65 @@ impl TextEditor {
...
+        let text_editor = Rc::downgrade(&self.rc);
+        let closure     = move |event:KeyboardEvent| {
+            const S_KEY : u32 = 83;
+            if event.ctrl_key() && event.key_code() == S_KEY {
+                if let Some(text_editor) = text_editor.upgrade() {
+                    text_editor.borrow().save();
+                }
+            }
+        };
+        let closure : Box<dyn FnMut(KeyboardEvent)> = Box::new(closure);
```
mwu-tow: I'd try moving the type (`Box::new::<TYPE>(...)`) to right hand side, to avoid this awful alignment issue.
Or just don't align line below. 
Or, if you really, really want to stick to rules, you should align also the next line (key_listener) `=`  as well. :P

lib/ide/src/view/text_editor.rs
```diff
@@ -13,25 +16,65 @@ use basegl::display::world::*;
...
+    controller   : Handle,
+    key_listener : Option<KeyboardListener>
 }
 
+impl {
+    /// Updates the underlying display object.
+    pub fn update(&self) {
+        let padding  = self.padding;
+        let position = self.position;
+        let position = Vector3::new(position.x + padding.left, position.y + padding.bottom, 0.0);
```
mwu-tow: Is the last coordinate always zero in such patterns? We might want to make it a named constant, or if the pattern repeats, have a special function that creates Vector3 from two values and filling last value with zero.

lib/ide/src/view.rs
```diff
@@ -13,3 +13,45 @@ pub mod temporary_panel;
...
+
+#[derive(Debug)]
+struct KeyboardListener {
+    callback   : KeyboardClosure,
+    element    : HtmlElement,
+    event_type : String
+}
+
+impl KeyboardListener {
+    fn new(event_type:String, callback:KeyboardClosure) -> Self {
```
mwu-tow: I don't like how this API requires passing some magic strings.
If their set is closed, I'd consider wrapping wrapped as some kind of enum.
Otherwise,  to avoid `.into` on arguments in call place, consider taking argument as `impl Str` and using `.into() in this functions body.

lib/ide/src/view/text_editor.rs
```diff
@@ -45,54 +88,65 @@ impl TextEditor {
...
+            const S_KEY : u32 = 83;
+            if event.ctrl_key() && event.key_code() == S_KEY {
+                if let Some(text_editor) = text_editor.upgrade() {
+                    text_editor.borrow().save();
+                }
+            }
+        };
+        let closure : Box<dyn FnMut(KeyboardEvent)> = Box::new(closure);
+        let callback                                = Closure::wrap(closure);
+        let key_listener = KeyboardListener::new("keydown".into(), callback);
```
mwu-tow: Literal here. Either replace it with enum or consider making a function like `KeyboardListener::new_key_down` so KeyboardListener does not force its users to use literal on every instantiation.

lib/ide/src/executor/web.rs
```diff
@@ -50,7 +49,7 @@ impl EventLoopExecutor {
     /// Returns a callback compatible with `EventLoop` that once called shall
     /// attempt achieving as much progress on this executor's tasks as possible
     /// without stalling.
-    pub fn runner_callback(&self) -> impl EventLoopCallback {
+    pub fn runner_callback(&self) -> impl FnMut(&f64) + 'static {
```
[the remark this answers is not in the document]
wdanilo: We agreed to create type aliases for such things, so this change is bad.

lib/ide/src/view.rs
```diff
@@ -13,3 +13,45 @@ pub mod temporary_panel;
...
+type KeyboardClosure = Closure<dyn FnMut(KeyboardEvent)>;
+
+#[derive(Debug)]
+struct KeyboardListener {
+    callback   : KeyboardClosure,
+    element    : HtmlElement,
+    event_type : String
+}
+
+impl KeyboardListener {
```
wdanilo: 2 questions:

1. Why KeyboardListener is defined in `view`?
2. What is the relation between this keyboard listener and the one created by Adam which is bound to FRP?

lib/ide/src/view.rs
```diff
@@ -13,3 +13,45 @@ pub mod temporary_panel;
...
+        let js_function : &Function = callback.as_ref().unchecked_ref();
+        element.add_event_listener_with_callback(&event_type,js_function).ok();
+        Self {callback,element,event_type}
+    }
+}
+
+impl Drop for KeyboardListener {
+    fn drop(&mut self) {
+        let callback : &Function = self.callback.as_ref().unchecked_ref();
+        self.element.remove_event_listener_with_callback(&self.event_type, callback).ok();
```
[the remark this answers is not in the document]
wdanilo: As I've written in some comments to old PRs - we should never crash app, ut we should log errors.

lib/ide/src/view/layout.rs
```diff
@@ -142,8 +111,8 @@ impl ViewLayoutData {
 
 impl ViewLayout {
     /// Creates a new ViewLayout with a single TextEditor.
-    pub fn default(world:&World) -> Self {
-        let text_editor  = TextEditor::new(&world);
+    pub fn new(world:&World, controller:Handle) -> Self {
```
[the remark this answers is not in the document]
wdanilo: I do think that importing a lot of things to the scope is bad idea. Polluting scope is generally bad idea. This should be `text::controller::Handle`. 

@mwu-tow writing `text::controller::Handle` is better than having a type `TextControllerHandle` :) You can this way easier track imports, and do not use `use something::*` everywhere.

lib/ide/src/view/text_editor.rs
wdanilo: We should not use literals in code until they are refactored to variables. Michal's question is very valid.

lib/ide/src/view/text_editor.rs
```diff
@@ -13,25 +16,65 @@ use basegl::display::world::*;
...
+impl {
+    /// Updates the underlying display object.
+    pub fn update(&self) {
+        let padding  = self.padding;
+        let position = self.position;
+        let position = Vector3::new(position.x + padding.left, position.y + padding.bottom, 0.0);
+        self.text_field.set_position(position);
+        // TODO: set text field size once the property change will be supported.
+        // let padding  = Vector2::new(padding.left + padding.right, padding.top + padding.bottom);
+        // self.text_field.set_size(self.dimensions - padding);
```
wdanilo: What does mean `once the property change will be supported.` ? Please link it to a ticket on zenhub

lib/ide/src/view/text_editor.rs
```diff
@@ -13,25 +16,65 @@ use basegl::display::world::*;
...
+            for string in &strings {
+                text += string;
+                text += "\n"
+            }
+            text.pop(); // removes last \n
+        });
+
+        executor::global::spawn(async move {
+            if controller.store_content(text).await.is_err() {
+                println!("Failed to write file")
```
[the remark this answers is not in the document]
wdanilo: Plus using println to report errors is wrong

lib/core/src/animation/animator/continuous.rs
```diff
@@ -114,8 +114,8 @@ impl ContinuousAnimator {
         let event_loop      = EventLoop::new();
         let data            = ContinuousAnimatorData::new(f);
         let weak_data       = Rc::downgrade(&data);
-        let callback_handle = event_loop.add_callback(move |absolute_time_ms| {
-            weak_data.upgrade().map(|data| data.on_animation_frame(absolute_time_ms));
+        let callback_handle = event_loop.add_callback(move |absolute_time_ms:&f64| {
```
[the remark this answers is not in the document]
notdanilo: This is probably a compiler error, because it should be capable of inferring the type.

There is a discussion about it here: https://github.com/rust-lang/rust/issues/58639

lib/core/src/control/event_loop.rs
```diff
@@ -58,7 +59,7 @@ impl EventLoop {
 
     /// Add new callback. Returns `CallbackHandle` which when dropped, removes
     /// the callback as well.
-    pub fn add_callback<F:EventLoopCallback>(&self, callback:F) -> CallbackHandle {
+    pub fn add_callback<F:CallbackMut1Fn<f64>>(&self, callback:F) -> CallbackHandle {
```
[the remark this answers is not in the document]
notdanilo: It was fixed. EventLoopCallback now has the right signature and it's used in the right places.

lib/ide/src/view.rs
```diff
@@ -13,3 +13,45 @@ pub mod temporary_panel;
...
+use js_sys::Function;
+use web_sys::HtmlElement;
+use web_sys::KeyboardEvent;
+use wasm_bindgen::JsCast;
+
+
+type KeyboardClosure = Closure<dyn FnMut(KeyboardEvent)>;
+
+#[derive(Debug)]
+struct KeyboardListener {
```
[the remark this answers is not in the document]
notdanilo: Because they were only created to be used inside layout and text editor. They aren't intended to be used anywhere else.

lib/core/src/display/shape/text/text_field.rs
```diff
@@ -1,4 +1,7 @@
 //! A module defining TextField. TextField is a basegl component displaying editable block of text.
+//FIXME:unused_parens is triggered in this file with no indication of where it happens. It's
+// possibly happening inside shared! macro.
+#![allow(unused_parens)]
```
wdanilo: This should be fixed in the macro and needs to be investigated where. Consult with me on Discord - I can help with it.

lib/ide/js/notification.js
```diff
@@ -0,0 +1,58 @@
...
+        document.body.appendChild(this.dom);
+    }
+}
+
+class Notification {
+    constructor(service,content,duration,fade_out) {
+        var duration   = duration   || 1.0;
+        var fade_out = fade_out || 1.0;
+        this.dom       = document.createElement("div");
+        var style      = `
```
wdanilo: alignment

lib/ide/src/view/layout.rs
```diff
@@ -136,43 +105,39 @@ impl ViewLayoutData {
...
     /// Creates a new ViewLayout with a single TextEditor.
-    pub fn default(world:&World) -> Self {
-        let text_editor  = TextEditor::new(&world);
-        let key_listener = None;
+    pub fn new
+    ( key_actions          : &mut KeyboardActions
+    , notification_service : &NotificationService
+    , logger               : &Logger
+    , world                : &World
+    , controller           : controller::text::Handle) -> Self {
```
wdanilo: minor one - `) -> Self {` should be in next line :)

lib/ide/src/view/layout.rs
```diff
@@ -136,43 +105,39 @@ impl ViewLayoutData {
...
     }
 }
 
 impl ViewLayout {
     /// Creates a new ViewLayout with a single TextEditor.
-    pub fn default(world:&World) -> Self {
-        let text_editor  = TextEditor::new(&world);
-        let key_listener = None;
+    pub fn new
+    ( key_actions          : &mut KeyboardActions
```
wdanilo: either use `keyboard_actions` or `kb_actions`. "key" is not a valid abbreviation from "keyboard" :)

lib/ide/src/view/project.rs
```diff
@@ -40,16 +62,27 @@ shared! { ProjectView
...
+        let keyboard_bindings    = KeyboardFrpBindings::new(&logger,&keyboard);
+        let mut keyboard_actions = KeyboardActions::new(&keyboard);
+        let resize_callback      = None;
+        let layout               = ViewLayout::new(
+            &mut keyboard_actions,
+            &notification,
+            &logger,
+            &world,
+            text_controller);
+        let data          = ProjectViewData
```
wdanilo: remove alignment as its not glued to previous line

lib/ide/src/view/project.rs
```diff
@@ -40,16 +62,27 @@ shared! { ProjectView
...
+        let keyboard             = Keyboard::default();
+        let keyboard_bindings    = KeyboardFrpBindings::new(&logger,&keyboard);
+        let mut keyboard_actions = KeyboardActions::new(&keyboard);
+        let resize_callback      = None;
+        let layout               = ViewLayout::new(
+            &mut keyboard_actions,
+            &notification,
+            &logger,
+            &world,
+            text_controller);
```
wdanilo: In the next line you use 
```rust
{world,layout,resize_callback,controller,notification,keyboard,keyboard_bindings,
             keyboard_actions};
```
- lets do. the same here :)

lib/ide/src/view/text_editor.rs
```diff
@@ -2,36 +2,70 @@
...
-    position   : Vector2<f32>,
-    size       : Vector2<f32>
+#[derive(Debug)]
+pub struct TextEditorData {
+    text_field           : TextField,
+    padding              : TemporaryPadding,
+    position             : Vector2<f32>,
+    size                 : Vector2<f32>,
+    controller           : controller::text::Handle,
+    notification_service : NotificationService
```
wdanilo: This is important. `NotificationService` duplicates the logic of logger. Instead of using a notification service explicitly, "warning", and "error" logs should be duplicated to it.

lib/ide/src/view/text_editor.rs
```diff
@@ -2,36 +2,70 @@
...
+    /// Saves text editor's content to file.
+    pub fn save(&self) {
+        let controller   = self.controller.clone();
+        let file_path    = controller.file_path();
+        let text         = self.text_field.get_content();
+        let store_fut    = controller.store_content(text);
+        let notification = self.notification_service.clone();
+        let duration     = 1.0;
+        let fade_out     = 1.0;
+        notification.info("Saving file", duration, fade_out);
```
wdanilo: Client of notification service should never control the apperance. Apperance for "info" should be defined in the notification service.

One more thing - again important - I really, really hate notifications like "saving a file". This just makes the GUI ugly. I believe that notifications should be very very rare in the GUI - only in case of serious error. In normal workflow, I do not want our GUI to display **any notifications**. The GUI should be clean and do not disturb constantly with popups. Only warnings and errors duplicated from the logger.

lib/ide/src/view/layout.rs
```diff
@@ -136,43 +104,39 @@ impl ViewLayoutData {
...
     /// Creates a new ViewLayout with a single TextEditor.
-    pub fn default(world:&World) -> Self {
-        let text_editor  = TextEditor::new(&world);
-        let key_listener = None;
+    pub fn new
+    ( logger               : &Logger
+    , kb_actions           : &mut KeyboardActions
+    , world                : &World
+    , controller           : controller::text::Handle
+    ) -> Self {
```
wdanilo: Alignment

lib/ide/src/view/text_editor.rs
```diff
@@ -2,97 +2,155 @@
...
 }
 
+impl {
+    /// Saves text editor's content to file.
+    pub fn save(&self) {
+        let controller   = self.controller.clone();
+        let file_path    = controller.file_path();
+        let text         = self.text_field.get_content();
+        let store_fut    = controller.store_content(text);
+        let logger       = self.logger.clone();
```
wdanilo: alignment

lib/ide/src/view/text_editor.rs
```diff
@@ -2,97 +2,155 @@
...
+        let size       = Vector2::new(screen.width, screen.height);
+        let black      = Vector4::new(0.0,0.0,0.0,1.0);
+        let base_color = black;
+        let text_size  = 16.0;
+        let properties = TextFieldProperties {font,text_size,base_color,size};
+        let text_field = TextField::new(&world,properties);
+
+        let content_future       = controller.read_content();
+        let text_field_weak      = text_field.downgrade();
+        let logger_ref           = logger.clone();
```
wdanilo: alignment

lib/ide/src/view/layout.rs
```diff
@@ -136,43 +104,39 @@ impl ViewLayoutData {
...
-                event.prevent_default();
-            }
-        };
-        let closure : Box<dyn FnMut(KeyboardEvent)> = Box::new(closure);
-        let callback                                = Closure::wrap(closure);
-        let body                                    = document().unwrap().body().unwrap();
-        let key_listener = KeyboardListener::new(&body, "keydown".into(), callback);
-        self.rc.borrow_mut().key_listener = Some(key_listener);
+    fn init_keyboard(self, keyboard_actions:&mut KeyboardActions) -> Self {
+        let save_keys:KeyMask = [Key::Control, Key::Character("f".to_string())].iter().collect();
```
mwu-tow: Is the name correct? `save_keys`
Should be rather about switching layout mode

mwu-tow: Also, I'd consider having some helper to create KeyMask for "Control+Key", it currently looks like a handful and will continue reappear.

notdanilo: I introduced:

```rust
    /// Creates Key::Control + Key::Character.
    pub fn new_control_character(character:char) -> Self {
        Self::from_vec(vec![Key::Control, Key::Character(character.to_string())])
    }

    /// Creates KeyMask from Vec<Key>.
    pub fn from_vec(keys:Vec<Key>) -> Self {
        keys.iter().collect()
    }
```

notdanilo:
https://github.com/luna/ide/pull/219 this PR was already merged to master :)

PR #212, opened 2020-02-24
Pass IDMap to scala parser.

Josef-Vonasek:
### Pull Request Description

Git ignored my last commit, so I had to create new PR.

### Checklist
Please include the following checklist in your PR:

- [x] The documentation has been updated if necessary.
- [x] All code conforms to the [Rust](https://github.com/luna/enso/blob/master/doc/rust-style-guide.md), [Scala](https://github.com/luna/enso/blob/master/doc/scala-style-guide.md), [Java](https://github.com/luna/enso/blob/master/doc/java-style-guide.md) or [Haskell](https://github.com/luna/enso/blob/master/doc/haskell-style-guide.md) style guides as appropriate.
- [x] All code has been tested where possible.

PR #215, opened 2020-02-24
Generate idmap from AST.

Josef-Vonasek:
### Pull Request Description

Waiting for merge of #212 

### Checklist


lib/ide/ast/impl/src/lib.rs
```diff
@@ -152,6 +152,28 @@ impl Ast {
         self
     }
 
+    // TODO use strongly typed IDMap, after it's merged to master
```
mwu-tow: Not a request (as the dependency is currently extremely small and we will soon merge it anyway) but you can also create PRs against branches, so that way you can develop features depending on not yet merged features.
(though merges can get really painful because we are squashing the PRs into single commit)

lib/ide/ast/impl/src/lib.rs
```diff
@@ -152,6 +152,28 @@ impl Ast {
...
+
+    fn collect_ids(&self, index:&mut usize, ids:&mut IDMap) {
+        self.id.map( |id| {
```
mwu-tow: There are two issues:
1. We count span twice  — once for `self`, once for its children. But self's span already includes the children.
2. We can't count index by adding subsequent children's spans, as they are not continuous.
Eg. `a + b` must also take into account spacing around the operator and it is not included in any of the infix application child.

I think we need to start with implementing a new algorithm like `zipWithOffset`.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -19,7 +19,7 @@ use uuid::Uuid;
 /// A sequence of AST nodes, typically the "token soup".
 pub type Stream<T> = Vec<T>;
 
-
+type IDMap = Vec<((usize,usize),ID)>;
```
wdanilo: Please use a struct instead of nested tuples. We have banned using tuples almost everywhere as it is really hard to understand what they are. For example, what are these `usizes`?

lib/ide/ast/impl/src/lib.rs
wdanilo: + Please fix linter warnings - run `cargo clippy` localy to get them :)

lib/ide/ast/impl/src/lib.rs
```diff
@@ -152,6 +152,28 @@ impl Ast {
...
+        for child in self {
+            child.collect_ids(index, ids);
+        }
+    }
+
+    pub fn id_map(&self) -> IDMap {
+        let mut idx = 0;
+        let mut ids = Vec::new();
+        self.collect_ids(&mut idx, &mut ids);
+
```
wdanilo: remove this nl pls :)

lib/ide/ast/impl/src/lib.rs
```diff
@@ -152,6 +152,28 @@ impl Ast {
...
+            ids.push(((*index, self.span), id));
+            *index += self.span;
+        });
+
+        for child in self {
+            child.collect_ids(index, ids);
+        }
+    }
+
+    pub fn id_map(&self) -> IDMap {
```
wdanilo: No docs - please be sure that all lints are enabled in this codebase (lints like `warn(missing_docs)`). Ask @mwu-tow for help in case of doubt.

mwu-tow: @Josef-Vonasek Take a look at https://github.com/luna/ide/blob/master/lib/ide/utils/src/lib.rs

Unfortunately we have enable lints manually in main file of each library crate.

lib/ide/ast/macros/src/repr.rs
```diff
@@ -1,146 +0,0 @@
-use crate::prelude::*;
```
Josef-Vonasek: Oops, renaming this file wasn't that smart.

mwu-tow: I will manage, Meld here comes helpful.

lib/ide/ast/macros/src/token.rs
```diff
@@ -0,0 +1,98 @@
...
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::Expr;
+use syn::GenericArgument;
+use syn::PathSegment;
+use syn::Token;
+use syn::punctuated::Punctuated;
+
+/// Generates `Tokenizer` that just panics when called.
+pub fn no_tokenizer
```
mwu-tow: `panicking_tokenizer`?
I don't like "no_" as that suggests to me as compile-time error is user tries to call it.

lib/ide/ast/impl/src/repr.rs
```diff
@@ -287,88 +253,58 @@ impl<T: HasRepr> HasRepr for TextBlockFmt<T> {
...
-    }
-}
 
-impl<T: HasRepr> HasRepr for TextUnclosed<T> {
-    fn write_repr(&self, target:&mut String) {
-        self.line.write_repr(target);
-        target.pop();  // remove missing quote
+// TODO: [jv] this implementation is wrong since we cannot `pop` from TokenBuilder
+//            either redesign TextUnclosed, so that it can use Tokenizer,
+//            or come up with some smart/ugly hack
```
mwu-tow: Please fix this before merge.
If smart solution doesn't come up quickly, just brute-force this and pattern match on the Line type, then emit proper opening quote and line segments.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+
+impl TokenBuilder for IdMapBuilder {
+    fn with_str(&mut self, val:&str ) { self.offset += val.len() }
+    fn with_int(&mut self, val:usize) { self.offset += val       }
+    fn with_chr(&mut self,_val:char ) { self.offset += 1         }
+    fn with_ast(&mut self, val:&Ast ) {
+        let begin = self.offset;
+        val.shape().tokenize(self);
+        if let Some(id) = val.wrapped.id {
+            let span = Span::new(Index::new(begin), Size::new(self.offset));
```
mwu-tow: Perhaps we should have `Span::from_index_size(usize,usize)` method to reduce boilerplate in such cases while keeping on the usage place clear notion of what is what.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+            self.id_map.0.push((span, id));
```
mwu-tow: We shouldn't access `.0`. `IdMap` should get a method `insert(span,id)`.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
-            el.write_repr(target)
+struct IdMapBuilder { id_map:IdMap, offset:usize }
+
+impl TokenBuilder for IdMapBuilder {
+    fn with_str(&mut self, val:&str ) { self.offset += val.len() }
+    fn with_int(&mut self, val:usize) { self.offset += val       }
+    fn with_chr(&mut self,_val:char ) { self.offset += 1         }
+    fn with_ast(&mut self, val:&Ast ) {
+        let begin = self.offset;
+        val.shape().tokenize(self);
```
mwu-tow: why not just `val.tokenize(self)`? 
(same goes for 3 other occurrences of this pattern)

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+        if let Some(id) = val.wrapped.id {
```
mwu-tow: Use `val.id`, we can thanks to deref coercion.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+trait HasIdMap {
+    /// Extracts IdMap from `self`.
+    fn id_map(&self) -> IdMap;
 }
 
-impl<T: HasRepr> HasRepr for Option<T> {
-    fn write_repr(&self, target:&mut String) {
-        for el in self.iter() {
-            el.write_repr(target)
+struct IdMapBuilder { id_map:IdMap, offset:usize }
```
mwu-tow: For each defined struct derive `Debug` and `Default`, if possible (and sensible). (and `Copy` if trivial, but that happens rarely).

This comment applies to all struct introduced in this PR.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
 }
 
-impl<T: HasRepr> HasRepr for Vec<T> {
-    fn write_repr(&self, target:&mut String) {
-        for el in self.iter() {
-            el.write_repr(target)
-        }
+impl<T:Tokenizer> HasIdMap for T {
+    fn id_map(&self) -> IdMap {
+        let mut builder = IdMapBuilder {id_map:default(),offset:default()};
```
mwu-tow: once `IdMapBuilder ` derives `Default`, you'll be able to just default() it.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+    /// for `t:T` the following must hold: `t.size() == t.repr().len()`.
+    fn size(&self) -> usize;
 }
 
-impl<T: HasRepr, U: HasRepr> HasRepr for (T,U) {
-    fn write_repr(&self, target:&mut String) {
-        self.0.write_repr(target);
-        self.1.write_repr(target);
-    }
+struct SpanBuilder { offset:usize }
```
mwu-tow: derives (see IdMapBuilder comment)

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+
+///// Things that can be asked about their textual representation.
+/////
+///// See also `HasSize`.
+pub trait HasRepr {
+    /// Obtain the text representation for the This type.
+    fn repr(&self) -> String;
+}
+
+struct ReprBuilder { repr:String }
```
mwu-tow: derives (see IdMapBuilder comment)

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
 }
 
-impl<T: HasRepr, U: HasRepr, V: HasRepr> HasRepr for (T,U,V) {
-    fn write_repr(&self, target:&mut String) {
-        self.0.write_repr(target);
-        self.1.write_repr(target);
-        self.2.write_repr(target);
+impl<T:Tokenizer> HasSize for T {
+    fn size(&self) -> usize {
+        let mut builder = SpanBuilder {offset:default()};
```
mwu-tow: default() whole builder

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+    fn with_chr(&mut self, val:char ) { self.repr.push(val)                  }
+    fn with_ast(&mut self, val:&Ast ) { val.shape().tokenize(self)           }
 }
 
-impl<T: HasRepr> HasRepr for &T {
-    fn write_repr(&self, target:&mut String) {
-        self.deref().write_repr(target)
+impl<T:Tokenizer> HasRepr for T {
+    fn repr(&self) -> String {
+        let mut builder = ReprBuilder {repr:default()};
```
mwu-tow: default() whole builder

lib/ide/ast/impl/src/lib.rs
```diff
@@ -584,162 +573,189 @@ pub enum MacroPatternMatchRaw<T> {
...
+/// Can be only fed with char, string, integer and ast.
+pub trait TokenBuilder {
+    /// process &str
+    fn with_str(&mut self, val:&str);
+    /// process usize
+    fn with_int(&mut self, val:usize);
+    /// process char
+    fn with_chr(&mut self, val:char);
+    /// process &Ast
+    fn with_ast(&mut self, val:&Ast);
```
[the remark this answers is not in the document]
Josef-Vonasek: That was my first design, and it was significantly uglier. 
Mostly because of explicit lifetimes.

lib/ide/ast/impl/src/lib.rs
Josef-Vonasek: Because Ast also implements Tokenizer.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -588,158 +582,197 @@ pub enum MacroPatternMatchRaw<T> {
...
+// === Tokenizer ===
+
+/// An enum of valid Ast tokens.I
+#[derive(Debug)]
+pub enum Token<'a> { Off(usize), Chr(char), Str(&'a str), Ast(&'a Ast) }
+
+/// Things that can be turned into stream of tokens.
+pub trait HasTokens {
+    /// Feeds TokenBuilder with stream of tokens obtained from `self`.
+    fn feed(&self, consumer:&mut impl TokenConsumer);
```
wdanilo: This doesnt make sense. How "HasTokens" is connected to "feed" ? How the fact that I have tokens is connected to the fact that I can be feed with tokens? Oh I see, the method name `feed` is wrong. Because it takes my tokens and pass it to the consumer - I would definitely work here on naming more.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -225,7 +219,7 @@ impl Serialize for Ast {
         if self.id.is_some() {
             state.serialize_field(ID, &self.id)?;
         }
-        state.serialize_field(SPAN,  &self.span)?;
+        state.serialize_field(SPAN, &self.size)?;
```
wdanilo: This and some other places in code are inconsistent. Now we have "span" and "size" poinitng to the same thing/. Moreover, I dont like the word "size" - it could be memory size or something else. Span is more unambiguaous.

lib/ide/ast/impl/src/lib.rs
```diff
@@ -588,158 +582,197 @@ pub enum MacroPatternMatchRaw<T> {
...
 }
 
-/// Counts codepoints.
-impl HasSpan for String {
-    fn span(&self) -> usize {
-        self.as_str().span()
+
+impl HasTokens for &str {
+    fn feed(&self, consumer:&mut impl TokenConsumer) {
+        consumer.feed(Token::Str(self));
```
wdanilo: why not `consumer.feed(self.into())` ? There should be conversion instance defined

lib/ide/ast/impl/src/lib.rs
```diff
@@ -588,158 +582,197 @@ pub enum MacroPatternMatchRaw<T> {
...
+        for t in self { t.feed(consumer); }
     }
 }
-impl HasSpan for usize {
-    fn span(&self) -> usize {
-        *self
+
+impl<T:HasTokens> HasTokens for Rc<T> {
+    fn feed(&self, consumer:&mut impl TokenConsumer) {
+        self.deref().feed(consumer);
```
wdanilo: This line is very error prone. Do you know that after removing constraint `T:HasTokens` you will have hefre infinite loop? Use `.content()` instead of `.deref()` - and read docs agbout `content` to learn more why

lib/ide/ast/impl/src/lib.rs
```diff
@@ -775,42 +791,42 @@ Layer<T> for WithID<S> {
...
 /// purposes.
 #[derive(Eq, PartialEq, Debug, Shrinkwrap, Serialize, Deserialize)]
 #[shrinkwrap(mutable)]
-pub struct WithSpan<T> {
+pub struct WithSize<T> {
     #[shrinkwrap(main_field)]
     #[serde(flatten)]
     pub wrapped: T,
-    pub span: usize
+    pub size: usize
```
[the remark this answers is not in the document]
wdanilo: 👎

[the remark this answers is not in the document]
wdanilo: Especially in Rust, where `Sized` trait means that the memory size is known. In Rust size is bound to memory

lib/ide/ast/impl/src/lib.rs
mwu-tow: Span collides with concept `(position + length)` which seems to be the default meaning of span (and so it was in Scala codebase).
I much prefer `size` or `length`.

lib/ide/ast/impl/src/lib.rs
Josef-Vonasek: I have been told that on rust channel that implementing `From` for references isn't normal.

lib/ide/ast/impl/src/lib.rs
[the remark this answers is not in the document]
mwu-tow: Then my suggestion would be `len()` or `length()`.
Definitely not span, it means `from<->to` where both bear value, like https://doc.rust-lang.org/proc_macro/struct.Span.html

lib/ide/ast/impl/src/repr.rs
```diff
@@ -49,37 +49,27 @@ pub const FMT_BLOCK_QUOTES:&str = "'''";
 
 
 // ===============
-// === Builder ===
+// === consumer ===
```
mwu-tow: Unintended replace?

lib/ide/ast/impl/src/lib.rs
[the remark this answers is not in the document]
farmaazon: I don't think "size should be bound to memory" is a general rule of naming things.
But I think `length` is a better name.
And I would add, that `WithSpan` is silly in this context, because with span of what? Span definition is 
> the full extent of something from end to end; the amount of space that something covers.
What covers something what implements "WithSpan" trait?

lib/ide/ast/impl/src/lib.rs
wdanilo: I don't think so. We have a lot of such from s implemented.

lib/ide/ast/impl/src/lib.rs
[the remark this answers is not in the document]
wdanilo: I like `Length` and `HasLength` . Lets do it

lib/ide/ast/impl/src/lib.rs
Josef-Vonasek: Nevertheless, we don't use Token anywhere but here, so implementing `From` would just add 30LOC without any real benefit.

PR #221, opened 2020-02-27
Minimal Text API for Double Representation

farmaazon:
### Pull Request Description
Implemented Text API which updates id_map and ast on each code change.

### Important Notes
- This PR contains also an implementation for actual opening and initializing Module Controller.
- It adds also a "Text changed" callback to TextField which is fired on every user edit.
- The state of Module Controller is AST and id_map. The id_map should be removed once we will be able to read it from AST.

### Checklist


lib/core/src/display/shape/text/text_field/content.rs
```diff
@@ -260,17 +290,48 @@ impl TextFieldContent {
...
+                self.line_offsets.push(current_char_index);
+            }
+        }
+        self.line_offsets[location.line] + location.column
+    }
+
+    /// Converts range of locations in this text represented by `row:column` pair to absolute
+    /// char's position from document begin.
+    pub fn convert_location_range_to_char_index(&mut self, range:&Range<TextLocation>)
+    -> Range<usize> {
```
mwu-tow: Perhaps `Index` from `data::text` could replace here (and in some other API usage) `usize`.

lib/core/src/display/shape/text/text_field/content.rs
```diff
@@ -512,6 +573,28 @@ mod test {
         assert_eq!(two_lines_expected, two_lines_change.inserted_text_range());
     }
 
+    #[wasm_bindgen_test(async)]
+    async fn converting_location_to_offset() {
+        msdf_sys::initialized().await;
```
mwu-tow: Should this be needed to test conversions between (row,col) and (index) ?

lib/core/src/display/shape/text/text_field.rs
```diff
@@ -246,28 +237,100 @@ impl TextField {
...
+        let mut opt_callback    = self.with_borrowed(|this| std::mem::take(&mut this.text_change_callback));
+        for (cursor_id,to_insert) in cursor_id_with_text_to_insert {
+            let notification = self.with_borrowed(|this| {
+                this.apply_one_cursor_change(&mut location_change,cursor_id,to_insert)
+            });
+            if let Some(callback) = opt_callback.as_mut() {
+                callback(&notification);
+            }
+        }
+        self.with_borrowed(|this| {
```
mwu-tow: so careful 👍

lib/ide/parser/src/lib.rs
```diff
@@ -43,13 +62,55 @@ impl Parser {
...
+        Ok(SharedParser(parser))
+    }
+
+    /// Obtains a default parser implementation.
+    #[cfg(target_arch = "wasm32")]
+    pub fn new() -> api::Result<SharedParser> {
+        let client = jsclient::Client::new()?;
+        let parser = Rc::new(RefCell::new(client));
+        Ok(SharedParser(parser))
+    }
```
mwu-tow: Why not just wrap whatever `Parser::new` returns?
Or replace current `Parser` with this `SharedParser` altogether. We don't really have much reason to have "unshareable" parser.

lib/ide/parser/src/lib.rs
```diff
@@ -1,26 +1,45 @@
...
+
+#![feature(trait_alias)]
+#![warn(missing_docs)]
+#![warn(trivial_casts)]
+#![warn(trivial_numeric_casts)]
+#![warn(unused_import_braces)]
+#![warn(unused_qualifications)]
+#![warn(unsafe_code)]
+#![warn(missing_copy_implementations)]
+#![warn(missing_debug_implementations)]
```
mwu-tow: 👍

lib/ide/src/controller/module.rs
```diff
@@ -4,20 +4,48 @@
...
 /// Mappable to filesystem path.
-#[derive(Clone,Debug,Eq,Hash,PartialEq)]
+#[derive(Clone,Debug,Display,Eq,Hash,PartialEq)]
 pub struct Location(pub String);
 
 impl Location {
+    /// Get the module location from filesystem path. Returns None if path does not lead to
+    /// module file.
+    pub fn from_path(path:&fmc::Path) -> Option<Self> {
+        // TODO [mwu] See function `to_path`
```
mwu-tow: not me :P

lib/core/src/display/shape/text/text_field.rs
```diff
@@ -27,6 +27,25 @@ use crate::display::world::World;
...
+/// A notification about text change.
+///
+/// In essence, it's `TextChange` with some additional useful information.
+#[derive(Clone,Debug,Shrinkwrap)]
+pub struct TextChangedNotification {
+    /// A change which has occurred.
+    #[shrinkwrap(main_field)]
+    pub change : TextChange,
+    /// The replaced range as char positions from document begin, instead of row:column pairs.
+    pub replaced_chars : Range<usize>,
```
mwu-tow: Is it possible to use types from `data::text` here?
BTW if we can have Range<data::text::Index> then perhaps it should replace `data::text::Span`?

lib/ide/src/controller/module.rs
```diff
@@ -28,32 +56,161 @@ impl Location {
...
+        pub fn apply_code_change(&mut self,change:&TextChangedNotification) {
+            let mut code        = self.code();
+            let replaced_range  = change.replaced_chars.clone();
+            let inserted_string = change.inserted_string();
+            let replaced_size   = Size::new(replaced_range.end - replaced_range.start);
+            let replaced_span   = Span::new(Index::new(replaced_range.start),replaced_size);
+
+            code.replace_range(replaced_range,&inserted_string);
+            apply_code_change_to_id_map(&mut self.id_map,&replaced_span,&inserted_string);
+            self.ast = self.parser.parse(code, self.id_map.clone()).unwrap();
```
mwu-tow: We should not unwrap here. While parser "should never" fail, this can be triggered pretty easily. Never trust external components. 

We should complain as much as possible and leave previous AST intact, in hope that future edits will fix it. (unless we have idea for better behavior here)

lib/ide/src/controller/module.rs
```diff
@@ -28,32 +56,161 @@ impl Location {
...
+            self.logger.trace(|| format!("Applied change; Ast is now {:?}", self.ast));
+        }
+
+        /// Read module code.
+        pub fn code(&self) -> String {
+            self.ast.repr()
+        }
+
+        /// Check if current module state is synchronized with given code. If it's not, log error,
+        /// and update module state.
```
mwu-tow: > update module state"

This is very generic and largely can be already inferred from `&mut self`. Update how? What is changed? What is the postcondition?

lib/ide/src/controller/module.rs
```diff
@@ -28,32 +56,161 @@ impl Location {
...
     }
 }
 
 impl Handle {
-    /// Receives a notification call when file with this module has been
-    /// modified by a third-party tool (like non-IDE text editor).
-    pub async fn file_externally_modified(&self) {
-        // TODO: notify underlying text/graph controllers about the changes
-        todo!()
+    /// Create a module controller for given location.
```
mwu-tow: I'd suggest that for async functions we should also try to document what the function awaits for. 
(something to give user rough understanding what to expect in terms of remote communication when using this function)

lib/ide/src/controller/module.rs
```diff
@@ -28,32 +56,161 @@ impl Location {
...
+    use data::text::Index;
+    use data::text::Size;
+    use json_rpc::test_util::transport::mock::MockTransport;
+    use uuid::Uuid;
+    use parser::SharedParser;
+    use basegl::display::shape::text::text_field::content::TextChange;
+    use basegl::display::shape::text::text_field::location::TextLocation;
+    use ast;
+    use wasm_bindgen_test::wasm_bindgen_test;
+    use ast::BlockLine;
```
mwu-tow: Sorting would make this… a little nicer.

lib/ide/src/controller/module.rs
```diff
@@ -28,32 +56,161 @@ impl Location {
...
+
+    #[wasm_bindgen_test]
```
mwu-tow: Is there a reason why we wouldn't want to test this natively as well?
Is it possible to have both wasm_bindgen_test and test attributes?

lib/ide/src/controller/module.rs
```diff
@@ -28,32 +56,161 @@ impl Location {
...
+        let uuid2        = Uuid::new_v4();
+        let code         = "2+2";
+        let id_map       = IdMap(vec!
+            [ (Span::new(Index::new(0), Size::new(1)),uuid1.clone())
+            , (Span::new(Index::new(2), Size::new(1)),uuid2)
+            ]);
+
+        let controller   = Handle::new_mock(location,code,id_map,file_manager,parser).unwrap();
+
+        // Change number
```
mwu-tow: It's hard to grasp all these indices at a quick glance. Comment like "Change code to `22+2`" would help.

lib/ide/src/controller/project.rs
```diff
@@ -5,11 +5,19 @@
...
 
+use file_manager_client as fmc;
 use json_rpc::Transport;
+use parser::SharedParser;
+use shapely::shared;
 use weak_table::weak_value_hash_map::Entry::Occupied;
 use weak_table::weak_value_hash_map::Entry::Vacant;
-use file_manager_client as fmc;
-use shapely::shared;
+use crate::controller::FallibleResult;
```
mwu-tow: crate use declarations should go before external use decls

lib/ide/src/controller/project.rs
```diff
@@ -78,47 +76,183 @@ impl Handle {
...
+    pub async fn open_text_file(&self, path:fmc::Path)
+    -> FallibleResult<controller::text::Handle> {
+        let cached = self.with_borrowed(|data| data.text_cache.get(&path));
+        match cached {
+            Some(controller) => Ok(controller),
+            None => {
+                let loaded = self.create_text_controller(path.clone()).await?;
+                //TODO[ao] Here we should make a better solution for case where we simultaneously
+                // load one module twice.
+                // This is duplicated with open_module, but it's not worth refactoring as its
```
mwu-tow: its -> it's

lib/ide/src/controller/project.rs
```diff
@@ -78,47 +76,183 @@ impl Handle {
...
+    async fn create_text_controller(&self, path:fmc::Path)
+    -> FallibleResult<controller::text::Handle> {
+        match ModuleLocation::from_path(&path) {
+            Some(location) => {
+                let module = self.open_module(location).await?;
+                Ok(controller::text::Handle::new_for_module(module))
+            },
+            None => {
+                let fm = self.file_manager();
+                Ok(controller::text::Handle::new_for_plain_test(path,fm))
```
mwu-tow: This looks inconsistent to me. Why it "opens" a module (async operation) but creates a new text file?
I expect that there should be two methods for path: one that opens (i.e. awaits the download) target file (plain or module) and the other one that creates a new controller immediately (and whatever was in the file under path, if it existed, will be ignored and overwritten on save)

mwu-tow: The function should be named `new_for_plain_text`, not `new_for_plain_test`

lib/ide/src/controller/project.rs
```diff
@@ -78,47 +76,183 @@ impl Handle {
...
+    use futures::executor::LocalPool;
+    use wasm_bindgen_test::wasm_bindgen_test;
+    use wasm_bindgen_test::wasm_bindgen_test_configure;
+
+    wasm_bindgen_test_configure!(run_in_browser);
+
+
 
-    #[test]
+    #[wasm_bindgen_test]
```
mwu-tow: As previously and below. For debugging I'd really want to keep ability to run natively as many tests as possible.

lib/ide/src/controller/project.rs
```diff
@@ -78,47 +76,183 @@ impl Handle {
...
+            "id"      : 1,
+            "result"  :"3+3"
+        }"#);
+        executor.run_until_stalled();
+        assert!(*finished.borrow());
     }
 
-    #[test]
+    #[wasm_bindgen_test]
     fn obtain_text_controller() {
```
mwu-tow: I'd clarify in the name that this is about plain_text

lib/ide/src/controller/project.rs
```diff
@@ -78,47 +76,183 @@ impl Handle {
         let data = Controller::new_running(file_manager_transport);
         Self::new_from_data(data)
     }
+
+    /// Returns a text controller for given file path.
```
mwu-tow: I'd clarify that it works both for plain text files and modules.
Now I wondered if it was meant to say "plain text".

lib/ide/src/controller/project.rs
```diff
@@ -78,47 +76,183 @@ impl Handle {
...
+        let mut executor    = LocalPool::new();
+        let finished        = Rc::new(RefCell::new(false));
+        let finished_clone  = finished.clone_ref();
+        let mut transport   = MockTransport::new();
+        let transport_clone = transport.clone_ref();
+        set_spawner(executor.spawner());
+
+        spawn(async move {
+            let project_ctrl = controller::project::Handle::new_running(transport_clone);
+            let path         = Path(format!("test.{}", constants::LANGUAGE_FILE_EXTENSION));
```
mwu-tow: Why not use module::Location("Test".into()).to_path()?

We shouldn't hardcode provisional module->path logic in more than a single place.

lib/ide/src/controller/text.rs
```diff
@@ -14,7 +15,7 @@ use flo_stream::Publisher;
 use flo_stream::Subscriber;
 use json_rpc::error::RpcError;
 use shapely::shared;
-
+use basegl::display::shape::text::text_field::TextChangedNotification;
```
mwu-tow: this should go upper

lib/ide/src/controller/module.rs
```diff
@@ -4,20 +4,48 @@
...
 
 use crate::prelude::*;
 
+use crate::controller::FallibleResult;
+use crate::double_representation::apply_code_change_to_id_map;
+
+use ast::Ast;
+use ast::HasRepr;
+use ast::IdMap;
+use basegl::display::shape::text::text_field::TextChangedNotification;
```
mwu-tow: I don't like that controller is using something from view.
Can we make this type part of controller API, so always it is only view that uses controller?

lib/ide/src/controller/text.rs
```diff
@@ -88,18 +89,36 @@ impl Handle {
...
-            FileHandle::PlainText {path,mut file_manager} => file_manager.read(path),
-            FileHandle::Module {..}                       => todo!(),
+            PlainText {path,mut file_manager} => file_manager.read(path).await,
+            Module    {controller}            => Ok(controller.code())
         }
     }
 
     /// Store the given content to file.
-    pub fn store_content(&self, content:String) -> impl Future<Output=Result<(),RpcError>> {
+    pub async fn store_content(&self, content:String) -> FallibleResult<()> {
```
mwu-tow: Do we want to change this?
I believe that this keeps self borrowed for the whole duration of the call (that after obtaining handle to file manager before waiting, doesn't need it anymore)

Previously I had trouble with your version of the signature here,

lib/ide/src/double_representation.rs
```diff
@@ -0,0 +1,79 @@
...
+use data::text::Size;
+use data::text::Span;
+
+
+
+// ================
+// === Text API ===
+// ================
+
+/// Update IdMap to reflect the recent code change.
```
mwu-tow: If this is minimal, provisional algorithm, I'd like to see a TODO note with information what should be expected to work and what roughly is remaining.

lib/ide/src/double_representation.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        apply_code_change_to_id_map(&mut id_map, &Span::new(Index::new(12), Size::new(2)), "x");
+        let expected = IdMap(vec!
+            [ (Span::new(Index::new(0) , Size::new(3)), uuid1)
+            , (Span::new(Index::new(5) , Size::new(8)), uuid2)
+            , (Span::new(Index::new(14), Size::new(2)), uuid5)
+            ]);
+        assert_eq!(expected, id_map);
+    }
+}
```
mwu-tow: missing newline

lib/ide/src/double_representation.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    #[test]
+    fn applying_code_changes_to_id_map() {
+        let uuid1 = Uuid::new_v4();
+        let uuid2 = Uuid::new_v4();
+        let uuid3 = Uuid::new_v4();
+        let uuid4 = Uuid::new_v4();
+        let uuid5 = Uuid::new_v4();
+        let mut id_map = IdMap(vec!
+            [ (Span::new(Index::new(0) , Size::new(3)), uuid1)
```
mwu-tow: Not a request for change — it would take more time and perhaps it's not worthy of that at all…  but this indices are to me really unreadable and unmaintainable. I have no idea what's going on without spending much time on counting the characters.

In somewhat similar situation in Scala I've introduced a mini-dsl notation for program sourcecode that allowed annotating it with necessary information.

lib/core/src/display/shape/text/text_field.rs
```diff
@@ -27,6 +27,25 @@ use crate::display::world::World;
...
+
+/// A notification about text change.
+///
+/// In essence, it's `TextChange` with some additional useful information.
+#[derive(Clone,Debug,Shrinkwrap)]
+pub struct TextChangedNotification {
+    /// A change which has occurred.
+    #[shrinkwrap(main_field)]
+    pub change : TextChange,
+    /// The replaced range as char positions from document begin, instead of row:column pairs.
```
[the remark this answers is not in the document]
farmaazon: I don't agree. Because this struct is not a _class_ in terms of OOP - is just a data structure. There are no functions which are requiring this invariant so far. If one receives such structure as notification and change one field - it's only his responsibility

lib/core/src/display/shape/text/text_field.rs
```diff
@@ -143,10 +165,20 @@ shared! { TextField
...
         }
 
+        /// Discards all current content and replaces it with new one.
+        /// Whenever possible, tries to maintain cursor positions.
+        pub fn set_content(&mut self, text:&str) {
+            // FIXME [ao] It should check if cursors positions are valid.
+            //       See: https://github.com/luna/ide/issues/187
+            self.content.set_content(text);
+            self.assignment_update().update_after_text_edit();
+            self.rendered.update_glyphs(&mut self.content);
```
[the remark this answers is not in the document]
farmaazon: This is to be refactored anyway, as a part of task https://github.com/luna/ide/issues/178 I'll add a comment then

lib/core/src/display/shape/text/text_field.rs
```diff
@@ -246,28 +237,100 @@ impl TextField {
...
+
+impl TextField {
+    /// Edit text.
+    ///
+    /// All the currently selected text will be removed, and the given string will be inserted
+    /// by each cursor.
+    pub fn write(&self, text:&str) {
+        let trimmed                 = text.trim_end_matches('\n');
+        let cursor_ids              = self.with_borrowed(|this| this.cursors.sorted_cursor_indices());
+        let is_line_per_cursor_edit = trimmed.contains('\n') && cursor_ids.len() > 1;
```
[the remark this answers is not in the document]
farmaazon: It checks if we write one line per cursor (like multicursor pasting) or just put the whole text on each cursor.

You can see it in the usage a few lines below

lib/ide/src/controller/module.rs
farmaazon: because then you must manually run scala parser instance.

lib/ide/src/controller/project.rs
farmaazon: I agree. But I also would not make a fixture running scala parser as a part of this already big task.

lib/ide/src/controller/project.rs
farmaazon: But this is test, isn't it? And we need anyway testing when a file is treated as module, and when not.

lib/core/src/display/shape/text/text_field.rs
```diff
@@ -286,17 +349,17 @@ impl TextFieldData {
...
-            let replaced = location_change.apply_to_range(cursor.selection_range());
-            let change   = TextChange::replace(replaced,to_insert);
-            location_change.add_change(&change);
-            *cursor = Cursor::new(change.inserted_text_range().end);
-            self.content.apply_change(change);
-        }
+    fn apply_one_cursor_change
+    (&mut self, location_change:&mut TextLocationChange, cursor_id:usize, to_insert:&str)
+    -> TextChangedNotification {
+        let cursor         = &mut self.cursors.cursors[cursor_id];
```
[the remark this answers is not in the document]
farmaazon: but if I would use get I would unwrap it anyway. Because this is an issue of application, so I'd like to be informed about it

lib/core/src/display/shape/text/text_field/content.rs
```diff
@@ -166,10 +171,12 @@ impl TextChange {
 #[derive(Debug)]
 #[allow(missing_docs)]
 pub struct TextFieldContent {
-    pub lines       : Vec<Line>,
-    pub dirty_lines : DirtyLines,
-    pub font        : FontHandle,
-    pub line_height : f32,
+    pub dirty_lines  : DirtyLines,
```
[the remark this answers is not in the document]
farmaazon: Aren't they self-explaining?

lib/core/src/display/shape/text/text_field/content.rs
```diff
@@ -260,17 +290,48 @@ impl TextFieldContent {
...
+        }
+        if self.line_offsets.len() <= location.line {
+            let required_len           = location.line + 1;
+            let mut current_char_index = *self.line_offsets.last().unwrap();
+            for line in self.lines.iter().take(required_len).skip(self.line_offsets.len()-1) {
+                current_char_index += line.len() + 1;
+                self.line_offsets.push(current_char_index);
+            }
+        }
+        self.line_offsets[location.line] + location.column
```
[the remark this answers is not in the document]
farmaazon: It is public. But public apis also can panic, if user uses them wrongly? I add an info to docs about potential panics.

lib/core/src/display/shape/text/text_field/content.rs
farmaazon: It always needed when you use fonts, and each creation of content require font

lib/ide/src/controller/project.rs
farmaazon: But where it creates a new text file?

The main goal of this function is to provide TextController. TextController for module requires ModuleController, that's why we open it. For other files, it requires just file manager with path, so why we should open anything?

farmaazon: Well, I renamed the "open_module" to "get_module_controller" and similarly "open_text_file" to "get_text_controller"

lib/ide/src/controller/module.rs
farmaazon: No, because this is actually a basegl structure. And basegl should know nothing about IDE views and controllers.

lib/ide/src/controller/text.rs
farmaazon: You didn't end a sentence, did you?

And you are wrong. The only borrow in inside `self.file_handle()` call. Once that function returns a cloned_ref file handle, borrow is no more

lib/core/src/display/shape/text/text_field.rs
mwu-tow: Had a call: Deferred until future discussion about how data::text is used is concluded.

lib/core/src/display/shape/text/text_field.rs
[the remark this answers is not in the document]
mwu-tow: Would be nice to have clarifying comment them.

lib/ide/src/controller/module.rs
mwu-tow: Let's try moving this to data crate.

PR #225, opened 2020-02-29
Definition Discovery

mwu-tow:
### Pull Request Description
NOTE: This PR is missing still some functionalities. The file ` lib/ide/src/graph_api.rs` should be considered [WIP] (and likely will be named differently).

Other parts should be mostly review-ready. They include:
* utilities to work with AST, in significant part ported from Scala
* moved some functions from parser tests (and similar) to allow their reuse in other tests that need to use parser
* [WIP] definition discovery is missing tests (and, as such, there might be issues)
* [WIP] tests should be fixed for CI (marked as wasm tests, not native)

The missing piece is locating definition and retrieving its contents.

### Checklist

- [ ] All code conforms to the [Rust](https://github.com/luna/enso/blob/master/doc/rust-style-guide.md), [Scala](https://github.com/luna/enso/blob/master/doc/scala-style-guide.md), [Java](https://github.com/luna/enso/blob/master/doc/java-style-guide.md) or [Haskell](https://github.com/luna/enso/blob/master/doc/haskell-style-guide.md) style guides as appropriate.


Cargo.toml
```diff
@@ -28,7 +28,7 @@ members = [
 ]
 
 [profile.dev]
-opt-level = 2
+opt-level = 0
```
wdanilo: nooooooo

lib/ide/ast/impl/Cargo.toml
```diff
@@ -10,6 +10,8 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 derive_more  = { version = "0.15.0"                             }
 failure      = { version = "0.1.5"                              }
+lazy_static =  { version = "1.4.0"                              }
+regex        = { version = "1"                                  }
```
wdanilo: I believe this increases our package size drastically. I was trying to include it several times already and always I was reverting it. Why do you need it for?

lib/ide/ast/impl/src/assoc.rs
```diff
@@ -0,0 +1,85 @@
...
+use regex::Regex;
+
+/// Operator associativity.
+#[derive(Clone,Copy,Debug,PartialEq)]
+pub enum Assoc {
+    /// Operator left-associative, i.e. a~b~c is (a~b)~c
+    Left,
+    /// Operator right-associative, i.e. a~b~c is a~(b~c)
+    Right
+}
```
wdanilo: remove doc comments. In trivial cases we prefer to use `allow(missing_docs)` than just make the code unreadable.

lib/ide/ast/impl/src/known.rs
```diff
@@ -0,0 +1,130 @@
...
+            Err(error_matching)
+        } else {
+            Ok(KnownAst(ast,default()))
+        }
+    }
+}
+
+impl<T,E> Deref for KnownAst<T>
+where for<'t> &'t Shape<Ast> : TryInto<&'t T,Error=E>,
+                           E : Debug, {
```
wdanilo: in such a case just do not use alignment

lib/ide/ast/impl/src/known.rs
```diff
@@ -0,0 +1,130 @@
...
+    type Target = T;
+    fn deref(&self) -> &Self::Target {
+        let result = self.0.shape().try_into();
```
wdanilo: I really dont like using `.0`. It always feel hacky. Luckily, you can implement `HasContent` and `RefContent` from prelude to have method `.content()` here :) (and many more!)

lib/ide/ast/impl/src/known.rs
```diff
@@ -0,0 +1,130 @@
...
+pub type TextUnclosed  = KnownAst<crate::TextUnclosed<Ast>>;
+pub type Prefix        = KnownAst<crate::Prefix<Ast>>;
+pub type Infix         = KnownAst<crate::Infix<Ast>>;
+pub type SectionLeft   = KnownAst<crate::SectionLeft<Ast>>;
+pub type SectionRight  = KnownAst<crate::SectionRight<Ast>>;
+pub type SectionSides  = KnownAst<crate::SectionSides<Ast>>;
+pub type Module        = KnownAst<crate::Module<Ast>>;
+pub type Block         = KnownAst<crate::Block<Ast>>;
+pub type Match         = KnownAst<crate::Match<Ast>>;
+pub type Ambiguous     = KnownAst<crate::Ambiguous>;
```
wdanilo: This should be autogenerated. Talk with me to discuss how

lib/ide/ast/impl/src/opr.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use crate::Ast;
+use crate::assoc::Assoc;
+use crate::known;
+use crate::Shape;
+
+pub mod predefined {
+    /// Used to create bindings, e.g. `add a b = a + b` or `foo = 5`.
+    pub const ASSIGNMENT : &str = "=";
+    /// Used to create type paths (like
```
wdanilo: doc

lib/ide/ast/impl/src/opr.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    pub fn flatten(&self) -> Chain {
+        let target = self.target_operand();
+        let rest   = (self.opr.clone(),self.argument_operand());
+
+        let target_subtree_infix = target.clone().and_then(|ast| GeneralizedInfix::try_new(&ast));
+        let mut target_subtree_flat  = match target_subtree_infix {
+            Some(target_infix) if target_infix.name() == self.name() =>
+                target_infix.flatten(),
+            _ =>
```
wdanilo: - nl

lib/ide/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-ast                      = { version = "0.1.0"  , path = "ast/impl"         }
+ast                      = { version = "0.1.0"   , path = "ast/impl"        }
```
farmaazon: Got some dealignment here

lib/ide/ast/impl/src/known.rs
farmaazon: If one does not like `.0` you can write:
```rust
let KnownAst(ast,_) = self
```
It's more expressive. For me `self.content()` would look just, hmm, "ugly". If I would not know the code I would wonder "what is the content of Ast?". And when I alt+click to content I get to definition of content in some HasContent abstraction which explains exactly nothing.

That's why I'm against the abstractions like `HasContent`

lib/ide/ast/impl/src/known.rs
```diff
@@ -0,0 +1,130 @@
...
+
+
+
+// =============
+// === Tests ===
+// =============
+
+#[cfg(test)]
+mod tests {
+    #[allow(unused_imports)] // [MWU] otherwise I get false warning.
```
farmaazon: Are there warnings actually false? As I see you indeed don't use the import. (e.g. you have `super::Var` instead of just `Var`)

lib/ide/ast/impl/src/known.rs
```diff
@@ -0,0 +1,130 @@
...
+// =================
+
+/// Wrapper for an AST node of known shape type that we can access.
+/// Use `TryFrom<&Ast>` to obtain values.
+///
+/// Provides `Deref` implementation that allows accessing underlying shape `T` value.
+#[derive(Derivative)]
+#[derivative(Clone(bound=""))]
+#[derive(Debug)]
+pub struct KnownAst<T>(Ast, PhantomData<T>);
```
farmaazon: IMO We're getting a little messy here:
1. We have Shape enum for specific types AST nodes
2. Then we have `Ast` structure being polymorphic _any shape_ + some data
3. Then we create KnownAst which restores the specific types.

How I would like to write it (which IMO is more rusty, but less Haskellish):
1. specific Shape enumerations.
2. Ast<T> which adds id and span fields to T. 
3. AstNode trait with id and span getters.
4. AnyAst which is like Box<dyn AstNode>
I'm not really convinced to names, but this is just an example how to make types structure more clear.
I'm also not sure, if we really want to refactor this at our timeline - but I really need to share my thoughts about this code

lib/ide/src/graph_api.rs
```diff
@@ -0,0 +1,210 @@
+//! TODO TODO
```
farmaazon: I created double_representation module. We should call and agree some module structure.

lib/ide/ast/impl/Cargo.toml
mwu-tow: If we can trust `script/assert_max_wasm_size.sh`, then there is no visible size increase.

farmaazon: I checked it, the gui.wasm.gz from this branch is 7 bytes greater than from master.

It may be, that regex library is already pulled as dependency somewhere (perhaps by another thirdparty)

lib/ide/parser/tests/ast.rs
```diff
@@ -0,0 +1,62 @@
...
+use ast::HasRepr;
+use ast::opr;
+use ast::prefix;
+use parser::api::IsParser;
+use wasm_bindgen_test::wasm_bindgen_test;
+
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+#[wasm_bindgen_test]
+pub fn flatten_prefix_test() {
```
farmaazon: This looks wrong to me, to have tests for functions in AST crate in parser crate..

lib/ide/parser/tests/ast.rs
```diff
@@ -0,0 +1,62 @@
...
+        let ast = parser.parse(code.into(),default()).unwrap();
+        let ast = ast::test_utils::expect_single_line(&ast);
+        let flattened = opr::Chain::try_new(&ast).unwrap();
+        expect_pieces(&flattened,target,expected_pieces);
+    };
+
+    case("a+b+c",  "a",vec!["b","c"]);
+    case("a,b,c",  "c",vec!["b","a"]);
+    case("a+b*c+d","a",vec!["b*c","d"]);
+}
```
farmaazon: I see that we test only flatening in ast::opr module, and there is much more content which shoud be tested.

lib/ide/src/double_representation/definition.rs
```diff
@@ -0,0 +1,272 @@
...
+
+use crate::prelude::*;
+
+use ast::Ast;
+use ast::HasRepr;
+use ast::Shape;
+use ast::known;
+use ast::prefix;
+use ast::opr;
+
```
farmaazon: add vertical space

lib/ide/src/double_representation/definition.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// should be the binding's left-hand side.
+    ///
+    /// Returns `None` if is not name-like entity.
+    pub fn from_ast(ast:&Ast) -> Option<DefinitionName> {
+        let accessor_chain = opr::Chain::try_new_of(ast,opr::predefined::ACCESS);
+        let (extended_target,name) = match accessor_chain {
+            Some(accessor_chain) => {
+                let mut args = vec![identifier_name(&accessor_chain.target?)?];
+                for arg in accessor_chain.args.iter() {
+                    let arg_ast = arg.1.as_ref()?;
```
farmaazon: And now I have to check what the 1 means. Can we create structure with named field instead of tuple there?

lib/ide/src/double_representation/definition.rs
```diff
@@ -0,0 +1,272 @@
...
+
+impl DefinitionInfo {
+    /// Returns the definition body, i.e. Ast standing on the assignment's right-hand side.
+    pub fn body(&self) -> Ast {
+        self.ast.rarg.clone()
+    }
+}
+
+/// Tries to interpret given `Ast` as a function definition. `Ast` is expected to represent the
+/// whole line of the program.
```
farmaazon: What is "whole line of the program"?

lib/ide/parser/tests/ast.rs
mwu-tow: is assignment + generalized infix

Cargo.toml
mwu-tow: Yes, yes, yes! — as the youth would say.

After tests, it seems that generated wasm size difference (which was the reason for having opt level 2 here) is ~10% and does not block this change.
A nice side-effect is that it makes CI twice as fast.

PR #231, opened 2020-03-03
Implemented notification about text&graph changes brokering

farmaazon:
### Pull Request Description
Add possibility to subscribe for notifications about changes of one of module representations. Today, the notifications are sent after applying code changes (there are no other operations implemented yet).

### Important Notes
* Refactored tests a bit and add utilities to run test with executor, and with a transport mock.

### Checklist


lib/ide/src/transport/test_utils.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use json_rpc::test_util::transport::mock::MockTransport;
+
+
+
+#[derive(Debug)]
+pub struct TestWithMockedTransport {
+    with_executor_fixture : TestWithLocalPoolExecutor,
+    transport             : MockTransport,
+    next_response_id      : usize,
```
mwu-tow: use `json_rpc::handler::IdGenerator`

and make its `new` public

lib/ide/src/transport/test_utils.rs
```diff
@@ -0,0 +1,45 @@
...
+        self.with_executor_fixture.when_stalled(move ||
+            transport.mock_peer_message_text(format!(r#"{{
+                "jsonrpc" : "2.0",
+                "id"      : {},
+                "result"  : {}
+            }}"#,id,response_val))
+        );
+        self.next_response_id += 1;
+        self
+    }
```
mwu-tow: use our json-rpc goodies:

```suggestion
    pub fn when_stalled_send_response(&mut self, result:impl Serialize) -> &mut Self {
        let mut transport = self.transport.clone_ref();
        let id            = self.next_response_id.generate();
        let result        = json_rpc::messages::Message::new_success(id,result);
        self.with_executor_fixture.when_stalled(move ||
            transport.mock_peer_message(result)
        );
        self
    }
```

then `test.when_stalled_send_response(None);` becomes `test.when_stalled_send_response(());`, rest will just consume directly the result type. This will work with any supported json-rpc type, not just strings and units.

(haven't tested but if there are issues ping me, and I'll know how to deal with them)

lib/ide/src/transport/test_utils.rs
```diff
@@ -0,0 +1,45 @@
...
+            with_executor_fixture : TestWithLocalPoolExecutor::set_up(),
+            transport             : transport.clone_ref(),
+            next_response_id      : 0,
+        }
+    }
+
+    pub fn run_test<TestBody>(&mut self, test:TestBody) -> &mut Self
+    where TestBody : Future<Output=()> + 'static {
+        self.with_executor_fixture.run_test(test);
+        self
```
mwu-tow: Is it really useful to return self in all these methods?
Chaining is mostly useful when dealing with temporaries (like ad-hoc function arguments), but I don't think this is really the case with our test runners.

lib/ide/src/controller/text.rs
```diff
@@ -81,11 +78,17 @@ shared! { Handle
 impl Handle {
     /// Create controller managing plain text file (which is not a module).
     pub fn new_for_plain_text(path:fmc::Path, file_manager:fmc::Handle) -> Self {
-        Self::new(FileHandle::PlainText {path,file_manager})
+         Self::new(FileHandle::PlainText {path,file_manager})
```
mwu-tow:
```suggestion
        Self::new(FileHandle::PlainText {path,file_manager})
```

lib/ide/src/controller/text.rs
```diff
@@ -6,34 +6,30 @@
...
-use failure::_core::fmt::Formatter;
-use failure::_core::fmt::Error;
+use data::text::TextChangedInfo;
 use file_manager_client as fmc;
 use flo_stream::MessagePublisher;
-use flo_stream::Publisher;
 use flo_stream::Subscriber;
 use json_rpc::error::RpcError;
 use shapely::shared;
+use crate::executor::global::spawn;
```
mwu-tow: use decls order

lib/ide/src/controller/text.rs
```diff
@@ -6,34 +6,30 @@
...
-/// all subscribers read message from buffer. We don't expect much traffic on file notifications,
-/// therefore there is no need for setting big buffers.
-const NOTIFICATION_BUFFER_SIZE : usize = 36;
-
-/// A notification from TextController.
-#[derive(Clone,Debug)]
-pub enum Notification {
-    /// File contents needs to be set to the following due to synchronization with external state.
-    SetNewContent(String),
+async fn process_stream_with_handle<Stream,Weak,F>(mut stream:Stream, weak:Weak, mut function:F)
```
mwu-tow: First I thought it should be implemented using StreamExt::map, but now I'm not convinced.
Just give it a fleeting thought. Likely it is not a good idea.

mwu-tow:
In several places I left note `Note [Module's Graph]`: here it is:

Note [Module's Graph]
~~~~~~~~~~~~~~
We currently have no such term as module's graph. Graph contains a number of definitions, and each of them is a graph (and may contain any number of sub-graphs).
So on module's level it is more like "Graphs Changed" rather than "Graph changed". Only when we enter a particular definitoin and observe its graph, then "the graph changed" makes sense.

lib/ide/src/controller/notification.rs
```diff
@@ -0,0 +1,79 @@
...
+// === Double Representation Changes ===
+// =====================================
+
+// === Text ===
+
+/// A notification about changes of text representation or plain text file content.
+#[derive(Copy,Clone,Debug,Eq,PartialEq)]
+pub enum TextChanged {
+    /// A notification indicating that the whole content was changed and should be fully reloaded.
+    Entirely,
```
mwu-tow: Totally bikeshed, but I'd rather use `Invalidate` for a notification saying "everything might have changed".

lib/data/src/text.rs
```diff
@@ -256,15 +258,15 @@ impl TextChange {
...
+// ==========================
 
-/// A notification about text change.
+/// An information about text change.
 ///
-/// In essence, it's `TextChange` with some additional useful information.
+/// In essence, it's `TextChange` with some additional useful data.
 #[derive(Clone,Debug,Shrinkwrap)]
-pub struct TextChangedNotification {
+pub struct TextChangedInfo {
```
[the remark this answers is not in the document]
farmaazon: I understand your point, but I've done this change, because TextChangedNotification was very similar to notification::TextChanged, but those have a little different usages (one is from TextEditor to controllers, the other is in opposite direction.

lib/ide/src/controller/module.rs
```diff
@@ -119,27 +137,57 @@ shared! { Handle
...
     /// Create a module controller for given location.
     ///
     /// It may wait for module content, because the module must initialize its state.
-    pub async fn new(location:Location, mut file_manager:fmc::Handle, mut parser:Parser)
+    pub async fn new(location:Location, mut file_manager:fmc::Handle, parser:Parser)
     -> FallibleResult<Self> {
-        let logger  = Logger::new(format!("Module Controller {}", location));
+        let logger              = Logger::new(format!("Module Controller {}", location));
+        let path                = location.to_path();
+        let ast                 = Ast::new(ast::Blank{},None);
```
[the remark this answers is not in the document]
farmaazon: And indeed we do: we get the file contents before creating the controller (at least using the "official" constructors). But this new creates "uninitialized", because we will also need "(re)load_module" functionality, and this is for not duplicating code.

lib/ide/src/controller/module.rs
```diff
@@ -119,27 +137,57 @@ shared! { Handle
...
+        let id_map              = default();
+        let text_notifications  = default();
+        let graph_notifications = default();
+
+        file_manager.touch(path).await?;
```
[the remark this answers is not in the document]
farmaazon: But this is a mechanism we used previously. If we open the module and it does not exists, we should create it. 

Eventually we should just support "modules in memory", but this is a bit of task.

lib/ide/src/executor/test_utils.rs
```diff
@@ -0,0 +1,50 @@
...
+            test.await;
+            *is_finished_clone.borrow_mut() = true;
+        });
+        self.executor.run_until_stalled();
+        self
+    }
+
+    pub fn when_stalled<Callback>(&mut self, callback:Callback) -> &mut Self
+    where Callback : FnOnce() {
+        if !*self.is_finished.borrow() {
```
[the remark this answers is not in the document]
farmaazon: 1. But calling callbacks can always make some side-effects. And I would expect that "when_stalled" is actually about _stalled_, not finished, ergo I would expect that callback is not called if test is finished (for me finished != stalled)

lib/ide/src/controller/text.rs
farmaazon: Not, because I want to stop reading stream once we know that handle is no more. With `map` function I cannot say "stop reading this stream anymore". I could use something like "filter_map", but still this is stinky.

mwu-tow: true

lib/data/src/text.rs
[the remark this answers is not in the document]
farmaazon: Finally we agreed on call, that we should one parametrized TextChanged structure.

lib/ide/src/controller/module.rs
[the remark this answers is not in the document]
mwu-tow: Creating an empty Module manually is perfectly fine.

lib/ide/src/controller/module.rs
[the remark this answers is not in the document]
farmaazon: After call: Touch should be moved to ide setup and have a comment explaining that's a temporary solution

lib/core/src/display/shape/text/text_field/content.rs
```diff
@@ -485,20 +559,20 @@ pub(crate) mod test {
...
         let mut content      = TextFieldContent::new(text,&mock_properties());
 
-        assert_eq!(0 , content.convert_location_to_char_index(TextLocation{line:0, column:0}));
-        assert_eq!(2 , content.convert_location_to_char_index(TextLocation{line:0, column:2}));
-        assert_eq!(5 , content.convert_location_to_char_index(TextLocation{line:0, column:5}));
-        assert_eq!(6 , content.convert_location_to_char_index(TextLocation{line:1, column:0}));
-        assert_eq!(12, content.convert_location_to_char_index(TextLocation{line:1, column:6}));
-        assert_eq!(13, content.convert_location_to_char_index(TextLocation{line:2, column:0}));
-        assert_eq!(18, content.convert_location_to_char_index(TextLocation{line:2, column:5}));
+        assert_eq!(0 ,content.convert_location_to_char_index(TextLocation{line:0, column:0}).value);
```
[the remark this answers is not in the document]
farmaazon: But I think we use newtypes for _not_ having `PartialEq` with all numbers to avoid situation where someone compares index with size.

PR #264, opened 2020-03-16
Making tests on CI faster by running them in parallel

wdanilo:
### Pull Request Description
- Runs tests only on MacOS, which is fastest setup on GH Actions + saves the actions instances so we dont run out of its limits and dont get queued.
- Runs native and WASM tests in parallel to save the time.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ jobs:
 
   build:
     name: Build
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -110,10 +110,34 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2019-11-04
+          override: true
+
+      - name: Install wasm-pack
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: wasm-pack --version 0.8.1
```
farmaazon: Do we need to install wasm-pack if we run no-wasm tests?

PR #274, opened 2020-03-18
Graph API - Node position in metadata

Josef-Vonasek:
### Pull Request Description

Implements modification of node position in metadata.

### Checklist

### DIFF
--- a/src/rust/ide/src/controller/graph.rs
+++ b/src/rust/ide/src/controller/graph.rs
@@ -181,7 +181,7 @@ impl Handle {
 
     /// Removes the node with given Id.
     pub fn remove_node(&self, id:ast::ID) -> FallibleResult<()> {
-        self.module().remove_node_metadata(id);
+        self.module().pop_node_metadata(id)?;
         todo!()
     }
 
@@ -197,10 +197,11 @@ impl Handle {
 
     /// Modify metadata of given node.
     /// If ID doesn't have metadata, empty (default) metadata is inserted.
-    pub fn with_node_metadata<F>(&self, id:ast::ID, updater:F) -> FallibleResult<NodeMetadata>
-    where F : FnOnce(&mut NodeMetadata) {
-        self.module().with_node_metadata(id, updater);
-        self.node_metadata(id)
+    pub fn with_node_metadata(&self, id:ast::ID, fun:impl FnOnce(&mut NodeMetadata)) {
+        let     module = self.module();
+        let mut data   = module.pop_node_metadata(id).unwrap_or(default());
+        fun(&mut data);
+        module.set_node_metadata(id, data);
     }
 }
 
@@ -271,7 +272,7 @@ mod tests {
 
             let uid          = graph.all_node_infos().unwrap()[0].id();
 
-            graph.with_node_metadata(uid, |data| data.position = Some(pos)).unwrap();
+            graph.with_node_metadata(uid, |data| data.position = Some(pos));
 
             assert_eq!(graph.node_metadata(uid).unwrap().position, Some(pos));
         })
--- a/src/rust/ide/src/controller/graph.rs
+++ b/src/rust/ide/src/controller/graph.rs
@@ -7,11 +7,11 @@
 use crate::prelude::*;
 
 pub use crate::double_representation::graph::Id;
+use crate::controller::module::NodeMetadata;
 
 use flo_stream::MessagePublisher;
 use flo_stream::Subscriber;
 use utils::channel::process_stream_with_handle;
-
 
 
 // ==============
@@ -28,10 +28,6 @@ pub struct NodeNotFound(ast::ID);
 // ============
 // === Node ===
 // ============
-
-/// TODO: replace with usage of the structure to be provided by Josef
-#[derive(Clone,Copy,Debug)]
-pub struct NodeMetadata; // here goes position
 
 /// Description of the node with all information available to the graph controller.
 #[derive(Clone,Debug)]
@@ -184,7 +180,8 @@ impl Handle {
     }
 
     /// Removes the node with given Id.
-    pub fn remove_node(&self, _id:ast::ID) -> FallibleResult<()> {
+    pub fn remove_node(&self, id:ast::ID) -> FallibleResult<()> {
+        self.module().remove_node_metadata(id);
         todo!()
     }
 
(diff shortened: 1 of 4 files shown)


src/rust/ide/src/controller/module.rs
```diff
@@ -42,7 +53,7 @@ use shapely::shared;
 pub struct Metadata {
     /// Metadata used within ide.
     #[serde(default="default")]
-    pub ide : serde_json::Value,
+    pub ide : IdeMetadata,
```
farmaazon: I would name it IDEMetadata

src/rust/ide/src/controller/module.rs
```diff
@@ -170,6 +203,23 @@ shared! { Handle
         pub fn subscribe_graph_notifications(&mut self) -> Subscriber<notification::Graphs> {
             self.graph_notifications.subscribe()
         }
+
+        /// Returns metadata for given node, if present.
+        pub fn node_metadata(&mut self, id:ast::ID) -> FallibleResult<NodeMetadata> {
+            self.module.metadata.ide.node.get(&id).cloned()
+                .ok_or_else(|| NodeMetadataNotFound(id).into())
```
farmaazon: According to our guide we should avoid to make mutliline expressions, rather create aliases for each step for such chain.
```suggestion
            let metadata = self.module.metadata.ide.node.get(&id).cloned();
            metadata.ok_or_else(|| NodeMetadataNotFound(id).into())
```

src/rust/ide/src/controller/module.rs
```diff
@@ -170,6 +203,23 @@ shared! { Handle
...
+        }
+
+        /// Modify metadata of given node.
+        /// If ID doesn't have metadata, empty (default) metadata is inserted.
+        pub fn with_node_metadata(&mut self, id:ast::ID, fun:impl FnOnce(&mut NodeMetadata)) {
+            fun(self.module.metadata.ide.node.entry(id).or_default());
```
farmaazon: There is a little danger here: we are calling _some_ callback while having borrowed module controller state. So the user cannot make any operation which potentially may borrow the same module controller again. Usually this is hard to guarantee (we should check all calls for "with_node_metadata").

I propose to implement this function in Handle implementation (not inside shared! macro) and do it this way:
1. borrow data take the metadata entry from the map.
2. without borrow call the callback.
3. borrow and put the modified ast back

src/rust/ide/src/controller/module.rs
```diff
@@ -51,6 +62,28 @@ pub struct Metadata {
...
+    /// Position in x,y coordinates.
+    pub position: Option<Position>
+}
+
+/// Used for storing node position.
+#[derive(Clone,Copy,Debug,PartialEq,Serialize,Deserialize)]
+pub struct Position {
+    /// Vector storing coordinates of the visual position.
+    pub vector:Vector2<f32>
+}
```
farmaazon: This struct looks like a little bit more complicated that it needs to be. I would just define position as a Vector2<f32>, or, if you want to have a separate type for it, I would make it as a unnamed field. possibly with Shrikwrap to have direct access to x and y.
```suggestion
pub struct Position(Vector2<f32>);
```

src/rust/ide/src/controller/module.rs
Josef-Vonasek: I thought we don't use this naming in this repo.

mwu-tow: I think the recommended one is currently `IdeMetadata`, as Wojciech recently requested uniforming `ID` -> `Id`. Should be confirmed.

src/rust/ide/src/controller/graph.rs
```diff
@@ -194,17 +191,16 @@ impl Handle {
...
 
-    /// Update metadata for the given node.
-    pub fn update_node_metadata<F>(&self, _id:ast::ID, _updater:F) -> FallibleResult<NodeMetadata>
+    /// Modify metadata of given node.
+    /// If ID doesn't have metadata, empty (default) metadata is inserted.
+    pub fn with_node_metadata<F>(&self, id:ast::ID, updater:F) -> FallibleResult<NodeMetadata>
     where F : FnOnce(&mut NodeMetadata) {
-        todo!()
+        self.module().with_node_metadata(id, updater);
+        self.node_metadata(id)
```
mwu-tow: I'm confused why this method returns the metadata, while the module's one doesn't. I think they should either behave the same or be named differently.

src/rust/ide/src/controller/module.rs
mwu-tow: This is exact revert of the change previously requested by @wdanilo .
(personally, I'm fine with newtype-like single tuples, when the name cannot really introduce new information)

src/rust/ide/src/controller/module.rs
mwu-tow: Good point @farmaazon  👍

src/rust/ide/src/controller/module.rs
Josef-Vonasek: @mwu-tow Explained precisely why this design was chosen.

src/rust/ide/src/controller/graph.rs
```diff
@@ -257,6 +255,29 @@ mod tests {
...
+                (location,code,idmap,file_manager,parser).unwrap();
+            let pos          = module::Position {vector:Vector2::new(0.0,0.0)};
+            let crumbs       = vec![DefinitionName::new_plain("main")];
+            let graph        = graph::Handle::new(module, Id {crumbs}).unwrap();
+
+            let uid          = graph.all_node_infos().unwrap()[0].id();
+
+            graph.with_node_metadata(uid, |data| data.position = Some(pos));
+
+            assert_eq!(graph.node_metadata(uid).unwrap().position, Some(pos));
```
wdanilo: why these newlines?

src/rust/ide/src/controller/graph.rs
```diff
@@ -257,6 +255,29 @@ mod tests {
...
+            let location     = module::Location("Test".to_string());
+            let code         = "main = Hello World";
+            let idmap        = default();
+            let module       = module::Handle::new_mock
+                (location,code,idmap,file_manager,parser).unwrap();
+            let pos          = module::Position {vector:Vector2::new(0.0,0.0)};
+            let crumbs       = vec![DefinitionName::new_plain("main")];
+            let graph        = graph::Handle::new(module, Id {crumbs}).unwrap();
+
+            let uid          = graph.all_node_infos().unwrap()[0].id();
```
wdanilo: What is this `[0]` ? Thsi should be refactored to some variables telling what this 0 is for.

wdanilo: Plus, iss there a guarantee that this unwrap and [0] will not fail? If so, what is the guarantee? We should nto have non-total code.

src/rust/ide/src/controller/graph.rs
```diff
@@ -257,6 +255,29 @@ mod tests {
         }
     }
 
+    #[wasm_bindgen_test]
+    fn node_operations() {
+        TestWithLocalPoolExecutor::set_up().run_test(async {
+            let transport    = MockTransport::new();
+            let file_manager = file_manager_client::Handle::new(transport);
+            let parser       = Parser::new().unwrap();
```
wdanilo: Why Parser constructor can fail?

src/rust/ide/src/controller/graph.rs
```diff
@@ -257,6 +255,29 @@ mod tests {
...
+            let location     = module::Location("Test".to_string());
+            let code         = "main = Hello World";
+            let idmap        = default();
+            let module       = module::Handle::new_mock
+                (location,code,idmap,file_manager,parser).unwrap();
+            let pos          = module::Position {vector:Vector2::new(0.0,0.0)};
+            let crumbs       = vec![DefinitionName::new_plain("main")];
+            let graph        = graph::Handle::new(module, Id {crumbs}).unwrap();
```
wdanilo: Why this can fail? In what situation it can happen and what we do then? We should remove untotal functions here.

src/rust/ide/src/controller/module.rs
wdanilo: `IdeMetadata` is correct.

PR #277, opened 2020-03-18
Cache dependencies for CI.

Josef-Vonasek:
### Pull Request Description

This PR introduces cache for github actions. 

The job `Check` is now mac-only, to deal with the limited size of cache (5GB).

Wasm-pack is now download from github (faster by 3.5 minutes compared to cargo install).

Clippy was replaced with `clippy-review` due to https://github.com/rust-lang/rust-clippy/issues/4612.

There is also some issue with cache mechanism: https://github.com/actions/cache/issues/227. But so far it wasn't very frequent (it happened once in 15 runs).

### Checklist


.github/workflows/build.yml
```diff
@@ -142,17 +161,36 @@ jobs:
...
         with:
           toolchain: nightly-2019-11-04
           override: true
 
       - name: Install wasm-pack
-        uses: actions-rs/cargo@v1
-        with:
-          command: install
-          args: wasm-pack --version 0.8.1
+        env:
```
farmaazon:
```suggestion
        # Downloading takes 1 sec. cargo install wasm-pack takes 3.5 minutes.
        env:
```

build/run.js
```diff
@@ -183,7 +183,7 @@ commands.test.rust = async function(argv) {
 
 commands.lint = command(`Lint the codebase`)
 commands.lint.rust = async function() {
-    await run('cargo',['clippy','--','-D','warnings'])
+    await run('cargo',['clippy-preview','-Z','unstable-options','--','-D','warnings'])
```
farmaazon:
```suggestion
    // We run clippy-preview due to https://github.com/rust-lang/rust-clippy/issues/4612
    await run('cargo',['clippy-preview','-Z','unstable-options','--','-D','warnings'])
```

.github/workflows/build.yml
```diff
@@ -1,6 +1,8 @@
-name: Build (MacOS, Linux, Windows)
+name: Package (MacOS, Linux, Windows)
```
wdanilo: Why this name change? I believe that build is better for a step which builds and packages the app.

wdanilo: Moreover it will change the name on the badge in docs - which is bad. The badge sohuld be build, not "package OK".

.github/workflows/check.yml
```diff
@@ -0,0 +1,167 @@
...
+  #       uses: actions-rs/grcov@v0.1
+  #
+  #     - name: Show the coverage report
+  #       run: cat ${{ steps.coverage.outputs.report }}
+  #
+  #     - name: Upload the coverage report
+  #       uses: codecov/codecov-action@v1.0.2
+  #       with:
+  #         token: ${{ secrets.CODECOV_TOKEN }}
+  #         file: ./lcov.info
```
wdanilo: nl

.github/workflows/check.yml
```diff
@@ -0,0 +1,167 @@
+name: Check
```
wdanilo: This division is bad. None of the actions below check if the JS builds. The JS step can be tested only after WASM binary is generated, so we need the full "build" command here. How would it change the times if we would run `run build` next to scripts below? If it will change the times drastically we can live without JS being tested on CI - the chance it will not work are so small that wecan ignore them (it will onlyc check if dependencies resolve because js code is uncheckable).

PR #294, opened 2020-03-24
Update Contributing Guide to address test suite requirements.

MichaelMauderer:
### Pull Request Description
This PR updates the Contributing Guide to mention the Chrome requirements and potential Windows issues for running the test suite.

Relates to issue #293.

### Checklist


docs/contributing.md
```diff
@@ -76,8 +76,15 @@ scripts which maximally automate the process:
   usage.
   
 - **Testing**
+  For the test suite to run you need a current version of Chrome installed.
   Use `node ./run test` run both unit and web-based visual test.
   
+  - *Note for Windows users*:
+    there is a [known issue with wasm-pack](https://github.com/rustwasm/wasm-pack/issues/611) using the
```
wdanilo, line 83: The mentioned issue is closed. Does it mean it was fixed or considered not to be an issue?

docs/contributing.md
```diff
@@ -76,8 +76,15 @@ scripts which maximally automate the process:
...
+    wrong version of the chrome driver. There is
+    [a workaround](https://github.com/rustwasm/wasm-pack/issues/611#issuecomment-522093207) described in 
+    the issue. The easiest fix is to run `npm install -g chromedriver`.
```
wdanilo, line 86: I would not call a global installation of a dependency to be a fix. We should not encourage anyone to install deps globally in their system (for some people it could not even be possible btw if they do not have admin rights).

docs/contributing.md
MichaelMauderer: I think they fixed an issue in their tests, and I can run those just fine. 
Our test suite fails, however, if I don't have a manually installed chrome driver. 
The `wasm-pack` version on my system seems current though (`0.8.1`).

I think this is how we solve it in our CI: https://github.com/luna/ide/commit/70f082cdd5bda579990af82819144567e2bbc5f6
This also seems to download and install the working version of the chrome driver binary.

From the issue in the wasm-pack repo, it seems that both @farmaazon  and @mwu-tow  also ran into this at some point.

Maybe it is an issue with our test suite instead?

docs/contributing.md
MichaelMauderer: I can change it to recommend the more generic "install a compatible chrome driver and ensure it's in your `PATH`" advice instead.

docs/contributing.md
wdanilo: I might be wrong but I feel its because e we did not update `wasm-pack` on our side and we cant now cause we are patching it. Is that the case @MichaelMauderer @farmaazon ?

docs/contributing.md
mwu-tow: TBH I would strongly prefer the previous advice. It gives me at least actionable item. I can copy-paste command and proceed. With the current wording, I can't.

Can we keep both of them ? (if needed just add warning that npm command affects user's global state)

> (for some people it could not even be possible btw if they do not have admin rights).
Does isntalling npm package can ever require admin rights? The storage is user-specific, at least on Windows.

MichaelMauderer: I think having either both of them or none and just referring to the other issue, might be a good solution. The `npm` one is definitly the quickest way to resolve this.

PR #310, opened 2020-03-26
Add port drawing.

MichaelMauderer:
### Pull Request Description
Adds functionality to draw ports and some related utilities.
Related to #210  

### Important Notes
There is a demo scene that can be accessed on `http://localhost:8080/debug/ports`

This contains the first instance of a compound shape. We should evaluate whether the chosen API can be used as a template for others. 

### Checklist

### DIFF
--- a/src/rust/ensogl/src/display/shape/compound/port.rs
+++ b/src/rust/ensogl/src/display/shape/compound/port.rs
@@ -102,7 +102,6 @@ impl Port{
         if show_debug{
             let debug_circle = circle_inner.fill(Srgba::new(0.1,0.1,0.1,0.5));
             (triangle_rounded + debug_circle).into()
-
         } else{
             triangle_rounded.into()
         }
@@ -133,12 +132,11 @@ impl Port{
 
         let circle_outer = Circle(outer_radius.px());
 
-        let triangle_rounded = Intersection(triangle circle_outer.clone()).fill(spec.color);
+        let triangle_rounded = Intersection(triangle,circle_outer.clone()).fill(spec.color);
 
         if show_debug{
             let debug_circle = circle_outer.fill(Srgba::new(0.1,0.1,0.1, 0.5));
             (triangle_rounded + debug_circle).into()
-
         } else{
             triangle_rounded.into()
         }
--- a/src/rust/ensogl/src/display/shape/compound/port.rs
+++ b/src/rust/ensogl/src/display/shape/compound/port.rs
@@ -1,0 +1,146 @@
+//! This module defines the shapes required for drawing node ports.
+
+use crate::data::color::*;
+use crate::display::shape::*;
+use crate::display::shape::primitive::def::modifier::immutable::*;
+use crate::display::shape::primitive::def::primitive::*;
+use crate::prelude::*;
+
+use crate::display::shape::primitive::def::class::ShapeOps;
+use crate::math::circle_segment::CircleSegment;
+use crate::math::topology::unit::AngleOps;
+use crate::math::topology::unit::Distance;
+use crate::math::topology::unit::PixelDistance;
+use crate::math::topology::unit::Pixels;
+use crate::math::topology::unit::{Angle, Degrees};
+
+use nalgebra::Vector2;
+
+
+
+// ===========================
+// === Port Specification ===
+// ===========================
+
+/// Indicates whether a port is incoming or outgoing.
+#[derive(Clone,Copy,Debug,Eq,PartialEq)]
+pub enum PortDirection {
+    /// Indicates port facing towards the center of its inner circle.
+    Inwards,
+    /// Indicates port facing away from the center of its inner circle.
+    Outwards,
+}
+
+
+/// Defines the properties of a port shape and can then
+/// be used to build the port shape.
+///
+/// Ports are constructed around an inner circle, and thus
+/// most measurements are in degrees, which are measured around
+/// a inner circle, that is defined by the `inner_radius`.
+#[derive(Clone,Copy,Debug)]
+pub struct PortSpecification{
+    /// Height of the port.
+    pub height      : f32,
+    /// Width of the port in degrees.
+    pub width       : Angle<Degrees>,
+    /// Radius of the inner circle that the port is constructed around.
+    pub inner_radius: f32,
+    /// Direction the port is facing.
+    pub direction   : PortDirection,
+    /// Location of the port along the inner circle.
+    pub location    : Angle<Degrees>,
+    /// Fill colour of the port.
+    pub color       : Srgb<f32>,
+}
+
+
+// ============
+// === Port ===
+// =============
+
+#[derive(Clone,Copy,Debug)]
+/// Dummy struct for port construction.
+pub struct Port;
+
+impl Port{
+    /// Construct a port according to the given `PortSpecification`.
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new(spec:PortSpecification) -> AnyShape{
+        let show_debug = false;
+        match spec.direction{
+            PortDirection::Inwards  => Port::new_port_inwards(spec, show_debug),
+            PortDirection::Outwards => Port::new_port_outwards(spec, show_debug),
+
+        }
+    }
+
+    /// Construct an outwards facing port.
+    fn new_port_outwards(spec:PortSpecification, show_debug:bool) -> AnyShape{
+        debug_assert_eq!(spec.direction,PortDirection::Outwards);
+
+        let inner_radius = spec.inner_radius;
+
+        let segment = CircleSegment::new(inner_radius,spec.width.radians());
+
+        // Create the triangle (pointing up)
+        let tri_base   = inner_radius - segment.sagitta();
+        let tri_height = spec.height + segment.sagitta();
+        let tri_width  = segment.chord_length();
+
+        let triangle = Triangle(tri_width,tri_height);
+
+        // Move the triangle to its base position and rotate it to its final destination.
+        let offfset: Vector2<Distance<Pixels>> =  Vector2::new(Distance::new(0.0),Distance::new(tri_base));
+        let triangle = Translate(triangle,offfset);
+        let triangle = Rotation(triangle,spec.location.radians());
+
+        let circle_inner = Circle(spec.inner_radius.px());
+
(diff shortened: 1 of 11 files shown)


src/rust/ensogl/src/display/symbol/gpu/geometry/compound/sprite.rs
```diff
@@ -89,6 +89,16 @@ impl {
...
+    /// Modifies the rotation of the sprite.
+    pub fn mod_rotation<F:FnOnce(&mut Vector3<f32>)>(&self, f:F) {
+        self.display_object.mod_rotation(f);
+    }
+
+    /// Sets the rotation of the sprite.
+    pub fn set_rotation(&self, value:Vector3<f32>) {
+        self.display_object.set_rotation(value)
+    }
+
```
wdanilo: This is not needed anymore. `display::object` exports traits adding these methods to everything which is. display object. In fact we should remove some of the methods, like `position` as they are auto-implemented nowadays.

src/rust/ensogl/src/math.rs
```diff
@@ -1,5 +1,6 @@
 //! Root module for math-related utilities.
 
+pub mod circle_segment;
```
wdanilo: I dont really like the placement of this module. On top of the module we have `topology` and `types`. Having `circle_segment` next to it feels wrong. If this is part of geometry analysis, then it should definitely be somewhere inside of topology, like `topology::geometry::circle::segment` - lets think about it more.

src/rust/ensogl/src/math/circle_segment.rs
```diff
@@ -0,0 +1,55 @@
...
+
+// =====================
+// === CircleSegment ===
+// =====================
+
+/// Implements computations related to circle segments.
+/// For details and background on the formulas used, see
+/// https://en.wikipedia.org/wiki/Circular_segment
+#[derive(Clone,Copy,Debug,PartialEq)]
+pub struct CircleSegment{
```
wdanilo: A very minor thing, but we put spaces before `{` everywhere. Also before and after `:` in struct definitions (lines below).

src/rust/ensogl/src/math/circle_segment.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Constructor.
+    pub fn new(radius:f32, angle:Angle<Radians>) -> Self{
+        CircleSegment{radius, angle}
+    }
+
+    /// The length of the direct line between the segment end points.
+    pub fn chord_length(self) -> f32{
+        let r     = self.radius;
+        let theta = self.angle.value;
+
```
wdanilo: Why this newline? I understand the motivation, but Imo it jsut makes it harder to see where the next function begins. I would remove it. Over the time there would be less and less comments about code styling! :)

src/rust/ensogl/src/math/circle_segment.rs
```diff
@@ -0,0 +1,55 @@
...
+impl CircleSegment{
+
+    /// Constructor.
+    pub fn new(radius:f32, angle:Angle<Radians>) -> Self{
+        CircleSegment{radius, angle}
+    }
+
+    /// The length of the direct line between the segment end points.
+    pub fn chord_length(self) -> f32{
+        let r     = self.radius;
```
wdanilo: We are not using 1-letter variables if we can. This should be named `radius` instead.

src/rust/ensogl/src/math/circle_segment.rs
```diff
@@ -0,0 +1,55 @@
...
+        let theta = self.angle.value;
+
+        r * (1.0 - (theta * 0.5).cos())
+    }
+
+    /// The apothem (height) of the triangular portion.
+    pub fn apothem(self) -> f32{
+        self.radius - self.sagitta()
+    }
+
```
wdanilo: . -nl

src/rust/ensogl/src/math/topology/unit.rs
```diff
@@ -261,6 +261,24 @@ impl AngleOps for i32 {
...
+        self.clone()
+    }
+
+    fn radians(&self) -> Angle<Radians> {
+        Angle::new(self.value.to_radians())
+    }
+}
+
+impl AngleOps for Angle<Radians> {
+    fn degrees(&self) -> Angle<Degrees> { Angle::new(self.value.to_degrees()) }
```
wdanilo: Lets use multiline layout. We are using single-line layout only if we have many similar lines glued together and we want to show their similarity.

src/rust/ensogl/src/display/shape/compound/port.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+
+// ============
+// === Port ===
+// =============
+
+#[derive(Clone,Copy,Debug)]
+/// Dummy struct for port construction.
+pub struct Port;
```
wdanilo: Port should be a structure which is a display object. We should be able to move it, rotate it, etc. Here, the phantom port structure has just associated methods producing shapes which tell how ports sohuld be drawn. Drawing things is a small part of object description. We need to talk more about the proper structurization of that, but lets do it on monday. Now jsut leaving a note.

src/rust/ensogl/src/display/shape/compound/port.rs
```diff
@@ -0,0 +1,146 @@
...
+        let show_debug = false;
+        match spec.direction{
+            PortDirection::Inwards  => Port::new_port_inwards(spec, show_debug),
+            PortDirection::Outwards => Port::new_port_outwards(spec, show_debug),
+
+        }
+    }
+
+    /// Construct an outwards facing port.
+    fn new_port_outwards(spec:PortSpecification, show_debug:bool) -> AnyShape{
```
wdanilo: Why are you defining 2 shapes instead of 1 which is parametrized by the orientation (outwards / inwards)? Two separate shapes require 2 separate sprite systems, which cause 2 draw calls instead of 1 draw call. Plus, these 2 functions share a big part of very similar code.

src/rust/lib/debug-scenes/src/ports.rs
```diff
@@ -0,0 +1,99 @@
...
+        t.y += screen.height / 2.0;
+    });
+
+    let shape_system_2 =     ShapeSystem::new(world,&port_shape_2);
+    let sprite_2 = shape_system_2.new_instance();
+    sprite_2.size().set(Vector2::new(500.0,500.0));
+    sprite_2.mod_position(|t| {
+        t.x += screen.width / 2.0;
+        t.y += screen.height / 2.0;
+    });
```
wdanilo: If I understand correctly, you are drawing ports on big rectangular planes and tyou are rotating them inside of the shapes description? This is bad. Every shape shoudl be drawn on smallest possible rectangle and the rectangle should be moved - this way we do less computations in shaders.

wdanilo: Moreover there are fewthigns that you need to take care about. The input ports should have a transparent area around them which is clickable. So when user hovers the mouse over the ports, the prost will react even if the mouse is close to them, not directly above them. One more thing - when camera zooms out, the ports should grow a little bit . They should become larger. The best effect would be to imagine it this way - if a port is intersection of some circle around node and a "pie of a plane", then when zooming out the circle should grow and thus the intersections hould as well.

src/rust/lib/debug-scenes/src/ports.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    let port_shape_2 = Port::new(port_spec);
+
+    let shape_system =     ShapeSystem::new(world,&port_shape_1);
+    let sprite = shape_system.new_instance();
+    sprite.size().set(Vector2::new(500.0,500.0));
+    sprite.mod_position(|t| {
+        t.x += screen.width / 2.0;
+        t.y += screen.height / 2.0;
+    });
```
wdanilo: This sprite, display object etc should be part of "Port" structure. `Port` should be jsut an object representing port on the stage.

src/rust/ensogl/src/display/shape/compound/port.rs
```diff
@@ -0,0 +1,144 @@
...
+// ===========================
+
+/// Indicates whether a port is incoming or outgoing.
+#[derive(Clone,Copy,Debug,Eq,PartialEq)]
+pub enum PortDirection {
+    /// Indicates port facing towards the center of its inner circle.
+    Inwards,
+    /// Indicates port facing away from the center of its inner circle.
+    Outwards,
+}
```
wdanilo: In case where the names of enums are just 100% clear about what it is about, we are often putting `#[allow(missing_docs)]` on the enum declaration. In such cases docs make the code less readable and do not add anything important to it.

Moreover, why `Inwards` / `Outwards` instead of `In` and `Out` ?

src/rust/ensogl/src/display/shape/compound/port.rs
```diff
@@ -0,0 +1,144 @@
...
+
+
+
+// ===========================
+// === Port Specification ===
+// ===========================
+
+/// Indicates whether a port is incoming or outgoing.
+#[derive(Clone,Copy,Debug,Eq,PartialEq)]
+pub enum PortDirection {
```
wdanilo: We try to name things for qualified usage. So the API will be consumed as `use something::port` and then `port::PortDirection`, which does not make much sense. Thus such things should be renamed to `Direction`. I know that sometimes in our codebase the names are wrong, we are slowly fixing that.

src/rust/ensogl/src/math.rs
MichaelMauderer: I agree that this might be a better fit in a different namespace.
Would topology not sit next to topology, though? I see that a lot of geometric primitives are in the topology module, but my feeling is that topology and geometry are different things. It seems this is also how [Wikipedia structures those topics](https://en.wikipedia.org/wiki/Areas_of_mathematics#Pure_mathematics).

With that in mind, I would propose to put it into `math::geometry::circle::segment`.

wdanilo: `Would topology not sit next to topology` - I don't think so :D But regarding geometry, hmm, I need to check it out in more detail.

MichaelMauderer:
I'm closing this and will open a new Draft PR with a branch from this repo (instead of my fork).

PR #333, opened 2020-04-02
Port Drawing

MichaelMauderer:
### Pull Request Description

Adds functionality to draw ports and some related utilities.
Related to #210

### Important Notes

Incomplete at the moment.

### Checklist


src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,33 @@
+use core::f32::consts::PI;
+use crate::display::shape::Var;
+use crate::math::algebra::{Sin, Cos, Sqrt};
+use crate::math::topology::unit::{Radians, Angle};
+use std::ops::*;
+
+pub struct Triangle<S> {
```
wdanilo, line 7: I dont like this triangle definition TBH. It allows me to construct impossible triangles. Structs allowing to create impossible data are rather error prone.

src/rust/lib/graph/src/component/node/port.rs
```diff
@@ -0,0 +1,319 @@
...
+        let height       : Var<f32> = spec.height.clone().into();
+
+        let segment_width_rad: Var<f32>       = spec.width.clone().into();
+        let segment_radius   : Var<f32>       = inner_radius.clone().into();
+        let segment : CircleSegment<Var<f32>> = CircleSegment::new(segment_radius,segment_width_rad);
+
+        let tri_base             = segment.sagitta();
+        let tri_height: Var<f32> = &height + &tri_base;
+        let tri_width            = segment.chord_length();
+
```
wdanilo, line 167: I was already writing about it. I don't like that input / output ports are separate shapes. Why this design? Why they are not a single shape but parametrized?

src/rust/lib/graph/src/component/node/port.rs
```diff
@@ -0,0 +1,319 @@
...
+#[derive(Debug,Clone,Copy)]
+pub struct InputPortView {}
+impl component::ShapeViewDefinition for InputPortView {
+    type Shape = shape_in::Shape;
+    fn new(shape:&Self::Shape, _scene:&Scene, _shape_registry:&ShapeRegistry) -> Self {
+        shape.width.set(DEFAULT_WIDTH);
+        shape.inner_radius.set(DEFAULT_RADIUS);
+        shape.height.set(DEFAULT_HEIGHT);
+
+        shape.sprite.size().set(Vector2::new(200.0,200.0));
```
wdanilo, line 227: do we need such a big ports?

src/rust/lib/graph/src/component/node/port.rs
```diff
@@ -0,0 +1,319 @@
...
+    }
+}
+
+
+/// Port definition. Can be parametrised to be either
+/// an InputPort or OutputPort.
+#[derive(Debug,Clone)]
+pub struct Port<T:ShapeViewDefinition> {
+    spec         : Specification,
+    pub view     : Rc<component::ShapeView<T>>
```
wdanilo, line 253: aligment

src/rust/lib/graph/src/component/node/port.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+
+impl<T:ShapeViewDefinition> Port<T> {
+
+    pub fn new(spec:Specification) -> Self{
+        let logger = Logger::new("node");
+        let view = Rc::new(component::ShapeView::<T>::new(&logger));
```
wdanilo, line 261: aligment

src/rust/lib/graph/src/component/node/port.rs
```diff
@@ -0,0 +1,319 @@
...
+/// an InputPort or OutputPort.
+#[derive(Debug,Clone)]
+pub struct Port<T:ShapeViewDefinition> {
+    spec         : Specification,
+    pub view     : Rc<component::ShapeView<T>>
+}
+
+
+impl<T:ShapeViewDefinition> Port<T> {
+
```
wdanilo, line 258: nl

src/rust/ensogl/src/math/geometry/triangle.rs
wdanilo, line 7: Plus, where this sturcture is used? I don't see any usages now.

src/rust/lib/graph/src/component/node/port.rs
MichaelMauderer: Only temprarily. I need to implement some re-centring of the shape and calculate this dynamically, depending on the actual size of the shape. That way we can minimise the sprite size.

src/rust/ensogl/src/math/geometry/triangle.rs
MichaelMauderer: This might be used when using `Plane().cut_angle` instead of `Triangle`. I've removed it for now and will add a clean version if it is needed again.

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub fn from_sas(a:S, b:S, gamma:S) -> Triangle<S> {
+        let a2 = (a.clone() * a.clone());
+        let b2 = (b.clone() * b.clone());
+
+        let c2 = a2.clone() + b2.clone() - (S::from(2.0_f32) * a.clone() * b.clone() * gamma.cos());
+        let c = c2.sqrt();
+
+        let alpha = (b2.clone() + c2.clone() - a2.clone()) / (S::from(2.0_f32) * b.clone() * c.clone());
+
+        let beta = S::from(PI) - alpha.clone() - gamma.clone();
```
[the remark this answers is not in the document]
MichaelMauderer: I think I we don't have `Copy` for `Var<_>` types, which is why this was needed. 
Adding `Copy` to those did not feel right either, as they do copy around `String`s.

(Style will be fixed if I add this again)

src/rust/ensogl/src/math/geometry/circle/circle_segment.rs
```diff
@@ -0,0 +1,67 @@
...
+use core::fmt::Debug;
+
+
+// =====================
+// === CircleSegment ===
+// =====================
+
+/// Implements computations related to circle segments.
+/// For details and background on the formulas used, see
+/// https://en.wikipedia.org/wiki/Circular_segment
```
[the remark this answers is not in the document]
MichaelMauderer: The first line was meant as the "introductory sentence"  on its own line. While the second line is >100 chars with the link in it. I'll change it to continue after the first sentence though.

src/rust/lib/graph/src/component/node/port.rs
MichaelMauderer: The way the shapes are currently constructed, one of them needs to be a `Union` between a `Circle` and a `Triangle` and the other one a `Difference`. 
I think the only way to make this work, might be to use additional shapes like always having one `Circle` that computes the difference and another one that computes the `Union`, and then cleverly moving them around based on some flag. But that would mean more distance tests at runtime. Or maybe implementing a new `Curve` shape, that can be more flexible in its shape than the a circle. 
But both of these seem to me to be more complex and computationally expensive than having two shapes, but maybe I am missing something?

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -214,6 +220,11 @@ impl Node {
             selection_ref.set_target_position(0.0);
         });
 
+        // TODO this is sample functionality. Needs to be replaced with logic creating ports.
+        self.data.port_manager.set_parent(self.downgrade());
+        self.data.port_manager.create_input_port();
+        self.data.port_manager.create_output_port();
```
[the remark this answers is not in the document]
wdanilo: Plus, what is the API to rotate ports, now? How can I do it? Can I have a reference to a port and write `port.set_rotation(...)` ? The API should be like that.

src/rust/ensogl/src/math/geometry/circle/circle_segment.rs
```diff
@@ -0,0 +1,73 @@
...
+        self.radius.clone() - self.sagitta()
+    }
+
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    // TODO more testing
```
wdanilo, line 66: Do it before merge to master pls.

src/rust/ensogl/src/math/geometry/circle/circle_segment.rs
```diff
@@ -0,0 +1,73 @@
...
+    angle  : T,
+}
+
+impl<T> CircleSegment<T>
+where T: Field<T>
+         + Sin<Output=T>
+         + Cos<Output=T>
+         + From<f32>
+         + Clone
+         + Debug {
```
wdanilo, line 27: This should be a type alias instead of 6 lines of constraints

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -214,6 +220,11 @@ impl Node {
             selection_ref.set_target_position(0.0);
         });
 
+        // TODO this is sample functionality. Needs to be replaced with logic creating ports.
+        self.data.port_manager.set_parent(self.downgrade());
```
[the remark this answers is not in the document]
MichaelMauderer: The `PortManager` contains logic to create new ports. To be able to to that it needs to know who is supposed to be the parent of those nodes to call `parent.add_child`.

I had assumed that keeping around a `WeakNode` for that purpose would be unproblematic. It would only get upgraded when trying to add a child. Is there an issue with that?

src/rust/lib/graph-editor/src/component/node.rs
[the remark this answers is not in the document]
MichaelMauderer: Since the positioning and rotating of the ports isn't independent the idea was for the `PortManager` to handle that functionality internally.

I can refine the API to allow access to individual ports though and allow someting like `node.ports.input.add()`

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,357 @@
...
+            };
+            new_port(port_spec_val)
+        }
+    }
+}
+
+
+
+// =================
+// === Port Node ===
```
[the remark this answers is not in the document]
MichaelMauderer: This was meant to be the `Node` part of the Port implementation. It should probably be only "Port" now, I think, since that is the main struct here.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,357 @@
...
+    fn update(&mut self) {
+        self.update_sprite();
+        self.view.display_object.update();
+    }
+
+    /// Update the position of the sprite according to the Port specification.
+    /// The position is given along a circle, thus the position and rotation of the sprite
+    /// are tied together, so the Port always point in the right direction.
+    fn update_sprite(&mut self) {
+        let translation_vector = na::Vector3::new(0.0,self.spec.inner_radius,0.0);
```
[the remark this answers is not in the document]
MichaelMauderer: I think `Vector3`/`Rotation` is not in the prelude. `na` seems to be the standard prefix that is used in the general Rust community for nalgebra, but I've changed it to `nalgebra`.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,357 @@
...
+
+// ====================
+// === Port Manager ===
+// ====================
+
+/// Handles creation and layouting of ports around a node.
+/// TODO implement the layouting
+#[derive(Debug,Default)]
+pub struct PortManager {
+    parent       : RefCell<Option<WeakNode>>,
```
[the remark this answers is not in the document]
MichaelMauderer: The parent is the node that the ports belong to. This is here so we can set it as a parent to new `Port`s when they are created.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,357 @@
...
+    input_ports  : RefCell<Vec<InputPort>>,
+    output_ports : RefCell<Vec<OutputPort>>,
```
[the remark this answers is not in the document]
MichaelMauderer: Since the `PortManager` is going to be used from the `NodeData`, which can only be used immutably through the `Node`, all functionality here needs to use interior mutability, too.

src/rust/ensogl/src/math/algebra.rs
```diff
@@ -221,3 +297,14 @@ impl Into<Vector3<f32>> for Point3 {
...
+
+
+// ============================
+// === Algebraic Structures ===
+// ============================
+// TODO evaluate for correctness and usefulness.
+
+/// Trait that describes a set of numbers that define addition, subtraction, multiplication,
+/// and division.
+pub trait Field<T> = Add<T,Output=T> + Sub<T,Output=T> + Mul<T,Output=T> + Div<T,Output=T>;
```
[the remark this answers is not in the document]
MichaelMauderer: I've changed the prelude exports to be static at the current level (minus the new traits).

src/rust/ensogl/src/math/geometry/triangle.rs
MichaelMauderer: I've changed the struct so it can not be instantiated outside of this module. Here it is only meant to serve as a return type for the helper functions.

src/rust/ensogl/src/display/shape/primitive/def/var.rs
```diff
@@ -2,8 +2,9 @@
 
 use crate::prelude::*;
 
+use crate::math::algebra::{Acos,Asin,Cos,Sin,Sqrt};
 use crate::data::color;
-use crate::math::topology::unit::Unit;
+use crate::math::topology::unit::{Unit,Radians,Degrees,Distance,Angle,Pixels};
 use crate::system::gpu::shader::glsl::Glsl;
 use crate::system::gpu::types::*;
```
wdanilo, line 9: This import style is forbidden by the guide.

src/rust/ensogl/src/display/shape/primitive/def/var.rs
```diff
@@ -411,3 +412,134 @@ define_shape_data_string_operator! { Add add (+) }
...
+// ===================
+// === Square Root ===
+// ===================
+
+impl<T> Sqrt for Var<T>
+where T: Sqrt<Output=T> {
+
+    type Output = Var<T>;
+
+    fn sqrt(&self) -> Self {
```
wdanilo, line 478: I asked in a comment to previous PR to remove the spacing in simple above - they are removed, but here the spacing is still present. Could you please apply the changes here and everywhere else too? :)

src/rust/ensogl/src/display/shape/primitive/def/var.rs
```diff
@@ -411,3 +412,134 @@ define_shape_data_string_operator! { Add add (+) }
...
+            Var::Dynamic (t) => Var::Dynamic(format!("value({})",t.glsl()).into()),
+        }
+    }
+}
+
+impl From<Var<f32>> for Var<Angle<Radians>> {
+    fn from(other:Var<f32>) -> Self {
+        match other {
+            Var::Static  (t) => Var::Static(Angle::from(t)),
+            Var::Dynamic (t) => Var::Dynamic(format!("Radians({})",t.glsl()).into()),
```
wdanilo, line 506: This conversion is an internal magic. We have now this magic in two places, which is bad. This should be refactored to a function and this function should be used instead. Otherwise changing the magic in one place is an easy way to make something very bad.

src/rust/ensogl/src/math.rs
```diff
@@ -1,11 +1,50 @@
...
+    pub use super::algebra::HasComponents;
+    pub use super::algebra::Dim1;
+    pub use super::algebra::Dim2;
+    pub use super::algebra::Dim3;
+    pub use super::algebra::Abs;
+    pub use super::algebra::Magnitude;
+    pub use super::algebra::Normalize;
+
+    pub use super::algebra::Point3;
+
```
wdanilo, line 49: - nl

src/rust/ensogl/src/math/geometry/circle.rs
```diff
@@ -0,0 +1,3 @@
+//! Root module for circle-related functionality.
+
+pub mod circle_segment;
```
wdanilo, line 3: I do not like this naming tbh as we are repeating ourselves `circle::circle_segment`. it should be `circle::segment` instead.

src/rust/ensogl/src/math/geometry/circle/circle_segment.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// Implements computations related to circle segments. For details and background on the formulas
+/// used, see https://en.wikipedia.org/wiki/Circular_segment
+#[derive(Clone,Debug,PartialEq)]
+pub struct CircleSegment<T> {
+    radius : T,
+    angle  : T,
+}
+
+pub trait FloatLike<T> = Field<T> + Sin<Output=T> + Cos<Output=T> + From<f32> + Clone + Debug;
```
wdanilo, line 21: This name does not really tells what it is. If I was reading a code and I would see `FloatLike` I would never assume it means that it needs to implement `Sin` or `Cos`. Are all of these bounds required for the functions in `CircleSegment` impl? If not, only the required bounds should be left here. Not needed bounds are sometimes causing strange requirements.

src/rust/ensogl/src/math/geometry/circle/circle_segment.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /// The apothem (height) of the triangular portion.
+    pub fn apothem(self) -> T {
+        self.radius.clone() - self.sagitta()
+    }
+
+}
+
+
```
wdanilo, line 59: Plase add "Tests" section here :)

src/rust/ensogl/src/math/geometry/circle/circle_segment.rs
```diff
@@ -0,0 +1,172 @@
...
+        assert_approx_eq!(segment.apothem(), 1.0 - 0.2066467);
+
+        let segment = CircleSegment::new(1.0, 180_f32.to_radians());
+        assert_approx_eq!(segment.apothem(), 1.0 - 1.0);
+
+        let segment = CircleSegment::new(1.0, 359_f32.to_radians());
+        assert_approx_eq!(segment.apothem(), 1.0 - 1.9999619);
+
+        let segment = CircleSegment::new(1.0, 360_f32.to_radians());
+        assert_approx_eq!(segment.apothem(), 1.0 - 2.0);
```
wdanilo, line 170: amazing job with tests! ❤️

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,128 @@
...
+///                      -/ gamma --\
+///                   --/            --\
+///               b -/                  --\ a
+///              --/                       --\
+///            -/                             --\
+///          -/                                  --\
+///       --/                                       --\
+///     -/  alpha                                   beta-\
+///   -/--------------------------------------------------- B
+///  A                              c
```
wdanilo, line 27: Merry Christmas! :D

wdanilo: What do you think about changing the variables and the picture to something simpler?

```
///   C
///  / \
/// A---B
```

Plus changing names to `side_ab` (the distance between `A` and `B`), `side_bc`, `side_ac`, `angle_a`, `angle_b`, and `angle_c` ?

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,128 @@
...
+#[derive(Debug)]
+pub struct Triangle<S> {
+    pub side_length_a   : S,
+    pub side_length_b   : S,
+    pub side_length_c   : S,
+    pub angle_alpha_rad : S,
+    pub angle_beta_rad  : S,
+    pub angle_gamma_rad : S,
+    // Prevent instantiation as something other than a result type from this module.
+    dummy: (),
```
wdanilo, line 42: Oh no, dont do it. If you want to prevent initialization just don't make fields public.

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+pub trait FloatLike<S> = Field<S>
+                       + Sin<Output=S>
+                       + Cos<Output=S>
+                       + Acos<Output=S>
+                       + Sqrt<Output=S>
+                       + Clone
+                       + From<f32>;
```
wdanilo, line 51: I see it second time already.

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl<S> Triangle<S>
+where S:FloatLike<S> {
+    /// Compute a triangle from two sides and the included angle (SAS).
+    /// See https://en.wikipedia.org/wiki/Solution_of_triangles#Two_sides_and_the_included_angle_given_(SAS)
+    pub fn from_sides_and_angle
+    (side_length_a:S, side_length_b:S, angle_gamma_rad:S)
+    -> Triangle<S> {
```
wdanilo, line 59: move to previous line

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,128 @@
...
+        let a_squared = side_length_a.clone() * side_length_a.clone();
+        let b_squared = side_length_b.clone() * side_length_b.clone();
+
+        let two = S::from(2.0_f32);
+
+        let c_squared_minuend    = a_squared.clone() + b_squared.clone();
+        let c_squared_subtrahend = two.clone() * side_length_a.clone() * side_length_b.clone() * angle_gamma_rad.cos();
```
wdanilo, line 66: Please keep all lines below 100 chars

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        Triangle{
+            side_length_a,
+            side_length_b,
+            side_length_c,
+            angle_alpha_rad,
+            angle_beta_rad,
+            angle_gamma_rad,
+            dummy: ()
+        }
```
wdanilo, line 87: In such a case we format constructors this way (I believe it is as clear and takes less space, what do you thin about it?): 

```rust
Triangle
    { side_length_a,side_length_b,side_length_c,angle_alpha_rad,angle_beta_rad
    , angle_gamma_rad,...... }
```

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -179,10 +183,11 @@ impl Node {
...
 
-        let logger = Logger::new("node");
-        let view   = component::ShapeView::new(&logger);
-        let events = Events {select,deselect};
-        let data   = Rc::new(NodeData {logger,label,events,view});
+        let logger       = Logger::new("node");
+        let view         = component::ShapeView::new(&logger);
+        let events       = Events {select,deselect};
+        let ports        = PortManager::default();
+        let data         = Rc::new(NodeData {logger,label,events,view,ports});
```
wdanilo, line 190: why such alignment?

src/rust/lib/graph-editor/src/component/node/port.rs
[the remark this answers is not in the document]
wdanilo: Ok, thats right!

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+// ====================
+// === Port Manager ===
+// ====================
+
+/// Handles creation and layouting of ports around a node.
+/// TODO implement the layouting
+#[derive(Debug,Default)]
+pub struct PortManager {
+    /// The node that all ports will be placed around.
+    parent_node  : RefCell<Option<WeakNode>>,
```
wdanilo, line 354: This is very bad because:

1. It is not port's responsibility to know it's parent.
2. This creates a cycle, of course it is using `WeakNode` so the cycle is broken, but any cycle is ugly.

I cannot find in code what this is used for. What do you need it for?

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+        };
+
+        let port = Port::new(port_spec);
+        self.add_child_to_parent(&port);
+        self.output_ports.borrow_mut().push(port);
+        self.update_port_shapes();
+    }
+
+    /// Update the shapes of all ports with the currently set specification values.
+    fn update_port_shapes(& self) {
```
wdanilo, line 409: space after `&`?

src/rust/lib/graph-editor/src/component/node.rs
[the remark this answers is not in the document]
wdanilo: What you mean by `isn't independent`? I know they are placed on a circle, thus I was asking about function allowing ports to set their rotation (by providing the angle only).

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+    ///  *          *             |                   |     *    *    * ------------ inner radius
+    ///   *        *              |                   |    * *       * *     |
+    ///      *  *------------------ outer radius      |    *            *    -this is some extra
+    ///                                                                        space that depends
+    ///  \----------|                                      \------------|       on the radius
+    ///     width                                              width
+    ///
+    /// The shape description sets up both shapes and then applies only one of the variants:
+    /// either it limits of the inward facing angle with the outer ring or it cuts of the
+    /// angle with the inner ring.
```
wdanilo, line 113: WOW ❤️

wdanilo: This and the code below is purely amazing

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+/// Type that represents an input port.
+pub type InputPort = Port<InputPortView>;
+
+/// Type that represents an output port.
+pub type OutputPort = Port<OutputPortView>;
+
+/// Internal data of `Port.
+#[derive(Debug,Clone)]
+#[allow(missing_docs)]
+pub struct PortData<T:ShapeViewDefinition+Clone> {
```
wdanilo, line 280: Where the bound `Clone` comes from?

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+}
+
+impl<T:ShapeViewDefinition<Shape=shape::Shape>+Clone> Port<T> {
+
+    /// Constructor.
+    pub fn new(spec:Specification) -> Self {
+        let logger = Logger::new("node");
+        let view   = Rc::new(component::ShapeView::new(&logger));
+        let spec = RefCell::new(spec);
+        let data = PortData{spec,view};
```
wdanilo, line 292: alignment

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+        Self{data: Rc::new(data)}.init()
```
wdanilo, line 293: please do it as 

```rust
let data = ...
Self {data} . init()
```
instead.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    /// Update the position of the sprite according to the Port specification.
+    /// The position is given along a circle, thus the position and rotation of the sprite
+    /// are tied together, so the Port always point in the right direction.
+    fn update_sprite(&self) {
+        let spec = self.data.spec.borrow();
+        let translation_vector = nalgebra::Vector3::new(0.0,spec.inner_radius,0.0);
+        let rotation_vector    = -nalgebra::Vector3::new(0.0,0.0,spec.location.rad().value);
+        let rotation           = nalgebra::Rotation3::new(rotation_vector);
```
wdanilo, line 320: This code does not seem linted. prelude exports `Vector3` so this qualification is not needed and linter should tell about it.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+        let translation        = rotation * translation_vector;
+
+        let node = &self.data.view.display_object;
+        node.set_position(translation);
+        node.set_rotation(rotation_vector);
```
wdanilo, line 325: this is display object, so it inherits all API from `DisplayObjectOps` - please use this API isntead.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// Update the shape parameters and sprite location with values from our current specification.
+    fn update(&self) {
+        self.update_shape();
+        self.update_sprite();
+    }
+
+    /// Update the position of the sprite according to the Port specification.
+    /// The position is given along a circle, thus the position and rotation of the sprite
+    /// are tied together, so the Port always point in the right direction.
+    fn update_sprite(&self) {
```
wdanilo, line 316: Ok, it took me a while to understand what all of these updates do and TBH I don't like this idea. We need a simple API for rotating ports etc. A generic function "update" which goes trough all parameters and updates the view according to them is 1) very complex, 2) not very efficient. Instead, lets create API for each thing we can do with a port, like `set_rotation`. ok ?

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+#[allow(missing_docs)]
+pub struct PortData<T:ShapeViewDefinition+Clone> {
+        spec : RefCell<Specification>,
+    pub view : Rc<component::ShapeView<T>>
+}
+
+impl<T:ShapeViewDefinition<Shape=shape::Shape>+Clone> Port<T> {
+
+    /// Constructor.
+    pub fn new(spec:Specification) -> Self {
```
wdanilo, line 288: When we are designing API one thing is super important - to keep constructors simple. This constructor should not have any parameters and it should just created port with default params. Then you should have an API to modify each parameter you want separately (see other comments here)

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    /// Construct an inwards facing port.
+    ///
+    /// The port consists of an angle embedded between two rings. The corner of the angle can
+    /// either face towards the center of thr ring of away from it. If it faces towards the
+    /// center, the side facing the outer ring, have a bulging shape on that side. If the angle
+    /// faces away from the center, the angle gets limited by the inner ring and shows a convex
+    /// "hole" on that side.
+    ///
+    /// Illustrations (not to scale).
```
wdanilo, line 96: `not to scale` - ok, this made my day :D

src/rust/ensogl/src/math/geometry/triangle.rs
MichaelMauderer: I'm more used to the former, but I don't have a strong opinion in this case.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: I've added a default constructor and one that is `From<Specification>`. This function is now private.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: This is refactored now and the parent is only needed to create a new port. 
There it is used to set the new ports parent.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: There is an issue with the derive for `Clone` requiring all `T` to also implement `Clone`. In this case that is not required an I've now manually implemnted clone to avoid the bound.

src/rust/ensogl/src/math/geometry/triangle.rs
MichaelMauderer: It's shorter so I'll go for it.
I cleaned up the diagram, but I'd like to keep the side and angle descriptions in there for reference.

src/rust/ensogl/src/math/geometry/triangle.rs
MichaelMauderer: Replaced it with a struct with private fields and getters.

src/rust/ensogl/src/math/geometry/circle/circle_segment.rs
MichaelMauderer: Yes, these are specifically the bounds required here (and there is also a long, but different set of bounds in `triangle.rs`.
I don't think there is a general name for it (or any other sub-group of traits in here). I can try to be more specific, like `SegmentInput` though.

src/rust/ensogl/src/display/shape/primitive/def/var.rs
MichaelMauderer: I've refactored this code (and the one in `unit.rs`) into helper functions inside of `glrs.rs`

src/rust/lib/graph-editor/src/component/node.rs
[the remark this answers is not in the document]
MichaelMauderer: If ports are positioned around the node their position needs to be updated and their rotation needs to be updated to point in the right direction. But this is now handled inside of the `Ports` themselves.

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,169 @@
...
+///                     /    \
+///                   /       \
+///                 /  angle_c \
+///    side_ca    /             \  side_bc
+///             /                \
+///           /                   \
+///         /                      \
+///       /                         \
+///     /  angle_a           angle_b \
+///   /------------------------------- B
```
wdanilo, line 31: Its super cool now! One question - do we need the image in the docs so big now? I mean, if you see `side_ab` you know this is between corners `a` and `b` right? Big ascii-arts makes the generated docs harder to read. Not a big deal though.

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+pub trait TriangleInput<T> = Field<T>
+                           + Sin<Output=T>
+                           + Cos<Output=T>
+                           + Acos<Output=T>
+                           + Sqrt<Output=T>
+                           + Clone
+                           + From<f32>;
+
```
wdanilo, line 93: make it 2-liner please - just love several things to second line

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,169 @@
...
+        let side_bc_squared = side_bc.clone() * side_bc.clone();
+        let side_ca_squared = side_ca.clone() * side_ca.clone();
+
+        let two = T::from(2.0_f32);
+
+        let side_ab_squared_minuend    = side_bc_squared.clone() + side_ca_squared.clone();
+        let side_ab_squared_subtrahend = two.clone()
+                                         * side_bc.clone()
+                                         * side_ca.clone()
+                                         * angle_c.cos();
```
wdanilo, line 109: This is not the most efficient impl here. Consider `Var<f32>` it could be string under the hood. Cloning LHS and RHS in order to compute result and throw LHS and RHS away is not optimal. Instead you could multiply references, like `&two * &side_bc`. In context of numbers, it would be the same, but for `Var`s it would actually be faster. After these changes the expressions should get shorter and fit single lines.

src/rust/ensogl/src/math/geometry/triangle.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        let angle_a_cos_numerator   = side_ca_squared + side_ab_squared - side_bc_squared;
+        let angle_a_cos_denominator = two * side_ca.clone() * side_ab.clone();
+
+        let angle_a_cos = angle_a_cos_numerator / angle_a_cos_denominator;
+        let angle_a     = angle_a_cos.acos();
+
+        let angle_b = T::from(PI) - angle_a.clone() - angle_c.clone();
+
+        Triangle{side_bc,side_ca,side_ab,angle_a,angle_b,angle_c}
```
wdanilo, line 122: please remove the newlines and align `=`

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -168,6 +175,7 @@ pub struct NodeData {
     pub label  : frp::Dynamic<String>,
     pub events : Events,
     pub view   : component::ShapeView<NodeView>,
+    pub ports  : PortManager,
```
wdanilo, line 178: lets make it qualified, like `port::Manager`. Btw, I was talking in one of the comments that `Manager` is not the best word as it is like "utils" - everything can be "manager" and everything can be "utils". We try not to use neither of these words in our codebase. What do you think about `port::Registry` instead? :)

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+    ///  \----------|                                      \------------|       on the radius
+    ///     width                                              width
+    /// ```
+    /// The shape description sets up both shapes and then applies only one of the variants:
+    /// either it limits of the inward facing angle with the outer ring or it cuts of the
+    /// angle with the inner ring.
+    ///
+    fn new_port
+    (height:Var<f32>,width:Var<f32>,inner_radius:Var<f32>,is_inwards:Var<f32>) -> AnyShape {
+
```
wdanilo, line 119: reove this nl please :)

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+        let circle_inner              = Circle(circle_inner_radius);
+
+        // We don't want to move around the angle so we position the two circles relative to the
+        // angle.
+
+        let circle_outer_offset_y = Var::<Distance<Pixels>>::from(
+            &shape_height
+             - segment_outer.sagitta()
+             - &outer_radius
+        );
```
wdanilo, line 171: make more variables and make the exprs single lines pls.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+        let rotation_angle = is_inwards * pi;
+        let rotation_angle = Var::<Angle<Radians>>::from(rotation_angle);
+        let sculpted_shape = sculpted_shape.rotate(rotation_angle);
+
+        sculpted_shape.into()
+    }
+
+    ensogl::define_shape_system! {
+        (height:f32,width:f32,inner_radius:f32,is_inwards:f32) {
+            // NOTE: `is_inwards` should only be 0.0 or 1.0.
```
wdanilo, line 206: change it to `FIXME` - we should change it to bools in the future.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+    type Shape = shape::Shape;
+    fn new(shape:&Self::Shape, _scene:&Scene,_shape_registry:&ShapeRegistry) -> Self {
+        let spec  = Specification::default();
+        shape.update_from_spec(&spec);
+        shape.is_inwards.set(0.0);
+
+        let bbox = Vector2::new(2.0 * shape.height.get(), 2.0 * shape.height.get());
+        shape.sprite.size().set(bbox);
+
+        Self {}
```
wdanilo, line 261: This is a code repetition. Pls refactor it to a function.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+#[derive(Debug,Default,Clone,Copy)]
+pub struct OutputPortView {}
+impl ShapeViewDefinition for OutputPortView {
+    type Shape = shape::Shape;
+    fn new(shape:&Self::Shape, _scene:&Scene,_shape_registry:&ShapeRegistry) -> Self {
+        let spec  = Specification::default();
+        shape.update_from_spec(&spec);
+        shape.is_inwards.set(0.0);
+
+        let bbox = Vector2::new(2.0 * shape.height.get(), 2.0 * shape.height.get());
```
wdanilo, line 258: These 2.0 * etc are OK, but are not described anywhere. You should refactor `2.0` to a variable and name it in such way that it will explain to the reader what it is really used for (it is the scaling factor to make place for growing shapes when zooming out, right?)

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+pub struct Port<T:PortShapeViewDefinition> {
+    pub data : Rc<PortData<T>>
+}
+
+impl<T:PortShapeViewDefinition> Clone for Port<T>{
+    fn clone(&self) -> Self {
+        let data = Rc::clone(&self.data);
+        Self {data}
+    }
+}
```
wdanilo, line 293: Why not derive?

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+/// use graph_editor::component::node::port::OutputPort;
+///
+/// let input_port  = InputPort::default();
+/// let output_port = OutputPort::default();
+/// ```
+#[derive(Debug,CloneRef)]
+#[allow(missing_docs)]
+pub struct Port<T:PortShapeViewDefinition> {
+    pub data : Rc<PortData<T>>
+}
```
wdanilo, line 286: Plase read about `CloneRef` from `prelude` - this is one of the ost important traits in our codebase. Apply it in your PR whenever it could be applied (you can derive it)

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+#[derive(Clone,Copy,Debug)]
+pub struct Specification {
+    /// Height of the port.
+    pub height      : f32,
+    /// Width of the port in degrees.
+    pub width       : Angle<Degrees>,
+    /// Radius of the inner circle that the port is constructed around.
+    pub inner_radius: f32,
+    /// Location of the port along the inner circle.
+    pub position: Angle<Degrees>,
```
wdanilo, line 58: I would name it rotation instead.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+pub type InputPort = Port<InputPortView>;
+
+/// Type that represents an output port.
+pub type OutputPort = Port<OutputPortView>;
+
+/// Internal data of `Port.
+#[derive(Debug,Clone)]
+#[allow(missing_docs)]
+pub struct PortData<T:PortShapeViewDefinition> {
+        spec : RefCell<Specification>,
```
wdanilo, line 305: IMO this `RefCell` is too broad. What I would propose is to remove the `Specification` and include all of its fields here. Many of the fields are `Copy` so they can be converted into `Cell`. Some fields are not `Copy`, but I think you would be fine (with the sea performance) to use `CloneCell` (from prelude) for them instead - please confirm if Im right.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,468 @@
...
+        self
+    }
+
+    /// Sets the port's position.
+    ///
+    /// Note this shadows the `set_position` function of `ObjectOps`.
+    pub fn set_position(&self, position:Angle<Degrees>) {
+        self.data.spec.borrow_mut().position = position;
+        self.update_sprite_position();
+    }
```
wdanilo, line 332: `set_position` in DisplayObject` allwos setting 2D position. This is about setting rotation of the port. I would name it to `set_rotation`. Moreover, if we as user of this struct cannot use the properties of display object freely, then `Port` should not be a display object per-se, right? Of course its field `view` will be a `display::Object`, but we do not need to make `Port` one. Exposing all utils like `mod_position` or `mod_rotation` from display object doesn't make sense here. IMO you should remove `impl<'t,T:PortShapeViewDefinition> From<&'t Port<T>> for &'t display::object::Node` .

src/rust/ensogl/src/math/geometry/triangle.rs
MichaelMauderer: I've reduced the sizes some more, but I think it useful to have a visual overview of all the terms.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: There is a bug in the derive macro, so it doesn't work here: https://github.com/rust-lang/rust/issues/26925
It requires `T` to have a `Clone` bound, which we don't have and don't need.
I've added a comment.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: I think that could lead to confusion about what is meant here since the rotation could also just men the "orientation" of the object. I think a position expressed as `Angle` might be clearer.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: I did implement this for `Port`. From my understanding, the other structures (like `Registry`) should not need it as they have clearer ownerships. (One owned `Registry` per `Node`).

src/rust/ensogl/src/math/geometry/triangle.rs
MichaelMauderer: The cloning is solved now, but the trait bounds are a bit uglier now.
The lines are still too long, even now.

src/rust/lib/graph-editor/src/component/node/port.rs
wdanilo: This is not a bug. This is how Rust type system works. in such case use `Derivative` macro - it is exported from prelude.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+}
+
+/// Shape view for input port.
+#[derive(Debug,Default,Clone,Copy)]
+pub struct InputPortView {}
+impl ShapeViewDefinition for InputPortView {
+    type Shape = shape::Shape;
+    fn new(shape:&Self::Shape, _scene:&Scene,_shape_registry:&ShapeRegistry) -> Self {
+        init_shape(shape, Direction::In);
+
```
wdanilo, line 252: remove nl

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+            height       : Cell::new(spec.height),
+            width        : Cell::new(spec.width),
+            inner_radius : Cell::new(spec.inner_radius),
+            position     : Cell::new(spec.position),
+            direction    : Cell::new(spec.direction),
+            view
+        });
+        Self {data} . init()
+    }
+
```
wdanilo, line 340: This is important. We've been discussing it for long time already. `new` should **always** behave just as `default`. Otherwise it is very confusing. `Port::new()` should just create a port with default values which you can change later. Please try not to use "builder pattern here". Seriously, I think the `Specification` structure is not needed here.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+    }
+}
+
+
+
+// ============
+// === Port ===
+// ============
+
+/// Initialise the shape with sensible defaults..
```
wdanilo, line 230: two dots

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+fn init_shape(shape:&shape::Shape, direction:Direction){
+    let spec  = Specification::default();
+    shape.update_from_spec(&spec);
```
wdanilo, line 233: I dont follow it really. We are creating a new port shape. When creating the port it has all parameters already set, like height. Then we call this function, which then calls `update_parameters`, which then sets the parameters again. What is the purpose of that?

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+            t.shape.update_parameters(height,inner_radius,width,direction);
+        }
+    }
+}
+
+impl<T:PortShapeViewDefinition> Default for Port<T> {
+    fn default() -> Self {
+        Self::new(Specification::default())
+    }
+}
```
wdanilo, line 412: derive Default for Port

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+/// Data structure that creates and stores port shapes.
+#[derive(Debug,Default)]
+#[allow(missing_docs)]
+pub struct PortBuffer<T:PortShapeViewDefinition> {
+    ports: RefCell<Vec<Port<T>>>,
+}
+
+/// Helper type that represents a `PortBuffer` for `InputPorts`.
+type InputPortBuffer = PortBuffer<InputPortView>;
+/// Helper type that represents a `PortBuffer` for `OutputPorts`.
```
wdanilo, line 435: here should be a newline

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+type OutputPortBuffer = PortBuffer<OutputPortView>;
+
+impl<T:PortShapeViewDefinition> PortBuffer<T> {
+    /// create a new port in this buffer and sets it as a child node of the given parent.
```
wdanilo, line 439: Use upper case letter on the beginning of the sentence.

src/rust/lib/graph-editor/src/component/node/port.rs
```diff
@@ -0,0 +1,478 @@
...
+    pub fn create(&self, parent:&Node) -> Port<T> {
+        let port = Port::default();
+        parent.add_child(&port.data.view.display_object);
+        self.ports.borrow_mut().push(port.clone());
```
wdanilo, line 443: port is `CloneRef`, so you sohuld use it here instead to indicate the intention clearly.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: `Port` requires some initialisation logic (e.g, setting the initial position of the sprite), so there needs to be a call to `new` which requires the manual implementation.

src/rust/lib/graph-editor/src/component/node/port.rs
MichaelMauderer: Actually this is the correct place to initialise the shape. The initialisation in the constructor doesn't work if the shape is not accessible (which it isn't at the moment). I've removed the initialisation of the shape there.

src/rust/lib/graph-editor/src/component/node/port.rs
wdanilo: then do not name it `new`. `new` should be equal to `default`. Name it `new_with_something` or osmething

PR #340, opened 2020-04-03
Fixed PageUp and PageDown

notdanilo:
### Pull Request Description
PageUp and PageDown won't change the view if all the text is visible. Instead, it moves the cursor to the beginning of the document or to its end.

### Important Notes
* `TextField`'s `page_down` and `page_up` methods were removed.
* `PageDown` and `PageUp` were implemented as navigation commands.
* `TextEditor` now initializes `TextField` taking half of the screen height.

### Checklist


src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -147,9 +152,12 @@ pub struct CursorNavigation<'a> {
...
 impl<'a> CursorNavigation<'a> {
     /// Creates a new CursorNavigation with defaults.
-    pub fn default(content:&'a mut TextFieldContent) -> Self {
+    pub fn default
+    ( content         : &'a mut TextFieldContent
+    , text_field_size : Vector2<f32>
+    , scroll_position : Vector2<f32>) -> Self {
         let selecting     = default();
-        Self {content,selecting}
+        Self {content,selecting,text_field_size,scroll_position}
```
farmaazon, line 160: I wonder if it should be still called "default" function. I meant to have a function which as all fields in some default state except the content (which cannot be defaulted), but this looks silly when you have 4 fields and only one can be actually defaulted :)

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -462,20 +478,27 @@ impl Cursors {
...
         let filtered = self.cursors.iter_mut().filter(|c| predicate(c));
         filtered.for_each(|cursor| navigation.move_cursor(cursor, step));
+        let new_position = self.last_cursor().position;
         self.merge_overlapping_cursors();
+
+        let scroll      = navigation.scroll_position.y;
+        let line_height = navigation.content.line_height;
+        let height      = ((navigation.text_field_size.y - line_height) / line_height).floor() * line_height;
+        let position    = new_position.line as f32 * navigation.content.line_height - scroll;
+        -position.min(position.max(height) - height)
```
farmaazon, line 501: I don't see why we compute the value, and why this function returns anything. The docs also don't mention that. And also I see we don't use this returned value anywhere. Probably this is change to be reverted.

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
notdanilo: This is true. I think it makes sense to pass `TextField` to it instead of these three different `TextField`'s properties separately.

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
notdanilo: The responsibility of scrolling the text field was moved from CursorNavigation to TextField. Scrolling now happens when navigating and when typing a new line.

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -658,7 +646,7 @@ mod test {
         let properties     = TextFieldProperties::default(&mut fonts);
         let mut content    = TextFieldContent::new(text,&properties);
         let mut navigation = CursorNavigation
-            {selecting:true, ..CursorNavigation::default(&mut content)};
+            {selecting:true, ..CursorNavigation::new(&mut content)};
```
farmaazon, line 649: I doubt it compiles now. Please check tests before requesting for review.

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -635,7 +646,7 @@ mod test {
...
```
farmaazon, line 649: And I still don't see the test cases for PageUp/PageDown navigation. I suspect that you missed it because I wrote it on "overall review comment", so now I put it here :)

src/rust/ensogl/src/display/shape/text/text_field.rs
```diff
@@ -158,21 +112,58 @@ shared! { TextField
...
+            let scroll_position = self.scroll_position();
+            let page_height     = self.size().y;
+            let scrolling       = match step {
+                Step::PageUp   =>  page_height,
+                Step::PageDown => -page_height,
+                _              => 0.0
+            };
+
+            let scrolling = scrolling.min(scroll_position.y);
+            let text_height = self.content.line_height * (self.content.lines().len() + 1) as f32;
```
wdanilo, line 135: What is this `+1`?

src/rust/ensogl/src/display/shape/text/text_field.rs
```diff
@@ -158,21 +112,58 @@ shared! { TextField
...
+        }
+
+        /// Adjust the view to make the last cursor visible.
+        fn adjust_view(&mut self) {
+            let line_height         = self.content.line_height;
+            let last_cursor         = self.cursors.last_cursor();
+            let scroll_y            = self.scroll_position().y;
+            let view_size           = self.size();
+            let current_line        = last_cursor.position.line as f32 * line_height;
+            let next_line           = (last_cursor.position.line + 1) as f32 * line_height;
```
wdanilo, line 149: And this `+1`?

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -470,9 +481,9 @@ impl Cursors {
     ///
     /// If after this operation some of the cursors occupies the same position, or their selected
     /// area overlap, they are irreversibly merged.
-    pub fn navigate_cursors<Predicate>
-    (&mut self, navigation:&mut CursorNavigation, step:Step, mut predicate:Predicate)
-    where Predicate : FnMut(&Cursor) -> bool {
+    pub fn navigate_cursors<P>
+    (&mut self, navigation:&mut CursorNavigation, step:Step, mut predicate:P)
+    where P : FnMut(&Cursor) -> bool {
```
wdanilo, line 486: I was reading this code and I was wondering what is `P`. I learned by reading the removed code where it is named `Predicate`. I think the previous name is much better.

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -470,9 +481,9 @@ impl Cursors {
...
         let filtered = self.cursors.iter_mut().filter(|c| predicate(c));
```
wdanilo, line 487: Would it be possible to change it to `.filter(predicate)` instead?

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
notdanilo: Fixed!

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
notdanilo: I got a test for it :)

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
farmaazon: I just wnader why you're wondering, if the argument name is "Predicate"?

src/rust/ensogl/src/display/shape/text/text_field.rs
```diff
@@ -77,6 +150,16 @@ shared! { TextField
...
+            let scroll_position = self.scroll_position();
+            let offset_y        = offset.y.min(scroll_position.y);
+            let padding_lines   = 2;
+            let lines           = self.content.lines().len() + padding_lines;
+            let text_height     = self.content.line_height * lines as f32;
+            let view_height     = self.size().y;
+            let height          = (text_height - view_height).max(0.0);
+            let offset_y        = offset_y.max(scroll_position.y - height);
+            let offset          = Vector2::new(offset.x, offset_y);
+
```
farmaazon, line 162:
```suggestion
            let scroll_position = self.scroll_position();
            let padding_lines   = 2;
            let lines           = self.content.lines().len() + padding_lines;
            let text_height     = self.content.line_height * lines as f32;
            let view_height     = self.size().y;
            let height          = (text_height - view_height).max(0.0);
            let offset_y        = offset.y.min(scroll_position.y).max(scroll_position.y - height);
            let offset          = Vector2::new(offset.x, offset_y);
```
A little bit clearer for me.

src/rust/ensogl/src/display/shape/text/text_field/content/line.rs
```diff
@@ -99,11 +98,16 @@ pub struct LineFullInfo<'a> {
 }
 
 impl<'a> LineFullInfo<'a> {
+    /// Gets line y position.
```
farmaazon, line 101: Please add info if it is a middle of the line, a _baseline_ or maybe a top of the line area (as it seem to be from `baseline_start` function.

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
farmaazon: :+1:

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -624,16 +620,39 @@ mod test {
...
+    async fn page_scrolling() {
+        ensogl_core_msdf_sys::initialized().await;
+        let text              = "1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15\n16\n17\n18\n19";
+        let initial_cursor    = Cursor::new(TextLocation::at_document_begin());
+        let initial_cursors   = vec![initial_cursor];
+        let expected_position = TextLocation {line:6,column:0};
+
+        let mut fonts  = FontRegistry::new();
+        let properties = TextFieldProperties::default(&mut fonts);
+        assert_eq!(properties.size, Vector2::new(100.0,100.0));
```
farmaazon, line 645:
```suggestion
        let mut properties = TextFieldProperties::default(&mut fonts);
        properties.size = Vector2::new(100.0,100.0);
```
I think we should just set required size instead of counting for defaults.

src/rust/ensogl/src/display/shape/text/text_field/cursor.rs
```diff
@@ -599,10 +593,12 @@ mod test {
...
-        let mut content    = TextFieldContent::new(text,&properties);
-        let mut navigation = CursorNavigation::new(&mut content);
+        let mut fonts       = FontRegistry::new();
+        let properties      = TextFieldProperties::default(&mut fonts);
+        let content         = &mut TextFieldContent::new(text,&properties);
+        let text_field_size = properties.size;
+        let selecting       = false;
+        let mut navigation  = CursorNavigation{selecting,content,text_field_size};
 
         for step in &[/*Left,Right,Up,*/Down,/*LineBegin,LineEnd,DocBegin,DocEnd*/] {
```
farmaazon, line 603: Oh, now I see that I accidentally push commented cases! Please uncomment them and add cases for PageUp and PageDown. It covers more corner cases (like cursor in the first line, in the last line, at end of line, at end of document etc.)

PR #349, opened 2020-04-06
Connection discovery and IDE identifier resolving spec

mwu-tow:
### Pull Request Description
This PR describes language rules as I understand them in a scope necessary to implement proper connection discovery (#320). It covers general language rules (that are relevant) and assumptions and simplifications that IDE can currently make.

Also, a number of TODO notes are left for things that I don't understand yet or that haven't been clearly specified. They are not blocking the implementation for connections, rather something that we are allowed to fill out later.

### Checklist

### DIFF
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -1,80 +1,78 @@
 # General Rules
-This section attempts to describe how the identifiers are introduced and
-resolved in the language.
+This document describes language rules that are relevant to connection discovery
+and operations. The purpose is not to specify the whole language. Just enough
+for IDE team members to be able to reason where identifiers are introduced and
+what entity identifier usage refers to.
 
-The purpose is not to specify the whole language. Just enough for an IDE team
-members to be able to reason where identifiers are introduced and what entity
-identifier usage refers to.
-
-This is the base allowing IDE to describe what connections are in the displayed graph.
+This is the base allowing IDE to describe what connections are in the displayed
+graph. The covered topics are mostly identifiers, scopes and their interactions.
 
 
 ## Identifier
-Identifier is a name that may denote value of some type. Syntactically we recognize:
-* variables, being names that do not contain upper-cased characters, like `foo2`
+Identifier is a name that denotes a value, i.e. that is bound to a value. The
+compiler also keeps track of its type, i.e. the set of possible values. Syntactically we recognize:
+* variable names, being names that do not contain upper-cased characters, like `foo2`
   or `make_new`;
-* constructors, which are like variables but with first character and every
+* constructor names, which are like variables but with first character and every
   character directly following underscore upper-cased (e.g. `Foo2` or
   `Make_New`);
-* operators, being names consisting of operator symbols (e.g. `+` or `<$>`).
-  Specifically, operator name may contain following characters
-  `!$%&*+-/<>?^~|:\,.()[]{}=`. Not every sequence of these characters is a valid
-  operator name, as they could collide with other language constructs.
+* operator names, being built solely from operator symbols (e.g. `+` or `<$>`).
+  Specifically, an operator name may contain only characters from the following
+  set: `!$%&*+-/<>?^~|:\,.()[]{}=`. Not every sequence of these characters is a valid operator name, as they can collide with other language constructs.
   
-Any other names not matching requirements above, like `HTTP`, `foO` or
-`Make_new` are not allowed.
+Any other names that do not match requirements above, like `HTTP`, `foO` or
+`Make_new` are not allowed and their behavior will not be specified.
 
-In non-pattern contexts, referring to an existing identifier is
-case-insensitive. So `foo` can be referred to as `Foo`. Note that `fOo` or `FOO`
-are not valid identifiers, as upper-cased letter may appear only as the first
-letter or after underscore (e.g. `Make_Request`).
+In non-pattern contexts, referring to the existing identifier is
+case-insensitive. So `foo` can be referred to either as `foo` or `Foo` with no
+difference.  In pattern context, lower-cased names are used to introduce a
+binding (or a constraint), while upper-cased name will refer to an already bound
+identifier. In short, variable name is allowed to shadow, while constructor 
+always unambiguously refers to an externally visible identifier.
 
-In pattern context, lower-cased names are used to introduce a binding (or a
-constraint), while upper-cased name will refer to an already bound identifier.
+Operator names behave either as:
+* variable names when placed in a prefix position (e.g. `+` in `+ ab`);
+* or as constructor names when placed in an infix position (e.g. `,` in `a,b`).
 
-Binding means introducing an identifier into scope and associating it with some
-value. Identifier can be introduced also without binding it to any specific
-value (e.g. as type constraint).
+Number and text literals (like `5` or `"Hello"`) are treated as
+constructor names.
 
-Operators behave as variables in prefix position (e.g. `+` in `+ a b`) or as
-constructors in an infix position (e.g. `,` in `a,b`).
+Identifiers can be introduced by a binding (e.g. using assignment or lambda
+argument matching) or when adding type constraints (using type ascriptions). 
 
 ## Scope
-Scope is the span in the code which shares the available identifiers set. Scopes
-can be seen as a span-tree structure, covering the whole program code.
+Scope is the span in the code which shares the identifiers set of visible,
+available identifiers. Scopes are a span-tree structure, covering the whole program code.
 
 Nested scope is allowed to:
 * access identifiers from the outer scope;
 * shadow identifiers from nested scope with a new binding;
 * introduce new constraints on the identifiers from parent (or own) scopes.
 
-The same identifier may be bound to multiple times in the same scope
-(overloading). It is allowed only for method overloads that differ in the type of
-the `this` parameter. This limitation may be relaxed in the future, if proper
-motivating use-cases are found.
+The same identifier may be bound multiple times in the same scope (overloading).
+It is allowed only for method overloads that differ in the type of the `this`
+parameter. This limitation may be relaxed in the future, if proper motivating use-cases are found.
 
-The identifier is always accessible scope-wide, before and after the line
+The identifier is always accessible scope-wide, both before and after the line
 introducing it. Some monadic contexts (like IO) can introduce order-dependent
 behavior, however. This is not something that IDE is (or can be) concerned about
 when figuring out connections.
 
-Identifier is bound by using a variable-type identifier in the pattern context.
-Exact behavior depends on the language construction that was to introduce the identifier.
+Identifier is bound by using a variable name in the pattern context. Exact behavior depends on the language construction that was to introduce the identifier.
 
 Identifier introduced into a scope is visible only in the scope's subtree
 (lexical scoping). 
 
-Scopes are introduced by:
-* module/file (the root scope);
-* code blocks (i.e. the block that follows the line with an trailing operator);
-* `->` operator for its right-hand side.
+Scopes, in the core language, are introduced by:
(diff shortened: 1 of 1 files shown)


docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+`main` here is a function binding that is desugared to lambda. As such its body
+has its own, separate scope. That would not have been a case, if such lines
+appeared in any non-root code block.
+
+## Lambdas
+`arg -> value` is the syntax for lambdas. Left-hand side is a pattern for the
+argument (lambdas are always unary) and the right-hand side is its body.
+Right-hand side, i.e. the lambda body, introduces its own scope.
+
+If the lambda is *not* introduced in what is already a pattern context, the `->`'s pattern introduces identifiers into the scope of the right-hand
```
farmaazon, line 372: I think you introduced _pattern scopes_, and _pattern context_ is a term from some previous version of this doc.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+## Scope
+Scope is the span in the code which shares the identifiers set of visible,
+available identifiers. Scopes are a span-tree structure, covering the whole program code.
+
+Nested scope is allowed to:
+* access identifiers from the outer scope;
+* shadow identifiers from nested scope with a new binding;
+* introduce new constraints on the identifiers from parent (or own) scopes.
+
+The same identifier may be bound multiple times in the same scope (overloading).
```
iamrecursion, line 52: This is currently not supported outside the global scope. We have an agreement that methods can only be defined in the global scope, so this seems fine to me. If that design decision has changed please let me know.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+It is allowed only for method overloads that differ in the type of the `this`
+parameter. This limitation may be relaxed in the future, if proper motivating use-cases are found.
+
+The identifier is always accessible scope-wide, both before and after the line
```
iamrecursion, line 56: This isn't something that is (or can be) handled at the moment. Additionally, it doesn't make sense. The definition of the desugaring of monadic contexts (in terms of bind) introduces an ordering dependency. If we want to distinguish between expressions `in Pure` and lift those into mutually recursive let blocks, that is a whole other kettle of fish that will require the typechecker.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+
+Scopes, in the core language, are introduced by:
+* module (file), i.e. the root scope;
+* code blocks, i.e. indented blocks that follow a trailing operator;
+* arrow (lambda) `->` operator for its right-hand side.
+
+Some other constructs seemingly introduce scope (like function definitions) but
+this is because they are desugared into some construct that introduces scope (like lambdas).
+
+TODO: Consider if there are any special rules for signatures on definitions, or
```
iamrecursion, line 74: The identifiers defined in a signature are available in the body of the function, and vice-versa.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+
+The following spans are pattern contexts:
+* left-hand side of assignment operator, e.g. `main` in  `main = println
+  "Hello"`;
+* right-hand side of a colon operator, e.g. `a` in `foo:a`;
+* left-hand side of an arrow operator, e.g. `a` in `a -> a + 1`.
+
+Details will follow with description of these operators.
+
+// TODO What about `case … of` ? Are there any other pattern-introducing constructs?
```
iamrecursion, line 133: The scrutinee of a case expression is _not_ a pattern context, but the left hand side of the battern branches is.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+will be desugared into:
+
+```
+«value» = case »get_opt« of
+  »Some« b -> b
+  _      -> error 
+```
+
+Therefore, only `value` will be introduced into the parent scope. `Some` and
+`get_opt` must be defined, the former being an atom with at least single field.
```
iamrecursion, line 305: You should explain why you say "at least" here.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+
+Example:
+
+```
+«foo» this:«a» = »body1«
+
+«foo» this:«b» = »body2«
+```
+
+In this case `a` and `b` for each `foo` definition will be inferred by the
```
iamrecursion, line 351: This is true, but not for a while.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+```
+a -> a -> b
+```
+
+Here the first `a` and second `a` are separate identifiers, the latter shadowing
+the first one. If one wanted to express that both arguments are of the same
+type, `a -> A -> b` would have been used. `b` refers to an identifier from
+graph's scope (it is in the body's position, not pattern).
+
+OPEN QUESTION: actually it might be "nice" to have both `a` unified in such
```
iamrecursion, line 417: I can't think of a reliable rule for this that doesn't have some downside.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+case. 
+
+
+
+## Type ascription
```
iamrecursion, line 422: Context ascription (`in`) should also introduce a pattern context.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+For example:
+```
+5 : «a»?
+```
+
+This introduces constraint on the type `a` that its value set must include atom
+`5`. If the `a` is already visible in the scope, this constraint will be added.
+Otherwise, `a` will be introduced into the current scope with that constraint.
+
+// TODO: What if parent scope only ascribes identifier with type constraint but
```
iamrecursion, line 442: Then we'd likely have to do some complex bidirectional inference to make this work.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+only nested scope assigns to it? 
+
+// TODO: Open design question, if `a` should modify existing variable or should
```
iamrecursion, line 445: I'm of the opinion that it should always try and shadow it.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+always try to shadow it. What is the difference between `5 : a` and `5 : A`?
+(except the latter not being able to ever introduce a new identifier)
+
+// TODO: Is it legal or sensible to ascribe a variable when there is no binding?
```
iamrecursion, line 449: I don't think it should be legal. A type signature should always be associated with a binding for that name.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+
+The `type` in this expression is pattern context and can be used to constrain
+the type variables. It is legal to assign constraints on an identifier using `:`
+multiple times in any of the scopes where identifier is visible.
+
+Signatures are just type ascriptions that happen to precede the assignments.
+They have no special rules currently defined. This area needs further design
+work.
+
+TODO: Open question: does empty type (`Void`) exists in the language?
```
iamrecursion, line 474: We do have a `Unit` type, yes.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+
+TODO signatures and their relation with scoping. Difference for root and
```
iamrecursion, line 476: Can you expand on this?

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+add : a -> a -> a
+add a b = a + b
+```
+
+TODO: 
+* With current rules `a` from the signature gets introduced into parent scope
+  will be unified with other uses of `a` in other definitions.
+* or, actually, we want this to happen only in the root scope. When in
+  definition body, the `a` actually should be unified between signatures.
+  Doesn't sound that clean though.
```
iamrecursion, line 491: Not clean at all.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+* Does argument-introduced `a` shadows the signature-introduced `a`?
+
+TODO: `b:A` — does this add constraint to `a` or just `b` ?
```
iamrecursion, line 494: Complex question of inference that can't easily be answered at this stage given how nebulous the type system is. Does it matter for the purposes of IDE right now?

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+```
+a.foo = print "hello"
+```
+
+are equivalent, engine currently supports only the latter.
+
+IDE can assume that all extension methods will be introduced using the
+`Type.name` syntax sugar.
+
+// TODO what if non-first argument is named `this` ? Is the magic happening only
```
iamrecursion, line 520: It is an error to define `this` at any position other than the first in the argument list.

docs/connections.md
```diff
@@ -0,0 +1,597 @@
...
+for this particular name? Is it sensitive for its position in the arguments list?
+// TODO What happens if the `this`-taking function in defined in the root scope
```
iamrecursion, line 522: This will also be an error.

docs/connections.md
```diff
@@ -493,6 +549,104 @@ TODO:
...
+This should give a better understanding of why the rules are as presented.
+
+Note that the desugaring translated below is very low-level.
+
+All assignments and code blocks are removed.
+
+
+## Out of order variable usage in block
+If within a block an identifier is used in the line before it is assigned, the
+`fix` function appropriate for the block's monadic context will be introduced.
```
iamrecursion, line 563: I don't know how performantly we can do this without a typechecker. The implementation doesn't currently support it, and I'll need to talk to Marcin about it when he's back.

docs/connections.md
iamrecursion: Currently it's a tentative 2. There are technical obstacles given the current way method lookup is performed in the interpreter, but also there would be some typechecking obstacles to doing it which I'm a little unclear on given I'm somewhat divorced from that work right now. The design _currently_ forbids, it, and we'd have to solve these technical challenges to allow it.

docs/connections.md
iamrecursion: Yeah we definitely want it to work eventually, but this is very much _eventually_.

docs/connections.md
iamrecursion: That's fair. 

And yes, it _is_ tricky, but it's also necessary for the way we're planning on handling dependency in type ascription. We want the ability to be able to define a type function `foo` in the body that can then be used in the signature.

docs/connections.md
iamrecursion: Doing this requires a typechecker.

I should probably note here that we _do_, however currently support overloading of a method on a given type as long as the arity is different, so you can define

```
Foo.bar = x -> x + 1
Foo.bar = x y -> x + 1
```

docs/connections.md
iamrecursion: Okay, cool. I'd like to do that consolidation ASAP, as otherwise _those_ docs are likely to evolve without accounting for the information you have here.

docs/connections.md
iamrecursion: Correct. I would also disallow it, I'm just saying that it's _technically_ feasible.

docs/connections.md
iamrecursion: Ah, I see what you mean. I don't think I've addressed that question in the type system doc yet. When we merge docs I'll make sure to add it as a TODO item.

docs/connections.md
iamrecursion: I see what you mean. I think I've addressed it with other comments, but let me know if things are unclear.

farmaazon:
@iamrecursion could you please look at this document again? It was written before removing UFCS and likely outdated in many places.

wdanilo:
What is the status of this ?

farmaazon:
> What is the status of this ?

It waits for @iamrecursion after-UFCS-check. See my comment above. And because we have other priorities right now, it seem properly frozen in icebox.

PR #353, opened 2020-04-07
Added MacOS specific shortcuts

notdanilo:
### Pull Request Description
This PR adds MacOS shortcuts fot text editor.

### Important Notes
Here is the list of the shortcuts:
* Jump to beginning of a line – Command+Left Arrow
* Jump to end of a line – Command+Right Arrow
* Jump to beginning of current word – Option+Right Arrow
* Jump to end of current word – Option+Right Arrow
* Jump to beginning of all text – Command+Up Arrow
* Jump to end of all text – Command+Down Arrow

### Checklist


src/rust/lib/system/web/src/platform.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// Queries which platform we are on.
+    pub fn query() -> Self {
+        let window = window();
+        let navigator = window.navigator();
+        let platform  = navigator.platform().unwrap_or_else(|_| "Unknown".into());
+        let agent     = navigator.user_agent().unwrap_or_else(|_| "Unknown".into());
+
+        const MACOS_PLATFORMS   : [&str; 4] = ["Macintosh", "MacIntel", "MacPPC", "Mac68K"];
+        const WINDOWS_PLATFORMS : [&str; 4] = ["Win32", "Win64", "Windows", "WinCE"];
+        const IOS_PLATFORMS     : [&str; 3] = ["iPhone", "iPad", "iPod"];
```
farmaazon, line 26: Let use HashSet for better performance like `let macos_platforms : HashSet<&str> = [...].iter().collect();` and then check using "contains".

src/rust/lib/system/web/src/platform.rs
```diff
@@ -0,0 +1,42 @@
...
+        if MACOS_PLATFORMS.iter().any(|item| **item == platform) {
+            Platform::MacOS
+        } else if IOS_PLATFORMS.iter().any(|item| **item == platform) {
+            Platform::IOS
+        } else if WINDOWS_PLATFORMS.iter().any(|item| **item == platform) {
+            Platform::Windows
+        } else if agent.find("Android").is_some() {
+            Platform::Android
+        } else if platform.find("Linux").is_some() {
+            Platform::Linux
```
farmaazon, line 37: Let's keep all the "platform constants" in one place (so put the const (LINUX|ANDROID)_PLATFORM : &str = next to rest of platforms list.

notdanilo: We can't generalize this logic because `Android` needs to be searched in `agent` and `Linux` is searched in `platform`.

src/rust/lib/prelude/src/lib.rs
```diff
@@ -38,6 +38,11 @@ pub use ifmt::*;
 pub use itertools::Itertools;
 pub use lazy_static::lazy_static;
 pub use num::Num;
+pub use maplit::btreemap;
+pub use maplit::btreeset;
+pub use maplit::convert_args;
+pub use maplit::hashmap;
+pub use maplit::hashset;
```
wdanilo, line 45: I would not expose in prelude: `btreemap`, `btreeset`, and `convert_args`. The first 2 are so rarely used that they could be imported when needed. The `convert_args` name is way too generic to be exported in prelude.

src/rust/lib/system/web/src/platform.rs
wdanilo: Do we always have either `navigator.platform()` or `navigator.user_agent()`? Are both valid on Android / Linux or they are used there interchangeable. Im asking cause I also think this code should be more structured (as Adam suggested) and I have idea how to do it but I need the above info :)

src/rust/ensogl/src/display/shape/text/text_field/frp/keyboard.rs
```diff
@@ -167,6 +168,57 @@ impl TextFieldKeyboardFrp {
 }
 
 
+// === Keys combinations ===
```
wdanilo, line 171: These things are very hacky but for now they are ok. Just a note that using the new shortcut manager it should be done way nicer.

src/rust/lib/system/web/src/platform.rs
notdanilo: @wdanilo yes, we need both `platform` and `user_agent`. Please check the following links:

https://stackoverflow.com/questions/19877924/what-is-the-list-of-possible-values-for-navigator-platform-as-of-today

https://developer.chrome.com/multidevice/user-agent

wdanilo: Oh god. Based on the linked answer Im wondering why we just don't do everything in unified way - search if there was `win`, `mac`, `linux`, `ipad`, `iphone` string inside the `platform` string (I used lowercase letters intentionally - we want to catch when browser decides to return "windows" instead of "Windows". All these things are not specified anywhere so we will have hard time tracking down what happened when a browser changes its way of reporting it. so lets make it as generic as possible. What do you think about it?

PR #356, opened 2020-04-08
Add a debug view that renders shape IDs

MichaelMauderer:
### Pull Request Description
Add a view mode that shows shape IDs of shapes in the scene.
Uses a pseudo-random colour scheme to colour shapes according to their `symbol_id` and `instance_id`.
Closes issue #355 .

### Important Notes
This adds a second debug view mode.

### Checklist


src/rust/ensogl/src/display/shape/primitive/glsl/fragment_runner.glsl
```diff
@@ -1,21 +1,26 @@
...
+output_id.g *= alpha_no_aa;
+output_id.b *= alpha_no_aa;
 
 if (input_display_mode == 0) {
     output_color = srgba(unpremultiply(shape.color)).raw;
     output_color.rgb *= alpha;
 } else if (input_display_mode == 1) {
     Rgb col = distance_meter(shape.sdf.distance, 200.0 * input_zoom * input_pixel_ratio, 200.0/input_zoom * input_pixel_ratio);
     output_color = rgba(col).raw;
+}else if (input_display_mode == 2) {
```
wdanilo, line 20: inconsistent spacing.

src/rust/ensogl/src/display/shape/primitive/glsl/fragment_runner.glsl
```diff
@@ -1,21 +1,26 @@
...
+    float r = float(((int(input_symbol_id) * 79) % 360)) / 360.0;
+    float g = float(((int(input_instance_id) * 43) % 360)) / 360.0;
+    float b = float(((int(input_instance_id) * 97) % 360)) / 360.0;
```
wdanilo, line 23: I'm not convinced this is the best idea for this debug mode. When we've been chatting on Discord I provided you with `hsv2rgb` function and sample implementation on purpose (and my bad I did not explain it good enough). There are 2 important things when using HSV translation instead:

1. We can still get every ID visually distinct, but IDs close to each other have similar colors, and this is actually very helpful information sometimes.
2. We can sample the color on screen, put it in any RGB to HSV converter and based on the hue easily compute what was the original ID we are sampling. It is not very easy by using your equation here.

MichaelMauderer: Ah, I didn't know this was supposed to be invertible.
This has a similar feature: shapes with the same `input_instance_id` will look more similar than shapes with a different `input_instance_id`.
I'll do something invertible then.

src/rust/ensogl/src/display/shape/primitive/glsl/color.glsl
```diff
@@ -8,6 +8,9 @@
 ///   - A Rgba struct which contains `vec4 raw` field for each component.
 ///   - A rich set of smart constructors for each of the types.
 ///   - A rich set of simple conversions between them.
+///
+/// All component values are in range 0.0..1.0.
```
MichaelMauderer, line 12: @wdanilo Can you confirm that this is correct for all colors?

wdanilo: It could not be correct for the hue! If so (needs a check), could you update the docs here please? Thanks for catching1

MichaelMauderer: No there is no bug (that I encountered), this was just for reference, so someone reading this code knows the covnentions right away.

PR #357, opened 2020-04-09
async sleep function

notdanilo:
### Pull Request Description
This PR implements an async sleep function for both wasm and native targets.

### Checklist


src/rust/lib/system/web/src/lib.rs
```diff
@@ -481,3 +483,79 @@ pub mod traits {
...
+///
+/// This function is an async version of std::thread::sleep.
+#[cfg(target_arch = "wasm32")]
+pub async fn sleep(duration:Duration) {
+    use wasm_bindgen_futures::JsFuture;
+
+    let performance = performance();
+    let call_time   = performance.now();
+    let future : JsFuture = js_sys::Promise::new(&mut |resolve:Function,_| {
+        let time_elapsed_from_call = ((performance.now() - call_time) * 1000.0) as i32;
```
wdanilo, line 499: This is interesting. Is it really measurable? I was expecting it to be really close to 0. What are the results here? I think it is worth mentioning in the comment inside the code (not even doc comment) what was the offset measured as people inspecting this code would constantly wonder about it.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -481,3 +483,79 @@ pub mod traits {
...
+        let duration               = duration.as_millis() as i32;
+        let duration               = (duration - time_elapsed_from_call).max(0);
+        let window                 = window();
+        let err                    = "setTimeout failed";
+        window.set_timeout_with_callback_and_timeout_and_arguments_0(&resolve,duration).expect(err);
+    }).into();
+    future.await.expect("setTimeout future failed");
```
wdanilo, line 506: Errors should be real sentences! :)

src/rust/lib/system/web/src/lib.rs
```diff
@@ -481,3 +483,79 @@ pub mod traits {
...
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+pub use async_std::task::sleep;
+
+#[cfg(test)]
```
wdanilo, line 512: Please add section for tests! :)

src/rust/lib/system/web/src/lib.rs
```diff
@@ -481,3 +483,79 @@ pub mod traits {
...
+    }
+
+    #[wasm_bindgen_test(async)]
+    async fn async_sleep() {
+        let instant = helpers::now();
+        sleep(Duration::new(1,0)).await;
+        assert!(helpers::elapsed(instant) >= 1.0);
+        sleep(Duration::new(2,0)).await;
+        assert!(helpers::elapsed(instant) >= 3.0);
+    }
```
wdanilo, line 554: I would also check if the time was not too big, like `>=1.0` but also `<= 1.0 + epsilon`

notdanilo: @wdanilo I think it's problematic. `setTimeout` will always run after the specified duration, but the exact moment is not predictable. What if CI's load balancer makes `setTimeout` take much more time to be executed than expected? We can either leave the test as it is or, if you really want it, we can define an bigger threshold to test it, and not a small number (epsilon).

src/rust/lib/system/web/src/lib.rs
notdanilo: @wdanilo I know these messages looks generic, but web-sys' has error for almost every single call. The reason `setTimeout` and its future fails isn't clear to me. Do you have any better suggestion for it?

src/rust/lib/system/web/src/lib.rs
mwu-tow: I'm not sure about this, it wouldn't exactly test our code correctness but also the CI machine's load. Race conditions always scare me.
(if we wanted to be very-correct, we should mock the instant/set_timeout api, but it seems to be overkill at this stage)

src/rust/lib/system/web/src/lib.rs
notdanilo: Well, yes... If we use `sleep(duration).await`, the time elapsed is close to 0. But I think there might be cases we wan't start a request, instantiate the asynchronous sleep function as a timeout, make other operations and then await it.

This is required in the original task decription: https://github.com/luna/ide/issues/351

notdanilo: Oh, ok. @mwu-tow just responded at the same time :)

src/rust/lib/system/web/src/lib.rs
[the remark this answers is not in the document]
notdanilo: As I said on @wdanilo's comment. The reason why it fails is unclear. `web-sys` returns `Result`s from most of API functions.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -481,3 +483,79 @@ pub mod traits {
...
+///
+/// This function might sleep for slightly longer than the specified duration but never less.
+///
+/// This function is an async version of std::thread::sleep.
+#[cfg(target_arch = "wasm32")]
+pub async fn sleep(duration:Duration) {
+    use wasm_bindgen_futures::JsFuture;
+
+    let performance = performance();
+    let call_time   = performance.now();
```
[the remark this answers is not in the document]
notdanilo: I researched Instant's API and there is no way to create a new instance of it setting up its underlying properties.

src/rust/ensogl/Cargo.toml
```diff
@@ -43,7 +43,7 @@ smallvec                   = { version = "1.0.0"
 typenum                    = { version = "1.11.2"                                }
 # We require exact version of wasm-bindgen because we do patching final js in our build process,
 # and this is vulnerable to any wasm-bindgen version change.
-wasm-bindgen               = { version = "=0.2.58" , features = ["nightly"]      }
+wasm-bindgen               = { version = "=0.2.60" , features = ["nightly"]      }
```
farmaazon, line 46: All upgrades of wasm-bindgen version will cause CI to fail, because we do some patches on wasm-bindgen output and apparently those depends on exact version.

src/rust/lib/system/web/src/lib.rs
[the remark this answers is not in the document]
farmaazon: Definitely agreed.

src/rust/lib/system/web/src/lib.rs
farmaazon: Wojtek meant that sentence should start with capital letter and end with dot.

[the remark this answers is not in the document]
farmaazon: So you should put here comment that we don't expect any error from this Promise (as it's our rust promise, not js code which may fail).

PR #359, opened 2020-04-15
[WIP] Second checkpoint for alias analysis

mwu-tow:
### Pull Request Description
Code should be functionally complete with the exception of macro support.
PR against the previous accepted checkpoint for #322.

### Checklist


src/rust/ide/src/double_representation/alias_analysis.rs
```diff
@@ -65,26 +72,257 @@ pub struct IdentifierUsage {
...
 
+
+// === Helper Datatypes ===
+
+/// Says whether the identifier occurrence introduces it into scope or uses it from scope.
+#[allow(missing_docs)]
+#[derive(Clone,Copy,Debug,Display,PartialEq)]
+pub enum OccurrenceKind { Used, Introduced }
+
+/// If the current context in the AST processor is a pattern context.
```
farmaazon, line 83: I'd like to have a reference to a document which describes what pattern context is.

src/rust/ide/utils/src/vec.rs
```diff
@@ -16,6 +16,17 @@ pub trait VecExt<T> : AsMut<Vec<T>> {
...
+    /// Removes the last `n` elements from the vector.
+    ///
+    /// Returns Some if elements were succesfully removed.
+    /// Returns None and leaves self intact, if there were not enough elements.
+    fn remove_last_n(&mut self, n:usize) -> Option<()> {
+        let vec      = self.as_mut();
+        let new_size = vec.len().checked_sub(n)?;
+        vec.truncate(new_size);
+        Some(())
+    }
```
farmaazon, line 29: I find returning option for unit kinda, hmm... strange. Maybe return bool?
```suggestion
    /// Removes the last `n` elements from the vector.
    ///
    /// Returns Some if elements were succesfully removed.
    /// Returns None and leaves self intact, if there were not enough elements.
    fn remove_last_n(&mut self, n:usize) -> bool {
        {
            let vec      = self.as_mut();
            let new_size = vec.len().checked_sub(n)?;
            vec.truncate(new_size);
            Some(())
        }.is_some()
    }
```
Not sure if it actually compiles.

src/rust/Cargo.toml
```diff
@@ -47,4 +47,4 @@ debug     = false
 [profile.test]
 opt-level = 0
 lto       = false
-debug     = false
+debug     = true
```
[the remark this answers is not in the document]
mwu-tow: Ok, I need this to properly debug my test, it also provides line numbers in call stack when test is failing.

@wdanilo Can you comment why it was set to false? Is it still necessary?

src/rust/ide/src/double_representation/alias_analysis.rs
mwu-tow: Will add TODO comment, to be fixed when #349 lands.

PR #364, opened 2020-04-16
Enable alpha blending for ID pass

### LINKED ISSUE #331: Node selection not working
### General Summary
Selecting nodes in the debug/shapes scene does not work for me.
Related to #150 

#### Environment
Windows 10 Pro 1903
Chrome  80.0.3987.149 
Rev 1d2fceabbbd29684c082773bb31d952e0f7c1911

### Steps to Reproduce
Run `node run watch`.
Navigate to http://localhost:8080/debug/shapes .
Place cursor on canvas.
Press n to create node.
Click on node.

### Expected Result
Node should be highlighted.

### Actual Result
Nothing happens.

### Enso Version
Rev 1d2fceabbbd29684c082773bb31d952e0f7c1911

### Additional information
The callback in `world.rs:147` always receives `[0,0,0,0]`.
The cursor position in `pixel_reader:119` seems correct.
Hard coding the `output_id` in `fragment_runer.glsl` does not change the behaviour.

MichaelMauderer:
### Pull Request Description
This PR addresses issue #317 by changing the texture format of symbol IDs to `Rgba32f`. This enables correct selections of nodes close to each other.

This also seems to resolve #331 .

### Important Notes
This PR adds a requirement for the WebGL extension `EXT_color_buffer_float` to allow us the use of a float buffer.

This PR adds a new API to specify parameters for textures, which is required to set non-default parameters for the float texture.

As opposed to the requirements in isssue #317 this PR does not implement bit-reinterpretation between floats and integers, as these can lead to[ unexpected issues with `NaN`s](https://doc.rust-lang.org/1.19.0/std/primitive.f64.html#method.from_bits), which might lead to unexpected issues down the line. Instead, the numbers are cast between the datatypes.

### Checklist

### DIFF
--- a/src/rust/ensogl/src/display/render/composer.rs
+++ b/src/rust/ensogl/src/display/render/composer.rs
@@ -109,7 +109,8 @@ impl ComposerPass {
             let name    = format!("pass_{}",output.name());
             let args    = (self.width,self.height);
             let texture = uniform::get_or_add_gpu_texture_dyn
-                (&self.context,&self.variables,&name,output.internal_format,output.item_type,args);
+                (&self.context,&self.variables,&name,output.internal_format,output.item_type,args,
+                 Some(output.parameters));
             self.add_output(texture);
         }
     }
--- a/src/rust/ensogl/src/display/render/composer.rs
+++ b/src/rust/ensogl/src/display/render/composer.rs
@@ -110,7 +110,7 @@ impl ComposerPass {
             let args    = (self.width,self.height);
             let texture = uniform::get_or_add_gpu_texture_dyn
                 (&self.context,&self.variables,&name,output.internal_format,output.item_type,args,
-                 Some(output.parameters));
+                 Some(output.texture_parameters));
             self.add_output(texture);
         }
     }
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -33,8 +33,8 @@ impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
         let color_parameters = texture::Parameters::default();
         let id_parameters = texture::Parameters{
-            min_filter: Context::NEAREST,
-            mag_filter: Context::NEAREST,
+            min_filter : Context::NEAREST,
+            mag_filter : Context::NEAREST,
             ..Default::default()
         };
         vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8, color_parameters)
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -33,8 +33,8 @@ impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
         let color_parameters = texture::Parameters::default();
         let id_parameters = texture::Parameters{
-            min_filter : Context::NEAREST,
-            mag_filter : Context::NEAREST,
+            min_filter : texture::MinFilterValue::NEAREST,
+            mag_filter : texture::MagFilterValue::NEAREST,
             ..Default::default()
         };
         vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8, color_parameters)
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -33,9 +33,9 @@ impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
         let color_parameters = texture::Parameters::default();
         let id_parameters    = texture::Parameters {
-            min_filter : texture::MinFilterValue::NEAREST,
-            mag_filter : texture::MagFilterValue::NEAREST,
-            ..Default::default()
+            min_filter : texture::MinFilter::NEAREST,
+            mag_filter : texture::MagFilter::NEAREST,
+            ..default()
         };
         vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8,color_parameters)
             , RenderPassOutput::new("id",texture::Rgba,texture::item_type::u8,id_parameters)
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -38,7 +38,7 @@ impl RenderPass for SymbolsRenderPass {
             ..Default::default()
         };
         vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8,color_parameters)
-            , RenderPassOutput::new("id",texture::Rgba32f,texture::item_type::f32,id_parameters)
+            , RenderPassOutput::new("id",texture::Rgba,texture::item_type::u8,id_parameters)
             ]
     }
 
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -32,15 +32,14 @@ impl SymbolsRenderPass {
 impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
         vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8)
-            , RenderPassOutput::new("id",texture::Rgba32ui,texture::item_type::u32)
+            , RenderPassOutput::new("id",texture::Rgba32f,texture::item_type::f32)
             ]
     }
 
     fn run(&mut self, context:&Context, _:&UniformScope) {
         let arr = vec![0.0,0.0,0.0,0.0];
-        let arr2 = vec![0,0,0,0];
         context.clear_bufferfv_with_f32_array(Context::COLOR,0,&arr);
-        context.clear_bufferuiv_with_u32_array(Context::COLOR,1,&arr2);
+        context.clear_bufferfv_with_f32_array(Context::COLOR,1,&arr);
         self.target.set_camera(&self.views.main.camera);
         self.target.render_by_ids(&self.views.main.symbols());
     }
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -32,13 +32,13 @@ impl SymbolsRenderPass {
 impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
         let color_parameters = texture::Parameters::default();
-        let id_parameters = texture::Parameters{
+        let id_parameters    = texture::Parameters {
             min_filter : texture::MinFilterValue::NEAREST,
             mag_filter : texture::MagFilterValue::NEAREST,
             ..Default::default()
         };
-        vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8, color_parameters)
-            , RenderPassOutput::new("id",texture::Rgba32f,texture::item_type::f32, id_parameters)
+        vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8,color_parameters)
+            , RenderPassOutput::new("id",texture::Rgba32f,texture::item_type::f32,id_parameters)
             ]
     }
 
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -31,8 +31,14 @@ impl SymbolsRenderPass {
 
 impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
-        vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8)
-            , RenderPassOutput::new("id",texture::Rgba32f,texture::item_type::f32)
+        let color_parameters = texture::Parameters::default();
+        let id_parameters = texture::Parameters{
+            min_filter: Context::NEAREST,
+            mag_filter: Context::NEAREST,
+            ..Default::default()
+        };
+        vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8, color_parameters)
+            , RenderPassOutput::new("id",texture::Rgba32f,texture::item_type::f32, id_parameters)
             ]
     }
 
--- a/src/rust/ensogl/src/display/render/passes/symbols.rs
+++ b/src/rust/ensogl/src/display/render/passes/symbols.rs
@@ -33,8 +33,8 @@ impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
         let color_parameters = texture::Parameters::default();
         let id_parameters    = texture::Parameters {
-            min_filter : texture::MinFilter::NEAREST,
-            mag_filter : texture::MagFilter::NEAREST,
+            min_filter : texture::MinFilter::Nearest,
+            mag_filter : texture::MagFilter::Nearest,
             ..default()
         };
         vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8,color_parameters)
--- a/src/rust/ensogl/src/display/render/pipeline.rs
+++ b/src/rust/ensogl/src/display/render/pipeline.rs
@@ -57,17 +57,17 @@ pub struct RenderPassOutput {
     /// Item texture type of the pass framebuffer's attachment.
     pub item_type : AnyItemType,
     /// Texture parameters that will be set when binding the texture.
-    pub parameters: Parameters,
+    pub texture_parameters : Parameters,
 }
 
 impl RenderPassOutput {
     /// Constructor.
     pub fn new<Name:Str,F:Into<AnyInternalFormat>,T:Into<AnyItemType>>
-    (name:Name, internal_format:F, item_type:T, parameters:Parameters) -> Self {
+    (name:Name, internal_format:F, item_type:T, texture_parameters:Parameters) -> Self {
         let name            = name.into();
         let internal_format = internal_format.into();
         let item_type       = item_type.into();
-        Self {name,internal_format,item_type,parameters}
+        Self {name,internal_format,item_type, texture_parameters }
     }
 
     /// Getter.
--- a/src/rust/ensogl/src/display/render/pipeline.rs
+++ b/src/rust/ensogl/src/display/render/pipeline.rs
@@ -3,7 +3,7 @@
 use crate::prelude::*;
 
 use crate::system::gpu::types::*;
-use crate::system::gpu::texture::*;
+use crate::system::gpu::texture;
 
 
 
@@ -53,17 +53,17 @@ pub struct RenderPassOutput {
     /// Name of the pass.
     pub name : String,
     /// Internal texture format of the pass framebuffer's attachment.
-    pub internal_format : AnyInternalFormat,
+    pub internal_format : texture::AnyInternalFormat,
     /// Item texture type of the pass framebuffer's attachment.
-    pub item_type : AnyItemType,
+    pub item_type : texture::AnyItemType,
     /// Texture parameters that will be set when binding the texture.
-    pub texture_parameters : Parameters,
+    pub texture_parameters : texture::Parameters,
 }
 
 impl RenderPassOutput {
     /// Constructor.
-    pub fn new<Name:Str,F:Into<AnyInternalFormat>,T:Into<AnyItemType>>
-    (name:Name, internal_format:F, item_type:T, texture_parameters:Parameters) -> Self {
+    pub fn new<Name:Str,F:Into<texture::AnyInternalFormat>,T:Into<texture::AnyItemType>>
+    (name:Name, internal_format:F, item_type:T, texture_parameters:texture::Parameters) -> Self {
         let name            = name.into();
         let internal_format = internal_format.into();
         let item_type       = item_type.into();
--- a/src/rust/ensogl/src/display/render/pipeline.rs
+++ b/src/rust/ensogl/src/display/render/pipeline.rs
@@ -56,16 +56,18 @@ pub struct RenderPassOutput {
     pub internal_format : AnyInternalFormat,
     /// Item texture type of the pass framebuffer's attachment.
     pub item_type : AnyItemType,
+    /// Texture parameters that will be set when binding the texture.
+    pub parameters: Parameters,
 }
 
 impl RenderPassOutput {
     /// Constructor.
     pub fn new<Name:Str,F:Into<AnyInternalFormat>,T:Into<AnyItemType>>
-    (name:Name, internal_format:F, item_type:T) -> Self {
+    (name:Name, internal_format:F, item_type:T, parameters:Parameters) -> Self {
         let name            = name.into();
         let internal_format = internal_format.into();
         let item_type       = item_type.into();
-        Self {name,internal_format,item_type}
+        Self {name,internal_format,item_type,parameters}
     }
 
     /// Getter.
--- a/src/rust/ensogl/src/display/render/pipeline.rs
+++ b/src/rust/ensogl/src/display/render/pipeline.rs
@@ -67,7 +67,7 @@ impl RenderPassOutput {
         let name            = name.into();
         let internal_format = internal_format.into();
         let item_type       = item_type.into();
-        Self {name,internal_format,item_type, texture_parameters }
+        Self {name,internal_format,item_type,texture_parameters}
     }
 
     /// Getter.
--- a/src/rust/ensogl/src/display/scene.rs
+++ b/src/rust/ensogl/src/display/scene.rs
@@ -123,20 +123,19 @@ impl Target {
     fn to_internal(&self) -> Vector4<u32> {
         match self {
             Self::Background                     => Vector4::new(0,0,0,0),
-            Self::Symbol {symbol_id,instance_id} => Vector4::new(*symbol_id,*instance_id,0,1),
+            Self::Symbol {symbol_id,instance_id} => {
+                Vector4::new(*symbol_id,*instance_id / 255,*instance_id % 255,1)
+            },
         }
     }
 
     fn from_internal(v:Vector4<u32>) -> Self {
-        if v.z != 0 {
-            panic!("Wrong internal format for mouse target.")
-        }
         if v.w == 0 {
             Self::Background
         }
-        else if v.w == 1 {
-            let symbol_id   = v.x;
-            let instance_id = v.y;
+        else if v.w == 255 {
+            let symbol_id   = v.x ;
+            let instance_id = v.y * 255 + v.z;
             Self::Symbol {symbol_id,instance_id}
         } else {
             panic!("Wrong internal format alpha for mouse target.")
--- a/src/rust/ensogl/src/display/scene.rs
+++ b/src/rust/ensogl/src/display/scene.rs
@@ -124,7 +124,7 @@ impl Target {
     /// Decode the symbol_id and instance_id that was encoded in the `fragment_runner`.
     ///
     /// See the `encode` method for more information on the encoding.
-    fn decode(pack1:u32, pack2:u32, pack3:u32) -> (u32, u32){
+    fn decode(pack1:u32, pack2:u32, pack3:u32) -> (u32, u32) {
         let value1 = (pack1 << 4) + (pack2 >> 4);
         let value2 = pack3 + ((pack2 & 0x00FF) << 8);
         (value1, value2)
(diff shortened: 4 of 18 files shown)


src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,19 +45,65 @@ impl Drop for TextureBindGuard {
 
 
 
+// ==================
+// === Parameters ===
+// ==================
+
+/// Parameters that need to be set when binding a texture.
```
wdanilo, line 52: Please provide link to MDN or something explaining the concepts better.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,19 +45,65 @@ impl Drop for TextureBindGuard {
...
+#[derive(Copy, Clone, Debug)]
+pub struct Parameters{
+    /// Specifies the value for `Context::TEXTURE_MIN_FILTER`.
+    pub min_filter : u32,
+    /// Specifies the value for `Context::TEXTURE_MAG_FILTER`.
+    pub mag_filter : u32,
+    /// Specifies the value for `Context::TEXTURE_WRAP_S`.
+    pub wrap_s     : u32,
+    /// Specifies the value for `Context::TEXTURE_WRAP_T`.
+    pub wrap_t     : u32,
```
wdanilo, line 62: Please create newtypes for that. We try not to use magical `u32` values here.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,19 +45,65 @@ impl Drop for TextureBindGuard {
...
+impl Default for Parameters{
+    fn default() -> Self {
+        Parameters{
+            min_filter : Context::LINEAR,
+            mag_filter : Context::LINEAR,
+            wrap_s     : Context::CLAMP_TO_EDGE,
+            wrap_t     : Context::CLAMP_TO_EDGE,
+        }
+    }
+}
```
wdanilo, line 74: after you create newtypoes this should be derived automatically

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,19 +45,65 @@ impl Drop for TextureBindGuard {
...
+
+
+
 // ===============
 // === Texture ===
 // ===============
 
+
+
```
wdanilo, line 95: what is this offset for ?

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,19 +45,65 @@ impl Drop for TextureBindGuard {
...
 pub struct Texture<Storage,InternalFormat,ItemType>
 where Storage: StorageRelation<InternalFormat,ItemType> {
-    storage    : StorageOf<Storage,InternalFormat,ItemType>,
-    gl_texture : WebGlTexture,
-    context    : Context,
+        storage    : StorageOf<Storage,InternalFormat,ItemType>,
+        gl_texture : WebGlTexture,
+        context    : Context,
+    /// Texture parameters that need to be set when binding the texture.
+    pub parameters : Parameters
```
wdanilo, line 106: Why this needs to be public here?

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -161,16 +207,14 @@ where S:StorageRelation<I,T> {
         let storage    = storage.into();
         let context    = context.clone();
         let gl_texture = context.create_texture().unwrap();
-        Self {storage,gl_texture,context}
+        let parameters = Parameters::default();
```
wdanilo, line 210: why not just `parameters = default()` ?

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -161,16 +207,14 @@ where S:StorageRelation<I,T> {
...
 
-    /// Sets the texture wrapping parameters.
-    pub fn set_texture_parameters(context:&Context) {
-        let target = Context::TEXTURE_2D;
-        let wrap   = Context::CLAMP_TO_EDGE as i32;
-        context.tex_parameteri(target,Context::TEXTURE_MIN_FILTER,Context::LINEAR as i32);
-        context.tex_parameteri(target,Context::TEXTURE_WRAP_S,wrap);
-        context.tex_parameteri(target,Context::TEXTURE_WRAP_T,wrap);
+
+    /// Sets the texture wrapping parameters without requiring self.
```
wdanilo, line 215: I dont understand the comment. What does "without requiring self" mean?

src/rust/ensogl/src/system/gpu/data/uniform.rs
```diff
@@ -334,12 +334,16 @@ macro_rules! define_get_or_add_gpu_texture_dyn {
         , internal_format : AnyInternalFormat
         , item_type       : AnyItemType
         , provider        : P
+        , parameters      : Option<Parameters>
```
wdanilo, line 337: Please never import names to scope which does not make sense here. What are uniform `Parameters` ? We should use names in qualified manner so the code is understandable. If Im in a file `uniform` and I see name `Parameters` it means "uniform parameters". This should be `texture::Parameters` instead.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
wdanilo: Moreover, please write more in the docs that these parameters are about filtering etc. When reading the source code it should be more easy to follow what it is about.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,19 +45,65 @@ impl Drop for TextureBindGuard {
...
 
 
+// ==================
+// === Parameters ===
+// ==================
+
+/// Parameters that need to be set when binding a texture.
+#[derive(Copy, Clone, Debug)]
+pub struct Parameters{
+    /// Specifies the value for `Context::TEXTURE_MIN_FILTER`.
```
wdanilo, line 55: This doc is not understandable. In order to understand it you need to know what `Context::TEXTURE_MIN_FILTER`. All docs should explain semantically the meaning of things. It should be described here really short what min filter is used for actually. On the other hand, after you create newtypes and add docs there, documentation of these fields could be removed and we can even use `allow(missing_docs)` as the docs would already be there.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -203,7 +203,11 @@ pub fn get_webgl2_context(canvas:&HtmlCanvasElement) -> WebGl2RenderingContext {
     let options = js_sys::Object::new();
     js_sys::Reflect::set(&options, &"antialias".into(), &false.into()).unwrap();
     let context = canvas.get_context_with_context_options("webgl2",&options).unwrap().unwrap();
-    context.dyn_into().unwrap()
+    let context : WebGl2RenderingContext =  context.dyn_into().unwrap();
+    context
+        .get_extension("EXT_color_buffer_float")
```
wdanilo, line 208: Add comment here what it is used for.

src/rust/ensogl/src/display/render/composer.rs
```diff
@@ -109,7 +109,8 @@ impl ComposerPass {
             let name    = format!("pass_{}",output.name());
             let args    = (self.width,self.height);
             let texture = uniform::get_or_add_gpu_texture_dyn
-                (&self.context,&self.variables,&name,output.internal_format,output.item_type,args);
+                (&self.context,&self.variables,&name,output.internal_format,output.item_type,args,
+                 Some(output.parameters));
```
wdanilo, line 113: It took me a while to understand what a field "output pass . parameters" mean. It should be named `texture_parameters` instead.

src/rust/ensogl/src/display/render/passes/symbols.rs
```diff
@@ -31,16 +31,21 @@ impl SymbolsRenderPass {
 
 impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
-        vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8)
-            , RenderPassOutput::new("id",texture::Rgba32ui,texture::item_type::u32)
+        let color_parameters = texture::Parameters::default();
+        let id_parameters = texture::Parameters{
```
wdanilo, line 35: align

wdanilo: Plus space before {

src/rust/ensogl/src/display/render/passes/symbols.rs
```diff
@@ -31,16 +31,21 @@ impl SymbolsRenderPass {
...
+            min_filter : Context::NEAREST,
+            mag_filter : Context::NEAREST,
+            ..Default::default()
+        };
+        vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8, color_parameters)
+            , RenderPassOutput::new("id",texture::Rgba32f,texture::item_type::f32, id_parameters)
```
wdanilo, line 41: we do not use spaces between args when calling fns

src/rust/ensogl/src/display/render/pipeline.rs
```diff
@@ -56,16 +56,18 @@ pub struct RenderPassOutput {
     pub internal_format : AnyInternalFormat,
     /// Item texture type of the pass framebuffer's attachment.
     pub item_type : AnyItemType,
+    /// Texture parameters that will be set when binding the texture.
+    pub parameters: Parameters,
```
wdanilo, line 60: What are "pipeline parameters"? See later comments - they explain the issue better.

src/rust/ensogl/src/display/render/passes/symbols.rs
wdanilo, line 41: How is it working? Seriously, I need proper explanation here. Accoridng to the table https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D , the `Rgba32f` format is not color renderable. It should not work, right?

src/rust/ensogl/src/display/symbol/gpu.rs
```diff
@@ -348,6 +348,12 @@ impl Symbol {
                 let count          = self.surface.point_scope().size()    as i32;
                 let instance_count = self.surface.instance_scope().size() as i32;
 
+                debug_assert_eq!(
+                    self.context.check_framebuffer_status(Context::FRAMEBUFFER),
+                    Context::FRAMEBUFFER_COMPLETE,
+                    "Framebuffer does not have status FRAMEBUFFER_COMPLETE."
+                );
+
```
wdanilo, line 356: Super cool! One request - please make the error more desriptive. If someone gets this error it should be understandable for us what happened there.

src/rust/ensogl/src/display/shape/primitive/glsl/fragment_runner.glsl
```diff
@@ -4,9 +4,9 @@ Env   env        = Env(1);
 vec2  position   = input_local.xy ;
 Shape shape      = run(env,position);
 float alpha      = shape.color.color.raw.a;
-uint alpha_no_aa = alpha > 0.5 ? uint(1) : uint(0);
+float alpha_no_aa = alpha > 0.5 ? 1.0 : 0.0;
 
-output_id = uvec4(input_symbol_id,input_instance_id,0,alpha_no_aa);
+output_id = vec4(float(input_symbol_id),float(input_instance_id),0,alpha_no_aa);
```
wdanilo, line 9: This conversion is wrong. We need bit reinterpretation here and it is needed for this task to be merged.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
MichaelMauderer: This doc should no longer be there. Removed.

src/rust/ensogl/src/system/gpu/data/uniform.rs
MichaelMauderer: The texture module was already blanked imported in this file and used all over this place. Messing with this import is going to be a bit of a time sink.

src/rust/ensogl/src/display/render/passes/symbols.rs
```diff
@@ -31,16 +31,21 @@ impl SymbolsRenderPass {
 
 impl RenderPass for SymbolsRenderPass {
     fn outputs(&self) -> Vec<RenderPassOutput> {
-        vec![ RenderPassOutput::new("color",texture::Rgba,texture::item_type::u8)
-            , RenderPassOutput::new("id",texture::Rgba32ui,texture::item_type::u32)
+        let color_parameters = texture::Parameters::default();
+        let id_parameters    = texture::Parameters {
+            min_filter : texture::MinFilterValue::NEAREST,
+            mag_filter : texture::MagFilterValue::NEAREST,
+            ..Default::default()
```
wdanilo, line 38: Please clean the codebase and then run Clippy on it. It should catch that you don't need qualification here :)

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,6 +45,109 @@ impl Drop for TextureBindGuard {
...
+    fn default() -> Self {
+        Self::LINEAR
+    }
+}
+
+/// Valid Parameters for the `gl.TEXTURE_MIN_FILTER` texture setting.
+/// Specifies how values are interpolated if the texture is rendered at a resolution that is
+/// lower than its native resolution.
+#[derive(Copy,Clone,Debug)]
+pub struct MinFilterValue(u32);
```
wdanilo, line 111: Implement it as enum pls. Enum has advantage of being checked by the compiler when matching over - so you can be sure that you handle all possible cases. The same applies to `Wrap` below.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,6 +45,109 @@ impl Drop for TextureBindGuard {
...
+/// The essential parameters that need to be set are about how the texture will be samples, i.e.,
+/// how the values of the texture are interpolated at various resolutions, and how out of bounds
+/// samples are handled.
+///
+/// For more background see: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texParameter
+///
+#[derive(Copy,Clone,Debug,Default)]
+pub struct Parameters {
+    /// Specifies the setting for the texture magnification filter (`Context::TEXTURE_MIN_FILTER`).
+    pub min_filter : MinFilterValue,
```
wdanilo, line 63: Small naming issue - we try not to add suffixes like `Value` or `Trait` to our names.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,6 +45,109 @@ impl Drop for TextureBindGuard {
...
+}
+
+impl Parameters {
+    /// Sets the context parameters in the given context.
+    pub fn set_parameters(&self, context:&Context) {
+        let target = Context::TEXTURE_2D;
+        context.tex_parameteri(target,Context::TEXTURE_MIN_FILTER,self.min_filter.0 as i32);
+        context.tex_parameteri(target,Context::TEXTURE_MIN_FILTER,self.mag_filter.0 as i32);
+        context.tex_parameteri(target,Context::TEXTURE_WRAP_S,self.wrap_s.0 as i32);
+        context.tex_parameteri(target,Context::TEXTURE_WRAP_T,self.wrap_t.0 as i32);
```
wdanilo, line 81: Code using `.0` is hacky as hell and hard to understand. Please name the field accordingly, like `gl_value`. Seeing `min_filter.gl_value` is much better than seeing `min_filter.0`.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,6 +45,109 @@ impl Drop for TextureBindGuard {
...
+    }
+}
+
+
+// === Parameter Types ===
+
+/// Valid Parameters for the `gl.TEXTURE_MAG_FILTER` texture setting.
+/// Specifies how values are interpolated if the texture is rendered at a resolution that is
+/// lower than its native resolution.
```
wdanilo, line 90: Please format the docs. If the first line is meant to be the caption, it needs a blank line below (Rust doc comments require it). If it is not meant to be caption, use all 100 chars wide in all lines.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -58,6 +161,8 @@ where Storage: StorageRelation<InternalFormat,ItemType> {
     storage    : StorageOf<Storage,InternalFormat,ItemType>,
     gl_texture : WebGlTexture,
     context    : Context,
+    /// Texture parameters that need to be set when binding the texture.
```
wdanilo, line 164: This is a self-explanatory field imo. These docs might be removed.

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -161,16 +283,13 @@ where S:StorageRelation<I,T> {
...
 
     /// Sets the texture wrapping parameters.
-    pub fn set_texture_parameters(context:&Context) {
-        let target = Context::TEXTURE_2D;
-        let wrap   = Context::CLAMP_TO_EDGE as i32;
-        context.tex_parameteri(target,Context::TEXTURE_MIN_FILTER,Context::LINEAR as i32);
-        context.tex_parameteri(target,Context::TEXTURE_WRAP_S,wrap);
-        context.tex_parameteri(target,Context::TEXTURE_WRAP_T,wrap);
+    pub fn set_texture_parameters(&self, context:&Context) {
+        self.parameters.set_parameters(context);
```
wdanilo, line 292: Something is bad with naming here. We have functions `Texture::set_parameters` and `Texture::set_texture_parameters`. Moreover the later is defined as `self.parameters.set_parameters(context);` which is totally not understandable. If it was meant to apply the parameters to context maybe using `apply_parameters` would make more sense?

src/rust/ensogl/src/system/gpu/data/uniform.rs
wdanilo: I bet it will not take more than couple of minutes :) If it was blanket imported then it was a bad design, but also I believe that no types which require qualification were used, only such as `Texture` which should be imported separately. Anyway, we should not merge a code which is confusing, and this one is, as `Uniform::Parameters` does not make sense.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -203,7 +203,13 @@ pub fn get_webgl2_context(canvas:&HtmlCanvasElement) -> WebGl2RenderingContext {
     let options = js_sys::Object::new();
     js_sys::Reflect::set(&options, &"antialias".into(), &false.into()).unwrap();
     let context = canvas.get_context_with_context_options("webgl2",&options).unwrap().unwrap();
-    context.dyn_into().unwrap()
+    let context : WebGl2RenderingContext =  context.dyn_into().unwrap();
+    // Add this extension so we can use RGB32f textures for rendering object ids in our picking
+    // pipeline.
+    context
+        .get_extension("EXT_color_buffer_float")
```
wdanilo, line 210: Could you please update comment in `texture/types/relations.rs` that this extension makes some changes to the table provided there?

src/rust/ensogl/src/display/render/pipeline.rs
```diff
@@ -56,16 +56,18 @@ pub struct RenderPassOutput {
...
     pub item_type : AnyItemType,
+    /// Texture parameters that will be set when binding the texture.
+    pub texture_parameters : Parameters,
 }
 
 impl RenderPassOutput {
     /// Constructor.
     pub fn new<Name:Str,F:Into<AnyInternalFormat>,T:Into<AnyItemType>>
-    (name:Name, internal_format:F, item_type:T) -> Self {
+    (name:Name, internal_format:F, item_type:T, texture_parameters:Parameters) -> Self {
```
wdanilo, line 66: `Pipeline::Parameters` does not make sense. Should be qualified.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +120,63 @@ pub enum Target {
...
 impl Target {
+
+    /// Decode the symbol_id and instance_id that was encoded in the `fragment_runner`.
+    ///
+    /// See the `encode` method for more information on the encoding.
+    fn decode(pack1:u32, pack2:u32, pack3:u32) -> (u32,u32) {
+        let value1 = (pack1 << 4) + (pack2 >> 4);
+        let value2 = pack3 + ((pack2 & 0x000F) << 8);
+        (value1, value2)
+    }
```
wdanilo, line 131: a very minor thing - I would rename `pack` to `chunk` to better illustrate what it is about.

wdanilo: Plus, encode provides proper docs here. In such a case, move `decode` under `encode` so when reading the code it is easier to first read proper docs :)

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +120,63 @@ pub enum Target {
...
+
+    /// Encode two u32 values into three u8 values.
+    /// This is the same encoding that is used in the `fragment_runner`.
+    /// This encoding is lossy.
```
wdanilo, line 135: Please use all 100 chars wide in doc sentences.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +120,63 @@ pub enum Target {
...
+    ///
+    /// We use 12 bits from each value and pack them into the 3 output bites like
```
wdanilo, line 137: I believe there is a typo `bites` -> `bytes`.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +120,63 @@ pub enum Target {
...
+    ///  value1 as bytes                    value2 as bytes
+    /// -------------------------           -------------------------
+    /// | v1a | v1b | v1c | v1d |           | v2a | v2b | v2c | v2d |
+    /// -------------------------           -------------------------
+    ///
+    ///  Output
+    ///
+    /// byte1                      byte2                      byte3
+    /// -----------------------    -----------------------     -------
+    /// | v2c[4..8] v1d[0..4] |    | v1d[4..8] v2c[4..8] |     | v2d |
```
wdanilo, line 151: The arrays do not show how bounds behave, please use Rust (`[x..=y]`) notion to show that. Plus, are you sure the drawings are correct? I cannot get my head around how they work (there is for example `v2c[4..8]` two times and they are strangely layouted).

src/rust/ensogl/src/display/scene.rs
wdanilo: Plus, as I normally love your docs because they contain every information that is needed, here I'm missing some important info, namely info about max ID that could be encoded as `symbol_id` and `instance_id`. It should be told very explicitly. Moreover, when trying to encode too big id, an warning should be emitted to console by using our logger.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +120,63 @@ pub enum Target {
...
+    /// -----------------------    -----------------------     -------
+    ///
+    fn encode(value1:u32, value2:u32) -> (u8,u8,u8) {
+        let pack1 = (value1 >> 4u32) & 0x00FFu32;
+        let pack2 = (value1 & 0x000Fu32) << 4u32;
+        let pack2 = pack2 | ((value2 & 0x0F00u32) >> 8u32);
+        let pack3 = value2 & 0x00FFu32;
+        (pack1 as u8, pack2 as u8 , pack3 as u8)
```
wdanilo, line 159: no space before `,`

src/rust/ensogl/src/display/scene.rs
```diff
@@ -151,6 +191,57 @@ impl Default for Target {
...
+#[cfg(test)]
+mod target_tests {
+    use super::*;
+
+    fn assert_valid_roundtrip(value1:u32, value2:u32) {
+        let pack   = Target::encode(value1,value2);
+        let unpack = Target::decode(pack.0.into(),pack.1.into(),pack.2.into());
+        assert_eq!(unpack.0,value1);
+        assert_eq!(unpack.1,value2);
+    }
```
wdanilo, line 205: It would be cool to have docs here. I don't understand why you want this to compile now: `assert_valid_roundtrip(1024, 999)`

src/rust/ensogl/src/display/shape/primitive/glsl/fragment_runner.glsl
```diff
@@ -4,13 +4,34 @@ Env   env        = Env(1);
...
+// src\rust\ensogl\src\display\scene.rs
+// See there for more explanation.
+uint pack1 = (uint(input_symbol_id) >> 4u) & 0x00FFu;
+uint pack2 = (uint(input_symbol_id) & 0x000Fu) << 4u;
+     pack2 = pack2 + ((uint(input_instance_id) & 0x0F00u) >> 8u);
+uint pack3 = uint(input_instance_id) & 0x00FFu;
+
+float alpha_no_aa = alpha > 0.5 ? 1.0 : 0.0;
+
+output_id = vec4(float(pack1) / 255.0, float(pack2) / 255.0, float(pack3) / 255.0, alpha_no_aa);
```
wdanilo, line 24: Could you please refactor that to a function in `glsl/math.glsl` please?

src/rust/ensogl/src/display/shape/primitive/system.rs
```diff
@@ -58,7 +58,7 @@ impl ShapeSystem {
         material.add_input  ("time"         , 0.0);
         material.add_input  ("symbol_id"    , 0);
         material.add_input  ("display_mode" , 0);
-        material.add_output ("id"           , Vector4::<u32>::new(0,0,0,0));
+        material.add_output ("id"           , Vector4::<f32>::new(0.0,0.0,0.0,0.0));
```
wdanilo, line 61: I know it was already ugly, but we can do better now. You can use `::zero()` instead :)

src/rust/ensogl/src/display/symbol/gpu.rs
```diff
@@ -348,6 +348,18 @@ impl Symbol {
...
                 let instance_count = self.surface.instance_scope().size() as i32;
 
+                // Check if we are ready to render. If we don't assert here we wil only get a warning
+                // that won't tell us where things went wrong.
+                {
+                    let framebuffer_status = context.check_framebuffer_status(Context::FRAMEBUFFER);
+                    debug_assert_eq!(
+                        framebuffer_status,
+                        Context::FRAMEBUFFER_COMPLETE,
+                        "Framebuffer does not have status FRAMEBUFFER_COMPLETE, but {}.",
```
wdanilo, line 358: A very minor thing. I would make the error sentence a little bit more standard (as in other implementations). I would use here `"Framebuffer incomplete (status: {})."` instead :) This is shorter and provides the same information.

src/rust/ensogl/src/display/render/passes/symbols.rs
MichaelMauderer: `node run lint` does not show anything here. Neither does the CI. But it's removed now.

src/rust/lib/system/web/src/lib.rs
MichaelMauderer: Actually we don't need that extension anymore! Is removed now.

wdanilo: Still, please add the comment to `relations.rs`. It might be needed in future and we will forget about it.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +129,84 @@ pub enum Target {
...
+    ///   |     |     |     |     |           |     |     |     |     |
+    ///   +-----+-----+-----+-----+           +-----+-----+-----+-----+
+    /// 32    24    16     8      0         32    24    16     8      0   <- Bit index
+    ///
+    ///
+    /// Output
+    ///
+    /// byte1            byte2                     byte3
+    /// +-----------+    +-------------------+     +-----------+
+    /// | v1[12..4] |    | v1[4..0] v2[4..0] |     | v2[12..4] |
```
wdanilo, line 155: Please use the notion of inclusive / non-incliusive ranges (I asked about it in my previous PR review). I believe it should be `v1[4=..=0]` and `v1[12=..4]` right? We can of course use mathematical notion instead, so `v1[12..4[` - according to your preferences.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +129,84 @@ pub enum Target {
...
             Self::Background                     => Vector4::new(0,0,0,0),
-            Self::Symbol {symbol_id,instance_id} => Vector4::new(*symbol_id,*instance_id,0,1),
+            Self::Symbol {symbol_id,instance_id} => {
+                match Self::encode(*symbol_id,*instance_id) {
+                    DecodingResult::Truncated(pack0,pack1,pack2) => {
+                        if let Some(logger) = logger{
+                            logger.warning(|| {
+                                format!("Target values too big to encode: ({},{}).",
+                                        *symbol_id,*instance_id)
+                            });
```
wdanilo, line 190: use `warning!` macro instead:

```rust
warning!(logger,"Target values too big to encode: ({symbol_id},{instance_id}).")
```

src/rust/ensogl/src/display/scene.rs
```diff
@@ -120,23 +129,84 @@ pub enum Target {
...
+    /// Decode the symbol_id and instance_id that was encoded in the `fragment_runner`.
+    ///
+    /// See the `encode` method for more information on the encoding.
+    fn decode(chunk1:u32, chunk2:u32, chunk3:u32) -> (u32, u32) {
+        let value1 = (chunk1 << 4) + (chunk2 >> 4);
+        let value2 = chunk3 + ((chunk2 & 0x000F) << 8);
+        (value1, value2)
+    }
+
+    fn to_internal(&self, logger:Option<&Logger>) -> Vector4<u32> {
```
wdanilo, line 180: Why logger is in option? Just lets assume you need to pass it always, the code will be simpler. Moreover, I don't believe  that allowing not passing logger and dropping warnings then is a good idea.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -151,6 +221,58 @@ impl Default for Target {
...
+#[cfg(test)]
+mod target_tests {
+    use super::*;
+
+    /// Asserts that decoding encoded the given values returns the correct initial values again.
+    /// That means that `decode(encode(value1,value2)) == (value1,value2)`.
+    fn assert_valid_roundtrip(value1:u32, value2:u32) {
+        let pack   = Target::encode(value1,value2);
+        match pack {
+            DecodingResult::Truncated(_, _, _) => {
```
wdanilo, line 235: Replace with `Truncated {..} =>`

src/rust/ensogl/src/system/gpu/data/texture/class.rs
```diff
@@ -45,6 +45,111 @@ impl Drop for TextureBindGuard {
...
+    NearestMipmapNearest = Context::NEAREST_MIPMAP_NEAREST as isize,
+    LinearMipmapNearest  = Context::LINEAR_MIPMAP_NEAREST  as isize,
+    NearestMipmapLinear  = Context::NEAREST_MIPMAP_LINEAR  as isize,
+    LinearMipmapLinear   = Context::LINEAR_MIPMAP_LINEAR   as isize,
+}
+
+#[allow(missing_docs)]
+impl MinFilter {
+
+}
```
wdanilo, line 124: This should be removed.

PR #365, opened 2020-04-16
FRP Improvements.

wdanilo:
# Pull Request Description
This PR improves the FRP implementation and introduces a new FRP mode. Now, FRP network can be either `static` or `dynamic` (it was only `static` in the past).

# Details

## Introduction

**READ THIS PAGE CAREFULLY.**

This module implements a sophisticated event handling framework, known as Functional Reactive
Programming system. It allows creating data flow diagrams from predefined nodes. Over years many
different FRP implementations appeared, especially in the Haskell community. The implementations
differ both in concepts they use, as well as the provided functionality. The following
implementation focuses on providing high performance and maintainability (ability to reason
about correctness of complex data flow graphs).

In order to grasp some of the main ideas we strongly encourage you to read the following FRP
introduction. Although this implementation differs in few details, understanding these concepts
would be helpful: https://github.com/hansroland/reflex-dom-inbits/blob/master/tutorial.md

## Usage
 
The FRP system exposed by this library can be used in two modes - network and dynamic. The
former was designed as a highly maintainable framework and it is what you should choose almost
always. The API was designed mainly with this use case in mind. In order to use the network
mode, just use the API exposed by `Network` and all node types without "Owned" prefix. In order
to use dynamic mode, do not use `Network` and use nodes with the "Owned" prefix. The modes can
be mixed, but unless you really know what you are doing, you are advised not to do it.

### Network mode
In this mode you define immutable FRP graphs called `Network`s. You are not allowed to add or
remove nodes from `Network` after it was defined. You are, however, allowed to create, remove,
connect, and reconnect networks dynamically. The `Network` manages the lifetime of all of
its nodes, and you are always working on weak references to the nodes. You can also define a
`BridgeNetwork` between two or more `Network`s which will be responsible for converting or
tagging data. The lifetime of `BridgeNetwork` is always managed by lifetimes of the `Network`s
it is connected to. It is dropped as soon as any of the networks gets dropped.

An example use case is a slider and scene definition. The slider component can contain its own
internal FRP network which manages its state and mouse interactions. The scene FRP network can
connect to slider network whenever a new slider is created. In order to tag the value change
events with information which slider they originate from, a `BridgeNetwork` should be used.

### Dynamic mode.
Please note, that although there are valid use cases when this mode is useful, you should always
default to network mode instead. In this mode the `Network` is not used and you are using strong
references to nodes instead (types prefixed with "Owned" in the API, like `frp::OwnedSource`).
Nodes are dropped as soon as no references exist to them anymore.

## The new FRP graph architecture

The architecture of the FRP engine is pretty complex and it is hard to understand all the
dependencies by reading the code only. In order to make it easier, the following diagram was
prepared. It visualizes a simple FRP network of `source -> count -> map(|t| t.to_string())`.
You can view the diagram in any GraphViz tool, for example this one:
https://dreampuf.github.io/GraphvizOnline

Meaning of the graph elements:
- On the left there are two `Network` nodes which will be missing if used in dynamic mode.
- Colors indicate to which network nodes belong (nodes of that color will be dropped as
  soon as the network gets dropped). Nodes with double color will be dropped as soon as an event
  will be emitted to nodes belonging to the dropped network.
- Black solid edges without arrows are just fields inside structure.
- Black solid edges with arrows are `Rc` pointers.
- Black dashed edges with arrows are `Weak` pointers.
- Red solid edges with arrows are trait object interfaces.
- The gray arrows are possible connections depending on which mode was used. In case of network
  mode, either an edge to `Stream` or a node without the "Owned" prefix will be created
  (depending on what will be provided to node constructor). In dynamic mode, as you will be
  providing strong node references to their constructors, the arrow to the node with "Owned"
  prefix will be created.

image

### Checklist


src/rust/lib/frp/src/macros.rs
```diff
@@ -68,35 +84,35 @@ macro_rules! new_bridge_network {
 
 /// Extends the provided network with new rules. See documentation of `new_network` to learn more.
 #[macro_export]
-macro_rules! _extend_network {
+macro_rules! _extend {
```
farmaazon, line 87: The documentation for `_extend` and `extend` is the same. Please add note which will distinct those two.

PR #381, opened 2020-04-21
Crumb implementation for Match and Ambiguous.

Josef-Vonasek:
### Checklist

### DIFF
--- a/src/rust/ide/ast/impl/src/crumbs.rs
+++ b/src/rust/ide/ast/impl/src/crumbs.rs
@@ -4,7 +4,6 @@
 use crate::prelude::*;
 
 use crate::known;
-use crate::Shifted;
 use crate::HasTokens;
 use crate::Shape;
 use crate::TokenConsumer;
@@ -242,12 +241,6 @@ pub enum PatternMatchCrumb {
 }
 
 
-// === Ambiguous ===
-
-#[allow(missing_docs)]
-#[derive(Clone,Copy,Debug,PartialEq,Eq,Hash)]
-pub struct AmbiguousCrumb;
-
 // === Conversion Traits ===
 
 macro_rules! from_crumb {
@@ -303,7 +296,7 @@ macro_rules! impl_crumbs {
         }
 
         /// Crumb identifies location of child AST in an AST node. Allows for a single step AST traversal.
-        #[derive(Clone,Copy,Debug,PartialEq,Eq,Hash)]
+        #[derive(Clone,Debug,PartialEq,Eq,Hash)]
         #[allow(missing_docs)]
         pub enum Crumb {
             $($id($crumb_id),)*
@@ -333,7 +326,6 @@ impl_crumbs!{
     ( Module        , ModuleCrumb        ),
     ( Block         , BlockCrumb         ),
     ( Match         , MatchCrumb         ),
-    ( Ambiguous     , AmbiguousCrumb     ),
     ( Import        , ImportCrumb        ),
     ( Mixfix        , MixfixCrumb        ),
     ( Group         , GroupCrumb         ),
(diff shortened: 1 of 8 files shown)


src/rust/ide/ast/impl/src/crumbs.rs
```diff
@@ -226,6 +246,30 @@ pub enum DefCrumb {
...
+#[derive(Clone,Debug,PartialEq,Eq,Hash)]
+pub enum MatchCrumb {
+    Pfx  {val:Vec<PatternMatchCrumb>        },
+    Segs {val:SegmentMatchCrumb, index:usize},
+}
+
+#[allow(missing_docs)]
+#[derive(Clone,Copy,Debug,PartialEq,Eq,Hash)]
+pub enum PatternMatchCrumb {
+    Seq  {left :bool },
```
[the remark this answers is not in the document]
Josef-Vonasek: Why? The boolean is always named, so there is no room for mistake.

Josef-Vonasek:
Your test works now.

src/rust/ide/ast/impl/src/crumbs.rs
```diff
@@ -677,6 +742,234 @@ impl Crumbable for crate::Block<Ast> {
...
+                _ => return missing("crumb"),
+
+            }
+        }
+        missing("elem")
+    }
+
+    fn set(&self, crumb:&Self::Crumb, new_ast:Ast) -> FallibleResult<Self> where Self: Sized {
+
+        fn set_pattern
```
mwu-tow, line 817: I'd place this in the root scope. It is well-defined function, no reason to hide it here.
Same for `iter_pat` and `get_pat`.

src/rust/ide/ast/impl/src/crumbs.rs
```diff
@@ -677,6 +742,234 @@ impl Crumbable for crate::Block<Ast> {
...
+                seg.head    = new_ast
+            }
+        }
+        Ok(new_self)
+
+    }
+
+    fn iter_subcrumbs<'a>(&'a self) -> Box<dyn Iterator<Item=Self::Crumb> + 'a> {
+
+        fn iter_pat(pat:&MacroPatternMatch<Shifted<Ast>>) -> Vec<Vec<PatternMatchCrumb>> {
```
mwu-tow, line 910: Could we have this return `Iterator<Item=Vec<PatternMatchCrumb>>` (in box if necessary?)

All `iter_subcrumbs` provide iterator-based API and that leads user to assumption that he will pay cost of traversing continually through traversal (and that cost for non-visisted subtrees is not paid).

Not a blocker (can be fixed later if ever relevant) but if feasible, would be nice.

src/rust/ide/ast/impl/src/crumbs.rs
```diff
@@ -677,6 +742,234 @@ impl Crumbable for crate::Block<Ast> {
...
+                    },
+                    Cls(pat) => {
+                        crumb.push(PatternMatchCrumb::Cls);
+                        patterns.push((crumb,&pat.elem))
+                    },
+                    Or(pat) => {
+                        crumb.push(PatternMatchCrumb::Or);
+                        match &pat.elem {
+                            crate::Switch::Left (pat) => patterns.push((crumb,&pat.value)),
+                            crate::Switch::Right(pat) => patterns.push((crumb,&pat.value)),
```
mwu-tow, line 949: I think that `Switch` should provide accessors to its contents.
Perhaps implement `Deref` for it. (esentially Switch<T> is T + flag, so it could provide access to the T)

src/rust/ide/ast/impl/src/lib.rs
```diff
@@ -79,7 +79,7 @@ pub type Stream<T> = Vec<T>;
 // ==============
 
 /// Exception raised by macro-generated TryFrom methods that try to "downcast"
-/// enum type to its variant subtype if different constructor was used.
+/// enum type to its vaShariant subtype if different constructor was used.
```
mwu-tow, line 82: Unintended edit.

src/rust/ide/ast/impl/src/crumbs.rs
```diff
@@ -247,6 +267,64 @@ pub enum DefCrumb {
...
+    Num,
+    Text,
+    Block,
+    Macro,
+    Invalid,
+    Except,
+    Tag,
+    Cls,
+    Or,
+    Seq  {right:bool },
```
mwu-tow, line 302: `seq.which==Left`  is so much nicer than `seq.right==false`.  "not right" is a very cumbersome way of expressing "left".
This is not a flag value, this is enum which happens to have two elements. 
Also, consistency with other code, like `Switch`.

PR #382, opened 2020-04-23
Style manager

wdanilo:
Partially implements [Theme manager](https://github.com/luna/ide/issues/207). To be implemented yet (in this or next PR):
- Example usage with shapes (incl. sugar for nice bindings creation).
- Ability to query which theme is used by shape under mouse.

Style sheet registry. Could be named "Cascading Style Sheets" but then the name will be
confusing with CSS used in web development. Defines a set of cascading style sheets. Each
style sheet can be assigned with a value of type `Data` or an expression to compute one. It
also allows creating variables which are automatically bound to the most specific style sheet.

Variables are associated with a style path like 'panel.button.size'
and are automatically bound to the most specific style sheet as soon as it gets defined. By
most specific, we mean the one with the longest path. For example, the 'panel.button.size' var
will be bound to one of 'panel.button.size', 'button.size', or 'size' if defined, in that order.

Style sheets are associated with a style path like 'panel.button.size'
and keep a `Data` value. The value can either be set explicitly, or computed automatically if
the style sheet is defined with en `Expression`.

An example style sheet registry (set of style sheets) is presented below:

image

### Checklist


src/rust/Cargo.toml
```diff
@@ -31,7 +31,7 @@ members = [
 ]
 
 [profile.dev]
-opt-level = 0
+opt-level = 2
```
farmaazon, line 34: changed opt-level

src/rust/lib/data/src/hash_map_tree.rs
```diff
@@ -0,0 +1,111 @@
...
+
+impl<K,T> HashMapTree<K,T>
+where K:Eq+Hash {
+    /// Constructor.
+    pub fn new() -> Self where T:Default {
+        default()
+    }
+
+    /// Constructor with explicit root value.
+    pub fn from_value(value:T) -> Self {
```
farmaazon, line 31: I would name it `single_node`

src/rust/lib/data/src/hash_map_tree.rs
```diff
@@ -0,0 +1,111 @@
...
+        let branches = default();
+        Self {value,branches}
+    }
+
+    /// Iterates over keys in `path`. For each key, traverses into the appropriate branch. In case
+    /// the branch does not exist, a default instance will be created.
+    #[inline]
```
farmaazon, line 38: Why we use `#[inline]` here? Usually in c++ manually inlining is disregarded, because if user want to have a fast application he should use lto, and lto usually optimizes much better that we could do.

src/rust/lib/data/src/hash_map_tree.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub fn focus<P,I>(&mut self, path:P) -> &mut HashMapTree<K,T>
```
farmaazon, line 39: `focus`... is this a term related to trees I haven't heard?

And doc could say what is returned.

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+}
+
+impl Mul<&Data> for &Data {
+    type Output = Data;
+    fn mul(self, rhs:&Data) -> Self::Output {
+        match(self,rhs) {
+            (Data::Invalid(t),_) => Data::Invalid(t.clone()),
+            (_,Data::Invalid(t)) => Data::Invalid(t.clone()),
+            (Data::Number(lhs),Data::Number(rhs)) => Data::Number(lhs*rhs),
+            // _ => Data::Invalid("Cannot multiply.".into())
```
farmaazon: commented code

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+}
+
+impl Add<&Data> for &Data {
+    type Output = Data;
+    fn add(self, rhs:&Data) -> Self::Output {
+        match(self,rhs) {
+            (Data::Invalid(t),_) => Data::Invalid(t.clone()),
+            (_,Data::Invalid(t)) => Data::Invalid(t.clone()),
+            (Data::Number(lhs),Data::Number(rhs)) => Data::Number(lhs+rhs),
+            // _ => Data::Invalid("Cannot multiply.".into())
```
farmaazon: commented code.

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+        }
+    }
+}
+
+impl Eq for Data {}
```
farmaazon: Why this is not put into `#[derive]`?

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+
+impl Path {
+    /// Builds the path from the provided segment iterator. Please note that the internal path
+    /// representation is reversed, as the style sheet dependencies in the style sheet map tree are
+    /// also kept in a reversed order. Please use the `visualize` utility to inspect the internal
+    /// structure and learn more.
+    pub fn from_segments<T,I,Item>(t:T) -> Self
+        where T : IntoIterator<IntoIter=I,Item=Item>,
+              I : Iterator<Item=Item>,
+              I : DoubleEndedIterator,
```
farmaazon:
```suggestion
              I : Iterator<Item=Item> + DoubleEndedIterator,
```

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+// ============
+// === Path ===
+// ============
+
+/// Path is a set of strings which describes how the variable or style sheet is nested in the
+/// cascading style sheet map.
+#[derive(Clone,Debug)]
+#[allow(missing_docs)]
+pub struct Path {
+    pub segments : Vec<String>
```
farmaazon: As I read below that _the internal path representation is reversed_ I think this field should be named "reversed_segments"

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+        where T : IntoIterator<Item=Item>,
+              Item : Into<String> {
+        let segments = t.into_iter().map(|s|s.into()).collect();
+        Self {segments}
+    }
+}
+
+impl From<&str> for Path {
+    fn from(t:&str) -> Self {
+        Self::from_rev_segments(t.rsplit("."))
```
farmaazon: I'm confused at this point.

So all the strings with dots we interpret as reversed segments. And we keep Path internally as reversed segments. And that's because also _sheet dependencies in the style sheet map tree are kept in a reversed order_. So why we cannot name the reversed segments as the normal ones?

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+#[derive(Debug)]
+pub struct Var {
+    /// Index of the var in the style var map.
+    pub index : Index<Var>,
+    /// Set of all `Sheet` indexes which are potential matches of this var. For example, for a var
+    /// 'panel.button.size', all of the following sheets will be included here: 'panel.button.size',
+    /// 'button.size', and 'size'.
+    pub matches : Vec<Index<Sheet>>,
+    /// Index of the most specific `Sheet` from `matches` which has a defined value if any.
+    pub binding : Option<Index<Sheet>>,
```
farmaazon, line 34: Do we really want to those two fields to being public, if they have somewhat "constrained" relation? One could just set something to `binding` without caring about the `matches` content.

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+    pub index : Index<Sheet>,
+    /// Current value of style sheet. Style sheets without value behave like if they do not exist.
+    pub value : Option<Data>,
+    /// Expression used to update the value.
+    pub expr : Option<Expression>,
+    /// Indexes of all `Var`s that are potential matches with this style sheet.
+    pub matches : HashSet<Index<Var>>,
+    /// Indexes of all `Var`s that are bound (best matches) with this style sheet.
+    pub bindings : HashSet<Index<Var>>,
+}
```
farmaazon, line 70: I'm constantly get confused of one thing: when I hear "style sheet" I see a whole set of properties (like size, color, border, margin, padding, etc). But here apparently we keep only one value. But I don't have any way how to fix this in mind.

src/rust/ensogl/src/display/style.rs
```diff
@@ -0,0 +1,686 @@
...
+        Self {vars,sheets,var_map,sheet_map}
+    }
+
+    /// Access variable by the given path or create new one if missing.
+    ///
+    /// Implementation note: under the hood, a `Sheet` for each sub-path will be created. For
+    /// example, when creating "panel.button.size" variable, three sheets will be created as well:
+    /// "panel.button.size", "button.size", and "size". This way we keep track of all possible
+    /// matches and we can create high-performance value binding algorithms.
+    pub fn var<P:Into<Path>>(&mut self, path:P) -> Index<Var> {
```
farmaazon, line 182: I don't see how this method fill into `constructors` section :)

src/rust/lib/data/src/hash_map_tree.rs
wdanilo: Stdlib uses it here as well. As agreed via chat, lets leave it.

src/rust/ensogl/src/display/style.rs
wdanilo: Will be fixed in next PR. This is just to mark that I want this branch to be present there.

src/rust/ensogl/src/display/style.rs
wdanilo: As previous comment.

src/rust/ensogl/src/display/style.rs
wdanilo: Because of f32, but as we've been chatting, this should not be there. Fixing.

src/rust/ensogl/src/display/style.rs
wdanilo: Style sheet is root of a style tree, so it has many values. I will fix docs.

src/rust/ensogl/src/display/style.rs
wdanilo: It uses `rsplit` so its ok :)

src/rust/ensogl/src/display/style.rs
wdanilo: Yeah, these (and other) fields should be private. Fixing.

src/rust/ensogl/src/display/style.rs
wdanilo: Its a constructor of new variable. I'm leaving it heard as it matches my taste but as soon as you recommend something better Id be happy to move it! :)

PR #383, opened 2020-04-24
Span Tree Operations

farmaazon:
### Pull Request Description
A part of Connection operations. 

Implemented Set and Erase operations on SpanTree. Also modified SpanTree, that it have much more Empty nodes, and those empty nodes now are used for inserting.

**Note:** Because we're approaching deadline, functions added to chains have not unit tests for their own, although they're covered by spantree operations tests.

### Checklist


src/rust/ide/span-tree/src/node.rs
```diff
@@ -132,6 +157,18 @@ impl<'a> Ref<'a> {
             None        => Some(self)
         }
     }
+
+    /// Get the node which exaclty matches the given Span:
```
mwu-tow, line 161:
```suggestion
    /// Get the node which exactly matches the given Span:
```

src/rust/ide/span-tree/src/node.rs
```diff
@@ -132,6 +157,18 @@ impl<'a> Ref<'a> {
...
+    pub fn find_by_span(self, span:&data::text::Span) -> Option<Ref<'a>> {
```
mwu-tow, line 162: We should document behavior when there is more than a single node with the same span.
(or document fact that such situation cannot happen — the question is how section Sides is dealt with)

src/rust/ide/span-tree/src/generate.rs
```diff
@@ -73,29 +55,35 @@ impl ChildGenerator {
...
             ast_crumbs          : vec![]
         };
         self.children.push(child);
         self.children.last().unwrap()
     }
+
+    fn reverse_children(&mut self) {
+        self.children.reverse();
+        for child in &mut self.children {
+            child.offset = self.current_offset - child.offset - child.node.size;
```
mwu-tow, line 84: Is it impossible to trigger usize wraparound?

src/rust/ide/ast/impl/src/known.rs
```diff
@@ -58,23 +46,48 @@ impl<T> KnownAst<T> {
...
+          E : Debug {
         let mut shape = self.shape().clone();
         f(&mut shape);
         self.ast = self.ast.with_shape(shape)
     }
+
+    /// Create new instance of KnownAst with mapped shape.
+    pub fn map_shape<S,E1>(&self, f:impl FnOnce(T) -> S) -> KnownAst<S>
+    where for<'t> &'t Shape<Ast> : TryInto<&'t S,Error=E1>,
+          T                      : Clone + Into<Shape<Ast>>,
```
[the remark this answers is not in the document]
farmaazon: Dunno, it haven't work without it

src/rust/ide/ast/impl/src/opr.rs
```diff
@@ -108,47 +110,77 @@ pub struct GeneralizedInfix {
...
+            },
+            Assoc::Right => GeneralizedInfix {opr,
+                left  : argument,
+                right : target,
+            },
+        }
+    }
+
+    /// Convert to AST node.
+    pub fn into_ast(self) -> Ast {
```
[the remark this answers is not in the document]
farmaazon: I could, but I didn't need it. And I would like to leave `into_ast` event having From implemented, because it's good to have such "strict" conversions (to avoid writing something like `<foo as From<GeneralizedInfix>>.from()`

src/rust/ide/ast/impl/src/opr.rs
```diff
@@ -219,27 +256,101 @@ impl Chain {
...
+
+    /// Consumes the chain and returns AST node generated from it. The ids of all Infixes and
+    /// Section don't preserve from any AST which was used to generate this chain.
+    pub fn into_ast(mut self) -> Ast {
+        while !self.args.is_empty() {
+            self.fold_arg()
+        }
+        // TODO[ao] the only case when target is none is when chain have no target and no arguments.
+        // But perhaps someone could thing that this is a valid chain. To consider returning error
+        // here.
```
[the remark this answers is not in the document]
farmaazon: But for `+` you still have some element in `args` vector. And this case will work. The only case it wont work is when target is None and args is empty list. We cannot generate such chain from AST, but anyway one could modify fields "by hand"

src/rust/ide/ast/impl/src/opr.rs
```diff
@@ -219,27 +256,101 @@ impl Chain {
...
+    /// Erase the current target from chain, and make the current first operand a new target.
+    /// Panics if there is no operand besides target.
+    pub fn erase_target(&mut self) {
+        let new_target = self.args.pop_front().unwrap().operand;
+        self.target = new_target
+    }
+
+    /// Replace the target and first argument with a new target being an proper Infix or Section
+    /// ast node. Does nothing if there are no more operands than target.
+    pub fn fold_arg(&mut self) {
```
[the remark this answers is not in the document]
farmaazon: I we have such invariant (so everyone taking this as an argument can assume that), then the fields of this struct should be also private.

src/rust/ide/ast/impl/src/opr.rs
```diff
@@ -67,26 +71,24 @@ pub fn is_assignment(ast:&Ast) -> bool {
 // ===========================
 
 /// Infix operator operand. Optional, as we deal with Section* nodes as well.
-pub type Operand = Option<Shifted<Located<Ast>>>;
+pub type Operand = Option<Shifted<Ast>>;
```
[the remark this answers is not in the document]
farmaazon: So this was one of reasons I created separated structure. But then you said to use Shifted :)

src/rust/ide/span-tree/src/node.rs
```diff
@@ -10,26 +10,40 @@ use data::text::Size;
...
     Root,
-    /// A node being a target (or "self") parameter of parent Infix, Section or Prefix.
-    Target,
+    /// A node chained with parent node. See crate's docs for more info about chaining.
+    Chained,
     /// A node representing operation (operator or function) of parent Infix, Section or Prefix.
     Operation,
+    /// A node being a target (or "self") parameter of parent Infix, Section or Prefix.
+    Target {
+        /// Indicates if you can erase this node from SpanTree.
```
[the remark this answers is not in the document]
farmaazon: https://github.com/luna/ide/blob/master/src/rust/ensogl/src/animation/loops.rs#L53
https://github.com/luna/ide/blob/master/src/rust/ensogl/src/data/container.rs#L10
And many more, just Ctrl+Shift+F and put "you"

But will change as you like. O pardon, as it is liked :)

src/rust/ide/span-tree/src/generate.rs
farmaazon: Yes, it is

src/rust/ide/ast/impl/src/known.rs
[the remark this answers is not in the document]
wdanilo: Adam, could you try it again? It's wrong. `shape` should be clone. If it was inferred to be `&T` maybe you were missing constraint `T:Clone` back then. It should work. Moreover, let's not close such comments by stating "I don't know" - we need to know why and adding constraints because compiler told so is a very bad thing and we will have problems because of that.

src/rust/ide/ast/impl/src/known.rs
```diff
@@ -58,23 +46,48 @@ impl<T> KnownAst<T> {
...
+            Ok(KnownAst {ast,phantom:default()})
+        }
+    }
 
     /// Returns the AST's shape.
-    pub fn shape<E>(&self) -> &T
-    where for<'t> &'t Shape<Ast> : TryInto<&'t T,Error=E>,
-          E                      : Debug, {
+    pub fn shape(&self) -> &T
+    where E : Debug {
```
wdanilo, line 67: Why this constraint is needed for deref? :O

src/rust/ide/ast/impl/src/opr.rs
```diff
@@ -219,30 +276,117 @@ impl Chain {
...
     /// arguments.
-    pub fn enumerate_operands<'a>(&'a self) -> impl Iterator<Item=&'a Shifted<Located<Ast>>> + 'a {
-        let this = std::iter::once(&self.target);
-        let args = self.args.iter().map(|elem| &elem.operand);
-        this.chain(args).flatten()
+    pub fn enumerate_operands<'a>(&'a self) -> impl Iterator<Item=Located<&'a Shifted<Ast>>> + 'a {
+        let this_crumbs = self.args.iter().rev().map(ChainElement::crumb_to_previous).collect_vec();
+        let this        = self.target.as_ref().map(|opr| Located::new(this_crumbs,opr));
+        let args        = self.args.iter().enumerate().map(move |(i,elem)| elem.operand.as_ref().map(|opr| {
+            let to_infix   = self.args.iter().skip(i+1).rev().map(ChainElement::crumb_to_previous);
```
wdanilo, line 294: could you make `i+1` as variable and describe what it means here please? Im trying to follow this chain of things but its not easy

src/rust/ide/ast/impl/src/opr.rs
```diff
@@ -219,30 +276,117 @@ impl Chain {
...
+    /// on. So inserting at index 0 will actually set the new operand as a new target, and the old
+    /// target will became the first argument.
+    ///
+    /// Indexing does not skip `None` operands. Function panics, if get index greater than operands
+    /// count.
+    pub fn insert_operand(&mut self, at_index:usize, operand:Shifted<Ast>) {
+        let offset      = operand.off;
+        let mut operand = Some(operand);
+        let operator    = self.operator.clone_ref();
+        if at_index == 0 {
```
wdanilo, line 322: Please extract `at_index == 0` to variable and describe semantically what it checks. Im trying to understand this code and Im failing to

src/rust/ide/ast/impl/src/prefix.rs
```diff
@@ -75,9 +75,30 @@ impl Chain {
...
         self.args.iter().map(move |arg| {
             crumbs.pop_front();
-            Located::new(&crumbs, arg)
+            Located::new(&crumbs,&arg.wrapped)
         })
     }
+
+    /// Replace the `func` and first argument with a new `func` being an proper Prefix ast node.
+    /// Does nothing if there are no arguments.
+    pub fn fold_arg(&mut self) {
```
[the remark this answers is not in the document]
wdanilo: I dont understand @mwu-tow comment, but it seems its not addressed.

src/rust/ide/span-tree/src/action.rs
```diff
@@ -1,21 +1,334 @@
...
+                    let has_target = infix.target.is_some();
+                    let has_arg    = infix.args.last().unwrap().operand.is_some();
+                    let last_elem  = infix.args.last_mut().unwrap();
+                    let last_arg   = &mut last_elem.operand;
+                    last_elem.offset = DEFAULT_OFFSET;
+                    match ins_type {
+                        BeforeTarget if has_target => infix.push_front_operand(item),
+                        AfterTarget  if has_target => infix.insert_operand(1,item),
+                        BeforeTarget               |
+                        AfterTarget                => infix.target = Some(item),
```
wdanilo, line 127: you can make it single line without braking layout

src/rust/ide/span-tree/src/action.rs
```diff
@@ -1,21 +1,334 @@
...
+                    Erase        => node.erase(&ast),
+                }.unwrap();
+                let result_repr = result.repr();
+                assert_eq!(result_repr,self.expected,"Wrong answer for case {:?}",self);
+            }
+        }
+
+        let cases:&[Case] = &
+            // Setting
+            [ Case{expr:"a + b"    , span:0..5, action:Set  , expected:"foo"            }
```
wdanilo, line 236: this should be made a very straightforward macro :P Then a line:

`Case{expr:"a + b"    , span:0..5, action:Set  , expected:"foo"            }`

Could be rewritten in:

`Set 0..5 on [a + b] = [foo]`

src/rust/ide/ast/impl/src/known.rs
farmaazon: I don't remember. if i checked it or just rewrite it from previous version

farmaazon: Without constraint I get:
```error[E0308]: mismatched types
  --> ide/ast/impl/src/known.rs:67:9
   |
67 |         self.deref()
   |         ^^^^^^^^^^^^ expected type parameter, found struct `known::KnownAst`
   |
   = note: expected type `&T`
              found type `&known::KnownAst<T>`
   = help: type parameters must be constrained to match other types
   = note: for more information, visit https://doc.rust-lang.org/book/ch10-02-traits.html#traits-as-parameters```
This is really strange to me

PR #390, opened 2020-04-29
Displaying visualizations on nodes

MichaelMauderer:
### Pull Request Description
Addresses features from #392 .
The debug/shapes scene now shows a dummy visualization that can be hidden by selecting the node and pressing "space". The visualization data can be changed by pressing "d". The data will affect the radius of the shown circle.

### Important Notes
Contains some dummy functionality that will be replaced once more issues form this project get implemented.

### Checklist


src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -198,33 +200,37 @@ impl component::ShapeViewDefinition for NodeView {
...
-        let view    = component::ShapeView::new(&logger);
-        let events  = Events {network,select,deselect};
-        let ports   = Registry::default() ;
-        let data    = Rc::new(NodeData {logger,label,events,view,ports});
+        let network       = node_network;
+        let logger        = Logger::new("node");
+        let view          = component::ShapeView::new(&logger);
+        let events        = Events {network,select,deselect,set_visualization};
+        let ports         = Registry::default() ;
+        let visualization = visualization::Container::default();
```
wdanilo, line 225: why not just `default()`? :)

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+/// Internal data of a `Container`.
+#[derive(Debug,Clone)]
+#[allow(missing_docs)]
+pub struct ContainerData {
+    pub logger : Logger,
+    pub events : Events,
+
+    node          : display::object::Instance,
+    size          : Cell<Vector2<f32>>,
+    position      : Cell<Vector3<f32>>,
```
wdanilo, line 166: You dont need to store position here, the display object stores it for you and gives you methods like `position` and `mod_position` as soon as you implement `display::Object for ContainerData`.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+                if let Some(vis) = weak_vis.upgrade() {
+                    vis.set_data(data.clone());
+                }
+            });
+        }
+        self
+    }
+
+    /// Toggle visibility on or off.
+    pub fn set_visibility(&self, visible: bool) {
```
wdanilo, line 252: very minor thing - the arg `visible` should be named `visibility`. "visible" suggests it is visible, not a change in state.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+        self.data.visible.set(visible)  ;
+        if let Some(vis) = self.data.visualization.borrow().deref() {
```
wdanilo, line 254: This can be written as `*self.data.visualization.borrow()` - and we use this style more often

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+    visible       : Cell<bool>,
+    visualization : RefCell<Option<Visualization>>,
+}
+
+impl Container {
+    /// Constructor.
+    pub fn new() -> Self {
+        let logger   = Logger::new("visualization");
+        let events   = Events::default();
+        let content  = RefCell::new(None);
```
wdanilo, line 176: use default :)

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+    fn init_frp(self) -> Self {
+        let network = &self.data.events.network;
+
+        frp::extend! { network
+            let weak_vis = self.downgrade();
+            def _f_show = self.data.events.show.map(move |_| {
+               if let Some(vis) = weak_vis.upgrade() {
+                    vis.set_visibility(true)
+               }
+            });
```
wdanilo, line 216: Here is a little bit ugly pattern - you need to do all this fancy stuff with weak and strong refs only because the lambdas in FRP are referencing the same object that owns the `frp::Network` (which owns the lambdas). Maybe you are able to somehow split it? So there would be struct owning `frp::Network` and another rc struct, and the lambdas could access only the another struct? In such case you would not need anymore to do all this upgrade / downgrade ping-pong. Maybe its impossible (I don't know yet, but in case it is possible, please do it).

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Toggle visibility on or off.
+    pub fn set_visibility(&self, visible: bool) {
+        self.data.visible.set(visible)  ;
+        if let Some(vis) = self.data.visualization.borrow().deref() {
+            // TODO use display object functionality
+            if visible {
+                vis.show();
+            } else {
+                vis.hide();
+            }
```
wdanilo, line 260: minor api request - `vis` should have `set_visibility` method which will accept bool. It is a popular need to do something like that and such method would make the API nicer.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+
+    /// Toggle visibility.
+    pub fn toggle_visibility(&self) {
+        self.set_visibility(!self.data.visible.get())
```
wdanilo, line 266: this should be `self.set_visibility(!self.data.is_visible())`

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+}
+
+impl WeakRef for WeakContainer {
+    type StrongRef = Container;
+    fn upgrade(&self) -> Option<Container> {
+        self.data.upgrade().map(|data| Container {data})
+    }
+}
+
+impl Object for Container {
```
wdanilo, line 297: this should be used qualified way, as `display::Object`

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+//! `Container` also provides the FRP API that allows internal interaction with the
+//! `Visualisation`. Data for a visualisation has to be provided wrapped in the `Data` struct.
+//!
+use crate::prelude::*;
+
+use crate::frp;
+
+use ensogl::display::DomSymbol;
+use ensogl::display::object::class::Object;
+use ensogl::display::object::class::ObjectOps;
```
wdanilo, line 15: You should not need this. It is imported by prelude.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+//! elements that facilitate generic interactions, for example, visualisation selection. The
+//! `Container` also provides the FRP API that allows internal interaction with the
+//! `Visualisation`. Data for a visualisation has to be provided wrapped in the `Data` struct.
+//!
+use crate::prelude::*;
+
+use crate::frp;
+
+use ensogl::display::DomSymbol;
+use ensogl::display::object::class::Object;
```
wdanilo, line 14: import `ensogl::display` instead. Then you can access it as `display::Object`.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+// === Wrapper for Visualisation Input Data ===
+// ============================================
+
+/// Wrapper for data that can be consumed by a visualisation.
+/// TODO replace with better typed data wrapper.
+#[derive(Clone,Debug)]
+#[allow(missing_docs)]
+pub enum Data {
+    JSON { content : String },
+    Empty,
```
wdanilo, line 31: I know there is TODO for it, but just a note. If in any case you need a data which can be empty, always use Option :)

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+// ============================================
+// === Wrapper for Visualisation Input Data ===
+// ============================================
+
+/// Wrapper for data that can be consumed by a visualisation.
+/// TODO replace with better typed data wrapper.
+#[derive(Clone,Debug)]
+#[allow(missing_docs)]
+pub enum Data {
+    JSON { content : String },
```
wdanilo, line 30: In case you would be looking for some JSON struct, use the one from serde.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+<svg>
+  <circle style="fill: #69b3a2" stroke="black" cx=50 cy=50 r={}></circle>
+</svg>
+"#, data.as_json()));
+     }
+
+    /// Make the visualisation visible in the scene.
+    pub fn show(&self) {
+        // TODO investigate how to do this through display_object functionality.
+        self.content.dom().set_style_or_panic("visibility", "visible");
```
wdanilo, line 84: FYI - all html elements (and webgl elements too) should appear on the scene as soon as you do something like `scene.add_child(&element)`. They should be hidden as soon as you do `element.unset_parent()`. See `display::ObjectOps` for more info. In case it would not work (it should, just in case as we were recently changing a lot of functionality around it), please ping me! :)

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+// === Visualization Events ===
+// ============================
+
+/// Visualization events.
+#[derive(Clone,CloneRef,Debug)]
+#[allow(missing_docs)]
+pub struct Events {
+    pub network           : frp::Network,
+    pub show              : frp::Source,
+    pub hide              : frp::Source,
```
wdanilo, line 118: A single frp node `frp::Source<bool>` would be much easier to use in frp network and it is more idiomatic to use.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub toggle_visibility : frp::Source,
+    pub update_visualization    : frp::Source<Option<Visualization>>,
```
wdanilo, line 120: Im not sure if "update" is the right word here. It took me a while to understand that this one replaces the visualization. You know "update visualization" might mean to change the data it displays. I would name it `replace_visualization` instead, but take it as a personal preference and if you have a strong pref to leave it as it, go ahead and ignore this comment :)

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+}
+
+impl From<DomSymbol> for Visualization {
+    fn from(symbol: DomSymbol) -> Self {
+        Visualization { content : Rc::new(symbol) }
+    }
+}
+
+impl From<Rc<DomSymbol>> for Visualization {
+    fn from(symbol: Rc<DomSymbol>) -> Self {
```
wdanilo, line 101: Please remember that there are no spaces after `:`

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Render the data as JSON.
+    pub fn as_json(&self) -> String {
+        match &self {
+            Data::JSON { content } => content.clone(),
+            Data::Empty => { "{}".to_string() },
+        }
+    }
+}
+
+impl Default for Data{
```
wdanilo, line 44: Please remember there are spaces before `{`

src/rust/lib/graph-editor/src/component/visualization.rs
MichaelMauderer: The prelude doesn't seem to be enough. I seem to need `use ensogl::display::traits::*;`, but the takes care of `use ensogl::display::object::class::Object;` and `use ensogl::display::object::class::ObjectOps;` .

src/rust/lib/graph-editor/src/component/visualization.rs
MichaelMauderer: Actually it should have been `as_ref()` here anyway...
It seems the `deref()` just got correctly coerced into an `Option<&Visualisation>`, too.

wdanilo: I believe that my line would do the same.

src/rust/lib/graph-editor/src/component/visualization.rs
wdanilo: Please add then `pub use ensogl::display::traits::*;` to prelude IN THIS CRATE only.

src/rust/lib/graph-editor/src/component/visualization.rs
wdanilo: FYI, there is `toggle` frp node. It changes from true to false (and vice versa) whenever event comes in. Its super useful

src/rust/lib/graph-editor/src/component/visualization.rs
MichaelMauderer: I've refactored this, so there are two structs: one for the data/logic (`ContainerData`) and one for the FRP (`ContainerFrp`), which are both held through `Rc`s by the `Container`/`WeakContainer`. If I'm seeing this right, that means there should be no cycle between the two, as the reference structure looks like this:
```text
                     `Container`/`WeakContainer`
                       /                      \
                      /                        \
                     V                          V
         `ContainerFrp`        ------>         `ContainerData`

```

src/rust/lib/graph-editor/src/component/visualization.rs
MichaelMauderer: Done.

src/rust/lib/debug-scenes/src/shapes.rs
```diff
@@ -26,6 +28,20 @@ pub fn run_example_shapes() {
     mem::forget(app);
 }
 
+fn set_up_visualization_demo_frp(graph_editor:&GraphEditor) {
+    let network       = &graph_editor.frp.network;
```
wdanilo, line 32: I know it doesnt matter too much here, but this is almost always a bad idea to extend a network not owned by you. Create a new network here instead and own it in a variable instead please .

src/rust/lib/debug-scenes/src/shapes.rs
```diff
@@ -26,6 +28,20 @@ pub fn run_example_shapes() {
...
+    let dummy_counter = Rc::new(Cell::new(1.0_f32));
+    frp::extend! { network
+        def _set_dumy_data = graph_editor.frp.inputs.set_visualization_data.map(move |node| {
+            let dc = dummy_counter.get();
+            dummy_counter.set(dc + 0.1);
+            let content = json!(format!("{}", 20.0 + 10.0 * dummy_counter.get().sin()));
+            let dummy_data = Some(visualization::Data::JSON { content });
+            node.visualization.frp.set_data.emit(dummy_data);
+        });
+    };
```
wdanilo, line 42: The shape of the FRP is a little complex, right now we have:

```rust
keyboard shortcut  ->
graph_editor.inputs.set_data_for_selected_node ->
graph_editor.inputs.set_visualization_data -> 
[external] _set_dumy_data (with internal counter) ->
node.visualization.frp.set_data
```

Few things are bad here:
1. `set_data_for_selected_node` should not become a public API. If you want to keep it for tests rename it to `DEBUG_set_data_for_selected_node` or something with fixme to remove it. Basically, we will never have use cases to support it and it should not make its way to public API - and lets consider everything on master to be a public API. OF course it changes, but lets minimize the amount of work we got there.

2. I don't like going out of the network to some external definition and getting back there.

3. Naming `set_visualization_data` and `set_data_for_selected_node` - are inconsistent. I would name them `set_output_data` and `DEBUG_set_output_data_for_selected_node` instead. Im not using the word visualization here on purpose - we do not need to suggest what this data will be used for (maybe it will be used for something else in the future?)

What do you think about changing it to the following:

```rust
keyboard shortcut ->
graph_editor.inputs.DEBUG_set_data_for_selected_node (with internal counter) (this should just act as an emitter) ->
graph_editor.inputs.set_visualization_data -> 
node.visualization.frp.set_data
```

In such scenario we have FRP which is much closer to final version than what we've got not.

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -391,8 +415,12 @@ impl application::command::Provider for GraphEditor {
...
     fn default_shortcuts() -> Vec<application::shortcut::Shortcut> {
         use keyboard::Key;
-        vec! [ Self::self_shortcut(&[Key::Character("n".into())] , "add_node_at_cursor")
-             , Self::self_shortcut(&[Key::Backspace]             , "remove_selected_nodes")
+        vec! [
+        Self::self_shortcut(&[Key::Character("n".into())]  , "add_node_at_cursor")
+      , Self::self_shortcut(&[Key::Backspace]              , "remove_selected_nodes")
+      , Self::self_shortcut(&[Key::Character(" ".into())]  , "toggle_visualization_visibility")
+        // TODO move to debug scene
+      , Self::self_shortcut(&[Key::Character("d".into())]  , "debug_set_data_for_selected_node")
```
wdanilo, line 423: USe indentation here please. Its hard to read it now.

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -488,15 +516,44 @@ impl application::View for GraphEditor {
...
+
             display_object.add_child(node);
+
+            let dummy_content = visualization::default_content();
+
+            let dom_layer = scene.dom.layers.front.clone_ref();
+            dom_layer.manage(&dummy_content);
+            let vis : Visualization = dummy_content.into();
+            node.events.set_visualization.emit(Some(vis));
+            // node.visualization.as_ref().unwrap().register_content(node, &dom_layer);
```
wdanilo, line 535: remove commented code

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,334 @@
...
+    JSON { content : serde_json::Value },
+    Binary,
+}
+
+impl Data {
+    /// Render the data as JSON.
+    pub fn as_json(&self) -> String {
+        match &self {
+            Data::JSON { content } => content.to_string(),
+            _ => { "{}".to_string() },
```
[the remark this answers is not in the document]
MichaelMauderer: `node run lint` doesn't highlight this. Are there some settings missing?

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,334 @@
...
+        &self.content.display_object()
+    }
+}
+
+impl Visualization {
+    /// Update the visualisation with the given data.
+    // TODO remove dummy functionality and use an actual visualisation
+    pub fn set_data(&self, data:Data){
+                self.content.dom().set_inner_html(
+                    &format!(r#"
```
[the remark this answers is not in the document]
MichaelMauderer: If it was not a dummy it probably should have been in its own file.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,334 @@
...
+    /// Toggle visibility.
+    pub fn toggle_visibility(&self) {
+        self.set_visibility(!self.is_visible())
+    }
+
+    /// Update the data in the inner visualisation.
+    pub fn set_data(&self, data:Data) {
+        if let Some(vis) = self.visualization.borrow().as_ref() {
+            vis.set_data(data)
+        }
```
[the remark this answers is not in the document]
MichaelMauderer: Clippy doesn't seem to like that style. I had a `map` here before and got shouted at to replace it with `if let`. But Clippy doesn't know about  `for_each_ref` it seems.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,334 @@
...
+impl Container {
+    /// Constructor.
+    pub fn new() -> Self {
+        let logger      = Logger::new("visualization");
+        let content     = default();
+        let size        = Cell::new(Vector2::new(100.0, 100.0));
+        let is_visible  = Cell::new(true);
+        let node        = display::object::Instance::new(&logger);
+
+        let data     = ContainerData {logger,visualization: content,size,is_visible,node};
```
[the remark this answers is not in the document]
MichaelMauderer: Must have happened during refactoring the variable name and I missed the initializer. Changed.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,334 @@
...
+
+// =============================================
+// === Internal Visualisation Representation ===
+// =============================================
+
+/// Inner representation of a visualisation.
+#[derive(Clone,CloneRef,Debug)]
+#[allow(missing_docs)]
+pub struct Visualization {
+    content : Rc<DomSymbol>
```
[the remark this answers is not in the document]
MichaelMauderer: Yeah, changing the `guard : Rc<Guard>` and ` size : Rc<Cell<Vector2<f32>>>` allows it to be `CloneRef` and can be held directly here.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,334 @@
...
+        let network = visualization_events;
+        Self {network,set_visibility,set_visualization,toggle_visibility,set_data }
+    }
+}
+
+
+
+// ================================
+// === Visualizations Container ===
+// ================================
```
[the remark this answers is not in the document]
MichaelMauderer: If you don't mind I'll make this change in the next PR. That would avoid some pain with rebasing the new changes. I wanted to do the split by then anyway.

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/// Internal data of a `Container`.
+#[derive(Debug,Clone)]
+#[allow(missing_docs)]
+pub struct ContainerData {
+    logger        : Logger,
+    display_object: display::object::Instance,
+    size          : Cell<Vector2<f32>>,
+    visualization : RefCell<Option<Visualization>>,
```
wdanilo, line 154: alignment

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,313 @@
...
+        self.visualization.replace(Some(visualization));
+        self.update_visualisation_properties();
+    }
+}
+
+
+impl Container {
+    /// Constructor.
+    pub fn new() -> Self {
+        let logger       = Logger::new("visualization_container");
```
wdanilo, line 210: alignment

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -0,0 +1,313 @@
...
+        let visualization  = default();
+        let size           = Cell::new(Vector2::new(100.0, 100.0));
+        let display_object = display::object::Instance::new(&logger);
+
+        let data     = ContainerData {logger,visualization,size,display_object};
+        let data     = Rc::new(data);
```
wdanilo, line 216: alignment

PR #391, opened 2020-04-30
File Manager Client

notdanilo:
### Pull Request Description
This PR updates the file manager client with newest API from the Enso protocol.

### Important notes

* The API was updated to the current protocol.
* The statement above makes the mock server deprecated, which is going to be completely removed anyway during the integration task.

### Checklist


src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -28,16 +28,23 @@ pub type Event = json_rpc::handler::Event<Notification>;
...
-    /// Wraps a `String`-like entity into a new `Path`.
-    pub fn new(s:impl Str) -> Path {
-        Path(s.into())
+impl Display for Path {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let result = write!(f, "{}", self.root_id);
+        for segment in &self.segments {
+            write!(f, "/{}", segment)?
+        }
+        result
```
wdanilo, line 47: why not `segments.join('/')`?

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -67,24 +74,37 @@ pub enum Notification {
 #[derive(Clone,Debug,PartialEq)]
 #[derive(Serialize, Deserialize)]
 pub struct FilesystemEvent {
-    /// Path of the file that the event is about.
-    pub path : Path,
     #[allow(missing_docs)]
+    // The serialization format requires the information to be wrapped around an "event" property.
```
wdanilo, line 78: "wrapped around" is not valid english. Use "wrapped into" instead.

wdanilo, line 78: Plus this documentation is semantically wrong. It doesn't tell anything about this field - it tells something about serialization format, not this field per se.

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -67,24 +74,37 @@ pub enum Notification {
...
+    pub event : FilesystemEventInfo
+}
+
+
+
+// ===========================
+// === FilesystemEventInfo ===
+// ===========================
+
+/// FilesystemEvent's kind and path of the file that the event is about.
```
wdanilo, line 88: If you are refering to entities from docs, use backticks to show it please ! :)

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -67,24 +74,37 @@ pub enum Notification {
...
+#[derive(Clone,Debug,PartialEq)]
+#[derive(Serialize, Deserialize)]
```
wdanilo, line 90: no spaces here! :)

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -67,24 +74,37 @@ pub enum Notification {
...
 #[derive(Serialize, Deserialize)]
 pub enum FilesystemEventKind {
-    /// A new file under path was created.
-    Created,
-    /// Existing file under path was deleted.
-    Deleted,
+    /// A new file under path was added.
+    Added,
+    /// Existing file under path was removed.
+    Removed,
```
wdanilo, line 105: I would use `allow(missing_docs)` on the whole enum, as this doc is really not necessary (it doesn't tell more than the names of the fields). Of course, leave it as is if you feel differently :)

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -94,96 +114,169 @@ pub enum FilesystemEventKind {
...
+#[serde(tag = "type")]
+pub enum FileSystemObject {
+    /// Represents a directory.
+    Directory(Object),
+    /// Represents a directory which contents have been truncated
+    DirectoryTruncated(Object),
+    /// Represents a file.
+    File(Object),
+    /// Represents unrecognized object. Example is a broken symbolic link.
+    Other(Object),
```
wdanilo, line 145: Then it should be named `Unrecognized` instead.

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -94,96 +114,169 @@ pub enum FilesystemEventKind {
...
-    /// File being a symbolic link.
-    SymbolicLink,
-    /// File being a regular file with opaque content.
-    RegularFile,
-    /// File being none of the above, e.g. a physical device or a pipe.
-    Other
+/// A representation of what kind of type a filesystem object can be.
+#[derive(Hash,Debug,Clone,PartialEq,Eq,Serialize,Deserialize)]
+#[serde(tag = "type")]
+pub enum FileSystemObject {
```
wdanilo, line 137: All these things should be in a module named `file::system::`. Plus this should be renam,ed from `Object` to `Node` as "file system node" is a popular term for that

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -94,96 +114,169 @@ pub enum FilesystemEventKind {
...
+    File(Object),
+    /// Represents unrecognized object. Example is a broken symbolic link.
+    Other(Object),
+    /// Represents a symbolic link that creates a loop.
+    SymlinkLoop(SymlinkLoop)
+}
+
+/// Represents an object.
+#[derive(Hash,Debug,Clone,PartialEq,Eq,Serialize,Deserialize)]
+pub struct Object {
```
wdanilo, line 152: What is `file_manager::Object`? The documentation does not tell what it is. I believe it should be moved to other module

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -94,96 +114,169 @@ pub enum FilesystemEventKind {
...
+    DirectoryTruncated(Object),
+    /// Represents a file.
+    File(Object),
+    /// Represents unrecognized object. Example is a broken symbolic link.
+    Other(Object),
+    /// Represents a symbolic link that creates a loop.
+    SymlinkLoop(SymlinkLoop)
+}
+
+/// Represents an object.
```
wdanilo, line 150: This documentation doesnt tell anything. I know that struct named `Object` represents an object :P

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -94,96 +114,169 @@ pub enum FilesystemEventKind {
...
+#[derive(Hash,Debug,Clone,PartialEq,Eq,Serialize,Deserialize)]
+pub struct Object {
+    /// Name of the object.
+    pub name : String,
+    /// Path to the object.
```
wdanilo, line 155: Lets agree that we write docs if they mean something. I mean, writing doc like "name of the object" on field "name" in "Object" doesn't make sense. The code is harder to read etc. Just remove the docs and add `#[allow(missing_docs)]` to the struct definition. It will make the code shorter and easier to read. Please apply everywhere where applicable.

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -94,96 +114,169 @@ pub enum FilesystemEventKind {
...
+    /// A target of the symlink. Since it is a loop, target is a subpath of the symlink.
+    pub target : Path
+}
+
+/// Response of `init_protocol_connection` method.
+#[derive(Hash,Debug,Clone,PartialEq,Eq,Serialize,Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct InitProtocolConnectionResponse {
+    /// List of Root IDs.
+    pub content_roots : Vec<Uuid>
```
wdanilo, line 175: A lot of docs are enigmatic. What are "Root IDs" ? It is not explained anywhere. IF it is explained somewhere, please link it in the docs, so the reader of the code could follow it.

src/rust/ide/enso-protocol/src/file_manager.rs
```diff
@@ -94,96 +114,169 @@ pub enum FilesystemEventKind {
...
+}
+
+/// Response of `file_read` method.
+#[derive(Hash,Debug,Clone,PartialEq,Eq,Serialize,Deserialize)]
+pub struct ReadResponse {
```
wdanilo, line 180: I believe there sohuld be a seciotn like `=== Responses ===` or something to group these things together.

wdanilo: Or maybe a module if its possible.

src/rust/ide/tests/file_manager.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        client.move_file(new_path,move_path.clone()).await.expect("Couldn't move file");
+        let read = client.read_file(move_path).await.expect("Couldn't read contents");
+        assert_eq!(contents,read.contents);
+    }
+
+    //#[wasm_bindgen_test::wasm_bindgen_test(async)]
+    #[allow(dead_code)]
+    async fn notifications() {
+        let ws         = WebSocket::new_opened("ws://localhost:30616").await;
```
wdanilo, line 75: We are hardcoding this socket eveyrwhere. I understand that these tests really connect via this socket to file manager? IF so, I believe this socket should be passed via env var instead.

farmaazon: Can you read env variables in browser?

If not, the socket should be moved to some constants anyway.

src/rust/ide/enso-protocol/src/file_manager.rs
wdanilo: Oh, ok, good point!

src/rust/ide/enso-protocol/src/file_manager.rs
wdanilo: @farmaazon Nope, I'm asking those who are doing the code here. If this is connected with engine, the person doing the PAI should talk with them, not the reviewer.

src/rust/ide/enso-protocol/src/file_manager.rs
wdanilo: @farmaazon :

1. If the name is dictated by a protocol, then all such things should be in a common module, like `engine::protocol::api`
2. Its not the responsibility of the reviewer to ask engine for a change ! This is the responsibility of people connecting to the API. So if we will come to a conclusion here to change it, it is the responsibility of Danilo or you as the team leader. Definitely not someone who tells that the code is ugly now.


Regarding: 
```
The module file::system would suggest, that we're implementing filesystem operations, but we just implement the client for an external service. 
```

I completely agree. Now it seems like we are implementing filesystem operations. If this is a protocol for engine, move it to a place where it looks like engine protocol, like `engine::protocol::api`

src/rust/ide/enso-protocol/src/file_manager.rs
wdanilo: This is good idea, yes.

src/rust/ide/tests/file_manager.rs
wdanilo: Good point. This should be in some constant or something then.

src/rust/ide/enso-protocol/src/file_manager.rs
farmaazon: > If this is connected with engine, the person doing the PAI should talk with them.
But nobody "is doing API" here, at least not the engine API. We're creating a client to connect with this API. Fixing bad names in engine API is not a part of this task. 

Here we want to keep name consistent with the API, and that is the first priority.

src/rust/ide/enso-protocol/src/file_manager.rs
farmaazon: @wdanilo but this is engo-protocol crate! Is it not sufficient to explain that those are clients using this API. But anyway, the information about "keeping names as in protocol" should be put in the module docs.

2. Yes it is. If you have concerns about names used in engine protocil, you're making yourself a reviewer of engine protocol. And the reviewer of engine protocol should request for changes in engine protocol, not a random developer who is using this API.

src/rust/ide/enso-protocol/src/file_manager.rs
wdanilo: I'm only telling this code has bad naming. Imo Danilo should chat with engine about changing the naming or create ticket for them about that in this case + write FIXME in code pointing to the ticket. Anyway, this is not task of the reviewer to involve in all codebase and discussions. Agreeing upon API and naming is between you and engine.

farmaazon: So, @notdanilo create a issue for discussing naming of protocol with engine.

notdanilo: @farmaazon is this issue created by @mwu-tow enough? https://github.com/luna/enso/issues/708

src/rust/ide/tests/file_manager.rs
mwu-tow: Split into a constant.

src/rust/ide/enso-protocol/src/file_manager.rs
notdanilo: `File System` is a concept. A module named `file_system` would be fine for me. But I agree that things should follow as much as possible the protocol documentation.

src/rust/ide/enso-protocol/src/language_server.rs
```diff
@@ -56,55 +59,45 @@ impl Display for Path {
...
-pub struct FilesystemEvent {
-    #[allow(missing_docs)]
-    // The serialization format requires the information to be wrapped around an "event" property.
-    pub event : FilesystemEventInfo
+    FileEvent {
+        /// The `file/event` notification input wrapper.
+        /// The serialization format requires the information to be wrapped into a field named "event".
+        /// This behavior is currently not specified by the specification and the issue has been raised
+        /// to address this: https://github.com/luna/enso/issues/707
+        /// TODO [mwu] Update as the issue is resolved on way or another.
```
farmaazon, line 72:
```suggestion
        // TODO [mwu] Update as the issue is resolved on way or another.
```
I think todos should be not a part of docs.

src/rust/ide/enso-protocol/src/language_server.rs
```diff
@@ -115,99 +108,112 @@ pub enum FilesystemEventKind {
...
 #[serde(tag = "type")]
 pub enum FileSystemObject {
     /// Represents a directory.
-    Directory(Object),
+    Directory {
+        /// A name of the directory.
+        name:String,
+        /// A path to the directory.
+        path:Path,
+    },
```
farmaazon, line 137: I'd rather put `#[allow(missing_docs)]` before enum, and remove the docs for all fields here (perhaps except the `target`).

mwu-tow: done

PR #403, opened 2020-05-04
Reorder nodes when connecting

mwu-tow:
### Pull Request Description
We want the lines of nodes follow the order of DAGs in the visual representation.
This PR makes it so that destination node's line will be placed after source's line node after connecting.

### Important Notes
We don't attempt to deal with cycles here. I don't think we can do anything about them here (other than just not failing).

### Checklist


src/rust/ide/src/controller/graph.rs
```diff
@@ -562,7 +581,13 @@ impl Handle {
...
         let source_identifier        = source_info.target_ast()?.clone();
         let updated_target_node_expr = destination_info.set(source_identifier)?;
-        self.set_expression_ast(connection.destination.node, updated_target_node_expr)
+        self.set_expression_ast(connection.destination.node,updated_target_node_expr)?;
+
+        // Reorder node lines, so the connection target is after connection source.
+        // Actually this is needed only in some order-dependant context. Once we have better
+        // information about the graph's underlying monadic context, we might want to constrain
+        // this operation.
+        self.place_node_line_after(connection.source.node,connection.destination.node)
```
farmaazon, line 590: So we sent notification about text change twice as I see. Can we do something with it?

If it is troublesome, add Fixme here for that.

PR #410, opened 2020-05-05
 Generate rust code from FlatBuffers specification.

Josef-Vonasek:
### Pull Request Description

There is a namespace bug in `flatc --rust`.

### Checklist


src/rust/ide/engine-api/build.rs
```diff
@@ -0,0 +1,138 @@
...
+
+/// Commit from `enso` repository that will be used to obtain artefacts from.
+const COMMIT: &str = "25494bbb4315d7d6d625424280dfdee0e49dd045";
+
+/// Obtains a URL where the artefacts can be downloaded.
+pub fn flatbuffers_url() -> reqwest::Url {
+    let url = format!("https://packages.luna-lang.org/fbs-schema/nightly/25494bbb4315d7d6d625424280dfdee0e49dd045/fbs-schema.zip");
+    let err = format!("{} is an invalid URL.",url);
+    reqwest::Url::parse(&url).expect(&err)
+}
```
farmaazon, line 31: As for docs: I don't know what the "artefacts" are. And `flatbuffers_url` suggests that this is url for downloading the whole library. I would rename it to `interface_description_url`

src/rust/ide/engine-api/build.rs
```diff
@@ -0,0 +1,138 @@
...
+// =========================
+// == Hardcoded constants ==
+// =========================
+
+
+/// The path where downloaded artefacts will be stored.
+const ZIP_PATH   : &str = "./pkg/fbs-schema.zip";
+/// The path where downloaded artefacts will be unziped.
+const UNZIP_PATH : &str = "./pkg/fbs-upload/fbs-schema/";
+
```
farmaazon, line 22: Usually generated files should be in $OUT directory according to https://doc.rust-lang.org/cargo/reference/build-scripts.html ; We don't do this with parser, because of flaw in wasm_bindgen library (it does not allow to put env variables to paths of js snippets).

I'm sure that I would download zip to the $OUT dir, and there wie should extract fbs files. However I'm not sure if I want to generate files there as well - I've seen that IDE does not work well with files generated in $OUT. But you can give it a try.

mwu-tow:
Encountered bugs in the generated Rust code.
Unfortunately, it seems that Rust is not a properly supported target language, see:
https://github.com/google/flatbuffers/issues/5275
https://github.com/google/flatbuffers/issues/5055

PR #419, opened 2020-05-11
Wip/ao/execution context

farmaazon:
### Pull Request Description
This PR introduces Executed Graph Controller, which is a Graph Controller with additional info about execution context. The creating and modifying execution context is handled by so-called "synchronized model" - that is a model which synchronises its state with Language Server after each change, creation or drop.

### Important Notes
At this point, the execution context is created without checking if graph actually exists.

This PR also introduces better diagnostics for mock errors, and utility to check if all set-up calls on mocks were actually called.

### Checklist


src/rust/ide/src/controller/graph/executed.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// This takes ownership of execution context which will be shared between all copies of this
+    /// handle; when all copies will be dropped, the execution context will be dropped as well
+    /// (and will then removed from LanguageServer).
+    pub fn new(graph:controller::Graph, execution_ctx:ExecutionContext) -> Self {
+        let execution_ctx = Rc::new(execution_ctx);
+        Handle{graph,execution_ctx}
+    }
+
+    //TODO[ao] Here goes the methods requiring ContextId
+}
```
notdanilo, line 30: Missing newline before eof.

src/rust/ide/src/model/synchronized.rs
```diff
@@ -0,0 +1,6 @@
+//! This module contains synchronising wrappers for models whose state are a reflection of
+//! Language Server state, e.g. modules, execution contexts etc. These wrappers synchronize both
+//! states by notifying Language Server of every change and listening on LanguageServer.
+pub mod execution_context;
+
+pub use execution_context::ExecutionContext;
```
notdanilo, line 6: Missing newline before eof.

src/rust/ide/enso-protocol/src/language_server.rs
```diff
@@ -130,14 +130,14 @@ trait API {
     fn destroy_execution_context(&self, context_id:ContextId) -> ();
 
     /// Move the execution context to a new location deeper down the stack.
-    #[MethodInput=PushExecutionContextInput,rpc_name="executionContext/push",
-    result=push_execution_context_result,set_result=set_push_execution_context_result]
-    fn push_execution_context(&self, context_id:ContextId, stack_item:StackItem) -> ();
+    #[MethodInput=ExecutionContextPushInput,rpc_name="executionContext/push",
+    result=execution_context_push_result,set_result=set_execution_context_push_result]
+    fn execution_context_push(&self, context_id:ContextId, stack_item:StackItem) -> ();
```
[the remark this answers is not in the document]
farmaazon: And that's confirmation that the naming change is good :) because for me it's not "pushing the execution context" but "push call to the execution context's stack"

src/rust/ide/src/model/synchronized/execution_context.rs
```diff
@@ -0,0 +1,230 @@
...
+
+
+// ==============
+// === Errors ===
+// ==============
+
+/// Error returned when module path is invalid, i.e. cannot obtain module name from it.
+#[derive(Clone,Copy,Debug,Fail)]
+#[fail(display="Invalid module path.")]
+pub struct InvalidModulePath {}
```
mwu-tow, line 20: TBH need for such error suggets that we might actually do want to have a newtype over Path for Module paths. (that we can represent that the path is a valid module path on the type level — handling such situation shouldn't be a code specific to execution context)

src/rust/ide/src/model/synchronized/execution_context.rs
```diff
@@ -0,0 +1,230 @@
...
+        trace!(logger,"Execution Context created. Id:{id}");
+        let this  = Self {id,module_path,model,language_server,logger};
+        this.push_root_frame().await?;
+        trace!(this.logger,"Pushed root frame");
+        Ok(this)
+    }
+
+    async fn push_root_frame(&self) -> FallibleResult<()> {
+        let module_name = controller::module::obtain_module_name_from_path(&*self.module_path);
+        let method_pointer = language_server::MethodPointer {
```
mwu-tow, line 60: unaligned `=`

src/rust/ide/src/controller/module.rs
```diff
@@ -25,6 +27,13 @@ use parser::Parser;
 /// Path identifying module's file in the Language Server.
 pub type Path = language_server::Path;
 
+/// Gets the module name from its path.
+pub fn obtain_module_name_from_path(path:&Path) -> Option<&str> {
+    let segment     = path.segments.last()?;
+    let name_length = segment.rfind('.').unwrap_or_else(|| segment.len());
+    Some(&segment[..name_length])
+}
```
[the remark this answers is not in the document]
mwu-tow: Also, aren't module names required to be upper-cased?
if so, this should be also, separately checked here/

src/rust/ide/src/model/synchronized/execution_context.rs
mwu-tow, line 60: I'd consider splitting out a method to create a MethodPointer from path+defname

src/rust/ide/src/model/synchronized/execution_context.rs
```diff
@@ -0,0 +1,230 @@
...
+        let positional_arguments_expressions = default();
+        let call = language_server::ExplicitCall {method_pointer,this_argument_expression,
+            positional_arguments_expressions};
+        let frame = language_server::StackItem::ExplicitCall(call);
+        self.language_server.execution_context_push(self.id,frame).await?;
+        Ok(())
+    }
+
+    /// Push a new stack item to execution context.
+    pub fn push(&self, stack_item:StackItem) -> impl Future<Output=Result<(),RpcError>> {
```
mwu-tow, line 75: Why only this method is not `async` but explicitly returns future?

src/rust/ide/src/model/synchronized/execution_context.rs
```diff
@@ -0,0 +1,230 @@
...
+            defined_on_type : module_name.ok_or(InvalidModulePath{})?.to_string(),
+            name            : self.model.root_definition.name.item.clone(),
+        };
+        let this_argument_expression         = default();
+        let positional_arguments_expressions = default();
+        let call = language_server::ExplicitCall {method_pointer,this_argument_expression,
+            positional_arguments_expressions};
+        let frame = language_server::StackItem::ExplicitCall(call);
+        self.language_server.execution_context_push(self.id,frame).await?;
+        Ok(())
```
mwu-tow, line 71: Wont this compile if we just leave  `self.language_server.execution_context_push(self.id,frame).await` as the last line?

src/rust/ide/src/model/synchronized/execution_context.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    /// Pop the last stack item from this context. It returns error when only root call
+    /// remains.
+    pub async fn pop(&self) -> FallibleResult<()> {
+        self.model.pop()?;
+        self.language_server.execution_context_pop(self.id).await?;
+        Ok(())
+    }
+
+    /// Create Execution Context for `root_def` with mocked connection to Language Server.
```
mwu-tow, line 91: What does `root_def` refer to?

src/rust/ide/json-rpc/src/macros.rs
```diff
@@ -112,15 +112,18 @@ macro_rules! make_rpc_methods {
...
 
         impl API for MockClient {
             $(fn $method(&self $(,$param_name:$param_ty)*)
             -> std::pin::Pin<Box<dyn Future<Output=Result<$result>>>> {
-                let mut result = self.$method_result.borrow_mut();
-                let result     = result.remove(&($($param_name),*)).unwrap();
-                Box::pin(async move { result })
+                let mut results = self.$method_result.borrow_mut();
+                let params      = ($($param_name),*);
+                let result      = results.remove(&params);
```
mwu-tow, line 124: Actually this remark is to the cleint — but should we remove entries from the map? 
With such we have no means to handle two repeated calls with the same arguments.

I'd either not remove (but that collides with this `expect_all_calls`) or use some kind of MultiMap

src/rust/ide/src/model/synchronized/execution_context.rs
```diff
@@ -0,0 +1,230 @@
...
+        ls_client.set_destroy_execution_context_result(context_id,Ok(()));
+        ls_client.expect_all_calls();
+        let connection = language_server::Connection::new_mock_rc(ls_client);
+
+        let mut test = TestWithLocalPoolExecutor::set_up();
+        test.run_task(async move {
+            let context = ExecutionContext::create(connection,path.clone(),root_def).await.unwrap();
+            assert_eq!(context_id             , context.id);
+            assert_eq!(path                   , context.module_path);
+            assert_eq!(Vec::<StackItem>::new(), context.model.stack_items().collect_vec());
```
mwu-tow, line 166: to assert against iterable values you might find `expect_tuple` useful

src/rust/ide/src/controller/module.rs
```diff
@@ -16,7 +17,23 @@ use data::text::*;
...
+// ==============
+// === Errors ===
+// ==============
+
+/// Error returned when module path is invalid, i.e. cannot obtain module name from it.
+#[derive(Clone,Copy,Debug,Fail)]
+#[fail(display="Invalid module path.")]
+pub struct InvalidModulePath {}
+
+/// Error returned when module path is invalid, i.e. cannot obtain module name from it.
```
mwu-tow, line 33: Should be different from the previous error.

src/rust/ide/src/controller.rs
```diff
@@ -25,3 +25,12 @@ pub use graph::executed::Handle as ExecutedGraph;
...
 pub use text::Handle            as Text;
+
+
+
+// ============
+// === Path ===
+// ============
+
+/// Path to a file on disc, used across all controllers
+pub type FilePath = enso_protocol::language_server::Path;
```
mwu-tow, line 36: missing newline

src/rust/ide/src/controller/module.rs
```diff
@@ -25,13 +42,56 @@ use crate::double_representation::definition::DefinitionName;
...
+}
+
+impl Path {
+    /// Create a path from the file path. Returns None if given path is not a valid module file.
+    pub fn from_file_path(file_path:FilePath) -> Option<Self> {
+        let has_proper_ext   = file_path.extension() == Some(constants::LANGUAGE_FILE_EXTENSION);
+        let first_char       = file_path.file_name().and_then(|name| name.chars().next());
+        let capitalized_name = first_char.map_or(false, |ch| ch.is_uppercase());
+        let is_module        = has_proper_ext && capitalized_name;
+        is_module.and_option_from(|| Some(Path{file_path}))
```
mwu-tow, line 57: Here, for every call that returns `Option`, I'd apply `?` operator immediately.
If there is no extension or no characters in name, then we should early return, rather than accumulating false flags.

mwu-tow, line 57: `.and_option_from(|| Some(…))` -> ` .as_some_from(|| …)`
(but I'd expect for this usage to go away altogether if previous comment is applied)

src/rust/ide/src/controller/module.rs
```diff
@@ -25,13 +42,56 @@ use crate::double_representation::definition::DefinitionName;
...
+        // The file stem existence should be checked during construction.
+        self.file_path.file_stem().unwrap()
+    }
 
-/// Gets the module name from its path.
-pub fn obtain_module_name_from_path(path:&Path) -> Option<&str> {
-    let segment     = path.segments.last()?;
-    let name_length = segment.rfind('.').unwrap_or_else(|| segment.len());
-    Some(&segment[..name_length])
+    /// Create a simple module path basing on module name. Used as test utility.
```
mwu-tow, line 73: I don't like "simple module path" term, as I'm not really sure what this exactly is.

Consider more like:
```
Create a module path consisting of a single segment, based on a given module name.
```

src/rust/ide/src/model/execution_context.rs
```diff
@@ -56,20 +58,20 @@ pub type Id  = language_server::ContextId;
...
+    stack           : RefCell<Vec<LocalCall>>,
     //TODO[ao] I think we can put here info about visualisation set as well.
 }
 
 impl ExecutionContext {
     /// Create new execution context
     pub fn new(root_definition:DefinitionName) -> Self {
         let stack = default();
-        Self {root_definition,stack}
+        Self { entry_point: root_definition,stack}
```
mwu-tow: Rename parameter to `entry_point`, will simplify the value creation back.

PR #425, opened 2020-05-13
Assign IDs to all nodes but few.

Josef-Vonasek:
### Pull Request Description

Waiting for enso CI to finish before updating scala-parser.js.

### Checklist


src/rust/ide/ast/impl/src/lib.rs
```diff
@@ -231,12 +231,18 @@ impl Ast {
         self
     }
 
-    /// Wraps given shape with an optional ID into Ast.
+    /// Wraps given shape with ID into Ast (uses random new ID if None).
```
mwu-tow, line 234: I'd like it much more if this was split to two separate functions: `new` that takes only `Shape` and the other one (like new_with_id) that takes Shape and non-optional Id.

src/rust/ide/ast/impl/src/lib.rs
```diff
@@ -231,12 +231,18 @@ impl Ast {
...
+        let shape  = shape.into();
+        let id     = id.unwrap_or_else(Id::new_v4);
         let length = shape.len();
-        Ast::new_with_length(shape,id,length)
+        Ast::from_ast_id_len(shape,Some(id),length)
+    }
+
+    pub fn module(shape:Module<Ast>) -> Ast {
+        let length = shape.len();
+        Ast::from_ast_id_len(shape.into(),None,length)
```
mwu-tow, line 245: This is not consistent with other constructors named after shapes. They take (usually generic) input that can be used to construct shape and then wrap it in Ast.
Here it should be something like IntoIterator<Item=BlockLine<Ast>> or sth. 
That'd also actually simplify the usage of this method.

src/rust/ide/ast/impl/src/lib.rs
```diff
@@ -231,12 +231,18 @@ impl Ast {
...
     pub fn new<S:Into<Shape<Ast>>>(shape:S, id:Option<Id>) -> Ast {
-        let shape: Shape<Ast> = shape.into();
+        let shape  = shape.into();
+        let id     = id.unwrap_or_else(Id::new_v4);
         let length = shape.len();
-        Ast::new_with_length(shape,id,length)
+        Ast::from_ast_id_len(shape,Some(id),length)
+    }
+
+    pub fn module(shape:Module<Ast>) -> Ast {
```
mwu-tow, line 243: Missing doc.

src/rust/ide/ast/impl/src/known.rs
```diff
@@ -98,6 +98,14 @@ impl<T:Into<Shape<Ast>>> KnownAst<T> {
     }
 }
 
+impl KnownAst<crate::Module<Ast>> {
+    /// Creates a new `KnownAst<Module>` from `shape`.
```
mwu-tow, line 102: True but this doesn't really explain the purpose of this function, i.e. that modules require some special treatment.
(if they do)

src/rust/ide/ast/impl/src/lib.rs
```diff
@@ -272,11 +271,6 @@ impl Ast {
     pub fn with_shape<S:Into<Shape<Ast>>>(&self, shape:S) -> Ast {
         Ast::new(shape.into(),self.id)
     }
-
-    /// Returns this AST node with removed ID.
-    pub fn without_id(&self) -> Ast {
-        Ast::from_ast_id_len(self.shape().clone(), None, self.len())
-    }
```
mwu-tow, line 279: I'd keep this as long as the Ast represents Id as an Option.  Particularly if we allow calling `from_ast_id_len` with `None`.
If this allows breaking our rules (assuming we adopt stricter take on them), make this private.

PR #429, opened 2020-05-14
Visualization Registry

MichaelMauderer:
### Pull Request Description
Adds a UI side visualization registry, as well as an FRP end-point, to register new visualizations from JS source files. 

### Checklist


src/rust/lib/graph-editor/src/component/visualization/registry.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// Register a new visualisation factory that's pre-wrapped in an `Rc` with the registry.
+    pub fn register_factory_rc(&self, factory:Rc<dyn Factory>) {
+        self.entries.borrow_mut().push(factory);
+    }
+
+    /// Return all `Factory`s that can create a visualisation for the given datatype.
+    pub fn valid_sources(&self, dtype:&EnsoType) -> Vec<Rc<dyn Factory>>{
+        // TODO: this is not super efficient. Consider building a HashMap from type to vis.
```
notdanilo, line 97: Could it be done in this PR?

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -1207,7 +1210,7 @@ impl application::View for GraphEditor {
         }));
 
          // === Vis Cycling ===
```
notdanilo, line 1212: I am just asking because it isn't clear for me. What's the meaning of visualisation cycling?

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -287,6 +286,7 @@ pub struct FrpInputs {
     pub translate_selected_nodes       : frp::Source<Position>,
     pub cycle_visualization            : frp::Source<NodeId>,
     pub set_visualization              : frp::Source<(NodeId,Option<Visualization>)>,
+    pub register_visualisation_source  : frp::Source<Option<Rc<dyn visualization::Factory>>>,
```
wdanilo, line 289: Inconsistency - source and factory. After reviewing almost all the code I believe this name should be class and we should have here `register_visualisation_class`. Moreover, frp types needs to be simple - please create newtype for every trait object you have!

src/rust/lib/graph-editor/src/lib.rs
wdanilo: We have different visualizations for a given type, this means that you switch to the next available visualization :)

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -1234,21 +1236,22 @@ impl application::View for GraphEditor {
...
+            let visualisations = visualization_registry.valid_sources(&"[[float;3]]".into());
+            cycle_count.set(cycle_count.get() % visualisations.len());
+            let vis  = &visualisations[cycle_count.get()];
+            let vis  = vis.instantiate(&scene);
+            let node = nodes.get_cloned_ref(node_id);
+            match (vis, node) {
+                (Ok(vis), Some(node))  => {
+                        node.view.visualization_container.frp.set_visualization.emit(Some(vis));
+                },
+                // TODO better error handling. Maybe use a logger?
```
wdanilo, line 1250: logger to the rescue! :D

src/rust/lib/graph-editor/src/component/visualization/registry.rs
MichaelMauderer: Done!

src/rust/lib/graph-editor/src/component/visualization/factory.rs
```diff
@@ -0,0 +1,149 @@
...
+
+// =========================
+// === JS Source Factory ===
+// =========================
+
+#[derive(CloneRef,Clone,Debug)]
+#[allow(missing_docs)]
+pub struct JsSourceFactory {
+    info   : Rc<Metadata>,
+    source : Rc<CowString>,
```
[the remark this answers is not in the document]
MichaelMauderer: I was doing this mostly as I was trying to be flexible in allowing the external API to provide us both an owned `String` or just a reference. We don't need really the full `String`, but maybe the caller can't own it either. I think the nicest thing would be to hide the fact that we use the `CowString internally and just provide a nice API to use either a `String` or a `&str` to create the `JsSourceFactory`

src/rust/lib/graph-editor/src/component/visualization/factory.rs
```diff
@@ -0,0 +1,149 @@
...
+//! use ensogl::display::Scene;
+//! use std::rc::Rc;
+//!
+//! // Create a factory from a JS source code snippet.
+//! let js_source_factory = JsSourceFactory::from_js_source_raw(r#"
+//! class BubbleVisualisation {
+//!     onDataReceived(root, data) {}
+//!     setSize(root, size) {}
+//! }
+//! return new BubbleVisualisation();
```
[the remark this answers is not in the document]
MichaelMauderer: I agree `Class` is nicer! I also changed the code to return the class object and use that instead of requiring a closure that returns an instance. This solves a few other issues, too.

src/rust/lib/graph-editor/src/component/visualization/factory.rs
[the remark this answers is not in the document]
MichaelMauderer: This has gone away.

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -20,11 +20,30 @@ pub struct EnsoCode {
...
+    }
+}
+
+/// Contains general information about a visualisation.
+#[derive(Clone,Debug)]
+#[allow(missing_docs)]
+pub struct Metadata {
+    pub name        : String,
+    pub input_types : Vec<EnsoType>,
+}
```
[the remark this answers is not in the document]
MichaelMauderer: I agree. Considerations: `ClassAttributes` (current), `VisualisationSpecification`, `VisualisationMetadata`

src/rust/lib/graph-editor/src/component/visualization/factory.rs
```diff
@@ -0,0 +1,149 @@
...
+
+
+impl Factory for NativeConstructorFactory {
+    fn metadata(&self) -> &Metadata {
+       &self.info
+    }
+
+    fn instantiate(&self, scene:&Scene) -> InstantiationResult {
+       self.constructor.call((scene,))
+    }
```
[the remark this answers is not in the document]
MichaelMauderer: Yes, this was required in some form to get the `Shape`s and `DomSymbol` set up. I'll make sure we have TODO's for those. 

A similar issue also exists with the JS visualizations though: if they want they can change arbitrary parts of the DOM, not just the part we want them to change.

src/rust/lib/graph-editor/src/component/visualization/class.rs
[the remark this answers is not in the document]
notdanilo: What about `VisualizationInformation`/`VisualizationInfo`?

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -1,29 +1,36 @@
...
 //!
 //! 2. The `Visualization` is a struct that wraps the `DataRenderer` and implements the generic
-//!    tasks that are the same for all visualisations. That is, interfacing with the other UI
+//!    tasks that are the same for all visualizations. That is, interfacing with the other UI
 //!    elements, the visualization registry, as well as propagating frp messages.
 //!
-//! 3. The `Container` wraps the `Visualisation` and provides the UI elements that facilitate
-//!    user interactions. For example, selecting a visualisation or connecting it to nodes in the
+//! 3. The `Class` is a struct that instantiates `Visualization` structs and provides data about
+//!    the visualization even before they are instantiated like input datatype and name.
```
wdanilo, line 14: Just a minor change request to make it more obvious "that instantiates" -> "that can instantiate multiple"

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -5,7 +5,10 @@ use crate::prelude::*;
 use crate::frp;
 use crate::visualization::*;
 
+use ensogl::display::Scene;
 use ensogl::display;
+use std::error::Error;
+use enso_prelude::CloneRef;
```
wdanilo, line 11: Use `use crate::prelude::*;` instead

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -20,11 +23,31 @@ pub struct EnsoCode {
...
+    }
+}
+
+/// Contains general information about a visualization.
+#[derive(Clone,Debug)]
+#[allow(missing_docs)]
+pub struct ClassAttributes {
+    pub name        : String,
+    pub input_types : Vec<EnsoType>,
+}
```
wdanilo, line 49: Im still tending to name it `visualization::class::Signature` instead. (btw you are in `visualization::class` so no need to repeat word `Class` in the name)

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
 }
+
+
+
+// ===========================
+// === Visualization Class ===
+// ===========================
+
+/// Result of the attempt to instantiate a `Visualization` from a `Class`.
+pub type InstantiationResult = Result<Visualization,Box<dyn Error>>;
```
wdanilo, line 187: If I'm correct, we are using different error handling method - please confirm it with @farmaazon and @mwu-tow before merging this PR. The way used here has 2 problems:

1. You cannot pattern match on the error after catching it.
2. The type `Box<dyn Error>` will appear in type errors during compilation because it is not a newtype, so users of this library can easily use it in their code etc - this makes the types in the app uglier.

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+/// use graph_editor::component::visualization::Visualization;
+/// use graph_editor::component::visualization::renderer::example::native::BubbleChart;
+/// use ensogl::display::Scene;
+/// use std::rc::Rc;
+///
+/// // Create a `visualization::Class` from a JS source code snippet.
+/// let js_source_class = visualization::JsSourceClass::from_js_source_raw(r#"
+///
+///    class BubbleVisualization {
+///         static inputTypes = ["[[float;3]]"]
```
wdanilo, line 213: These should be Enso types, so if you want to have a list of 3 floats, you write `[Float,Float,Float]` instead of `[float;3]` :) If you want to have a List of unknown length then you write `List Float`.

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+///
+///     return BubbleVisualization;
+///
+/// "#.into());
+///
+/// // Create a `visualization::Class` that instantiates a `BubbleChart`.
+/// let native_bubble_vis_class = visualization::NativeConstructorClass::new(
+///     visualization::ClassAttributes {
+///         name        : "Bubble Visualization (native)".to_string(),
+///         input_types : vec!["[[float;3]]".to_string().into()],
```
wdanilo, line 226: thy we need `.to_string()` here? We should have conversion from &str defined

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+/// );
+/// ```
+pub trait Class: Debug {
+    /// Provides additional information about the `Class`, for example, which `DataType`s can be
+    /// rendered by the instantiated visualization.
+    fn attributes(&self) -> &ClassAttributes;
+    /// Create new visualization, that is initialised for the given scene. This can fail if the
+    /// `visualization::Class` contains invalid data, for example, JS code that fails to execute,
+    /// or if the scene is in an invalid state.
+    // TODO consider not allowing failing here and require the checking on instantiation of the `Class`.
```
wdanilo, line 238: You cannot do it. JS can behave in such way that even if 2 instantiations were OK, the third could throw error.

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+    // TODO consider not providing the scene here, but hooking the the shapes/dom elements into the
+    // scene externally.
+    fn instantiate(&self, scene:&Scene) -> InstantiationResult;
+}
+
+#[derive(Clone,Debug,Default)]
+#[allow(missing_docs)]
+pub struct ClassHandle {
```
wdanilo, line 246: You are in `visualization::class`, so this sohuld be named `Handle`

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+    class : Option<Rc<dyn Class>>
+}
+
+impl ClassHandle {
+    /// Constructor.
+    pub fn new<T: Class + 'static>(class: T) -> ClassHandle {
+        let wrapped = Rc::new(class);
+        ClassHandle{class:Some(wrapped)}
+    }
```
wdanilo, line 255: Spacing everywhere! :)
It should be: 

```rust
    pub fn new<T:Class+'static>(class:T) -> ClassHandle {
        let class = Some(Rc::new(class));
        ClassHandle {class}
    }
```

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+
+    /// Return the inner class.
+    pub fn get_class(&self) -> Option<Rc<dyn Class>> {
+        self.class.clone()
+    }
+}
```
wdanilo, line 261: getters do not have `get_` prefix!

src/rust/lib/graph-editor/src/component/visualization/class.rs
wdanilo, line 246: I know we are using allow missing docs for the inner field, but docs regarding what actually this Handle is for should be there :)

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+// ================================
+// === Native Constructor Class ===
+// ================================
+
+/// Type alias for a function that can create a `Visualization`.
+pub trait VisualizationConstructor = Fn(&Scene) -> InstantiationResult;
+
+#[derive(CloneRef,Clone,Derivative)]
+#[derivative(Debug)]
+#[allow(missing_docs)]
```
wdanilo, line 276: We are using `allow(missing_docs)` only for fields. Pub structs have to always have docs.

src/rust/lib/graph-editor/src/component/visualization/class.rs
```diff
@@ -154,3 +176,126 @@ impl Visualization {
...
+}
+
+impl NativeConstructorClass {
+    /// Create a visualization source from a closure that returns a `Visualization`.
+    pub fn new<T>(info: ClassAttributes, constructor:T) -> Self
+    where T: VisualizationConstructor + 'static {
+        let info = Rc::new(info);
+        let constructor = Rc::new(constructor);
+        NativeConstructorClass { info,constructor }
+    }
```
wdanilo, line 290: spacing

src/rust/lib/graph-editor/src/component/visualization/js.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    fn attributes(&self) -> JsResult<ClassAttributes> {
+        let input_types = self.input_types().unwrap_or_default();
+        let name        = self.name()?;
+        Ok(ClassAttributes{name,input_types})
+    }
+
+    fn constructor(&self) -> js_sys::Function {
+        js_sys::Reflect::get(&self.prototype(),&"constructor".into()).unwrap().into()
```
wdanilo, line 47: JS world is very very strange. You can alter everything there. And we never want that user provided visualization crashes the application, thus you should never do `unwrap` when accessing properties in user-provided code. We can use `unwrap` when we own the behavior, like for example getting some webgl context, but on user-provided code we need to be very, very careful.

src/rust/lib/graph-editor/src/component/visualization/registry.rs
```diff
@@ -0,0 +1,97 @@
...
+        registry
+    }
+
+    /// Register a new visualization class with the registry.
+    pub fn register_class<T: Class + 'static>(&self, class:T) {
+        self.register_class_rc(Rc::new(class));
+    }
+
+    /// Register a new visualization class that's pre-wrapped in an `Rc` with the registry.
+    pub fn register_class_rc(&self, class:Rc<dyn Class>) {
```
wdanilo, line 82: Please, do not expose rc'd dyn traits in public API. Create newtype for them, always. Or make this function private.

src/rust/lib/graph-editor/src/component/visualization/registry.rs
```diff
@@ -0,0 +1,97 @@
...
+            },
+            |scene:&Scene| Ok(Visualization::new(BubbleChart::new(scene)))
+        ));
+        registry.register_class(get_bubble_vis_class());
+
+        registry
+    }
+
+    /// Register a new visualization class with the registry.
+    pub fn register_class<T: Class + 'static>(&self, class:T) {
```
wdanilo, line 77: spacing

src/rust/lib/graph-editor/src/component/visualization/renderer/example/js.rs
```diff
@@ -86,54 +44,8 @@ pub fn object_sample_js_bubble_chart() -> JsRenderer {
...
-                svgElem.setAttributeNS(null, "height" , height);
-            }
-        }
-
-        return new BubbleVisualisation();
-    "#;
-    JsRenderer::from_constructor(fn_constructor).unwrap()
-}
+    JsSourceClass::from_js_source_raw(fn_constructor).unwrap()
+}
```
wdanilo, line 51: nl

src/rust/lib/graph-editor/src/component/visualization/class.rs
MichaelMauderer: That is unfortunately true.

src/rust/lib/graph-editor/src/component/visualization/class.rs
MichaelMauderer: I fixed the spacing, but the compiler needs the `ClassHandle{class:Some(wrapped)}` otherwise it can't determine the types correctly without another anotation.

src/rust/lib/graph-editor/src/component/visualization/class.rs
MichaelMauderer: `Box<dyn Error>>` has been replaced with a proper error enum `InstantiationError`.

src/rust/lib/graph-editor/src/component/visualization/js.rs
MichaelMauderer: I fixed all the error handling here. There should be no `unwrap`s left.

PR #431, opened 2020-05-14
Zero-copy JSON-RPC methods

notdanilo:
### Pull Request Description
This PR introduces zero-copy JSON-RPC methods wherever it's possible.

### Important Notes
* Every parameter is translated as references so we don't need to move ownership.
* It's also true for copy types such as atomic types.

### Checklist

### DIFF
--- a/src/rust/ide/json-rpc/src/macros.rs
+++ b/src/rust/ide/json-rpc/src/macros.rs
@@ -1,7 +1,6 @@
 //! Helper macros to generate RemoteClient and MockClient.
 
-// FIXME[dg]: https://github.com/luna/ide/issues/401 We want to make the generated methods to
-// take references instead of ownership.
+// TODO[dg]: Make it possible to create an API which accepts pass both references and ownership.
 /// This macro reads a `trait API` item and generates asynchronous methods for RPCs. Each method
 /// should be signed with `MethodInput`, `rpc_name`, `result` and `set_result` attributes. e.g.:
 /// ```rust,compile_fail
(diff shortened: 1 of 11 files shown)


src/rust/ide/json-rpc/src/macros.rs
```diff
@@ -82,22 +82,29 @@ macro_rules! make_rpc_methods {
...
         $(
             /// Structure transporting method arguments.
-            #[derive(Serialize,Deserialize,Debug,PartialEq)]
+            #[derive(Serialize,Debug,PartialEq)]
             #[serde(rename_all = "camelCase")]
-            struct $method_input {
-                $($param_name : $param_ty),*
+            struct $method_input<'a> {
+                #[serde(skip)]
+                phantom : std::marker::PhantomData<&'a u8>,
```
farmaazon, line 103: When I saw this code I wondered "why u8?", perhaps unit gives a stronger suggestion that this is for lifetime alone :)
```suggestion
                phantom : std::marker::PhantomData<&'a()>,
```

src/rust/ide/src/lib.rs
```diff
@@ -153,13 +153,13 @@ pub async fn open_project
 /// Open most recent project or create a new project if none exists.
 pub async fn open_most_recent_project_or_create_new
 (project_manager:&impl project_manager::API) -> FallibleResult<controller::Project> {
-    let mut response = project_manager.list_recent_projects(1).await?;
+    let mut response = project_manager.list_recent_projects(&1).await?;
```
farmaazon, line 156: I don't know, just `&1` looks weird, especially in the actual (non-test) code.
```suggestion
    let projects_to_list = 1;
    let mut response     = project_manager.list_recent_projects(&projects_to_list).await?;
```

PR #433, opened 2020-05-18
Visualization Registry integration

notdanilo:
### Pull Request Description
This PR provides the Visualisation Controller, which lists files from 

### Checklist


src/rust/ide/src/controller.rs
```diff
@@ -19,6 +19,7 @@ pub mod graph;
 pub mod module;
 pub mod project;
 pub mod text;
+pub mod visualisation;
```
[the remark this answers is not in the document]
farmaazon: Additionally, I see that my browser makes a red underline below _visualisation_ word, but it's ok with _visuali*z*ation_. And as I read [here](https://ell.stackexchange.com/a/5745), the z is an american spelling, so I think it should be used here (to be consistent with other words, like connection).

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn new
+    ( language_server_rpc  : Rc<language_server::Connection>
+    , embedded_visualisers : HashMap<String,String>) -> Self {
+        Self {language_server_rpc,embedded_visualisers}
+    }
+
+    /// Get a list of all available visualisers.
+    pub async fn list_visualisers(&self) -> FallibleResult<Vec<VisualiserIdentifier>> {
+        let root_id    = self.language_server_rpc.content_root();
+        let vis_path   = language_server::Path{root_id,segments:vec!["visualisation".into()]};
```
farmaazon, line 67: The path to the visualizations in project could be a spearate function, and definitely the "visualisation" string should be a separate constant.

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+                path.segments.push(name.to_string());
+                let root_id                  = path.root_id;
+                let segments                 = path.segments;
+                let path                     = language_server::Path{root_id,segments};
+                let visualisation_identifier = VisualiserIdentifier::File(path);
+                Some(visualisation_identifier)
+            } else {
+                None
+            }
+        }));
```
farmaazon, line 84: A minor thing, regarding iterators: instead of creating vector and then extending it, you can use `chain` method to chain two iterators and do then the final collecting. That in theory is better, because we do only one allocation of vector.

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /// Load the source code of the specified visualiser.
+    pub async fn load_visualiser(&self, visualiser:&VisualiserIdentifier) -> FallibleResult<String> {
+        match visualiser {
+            VisualiserIdentifier::Embedded(identifier) => {
+                let result = self.embedded_visualisers.get(identifier);
+                let error  = || {
+                    failure::Error::from(VisualiserError::NotFound{identifier:visualiser.clone()})
+                };
```
farmaazon, line 95: Check if it will work:
```suggestion
                let error  = || VisualiserError::NotFound{identifier:visualiser.clone()}.into();
```

src/rust/ide/src/controller/project.rs
```diff
@@ -6,6 +6,7 @@
 use crate::prelude::*;
 
 use crate::controller::FilePath;
+use crate::controller::visualisation;
```
farmaazon, line 9: We usually do not do such imports, because we want to see in usages that the handle is a _controller_ handle (and not e.g. module). If you apply my suggestion about alias in src/controller.rs, then you can just write here `controller::Visualisation`

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+
+
+
+// ===============================
+// === VisualisationIdentifier ===
+// ===============================
+
+/// This enum is used to identify a visualiser both in the project folder or natively embedded in
+/// IDE.
+#[derive(Debug,Clone,PartialEq,Eq,Display)]
```
farmaazon, line 36: Sort the derives alphabetically.

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+                let segments                 = path.segments;
+                let path                     = language_server::Path{root_id,segments};
+                let visualisation_identifier = VisualiserIdentifier::File(path);
+                Some(visualisation_identifier)
+            } else {
+                None
+            }
+        }));
+        Ok(result)
+    }
```
farmaazon, line 86: Please split this function to two: one returning embedded visualizations and second returning file visualisations.

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    #[test]
+    fn list_and_load() {
+        let mock_client     = language_server::MockClient::default();
+
+        let root_id = uuid::Uuid::default();
+        let path    = Path{root_id,segments:vec!["visualisation".into()]};
+        let paths   = vec![
+            FileSystemObject::File{path:path.clone(),name:"histogram.js".into()},
+            FileSystemObject::File{path:path.clone(),name:"graph.js".into()},
```
farmaazon, line 132: Add the object which is not a `File` to cover all code branches.

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+impl Handle {
+    /// Creates a new visualisation controller.
+    pub fn new
+    ( language_server_rpc  : Rc<language_server::Connection>
+    , embedded_visualisers : HashMap<String,String>) -> Self {
+        Self {language_server_rpc,embedded_visualisers}
+    }
+
+    /// Get a list of all available visualisers.
+    pub async fn list_visualisers(&self) -> FallibleResult<Vec<VisualiserIdentifier>> {
```
wdanilo, line 65: What is visualizer? :O

src/rust/ide/src/controller/visualisation.rs
wdanilo: Yes, please, be sure it never ever happens in PRs . No hardcoded literals. Ever! :)

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+        let result     = result.map(|identifier| VisualiserIdentifier::Embedded(identifier.clone()));
+        let mut result = result.collect::<Vec<VisualiserIdentifier>>();
+        result.extend(file_list.paths.iter().filter_map(|object| {
+            if let language_server::FileSystemObject::File{name,path} = object {
+                let mut path = path.clone();
+                path.segments.push(name.to_string());
+                let root_id                  = path.root_id;
+                let segments                 = path.segments;
+                let path                     = language_server::Path{root_id,segments};
+                let visualisation_identifier = VisualiserIdentifier::File(path);
```
wdanilo, line 79: very minor thing - I would name it "identifier" as its obvious what it is - and it will make the code nicer

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+    use enso_protocol::language_server::Path;
+
+    fn result<T,F:Future<Output = FallibleResult<T>>>(fut:F) -> FallibleResult<T> {
+        let mut fut = Box::pin(fut);
+        poll_future_output(&mut fut).expect("Promise isn't ready")
+    }
+
+    #[test]
+    fn list_and_load() {
+        let mock_client     = language_server::MockClient::default();
```
wdanilo, line 126: alignment

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        let root_id = uuid::Uuid::default();
+        let path    = Path{root_id,segments:vec!["visualisation".into()]};
+        let paths   = vec![
+            FileSystemObject::File{path:path.clone(),name:"histogram.js".into()},
```
wdanilo, line 131: could you please make a constructor, like `File::new(...)` which accepts anything that is convertible to file path? Writing `.into()` on all call-sides is an ugly API

src/rust/ide/src/controller/visualisation.rs
```diff
@@ -0,0 +1,166 @@
...
+// ===============================
+// === VisualisationIdentifier ===
+// ===============================
+
+/// This enum is used to identify a visualiser both in the project folder or natively embedded in
+/// IDE.
+#[derive(Debug,Clone,PartialEq,Eq,Display)]
+#[allow(missing_docs)]
+pub enum VisualiserIdentifier {
+    Embedded(String),
```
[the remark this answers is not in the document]
farmaazon: Don't we agreed with thar? Embedded vis are as a workaround - they are vis provided by us to be able in each project, because we won't support reading vis from "common place for all projects" until alpha at least.

src/rust/ide/src/controller.rs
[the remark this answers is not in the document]
farmaazon: Now I see, that the enso protocol use "Visualisation" everywhere. Forget about it.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+
+
+// ===============================
+// === VisualizationIdentifier ===
+// ===============================
+
+/// This enum is used to identify a visualiser both in the project folder or natively embedded in
```
farmaazon, line 42: I see that in Michael code we don't use term "visualiser", but rather "visualization" and "visualization instance" for attached vis.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,262 @@
...
+// ==============
+
+const VISUALISATION_FOLDER : &str = "visualization";
+
+/// Visualization Controller's state.
+#[derive(Debug,Clone)]
+pub struct Handle {
+    language_server_rpc     : Rc<language_server::Connection>,
+    embedded_visualizations : EmbeddedVisualizations,
+    graph_editor            : RefCell<Option<GraphEditor>>
```
farmaazon, line 82: Controllers should not use the concrete view component. Instead you should add code integrating both worlds in some place in views, e.g. see `view::node_editor` module.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,262 @@
...
+            visualization.map(|visualization| {
+                if let Some(graph_editor) = graph_editor.as_ref() {
+                    let class_handle = &Some(visualization);
+                    graph_editor.frp.register_visualization_class.emit_event(class_handle);
+                }
+            })?;
+        }
+        *self.graph_editor.borrow_mut() = graph_editor;
+        Ok(())
+    }
```
farmaazon, line 109: This is code which should be moved to `view::node_editor` module (or some submodule of it).

src/rust/ide/enso-protocol/src/language_server/types.rs
```diff
@@ -37,6 +37,14 @@ impl Display for Path {
 }
 
 impl Path {
+    /// Splits path into name and segments. e.g.:
+    /// Path{root_id,segments:["foo","bar","qux"]} => ("qux",Path{root_id,segments:["foo","bar"]})
+    pub fn split_into_name_and_segments(&self) -> (String,Path) {
```
mwu-tow, line 42: If we return by value, we should take `self` by value as well. This way we'll avoid clones.

src/rust/ide/enso-protocol/src/language_server/types.rs
```diff
@@ -37,6 +37,14 @@ impl Display for Path {
...
+        let mut path = self.clone();
+        let name     = path.segments.pop().unwrap_or_default();
```
mwu-tow, line 44: I'd rather not unwrap but return return Option, for consistency with `file_name`.
If there are no segments in the path, splitting it into two pieces makes no sense.

mwu-tow: Also, for consistency, like we have `file_name` we could also introduce `parent` and then implement this function in terms of these two.

src/rust/ide/enso-protocol/src/language_server/types.rs
mwu-tow, line 42: I'd name rather like `split_into_file_name_and_parent` — the point is that new segments are the parent path. After all, we had segments all along.

src/rust/ide/src/controller/project.rs
```diff
@@ -37,13 +39,14 @@ impl Handle {
...
-            language_server_bin : Rc::new(language_server_binary),
-            logger              : parent.sub("Project Controller"),
-        }
+        let module_registry         = default();
+        let parser                  = Parser::new_or_panic();
+        let language_server_rpc     = Rc::new(language_server_client);
+        let language_server_bin     = Rc::new(language_server_binary);
+        let logger                  = parent.sub("Project Controller");
+        let embedded_visualizations = default();
+        let visualization = Visualization::new(language_server_rpc.clone(),embedded_visualizations);
```
mwu-tow, line 48: alignment before `=`

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
+//! Visualization controller.
+//!
+//! Ths Visualization Controller is Responsible identifying all the available visualization natively
+//! embedded in IDE and available within the project's `visualization` folder. The
+//! `Visualization Controller` lives as long as the `Project Controller`.
+
+use crate::prelude::*;
+
+use std::rc::Rc;
+use enso_protocol::language_server;
```
mwu-tow, line 10: not in lexicographical order

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
+//! Visualization controller.
+//!
+//! Ths Visualization Controller is Responsible identifying all the available visualization natively
```
mwu-tow, line 3:
```suggestion
//! Ths Visualization Controller is Responsible identifying all the available visualizations natively
```

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+//! embedded in IDE and available within the project's `visualization` folder. The
+//! `Visualization Controller` lives as long as the `Project Controller`.
```
mwu-tow, line 5: > The `Visualization Controller` lives as long as the `Project Controller`.

I'd say this is more of a Project Controller's property rather than Visualization Controller.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Enumeration of errors used in `Visualization Controller`.
+#[derive(Debug,Fail)]
+#[allow(missing_docs)]
+pub enum VisualizationError {
+    #[fail(display = "Visualization \"{}\" not found.", identifier)]
+    NotFound {
+        identifier : VisualizationIdentifier
+    },
+    #[fail(display = "JavaScript visualization \"{}\" failed to compile.", identifier)]
```
mwu-tow, line 28: Javascript is compiled? To what target?

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+
+
+// ==============================
+// === EmbeddedVisualizations ===
+// ==============================
+
+#[allow(missing_docs)]
+pub type EmbeddedVisualizationName   = String;
```
mwu-tow, line 56:
```suggestion
pub type EmbeddedVisualizationName = String;
```

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Embedded visualizations mapped from name to source code.
+#[derive(Shrinkwrap,Debug,Clone,Default)]
+#[shrinkwrap(mutable)]
+pub struct EmbeddedVisualizations {
+    #[allow(missing_docs)]
+    pub map : HashMap<EmbeddedVisualizationName,Rc<class::Handle>>
```
mwu-tow, line 63:
```suggestion
    pub map:HashMap<EmbeddedVisualizationName,Rc<class::Handle>>
```

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+
+
+// ===============================
+// === VisualizationIdentifier ===
+// ===============================
+
+/// This enum is used to identify a visualization both in the project folder or natively embedded in
+/// IDE.
```
mwu-tow, line 41:
```suggestion
/// This enum is used to identify a visualization either in the project folder or natively embedded in
/// IDE.
```

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+
+
+
+// ==============
+// === Handle ===
+// ==============
+
+const VISUALISATION_FOLDER : &str = "visualization";
```
mwu-tow, line 73:
```suggestion
const VISUALISATION_FOLDER:&str = "visualization";
```

mwu-tow: Also, despite value `"visualization"` the identifier here uses `VISUALISATION` (with `s` not `z`).

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+
+#[derive(Debug)]
+struct VisualizationControllerData {
+    language_server_rpc     : Rc<language_server::Connection>,
+    embedded_visualizations : EmbeddedVisualizations
+}
+
+/// Visualization Controller's state.
```
mwu-tow, line 81: I don't explaining `Handle` as `state`.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Creates a new visualization controller.
+    pub fn new
+    ( language_server_rpc     : Rc<language_server::Connection>
+    , embedded_visualizations : EmbeddedVisualizations) -> Self {
+        let data = VisualizationControllerData{language_server_rpc,embedded_visualizations};
+        let data = Rc::new(RefCell::new(data));
+        Self {data}
+    }
+
+    async fn list_file_visualizations(&self) -> FallibleResult<Vec<VisualizationIdentifier>> {
```
mwu-tow, line 97: I'd consider naming this more like "list_project_specific_visualizations".
I expect that embedded visualizations can (or in future will can) be recognized as files as well.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+        let data                = self.data.borrow();
+        let language_server_rpc = &data.language_server_rpc;
+        let root_id             = language_server_rpc.content_root();
+        let path                = language_server::Path::new(root_id,&[VISUALISATION_FOLDER]);
+        let file_list           = language_server_rpc.file_list(&path).await?;
```
mwu-tow, line 102: I'm not convinced if we should error-out if the visualization folder does not exist. Perhaps we should just return empty vector then.
(but I'm not sure if we can now tell the errors apart, so likely not feasible anyway)

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+        let result              = file_list.paths.iter().filter_map(|object| {
+            if let language_server::FileSystemObject::File{name,path} = object {
+                let mut path = path.clone();
+                path.segments.push(name.to_string());
+                let root_id    = path.root_id;
+                let segments   = path.segments;
+                let path       = language_server::Path{root_id,segments};
```
mwu-tow, line 109: We should make `From<FileSystemObject> for Path`.
This logic is not specific to visulization controller.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+                let identifier = visualization.clone();
+                let error      = || VisualizationError::NotFound{identifier}.into();
+                result.cloned().ok_or_else(error)
+            },
+            VisualizationIdentifier::File(path) => {
+                let language_server = &self.data.borrow().language_server_rpc;
+                let js_code         = language_server.read_file(&path).await?.contents;
+                let identifier      = visualization.clone();
+                let error           = |_| VisualizationError::CompileError{identifier}.into();
+                let js_class        = JsSourceClass::from_js_source_raw(&js_code).map_err(error);
```
mwu-tow, line 149: The `JsSourceClass` seems to refer to this as "Instantiation Error" rather than "Compile Error"

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use enso_protocol::language_server::FileSystemObject;
+    use enso_protocol::language_server::Path;
+    use graph_editor::component::visualization::{NativeConstructorClass, Signature, Visualization};
+    use graph_editor::component::visualization::renderer::example::native::BubbleChart;
+    use ensogl::display::Scene;
+    use json_rpc::expect_call;
```
mwu-tow, line 171: should be ordered

src/rust/lib/debug-scenes/src/visualization.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let mut was_rendered = false;
+    let mut loader_hidden = false;
+    world.on_frame(move |time_info| {
+        let _keep_alive = &navigator;
+
+        let data    = generate_data((time_info.local / 1000.0).into());
+        let data    = Rc::new(data);
+        let content = Rc::new(serde_json::to_value(data).unwrap());
+        let data    = Data::JSON{ content };
```
mwu-tow, line 108:
```suggestion
        let data    = Data::JSON {content};
```

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+    use wasm_bindgen_test::wasm_bindgen_test;
+
+    wasm_bindgen_test_configure!(run_in_browser);
+
+    #[wasm_bindgen_test(async)]
+    async fn list_and_load() {
+        let mock_client = language_server::MockClient::default();
+
+        let root_id = uuid::Uuid::default();
+        let path    = Path{root_id,segments:vec!["visualization".into()]};
```
mwu-tow, line 183: why not use Path::new like below?

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+        )));
+        embedded_visualizations.insert("PointCloud".to_string(),embedded_visualization.clone());
+        let vis_controller           = Handle::new(language_server,embedded_visualizations);
+
+        let visualizations = vis_controller.list_visualizations().await;
+        let visualizations = visualizations.expect("Couldn't list visualizations.");
+
+        assert_eq!(visualizations[0],VisualizationIdentifier::Embedded("PointCloud".to_string()));
+        assert_eq!(visualizations[1],VisualizationIdentifier::File(path0));
+        assert_eq!(visualizations[2],VisualizationIdentifier::File(path1));
```
mwu-tow, line 233: we should assert that there are no more visualizations listed

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        let javascript_vis0 = JsSourceClass::from_js_source_raw(&file_content0);
+        let javascript_vis1 = JsSourceClass::from_js_source_raw(&file_content1);
+        let javascript_vis0 = javascript_vis0.expect("Couldn't create visualization class.");
+        let javascript_vis1 = javascript_vis1.expect("Couldn't create visualization class.");
+        let javascript_vis0 = Rc::new(class::Handle::new(javascript_vis0));
+        let javascript_vis1 = Rc::new(class::Handle::new(javascript_vis1));
+
+        let content = vec![embedded_visualization,javascript_vis0,javascript_vis1];
```
mwu-tow, line 242: `content` is extremely generic name here. Perhaps something like `expected_js_classes` ? 
Also I don't think we anywhere else refer to `JsSourceClass` handle as to "content". (it more brings to my mind the "file contents", i.e. before the JsSourceClass instantiation)

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+        let zipped  = visualizations.iter().zip(content.iter());
+        for (visualization,content) in zipped {
+            let loaded_content   = vis_controller.load_visualization(&visualization).await;
+            let loaded_content   = loaded_content.expect("Couldn't load visualization's content.");
+            let loaded_signature = loaded_content.class();
+            let loaded_signature = loaded_signature.as_ref().expect("Couldn't get class.").signature();
+            let signature        = content.class();
+            let signature        = signature.as_ref().expect("Couldn't get class.").signature();
+            assert_eq!(signature,loaded_signature);
```
mwu-tow, line 251: prefer having expected value on the right side of assert_eq

src/rust/ide/src/view/layout.rs
```diff
@@ -87,26 +87,30 @@ impl ViewLayoutData {
...
+    , visualization_controller : controller::Visualization
+    , fonts                    : &mut font::Registry
+    ) -> FallibleResult<Self> {
         let logger        = logger.sub("ViewLayout");
         let world         = &application.display;
         let text_editor   = TextEditor::new(&logger,world,text_controller,kb_actions,fonts);
-        let node_editor   = NodeEditor::new(&logger,application,graph_controller.clone_ref());
         let node_searcher = NodeSearcher::new(world,&logger,graph_controller.graph.clone_ref(),fonts);
+        let graph_controller = graph_controller.clone_ref();
+        let node_editor      = NodeEditor::new
```
mwu-tow, line 104: `=` alignment

src/rust/ide/src/view/layout.rs
```diff
@@ -87,26 +87,30 @@ impl ViewLayoutData {
...
+            (&logger,application,graph_controller,visualization_controller).await?;
         world.add_child(&text_editor.display_object());
         world.add_child(&node_editor);
         world.add_child(&node_searcher);
         let size         = zero();
         let data         = ViewLayoutData {text_editor,node_editor,node_searcher,size,logger};
         let rc           = Rc::new(RefCell::new(data));
-        Self {rc}.init(world,kb_actions)
+        let rc = Self {rc}.init(world,kb_actions);
```
mwu-tow, line 112: `=` alignment

src/rust/lib/debug-scenes/src/shapes.rs
```diff
@@ -181,6 +181,8 @@ use graph_editor::component::node::port::Expression;
 
 
 pub fn expression_mock() -> Expression {
+    let pattern_cr       = vec![Seq { right: false }, Or, Or, Build];
```
mwu-tow, line 184: `cr` is not really a descriptive name
(and the situation here doesn't really justify us using shortcuts)

src/rust/lib/debug-scenes/src/visualization.rs
```diff
@@ -0,0 +1,124 @@
...
+use ensogl::system::web;
+use ensogl::application::Application;
+use wasm_bindgen::prelude::*;
+use ensogl_core_msdf_sys::run_once_initialized;
+use graph_editor::component::visualization::JsSourceClass;
+use graph_editor::component::visualization::Data;
+use graph_editor::component::visualization::Registry;
+use std::rc::Rc;
+use nalgebra::Vector2;
+use js_sys::Math::sin;
```
mwu-tow, line 13: imports order

src/rust/lib/debug-scenes/src/visualization.rs
```diff
@@ -0,0 +1,124 @@
+#![allow(missing_docs)]
```
mwu-tow, line 1: I'd welcome some documentation for the scene.
How is it working? What should I expect to see?

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,254 @@
...
+
+// ===============================
+// === VisualizationIdentifier ===
+// ===============================
+
+/// This enum is used to identify a visualization both in the project folder or natively embedded in
+/// IDE.
+#[derive(Clone,Debug,Display,Eq,PartialEq)]
+#[allow(missing_docs)]
+pub enum VisualizationIdentifier {
```
mwu-tow, line 44: I don't like that we have both `VisualizationIdentifier` and `VisualizationId` and that they mean totally different things.

src/rust/ide/src/controller/project.rs
notdanilo: @mwu-tow I have followed the rule for keeping it one line instead of separating it in multiple lines.

src/rust/ide/src/controller/visualization.rs
notdanilo: I am removing this statement from this module.

src/rust/ide/src/controller/visualization.rs
notdanilo: I am writing "JavaScript visualization \"{}\" failed to be evaluated." instead.

src/rust/ide/src/controller/visualization.rs
notdanilo: @mwu-tow some of these statements were copied from other controllers.

notdanilo:
```
/// Visualization Controller is responsible for listing and loading all the available
/// visualizations on the project and the native ones embedded on IDE.
```

src/rust/ide/src/controller/visualization.rs
notdanilo: I improved this logic by verifying it calling `file_exists(path)` first.

src/rust/ide/src/controller/visualization.rs
notdanilo: @mwu-tow how do you feel about `VisualizationPath`?

mwu-tow: `VisualizationPath` is fine.

src/rust/ide/src/controller/visualization.rs
mwu-tow: This fells a little race-y, as the directory can be removed between the check and read operation.
But it is correct here, as the fact of directory not existing at any point during this call, is a good enough reason for us to return empty vector.

Still, I'd leave a comment that once we have a better mechanism for recognizing Language Server errors one from another, we should use this capability here.

src/rust/ide/enso-protocol/src/language_server/types.rs
```diff
@@ -37,6 +49,25 @@ impl Display for Path {
 }
 
 impl Path {
+    /// Splits path into name and segments. e.g.:
```
farmaazon, line 52:
```suggestion
    /// Splits path into name and path to parent directory. e.g.:
```

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,256 @@
...
+    language_server_rpc     : Rc<language_server::Connection>,
+    embedded_visualizations : EmbeddedVisualizations
+}
+
+/// Visualization Controller is responsible for listing and loading all the available
+/// visualizations on the project and the native ones embedded on IDE.
+#[derive(Debug,Clone,CloneRef)]
+pub struct Handle {
+    data : Rc<RefCell<VisualizationControllerData>>
+}
```
farmaazon, line 85: The Rc for language_server connection does not have to be under RefCell, because it's already internal mutable. Therefore I suggest to simplify the structs this way:
```suggestion
/// Visualization Controller is responsible for listing and loading all the available
/// visualizations on the project and the native ones embedded on IDE.
#[derive(Debug,Clone,CloneRef)]
pub struct Handle {
    language_server_rpc     : Rc<language_server::Connection>,
    embedded_visualizations : Rc<RefCell<EmbeddedVisualizations>>
}
```
Then you need borrow only when obtaining embedded visualizations.

src/rust/ide/enso-protocol/src/language_server/types.rs
```diff
@@ -29,6 +29,18 @@ pub struct Path {
 
 }
 
+impl From<&FileSystemObject> for Path {
+    fn from(file_system_object:&FileSystemObject) -> Path {
+        match file_system_object {
+            FileSystemObject::Directory{name,path}          => path.clone_with_segment(name),
```
wdanilo, line 35: The name `clone_with_segment` is not understnadable. It is `append` like? Immutable one? If `Path` is designed to be always immutable, just name it `append`. If `Path` is designed to be mutable, name it `append_im` (like a lot of Rust libs do). I believe that `Path` should be immutable by design btw.

src/rust/ide/enso-protocol/src/language_server/types.rs
```diff
@@ -37,6 +49,25 @@ impl Display for Path {
 }
 
 impl Path {
+    /// Splits path into name and path to parent directory. e.g.:
+    /// Path{root_id,segments:["foo","bar","qux"]} => ("qux",Path{root_id,segments:["foo","bar"]})
+    pub fn split_into_name_and_parent(mut self) -> Option<(String, Path)> {
+        self.segments.pop().map(|name| (name,self))
+    }
```
wdanilo, line 56: Use shorter names. Actually names for path operations are so standard, we should use the same as are used in Python or JavaScript.

src/rust/ide/src/controller/project.rs
wdanilo: Extract `language_server_rpc.clone()` to variable `server_rpc` and you can apply Michals comment :)

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,251 @@
...
+#[derive(Shrinkwrap,Debug,Clone,Default)]
+#[shrinkwrap(mutable)]
+pub struct EmbeddedVisualizations {
+    #[allow(missing_docs)]
+    pub map:HashMap<EmbeddedVisualizationName,Rc<class::Handle>>
+}
+
+
+
+
```
wdanilo, line 67: spaces

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,251 @@
...
+// ==============
+// === Handle ===
+// ==============
+
+const VISUALIZATION_FOLDER:&str = "visualization";
```
wdanilo, line 72: Oh no, please no. You are defining a folder constant in a module `controller::visualization`. This is very, very bad. Such constants should be in some top-level file, easily accessible and visible.

src/rust/ide/src/controller/visualization.rs
```diff
@@ -0,0 +1,251 @@
...
+                let identifier              = visualization.clone();
+                let error                   = || VisualizationError::NotFound{identifier}.into();
+                result.cloned().ok_or_else(error)
+            },
+            VisualizationPath::File(path) => {
+                let js_code         = self.language_server_rpc.read_file(&path).await?.contents;
+                let identifier      = visualization.clone();
+                let error           = |_| VisualizationError::InstantiationError {identifier}.into();
+                let js_class        = JsSourceClass::from_js_source_raw(&js_code).map_err(error);
+                js_class.map(|js_class| Rc::new(class::Handle::new(js_class)))
```
wdanilo, line 141: alignemnt

PR #437, opened 2020-05-19
Calling Text API of Language Server on each module change

farmaazon:
### Pull Request Description
This introduces additional layer for module model, which synchronizes it with LanguageServer - both its state and the lifetime (it is opened in LS on creation and closed when dropped).

### Important Notes
This contains also some improvements in json-rpc client mocks.

### Checklist


src/rust/ide/src/model/synchronized/module.rs
```diff
@@ -0,0 +1,381 @@
...
+    async fn notify_language_server
+    ( &self
+    , ls_content        : &ContentSummary
+    , edits_constructor : impl FnOnce(SerializedSourceFile) -> Vec<TextEdit>
+    ) -> FallibleResult<ParsedContentSummary> {
+        let content = self.model.serialized_content()?;
+        let summary = ParsedContentSummary::from_source(&content);
+        let edit    = language_server::types::FileEdit {
+            path        : self.path.file_path().clone(),
+            edits       : edits_constructor(content),
```
mwu-tow, line 231: What are the rules for applying multiple edits in the same batch?
Do subsequent edits are described in positions relative to the original contents or relative to the result of previous edit? 

Have we tested this / do we comply with whatever engine is doing here?

src/rust/ide/json-rpc/src/macros.rs
```diff
@@ -116,52 +115,100 @@ macro_rules! make_rpc_methods {
...
+                    if self.require_all_calls.get() {
+                        $(assert!(self.expect.$method.borrow().is_empty(), "Didn't make expected call {}");)*
+                    }
+                }
+            }
+
+            /// A set of handlers of expected Mock Client calls. Handlers get call's parameters and
+            /// returns the value to be returned.
+            #[derive(Default)]
+            pub struct ExpectedCalls {
```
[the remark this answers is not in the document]
farmaazon: I leave a todo then.

src/rust/ide/src/model/synchronized/module.rs
```diff
@@ -0,0 +1,381 @@
...
+
+/// The minimal information about module's content, required to do properly invalidation of opened
+/// module in Language Server.
+#[derive(Clone,Debug)]
+struct ContentSummary {
+    digest      : Sha3_224,
+    end_of_file : TextLocation,
+}
+
+/// The information about module's content, which was parsed at least once. In addition to minimal
```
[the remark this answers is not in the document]
farmaazon: If we haven't parsed the code, we don't have information about code, idmap and metadata positions.

src/rust/ide/src/model/synchronized/module.rs
```diff
@@ -0,0 +1,381 @@
...
+    /// so any new update attempt should perform full invalidation.
+    fn new_ls_content_info
+    (&self, old_content:ContentSummary, new_content:FallibleResult<ParsedContentSummary>)
+    -> LanguageServerContent {
+        match new_content {
+            Ok(new_content) => LanguageServerContent::Synchronized(new_content),
+            Err(err)        => {
+                error!(self.logger,"Error during sending text change to Language Server: {err}");
+                LanguageServerContent::Desynchronized(old_content)
+            }
```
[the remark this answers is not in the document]
farmaazon: Yes, and I'd rather assume that the only errors will be regarding connectivity. Then we should discuss with engine better handling such situations.

src/rust/ide/src/controller/module.rs
```diff
@@ -116,34 +116,26 @@ impl Handle {
...
     /// Save the module to file.
     pub fn save_file(&self) -> impl Future<Output=FallibleResult<()>> {
-        let path    = self.path.file_path().clone();
+        let content = self.model.serialized_content();
+        let path    = self.path.clone_ref();
         let ls      = self.language_server.clone();
-        let content = self.model.source_as_string();
-        async move { Ok(ls.client.write_file(&path,&content?).await?) }
+        async move {
+            let version = Sha3_224::new(content?.string.as_bytes());
```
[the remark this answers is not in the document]
farmaazon: As for line endings: we read the file as-is, and then make the initial update. We do it mainly due to id_map update, but in the same operation we set all line endings to '\n'.

src/rust/ide/enso-protocol/src/language_server/types.rs
```diff
@@ -342,6 +336,15 @@ pub struct CapabilityRegistration {
     pub register_options: RegisterOptions
 }
 
+impl CapabilityRegistration {
+    /// Create "text/canEdit" capability for path
+    pub fn create_can_edit(path:Path) -> Option<Self> {
```
mwu-tow, line 341: I'd consider including `text` as part of this method name. Currently this does not collide (the exeuction context is fortunately "modified" rather than "edited") but still...
Though yes, I know that in some other places we similarly don't follow, so this change might be bad for consistency or troublesome to fully push through.

So writing this just for consideration.

src/rust/ide/enso-protocol/src/language_server/types.rs
```diff
@@ -342,6 +336,15 @@ pub struct CapabilityRegistration {
...
+        let method           = "text/canEdit".to_string();
+        let register_options = RegisterOptions::ReceivesTreeUpdates(ReceivesTreeUpdates {path});
```
mwu-tow, line 343: This is IMO incorrect. `ReceivesTreeUpdates` is the options for `file/receivesTreeUpdates`, not for `text/canEdit`.
The latter should get its own variant in the `RegisterOptions`.

This will likely work as-is, since both structures are identical (and the name is not serialized) but still this is by accident.

src/rust/ide/src/model/synchronized/module.rs
farmaazon: tested it and apparently subsequent edits' positions should be those "affected" by previous edits. I fixed that.

kustosz: I was sure this was documented! Apparently it's not, I'll add the doc for it. You are correct – multiple edits in one message behave the same as multiple messages with one edit each.

PR #439, opened 2020-05-20
Ability to change text on nodes.

wdanilo:
### Pull Request Description
Implements https://github.com/luna/ide/issues/440

### Important Notes
Please take into consideration that most changes in this PR are in the WIP-files. So not everything is as beautiful as it should be. Nodes, connections, graph editor implementations will be cleaned up gradually in the next PRs.

### Checklist

### DIFF
--- a/src/rust/ensogl/src/display/shape/text/glyph/font.rs
+++ b/src/rust/ensogl/src/display/shape/text/glyph/font.rs
@@ -234,12 +234,18 @@ impl RenderInfo {
 
 
 
-// ===================
-// === LoadedFonts ===
-// ===================
+// ==============
+// === Handle ===
+// ==============
 
 /// A handle for fonts loaded into memory.
 pub type Handle = Rc<RenderInfo>;
+
+
+
+// ================
+// === Registry ===
+// ================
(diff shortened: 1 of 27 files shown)


src/rust/ensogl/src/display/shape/text/glyph/font.rs
```diff
@@ -267,19 +280,50 @@ impl FontRegistry {
...
+        let model = Rc::new(RefCell::new(Registry::new()));
+        Self {model}
+    }
+
+    /// Get render font info from loaded fonts, and if it does not exists, load data from one of
+    /// embedded fonts. Returns None if the name is missing in both loaded and embedded font list.
+    pub fn get_or_load_embedded_font(&self, name:&str) -> Option<Handle> {
+        self.model.borrow_mut().get_or_load_embedded_font(name)
+    }
+}
```
farmaazon, line 319: I wonder if we could have just only one `Registry` which is internal mutable. Then everyone can decide if he want to keep it in Rc or not. And there will be only one structure to comprehend.

But it's up to you, do as you please.

PR #450, opened 2020-05-22
Add timeout to rpc handler.

Josef-Vonasek:
### Checklist


src/rust/ide/json-rpc/src/error.rs
```diff
@@ -28,6 +28,10 @@ pub enum RpcError {
     /// Failed to deserialize message from server.
     #[fail(display = "Failed to deserialize a message: {}.", _0)]
     DeserializationFailed(serde_json::Error),
+
+    /// Response timeout.
+    #[fail(display = "Response timeout after: {:?}.", _0)]
```
mwu-tow, line 33: Can we pretty print to something more like "after {} ms." ? 
When feasible, we want the errors to be pretty, as they might end up being displayed to user.

src/rust/ide/json-rpc/src/handler.rs
```diff
@@ -222,6 +227,7 @@ impl<Notification> Handler<Notification> {
     /// `Transport` must be functional (e.g. not in the process of opening).
     pub fn new(transport:impl Transport + 'static) -> Handler<Notification> {
         let data = HandlerData {
+            timeout         : Duration::new(1,0),
```
mwu-tow, line 230: Prefer `Duration::from_secs(1)`, it's at glance what the value menas

src/rust/ide/json-rpc/Cargo.toml
```diff
@@ -8,11 +8,12 @@ edition = "2018"
...
-shrinkwraprs = "0.3.0"
+enso-prelude      = { version = "0.1.0"  , path = "../../lib/prelude" }
+ensogl-system-web = { version = "0.1.0"  , path = "../../lib/system/web"   }
+shapely           = { version = "0.1.0"  , path = "../../lib/shapely/impl" }
+utils             = { version = "0.1.0"  , path = "../utils" }
+serde             = { version = "1.0", features = ["derive"] }
+futures           = "0.3.1"
+failure           = "0.1.6"
+serde_json        = "1.0"
+shrinkwraprs      = "0.3.0"
```
farmaazon, line 19: Make the dependency list looking similar to other Cargo.tomls:
* use `{ version = }` for all and align {}
* split to two groups (internal and external) and sort them alphabetically (move `serde`)

src/rust/ide/json-rpc/src/error.rs
```diff
@@ -9,7 +9,6 @@ use crate::messages::Response;
 use futures::channel::oneshot::Canceled;
 
 
-
```
farmaazon, line 12: We should not remove this. Before "big" section headers we require three empty lines.

src/rust/ide/json-rpc/src/error.rs
```diff
@@ -28,6 +27,11 @@ pub enum RpcError {
     /// Failed to deserialize message from server.
     #[fail(display = "Failed to deserialize a message: {}.", _0)]
     DeserializationFailed(serde_json::Error),
+
+    /// Response timeout.
+    #[allow(missing_docs)]
+    #[fail(display = "Response timeout after {} ms.", millis)]
+    TimeoutError{millis:u128},
```
farmaazon, line 34: Why we use u128 here? Do we expect timeout exceeding 18446744073709551615 (c.a. 500 millions years)?

src/rust/ide/json-rpc/src/handler.rs
```diff
@@ -195,6 +203,11 @@ impl<Notification> {
             channel::emit(event_transmitter,event)
         }
     }
+
+    /// All requests timeout after this duration.
+    pub fn timeout(&self) -> Duration {
+        self.timeout
+    }
```
farmaazon, line 210: I don't understand: what is "this duration"?

src/rust/ide/json-rpc/src/handler.rs
```diff
@@ -131,11 +134,16 @@ pub type OngoingCalls = HashMap<Id,oneshot::Sender<ReplyMessage>>;
 
 pub use shapely::shared;
 
+/// The default timeout for all responses given in seconds.
+const TIMEOUT:u64 = 1;
```
wdanilo, line 138: The info that it is in secodns should be encoded in type OR in name

src/rust/ide/json-rpc/src/handler.rs
```diff
@@ -276,7 +290,14 @@ impl<Notification> Handler<Notification> {
...
         }
-        ret
+
+        let timeout = self.timeout();
+        future::select(ret, sleep(timeout).boxed()).map(move |either|
+            match either {
+                future::Either::Left ((x, _)) => x,
+                future::Either::Right((_, _)) => Err(RpcError::TimeoutError{millis:timeout.as_millis()}),
+            }
+        )
```
farmaazon, line 300: Have you checked out `select!` macro? Perhaps it make a little bit prettier code here.

src/rust/ide/json-rpc/src/handler.rs
```diff
@@ -276,7 +290,14 @@ impl<Notification> Handler<Notification> {
...
             self.remove_ongoing_request(id);
         }
-        ret
+
+        let timeout = self.timeout();
+        future::select(ret, sleep(timeout).boxed()).map(move |either|
+            match either {
+                future::Either::Left ((x, _)) => x,
+                future::Either::Right((_, _)) => Err(RpcError::TimeoutError{millis:timeout.as_millis()}),
+            }
```
wdanilo, line 299: There has to be a function for that. It is too popular operation to require pattern matching

src/rust/ide/json-rpc/src/handler.rs
Josef-Vonasek: Yes, and it doesn't.

src/rust/ide/json-rpc/src/error.rs
Josef-Vonasek: It's the type that is returned by Duration.

src/rust/ide/json-rpc/src/handler.rs
Josef-Vonasek: It's a type from std. You can see it's docs for more info.

src/rust/ide/json-rpc/src/handler.rs
Josef-Vonasek: I am not aware of such function.

src/rust/ide/json-rpc/src/handler.rs
farmaazon: Oh, I got `timeout` as a noun, not a verb!
```suggestion
    /// A `Duration` after which requests are timed out.
    pub fn timeout(&self) -> Duration {
        self.timeout
    }
```

src/rust/ide/json-rpc/tests/test.rs
```diff
@@ -17,6 +17,8 @@ use serde::Serialize;
 use std::future::Future;
 use std::pin::Pin;
 use utils::test::traits::*;
+use std::thread::sleep;
+use ensogl_system_web::Duration;
```
mwu-tow, line 21: please keep this sorted

src/rust/ide/json-rpc/src/handler.rs
```diff
@@ -131,11 +134,16 @@ pub type OngoingCalls = HashMap<Id,oneshot::Sender<ReplyMessage>>;
 
 pub use shapely::shared;
 
+/// The default timeout for all responses given in seconds.
+const TIMEOUT_SECS:u64 = 1;
```
mwu-tow, line 138: I'm afraid that not but I have to ask — perhaps `Duration` supports `const` so we could properly type this?
(if not, leave as-is)

PR #459, opened 2020-05-25
Visualization active state

MichaelMauderer:
### Pull Request Description
Adds an active state, fullscreen state, transition animations, and double click FRP interaction.

### Checklist


src/rust/ensogl/src/application/shortcut.rs
```diff
@@ -5,9 +5,14 @@ use crate::prelude::*;
...
 use crate::control::io::keyboard::listener::KeyboardFrpBindings;
-use crate::frp::io::keyboard::Keyboard;
 use crate::frp::io::keyboard::KeyMask;
+use crate::frp::io::keyboard::Keyboard;
 use crate::frp;
+#[cfg(target_arch = "wasm32")]
+use crate::system::web;
+
+#[cfg(not(target_arch = "wasm32"))]
+use std::time;
```
wdanilo, line 15: I dont like such conditional compilation. Instead, the `crate::system::*` should work across all platforms. Sure, it panics now on native, but its better to leave it in this stage and improve `crate::system::*` in the future than conditionally hide methods on usage side.

src/rust/ensogl/src/application/shortcut.rs
```diff
@@ -75,7 +84,22 @@ impl Registry {
...
+    /// to detect DoublePress actions.
+    fn pre_process_action(&self, action_type:ActionType, key_mask:&KeyMask) -> ActionType {
+        let mut detector     =  self.double_press_detector.borrow_mut();
+        let is_double_press  = detector.process_action(action_type,key_mask);
+        if is_double_press {
+            ActionType::DoublePress
+        } else {
+            action_type
+        }
+    }
```
wdanilo, line 98: I really don't like this design. There was an FRP network that decides whether this is press or release event. The idea is that FRP is the controller (drives the whole logic), and then uses functions to change the model. Now, the FRP is part of the stateful controller, discovers action like press or release, then functions it calls can change the action by adding some more state and control logic. Instead, the discovery whether it was press / release / double press should be all in FRP.

src/rust/ensogl/src/application/shortcut.rs
```diff
@@ -348,3 +377,67 @@ impl<T:command::Provider> DefaultShortcutProvider for T {
...
+        let within_threshold = if let Some(prev_time) = self.prev_time {
+            (now - prev_time) < self.threshold_ms
+        } else {
+            false
+        };
+        self.prev_time = Some(now);
+        self.prev_key  = Some(*key_mask);
+        is_same_key && within_threshold
+    }
+}
```
wdanilo, line 443: This whole thing should be encoded as FRP network.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -702,28 +702,44 @@ impl ViewData {
...
+        let overlay_cursor         = View::new(&logger,width,height);
+        let label                  = View::new_with_camera(&logger,&main.camera);
+        let all                    = vec![main.downgrade(),
+                                          visualisation.downgrade(),
+                                          cursor.downgrade(),
+                                          label.downgrade(),
+                                          overlay.downgrade(),
+                                          overlay_visualisation.downgrade(),
+                                          overlay_cursor.downgrade(),
+                                     ];
```
wdanilo, line 739: Break alignment and make it 2 lines instead pls :)

src/rust/ensogl/src/display/symbol/dom.rs
```diff
@@ -28,7 +28,7 @@ mod js {
 
         export function set_object_transform(dom, matrix_array) {
             let css = arr_to_css_matrix3d(matrix_array);
-            dom.style.transform = 'translate(-50%, -50%)' + css;
+            dom.style.transform = css;
```
wdanilo, line 31: Why this change? Wasn't it just placing DOM elements in such way that their origin is in the middle? If so, that was intentional - to make these shapes behave the same way as our webgl shapes.

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -379,13 +379,14 @@ impl NodeModel {
...
 
-        let visualization_container = visualization::Container::new();
+        let visualization_container = visualization::Container::new(&scene);
         visualization_container.mod_position(|t| {
             t.x = 60.0;
             t.y = -120.0;
         });
 
         display_object.add_child(&visualization_container);
+        visualization_container.frp.set_visibility.emit(false);
```
wdanilo, line 389: The container should not be visible by default. We should not need to emit event to hide it.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
+//! Provides generic operations that can be applied to UI components.
```
wdanilo, line 1: I dont like `component::operator` name. I was not expecting this content when reading the file name. Im still not sure what the proper name is tohugh.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// Return the size of the UI element.
+    fn size(&self) -> Vector3<f32>;
+}
+
+/// A component that owns a Network. Can be used to create external animations.
+pub trait Networked {
+    /// Return a reference to the components network.
+    fn network(&self) -> &frp::Network;
+}
+
```
wdanilo, line 85: nls

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+/// Indicates the required target layer.
+// FIXME this is a layer management hack. Remove this once we have nicer scene layer management.
+#[derive(Debug)]
+pub enum SymbolType {
+    /// A symbol that goes onto the `Main` layer.
+    Main (Symbol),
+    /// A visualisation symbol that goes above the `Main` layer, but below the cursor.
+    Visualisation (Symbol),
+}
```
wdanilo, line 141: This is confusing. This is meant to be symbol type, but in fact it contains a symbol! Either the name is bad or this should be refactored to separate struct (preferably the later).

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+        });
+        resize_animation.set_target_position(1.0);
+
+        self.scene.views.toggle_overlay_cursor();
+
+        self
+    }
+
+    /// Undo the fullscreen operation and restore the previous state exactly as it was.
+    pub fn undo(self) {
```
wdanilo, line 236: What does `my_object.undo()` does? If this is designed to be a method, it should be named `undo_fullscreen`.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+/// symbols some helper methods for working with them.
+pub trait HtmlComponent {
+    /// Return all `DomSymbol`s that make up this component.
+    fn elements(&self) -> Vec<DomSymbol>;
+
+    /// Change the `DomScene` of all `DomSymbol`s.
+    fn set_layer(&self, scene:&DomScene) {
+        self.elements().iter().for_each(|element| scene.manage(&element))
+    }
+}
```
wdanilo, line 69: Lets remove it. After we introduce cutting holes to the webgl layer, there will be a single HTML scene, so this will never be used.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+pub trait Fullscreenable = display::Object+Resizable+NativeComponent+Networked+CloneRef+'static;
+
+/// A `FullscreenOperator` can be used to apple fullscreen mode to a UI element as well as undo the
+/// undo the fullscreen operation and restore the previous state. The  `FullscreenOperator` can be
+/// applied to any target that implements `display::Object`, `Resizable` and `NativeUiElement`.
+// TODO consider incorporating these traits into display::Object or another common "SceneElement"
+// type. But it is important that complex UI components can provide information about their
+// sub-components (for example, multiple sub-shapes or HTML components).
+#[derive(Debug)]
+pub struct FullscreenOperator<T> {
```
wdanilo, line 184: I dont really understand this "Operator" naming here. If I understand correctly, this is just `FullscreenStateModel` while the current `FullscreenOperatorHandle` is just `FullscreenState`. Is this correct?

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+        self.target.display_object().set_parent(self.scene.display_object());
+        self.target.unset_layers_all(&self.scene);
+        set_layers_fullscreen(&self.target, &self.scene);
+
+        let margin = 0.0;
+        let scene_shape = self.scene.shape();
+        let size_new    = Vector3::new(scene_shape.width(), scene_shape.height(),0.0) * (1.0 - margin);
+
+        // TODO Currently we assume objects are center aligned, but this needs to be properly
+        // accounted for here.
```
wdanilo, line 215: What does "properly accounted" mean here? And what are the "objects"? When can objects not be center aligned? I mean - I really do not understanding this TODO.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+        let target_size      = size_new;
+        let target           = self.target.clone_ref();
+        let resize_animation = animation(frp_network, move |value| {
+            let pos  = original_pos  * (1.0 - value) + target_pos * value;
+            let size = original_size * (1.0 - value) + target_size * value;
+            target.set_position(pos);
+            target.set_size(size);
+        });
+        resize_animation.set_target_position(1.0);
+
```
wdanilo, line 229: 1. Code repetition. Below you have the same code.
2. You are leaking memory here. `animation(frp_network, ...)` creates a new node in `frp_network`. So if you call multiple times this function, the network will contain multiple nodes and they will not be dropped until network is dropped.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+        let parent_pos     = parent_pos.unwrap_or_else(Vector3::zero);
+        let mut source_pos = self.target.position();
+        source_pos        += global_pos_start ;
+        source_pos        -= parent_pos ;
+        let source_pos     = source_pos;
+
+        self.target.set_position(source_pos);
+
+        let original_pos     = self.target.position();
+        let target_pos       = self.position_original;
```
wdanilo, line 256: Here also, a lot of repetitive code.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+#[derive(Debug)]
+pub struct FullscreenOperator<T> {
+    target            : T,
+    scene             : Scene,
+    size_original     : Vector3<f32>,
+    position_original : Vector3<f32>,
+    parent_original   : Option<display::object::Instance>,
+}
+
+impl<T:Fullscreenable> FullscreenOperator<T> {
```
wdanilo, line 192: Just a sanity check - does it work with scene panned?

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+        let target_size      = self.size_original;
+        let target           = self.target.clone_ref();
+        let frp_network      = &self.target.network().clone_ref();
+        let resize_animation = animation(frp_network, move |value| {
+            let pos  = original_pos  * (1.0 - value) + target_pos * value;
+            let size = original_size * (1.0 - value) + target_size * value;
+            target.set_position(pos);
+            target.set_size(size);
+        });
+        resize_animation.set_target_position(1.0);
```
wdanilo, line 267: This is very unsafe. What happens here is complex:

1. You define FRP network above and you move to lambda `target`, which has interior mutability.
2. You are starting the FRP network here, which means the lambda could be evaluated. In particular, `target.set_position` or something could be fired. Other things could listen to this frp network and could potentially trigger undoing or doing fullscreen.
3. If they do so, there ow uldbe runtime crash because you are now in non-interial mutable thing (you have mutable borrowed it).

Thats why we often use `RefCells` only on some fields, not on whole structs like here - just to borrow things for the shortest possible time. If you borrow thing and you eval method on borrowed thing, you are not allowed to run any FRP that changes things that can be observed from outside of this borrow.

src/rust/lib/graph-editor/src/component/operator.rs
wdanilo, line 184: This whole thing works currently on visualizations only, lets move it to `component/visualization/fullscreen` for now, until we see a generalization pattern.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+    for symbol in target.symbols() {
+        match symbol {
+            SymbolType::Main(symbol)          => scene.views.main.add(&symbol),
+            SymbolType::Visualisation(symbol) => scene.views.visualisation.add(&symbol),
+        }
+    }
+}
+
+/// Moves the given components shapes to the fullscreen scene layers.
+// FIXME This is an ugly hack for layer management.
```
wdanilo, line 157: Describe the hack please. What is the hack? Or, why you consider it to be a hack?

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub fn get_element(&self) -> Option<T>{
+        self.operator.borrow().as_ref().map(|op| op.target.clone_ref())
+    }
+}
+
+
+
+// ===============================
+// === Layer Management Helper ===
+// ===============================
```
wdanilo, line 131: Please keep related implemetations close. Now e have `FullscreenOperatorHandle`, utils, `FullscreenOperator`. Its hard to read this code this way.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+    fn network(&self) -> &frp::Network;
+}
+
+// ==================================
+// === Fullscreen Operator Handle ===
+// ==================================
+
+/// FullscreenOperatorCellHandle is a helper type that wraps a `FullscreenOperator` and applies an
+/// undos the operator on the inner component. This can be used to ensure that only a single
+/// component is made fullscreen at any time.
```
wdanilo, line 92: If your docs refer to other entity in the code, it would be cool if the entity was defined earlier in the file so its easier to read the codebase without jumping the file up and down

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+        let operator = FullscreenOperator::apply(target,scene);
+        self.operator.set(operator);
+    }
+
+    /// Disables fullscreen mode for the given component.
+    pub fn disable_fullscreen(&self) {
+        if let Some(old) = self.operator.borrow_mut().take() {
+            old.undo();
+        }
+    }
```
wdanilo, line 119: then aming is inconsistent. `disable_fullscreen` = `undo`.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+
+
+
+// ==================================
+// === UI Component Helper Traits ===
+// ==================================
+
+/// Should be implemented by UI component that consist of `Symbol`. Provides access to the shapes
+/// and some helper methods for working with those shapes.
+pub trait NativeComponent {
```
wdanilo, line 34: After readign the code for 1 hour I un derstood that the logic for `NativeComponent` is in fact to manage its z-index only. Please rename it accordingly to this knowledge. hen it ill grow bigger and will have more responsibilities, we can rename it again. No its only for z-indexing.

wdanilo: Moreover, until we got proper z-indexing management, I would move it for now to `component::visualization::` space

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/// Provides information and functionality to resize an element. A complex UI component should
+/// implement this and propagate size and layout changes to all its sub-components.
+pub trait Resizable {
+    /// Set the size for the UI component.
+    fn set_size(&self, size:Vector3<f32>);
+    /// Return the size of the UI element.
+    fn size(&self) -> Vector3<f32>;
+}
```
wdanilo, line 78: Lets move it for now to `visualization::` space. Resizing elements in a generic fashion ill be there, but its complex and **until we implement a generic mechanism that is ready to be adopted by everything we have now, lets keep it in local module**.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+
+/// A component that owns a Network. Can be used to create external animations.
+pub trait Networked {
```
wdanilo, line 81: We have naming convention for that - it should be named `HasNetwork` (and should be placed in `frp::` space). BUT, it was not defined on purpose. I believe that defining it encourages you to actually leak memory accidently. See comments below. You should not access frp network on demand - only when initializing an object. Otherwise its too easy to leak memory. I would remove it.

src/rust/lib/graph-editor/src/component/operator.rs
```diff
@@ -0,0 +1,271 @@
...
+/// undos the operator on the inner component. This can be used to ensure that only a single
+/// component is made fullscreen at any time.
+#[derive(Debug,CloneRef,Derivative)]
+#[derivative(Clone(bound=""))]
+#[derivative(Default(bound=""))]
+pub struct FullscreenOperatorHandle<T> {
+    operator: Rc<RefCell<Option<FullscreenOperator<T>>>>
+}
+
+impl<T:display::Object+Resizable+ NativeComponent +CloneRef+Networked+'static> FullscreenOperatorHandle<T> {
```
wdanilo, line 100: This bound is huge! And sapcign is bad as well

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -2,11 +2,93 @@
...
+            let width_bg  : Var<Distance<Pixels>> = width_bg.into();
+            let height_bg : Var<Distance<Pixels>> = height_bg.into();
+            let radius    : Var<Distance<Pixels>> = padding.clone().into();
+            let color_bg      = color::Lcha::new(0.2,0.013,0.18,1.0);
+            let background    = Rect((&width_bg,&height_bg)).corners_radius(&radius);
+            let background    = background.fill(color::Rgba::from(color_bg));
+
+            let frame_outer = Rect((&width_bg,&height_bg)).corners_radius(&radius);
+
+            // +1 at the end to avoid aliasing artifacts.
```
wdanilo, line 51: name the variable `aliasing_overlay` instead.

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -2,11 +2,93 @@
...
+            let width_frame_inner       =  &width  - &padding * Var::<f32>::from(2.0) * &selected + Var::<f32>::from(1.0);
+            let height_frame_inner      =  &height - &padding * Var::<f32>::from(2.0) * &selected + Var::<f32>::from(1.0);
```
wdanilo, line 53: lines too long, here and in other places. Refactor to more vars.

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -2,11 +2,93 @@
...
+            let width_frame_inner  : Var<Distance<Pixels>> = width_frame_inner.into();
+            let height_frame_inner : Var<Distance<Pixels>> = height_frame_inner.into();
+            let inner_radius = &radius * (Var::<f32>::from(1.0) - &selected);
+            let frame_inner = Rect((&width_frame_inner,&height_frame_inner)).corners_radius(&inner_radius);
+
+            let frame = frame_outer.difference(frame_inner);
+            let color_frame    = color::Lcha::new(0.72,0.5,0.22,1.0);
+            let frame = frame.fill(color::Rgba::from(color_frame));
+
```
wdanilo, line 62: alignment

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -2,11 +2,93 @@
...
+    ensogl::define_shape_system! {
+        (width:f32,height:f32,selected:f32,padding:f32) {
+            let width_bg       = width.clone();
+            let height_bg      = height.clone();
+            let width_bg  : Var<Distance<Pixels>> = width_bg.into();
+            let height_bg : Var<Distance<Pixels>> = height_bg.into();
+            let radius    : Var<Distance<Pixels>> = padding.clone().into();
+            let color_overlay = color::Rgba::new(1.0,0.0,0.0,0.000_000_1);
+            let overlay    = Rect((&width_bg,&height_bg)).corners_radius(&radius);
+            let overlay       = overlay.fill(color_overlay);
```
wdanilo, line 84: alignment - here and other places

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -47,49 +141,64 @@ impl Default for ContainerFrp {
...
     // network: the FRP network holds `Rc`s to the `ContainerData` and thus must not live in the
     // same struct.
-
+    #[derivative(PartialEq(compare_with="ref_eq"))]
     #[shrinkwrap(main_field)]
-        data : Rc<ContainerData>,
-    pub frp  : ContainerFrp,
+    pub data              : Rc<ContainerData>,
+    #[derivative(PartialEq="ignore")]
+    pub frp               : ContainerFrp,
```
wdanilo, line 157: alignment

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -47,49 +141,64 @@ impl Default for ContainerFrp {
...
+}
+
+fn ref_eq(obj_a:&Rc<ContainerData>, obj_b:&Rc<ContainerData>) -> bool {
+    let ptr_a:*const _ = obj_a.as_ref();
+    let ptr_b:*const _ = obj_b.as_ref();
+    ptr_a == ptr_b
```
wdanilo, line 163: Rc has `ptr_eq`, why not using it instead?

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -122,27 +262,52 @@ impl display::Object for ContainerData {
...
+        let display_object_internal = display::object::Instance::new(&logger);
+        let display_object_visualisation = display::object::Instance::new(&logger);
+
+        let padding                 = Cell::new(10.0);
+        let shape_frame             = component::ShapeView::<frame::Shape>::new(&logger,scene);
+        let shape_overlay           = component::ShapeView::<overlay::Shape>::new(&logger,scene);
+        let scene                   = scene.clone_ref();
+        let data                    = ContainerData {logger,visualization,size,display_object,shape_frame,
+                                                     display_object_internal,padding,scene,shape_overlay,
+                                                     display_object_visualisation};
```
wdanilo, line 281: in such cases, just use next line with 4 spaces indentation

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -122,27 +262,52 @@ impl display::Object for ContainerData {
...
+        let data                    = Rc::new(data);
+        let frp                     = default();
+        Self {data,frp} . init() . init_frp()
```
wdanilo, line 284: init_frp should be part of init

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -162,14 +327,52 @@ impl Container {
                  container_data.visualization.borrow()
                     .for_each_ref(|vis| vis.frp.set_data.emit(data));
             }));
+
+            def _select = frp.select.map(f!([selection](_) {
```
wdanilo, line 331: use `f_!` instead. Plus, drop `[selection]`. Just simplify all usages below. These 3 lines are the same as:

```rust
eval frp.select (selection.set_target_position(1.0))
```

src/rust/lib/graph-editor/src/component/operator.rs
wdanilo: Actually the logic is even moe complex here. `NativeComponent` is both something that provides you with symbols and also layouts them. I would extract `HasSymbols` as separate trait and combine it with other traits to get the desired effect.

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -627,6 +633,88 @@ impl Nodes {
 
 
 
+// =========================
+// === Entity Collection ===
+// =========================
+
+#[derive(Clone,CloneRef,Debug)]
+#[allow(missing_docs)]
+struct EntityFrp {
```
wdanilo, line 642: This name is absolutely enigmatic.

src/rust/ensogl/src/display/scene.rs
MichaelMauderer: I think this one make more sense vertical: the layer ordering indicates the occlusion of the layers. Scanning it top to bottom here and seeing the order is easier than if it is inline.

src/rust/ensogl/src/display/symbol/dom.rs
MichaelMauderer: I think I had to do this for them to be consistent with the shapes. I will revisit this and double check the behaviour.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -737,6 +754,21 @@ impl Views {
...
     }
+
+    /// Move cursor from/to cursor/cursor_overlay layer.
+    pub fn toggle_overlay_cursor(&self) {
+        let mut bottom_symbols = self.cursor.symbols.borrow_mut();
+        let mut top_symbols    = self.overlay_cursor.symbols.borrow_mut();
+        mem::swap(bottom_symbols.deref_mut(), top_symbols.deref_mut());
+    }
+
+    pub fn remove_symbol(&self, symbol:&Symbol) {
```
wdanilo, line 765: missing docs

src/rust/lib/graph-editor/src/component.rs
```diff
@@ -1,7 +1,7 @@
 //! Root module for graph component definitions.
 
-pub mod connection;
 pub mod connection2;
+pub mod connection;
 pub mod cursor;
 pub mod node;
-pub mod visualization;
+pub mod visualization;
```
wdanilo, line 7: nl

src/rust/lib/graph-editor/src/component/visualization.rs
```diff
@@ -25,8 +25,11 @@ pub mod class;
 pub mod container;
 pub mod data;
 pub mod js;
+pub mod fullscreen;
 pub mod registry;
 pub mod renderer;
+pub mod traits;
```
wdanilo, line 31: I don't like the name of this module. This is a name that I call "syntactic name". It is like calling variable "var". Sure, this module contains traits, but yeah, every module contains structs, traits and funcs, but we don't name them "structs.rs" or "funcs.rs". We need to use semantic naming always :)

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -2,11 +2,98 @@
 
 use crate::prelude::*;
 
+use crate::component::visualization::traits::{HasSymbols, HasFullscreenDecoration};
```
wdanilo, line 5: let not use this import style :)

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -47,49 +146,58 @@ impl Default for ContainerFrp {
...
+    padding                      : Cell<f32>,
+    /// Topmost display object in the hierarchy. Used for global positioning.
+    display_object               : display::object::Instance,
+    /// Internal display object that will be sole child of `display_object` and can be attached and
+    /// detached from its parent for showing/hiding all child shapes.
+    display_object_internal      : display::object::Instance,
+    /// Parent of the visualisation. Allows adding/removing of visualisations without affecting
+    /// the order of other container shapes.
+    display_object_visualisation : display::object::Instance,
+
```
wdanilo, line 180: The documentation for these 3 display objexcts is complex and tbh, I don't fully understand it. Is there a way to make the relations and responsibilities of the objects easier to understand?

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -108,12 +216,52 @@ impl ContainerData {
...
+        frame_shape.height.set(height);
+        frame_shape.padding.set(padding);
+        frame_shape.sprite.size().set(Vector2::new(width, height));
+        overlay_shape.width.set(width);
+        overlay_shape.height.set(height);
+        overlay_shape.padding.set(padding);
+        overlay_shape.sprite.size().set(Vector2::new(width, height));
+
+
+        let pos = Vector3::new(width/2.0,height/2.0, 0.0);
```
wdanilo, line 244: too many whitespaces

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -108,12 +216,52 @@ impl ContainerData {
...
+    fn set_size(&self, size:Vector3<f32>) {
+        if let Some(vis) = self.visualization.borrow().as_ref() {
+            vis.set_size(size.xy());
+        }
+
+        self.size.set(size.xy());
+        self.update_shape_sizes();
+    }
+
+    fn set_corner_roundness(&self, value:f32) {
```
wdanilo, line 258: rename to set_corner_radiuses as this is what we use in other places, like in shapes definitions. We can alternatively rename it there, but I believe that using "radius" is more common (CSS like things)

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -122,27 +270,55 @@ impl display::Object for ContainerData {
...
+    }
+
+    fn init_shape(self) -> Self {
+        self.update_shape_sizes();
+        self.set_corner_roundness(1.0);
+        self.data.display_object_internal.add_child(&self.data.shape_overlay);
+        self.data.display_object_internal.add_child(&self.data.shape_frame);
+        self.data.display_object_internal.add_child(&self.data.display_object_visualisation);
+        // Remove default parents to stay hidden on init.
+        self.data.display_object.add_child(&self.display_object_internal);
```
wdanilo, line 306: refactor it to function named `hide_on_init` :)

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -162,14 +338,53 @@ impl Container {
                  container_data.visualization.borrow()
                     .for_each_ref(|vis| vis.frp.set_data.emit(data));
             }));
+
+            eval frp.select   ((_) selection.set_target_position(1.0));
+            eval frp.deselect ((_) selection.set_target_position(0.0));
+
+            def _output_hide = container_data.shape_overlay.events.mouse_down.map(f!([frp](_) {
+                frp.on_click.emit(())
+            }));
```
wdanilo, line 347: simplify it with `eval` and `f_!` what you have here is the same as:
```rust
eval container_data.shape_overlay.events.mouse_down (frp.on_click.emit());
```

src/rust/lib/graph-editor/src/component/visualization/traits.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// A symbol that goes onto the main layer.
+    Main,
+    /// A visualisation symbol that goes into the visualisation layer.
+    Visualisation,
+}
+
+/// Contains a symbol and information about which layer it should be placed on.
+#[derive(Clone,Debug)]
+#[allow(missing_docs)]
+pub struct SymbolLayoutData {
```
wdanilo, line 26: I still have problem with the name. It doesn't tell me what it is. I would vote for `SymbolWithLayout` instead.

src/rust/lib/graph-editor/src/component/visualization/container.rs
```diff
@@ -178,3 +393,13 @@ impl display::Object for Container {
...
+
+impl HasFullscreenDecoration for Container {
+    fn enable_fullscreen_decoration(&self) {
+        self.data.set_corner_roundness(0.0);
+    }
+
+    fn disable_fullscreen_decoration(&self) {
+        self.data.set_corner_roundness(1.0);
+    }
+}
```
wdanilo, line 405: nl

src/rust/lib/graph-editor/src/component/visualization/fullscreen.rs
```diff
@@ -0,0 +1,338 @@
...
+            },
+            _  => ()
+        }
+        // Check for end of animation and update the state.
+        if value >= 1.0 {
+           state.borrow_mut().animation_end_transition();
+        }
+    }
+}
+
```
wdanilo, line 83: nls

src/rust/lib/graph-editor/src/component/visualization/fullscreen.rs
```diff
@@ -0,0 +1,338 @@
...
+// ==================
+// === StateModel ===
+// ==================
+
+/// Represents the internal state of the FullscreenState. This is used to ensure animations always
+/// finish without interruption.
```
wdanilo, line 89: This is bad. We do want to interrupt animations - look at the cursor press animation. If you press and hold it gets slowly smaller. But if you release the animation starts in the middle of previous naimation.

src/rust/lib/graph-editor/src/component/visualization/fullscreen.rs
```diff
@@ -0,0 +1,338 @@
...
+
+/// The `FullscreenState` manages the state changes between fullscreen mode and non-fullscreen mode
+/// for a UI component. It creates animations for the state changes and ensure that the component
+/// cannot come into an illegal state during the transition.
+#[derive(Debug,CloneRef,Derivative)]
+#[derivative(Clone(bound=""))]
+pub struct FullscreenState<T> {
+    network   : frp::Network,
+    state     : Rc<RefCell<StateModel<T>>>,
+    animation : DynSimulator<f32>,
```
wdanilo, line 157: Please code all UI animations using FRP animations system only. Doing that using raw DynSimulator leads to a very complex code (like here). See for example animations in `cursor.rs`.

src/rust/lib/graph-editor/src/component/visualization/fullscreen.rs
```diff
@@ -0,0 +1,338 @@
...
+            if let StateModel::Fullscreen {data, ..} = state {
+                Some(data.clone())
+            } else {
+                None
+            }
+        };
+        if let Some(data) = fullscreen_data {
+            self.transition_to_non_fullscreen(&data)
+        }
+    }
```
wdanilo, line 202: If that was done in frp, the implementation would be one line - `fullscreen.set_target(1.0)`.

src/rust/lib/graph-editor/src/component/visualization/fullscreen.rs
```diff
@@ -0,0 +1,338 @@
...
+    },
+    /// There is an animation running from non-fullscreen mode to fullscreen mode.
+    TransitioningToFullscreen {
+        animation_data : AnimationTargetState<T>,
+        target_state   : FullscreenStateData<T>
+    },
+    /// There is no UI component in fullscreen mode.
+    NotFullscreen
+}
+
```
wdanilo, line 109: I believe that after using frp this whole struct would not be needed anymore

src/rust/lib/graph-editor/src/component/visualization/fullscreen.rs
```diff
@@ -0,0 +1,338 @@
...
+/// the UI component (setting of size/layers) outside of the animation and and provides the target
+/// state for the animation.
+#[derive(Debug,Clone)]
+pub struct FullscreenStateData<T> {
+    target            : T,
+    scene             : Scene,
+    size_original     : Vector3<f32>,
+    position_original : Vector3<f32>,
+    parent_original   : Option<display::object::Instance>,
+}
```
wdanilo, line 251: I want to review it again after the animation will be coded in frp. I believe it will be simplified drastically.

wdanilo:
Merged as other PR

PR #485, opened 2020-05-26
Make it possible to disable logging.

Josef-Vonasek:
### Checklist

### DIFF
--- a/src/rust/ensogl/src/lib.rs
+++ b/src/rust/ensogl/src/lib.rs
@@ -65,6 +65,7 @@ pub use enso_frp as frp;
 pub mod prelude {
     pub use enso_prelude::*;
     pub use logger::*;
+    pub use logger::enabled::Logger;
     pub use shapely::CloneRef;
     pub use shapely::newtype_copy;
     pub use shapely::shared;
--- a/src/rust/lib/logger/src/disabled.rs
+++ b/src/rust/lib/logger/src/disabled.rs
@@ -1,0 +1,44 @@
+use crate::LogMsg;
+use enso_prelude::*;
+use shapely::CloneRef;
+use std::fmt::Debug;
+
+
+
+// ==============
+// === Logger ===
+// ==============
+
+#[derive(Clone,CloneRef,Debug,Default)]
+pub struct Logger {
+    pub path: Rc<String>,
+}
+
+#[allow(dead_code)]
+impl Logger {
+    pub fn new<T:Str>(path:T) -> Self {
+        Self {path:Rc::new(path.into())}
+    }
+
+    pub fn sub<T:Str>(&self, path:T) -> Self {
+        if self.path.is_empty() {
+            Self::new(path)
+        } else {
+            Self::new(format!("{}.{}", self.path, path.as_ref()))
+        }
+    }
+
+    pub fn group<M:LogMsg, T, F:FnOnce() -> T>(&self, _msg:M, f:F) -> T {
+        f()
+    }
+}
+
+impl Logger {
+    pub fn trace      <M:LogMsg>(&self, _msg:M) {}
+    pub fn debug      <M:LogMsg>(&self, _msg:M) {}
+    pub fn info       <M:LogMsg>(&self, _msg:M) {}
+    pub fn warning    <M:LogMsg>(&self, _msg:M) {}
+    pub fn error      <M:LogMsg>(&self, _msg:M) {}
+    pub fn group_begin<M:LogMsg>(&self, _msg:M) {}
+    pub fn group_end            (&self        ) {}
+}
--- a/src/rust/lib/logger/src/lib.rs
+++ b/src/rust/lib/logger/src/lib.rs
@@ -4,13 +4,10 @@
 #![warn(missing_copy_implementations)]
 #![warn(missing_debug_implementations)]
 
-use enso_prelude::*;
-use shapely::CloneRef;
-use std::fmt::Debug;
-use wasm_bindgen::JsValue;
+pub mod disabled;
+pub mod enabled;
 
-#[cfg(target_arch = "wasm32")]
-use web_sys::console;
+use enso_prelude::*;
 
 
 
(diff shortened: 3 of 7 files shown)


src/rust/ensogl/src/display/camera/camera2d.rs
```diff
@@ -322,7 +322,7 @@ impl Camera2d {
     /// Creates new Camera instance.
     pub fn new(logger:&Logger, width:f32, height:f32) -> Self {
         let logger         = logger.sub("camera");
-        let display_object = display::object::Instance::new(&logger);
+        let display_object = display::object::Instance::new(logger.clone());
```
mwu-tow, line 325: We want to keep ability of passing `&logger`

src/rust/ensogl/src/display/object/class.rs
```diff
@@ -162,8 +162,8 @@ pub struct NodeData {
 }
 
 impl NodeData {
-    pub fn new<L:Into<Logger>>(logger:L) -> Self {
-        let logger           = logger.into();
+    pub fn new(logger:Logger) -> Self {
```
mwu-tow, line 165: We don't want callers to make conversions, this should stay generic.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -539,9 +539,11 @@ pub struct Callbacks {
 // === Renderer ===
 // ================
 
+pub type RendererLogger = disabled::Logger;
+
 #[derive(Clone,CloneRef,Debug)]
 pub struct Renderer {
-    logger    : Logger,
+    logger    : RendererLogger,
```
mwu-tow, line 546: Use `Logger` when storing fields.
`RendererLogger` on the other hand may be defined somewhere upper in the crate, so we have a place where we can easily toggle different logger types behavior.

src/rust/ensogl/src/display/scene.rs
```diff
@@ -551,7 +553,14 @@ pub struct Renderer {
...
 impl Renderer {
-    pub fn new(logger:&Logger, dom:&Dom, context:&Context, variables:&UniformScope) -> Self {
+
+    pub fn new
+    (logger:impl Into<RendererLogger>, dom:&Dom, context:&Context, variables:&UniformScope) -> Self {
+        Self::new_impl(logger.into(),dom,context,variables)
+    }
+
+    /// Separate implementation of `new` to avoid function body duplication after logger monorphization.
+    fn new_impl(logger:RendererLogger, dom:&Dom, context:&Context, variables:&UniformScope) -> Self {
```
mwu-tow, line 563: By default We should not worry about code size of generics.

Unless this split actaully does make any difference. I doubt, but check the actual wasm size after minimization and if there is visible difference, explain this in comment.

src/rust/lib/logger/src/lib.rs
```diff
@@ -35,106 +32,39 @@ impl<F: Fn() -> S, S:Str> LogMsg for F {
...
-
-
-// ====================
-// === Logger Utils ===
-// ====================
+pub trait LoggerApi {
+    /// Creates a new logger. Path should be a unique identifier of this logger.
+    fn new<T:Str>(path:T) -> Self;
+    /// Creates a new logger with this logger as a parent.
+    fn sub<T:Str>(&self, path:T) -> Self;
```
mwu-tow, line 44: This should be a static method.

src/rust/ensogl/src/lib.rs
```diff
@@ -65,6 +65,8 @@ pub use enso_frp as frp;
 pub mod prelude {
     pub use enso_prelude::*;
     pub use logger::*;
+    pub use logger::LoggerApi;
+    pub use logger::enabled::Logger;
```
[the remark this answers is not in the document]
Josef-Vonasek: Do we? Or is renderer enough.

src/rust/lib/logger/src/disabled.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    fn new(path:impl Str) -> Self {
+        Self {path:Rc::new(path.into())}
+    }
+
+    fn trace      <M: LogMsg>(&self, _:M){}
+    fn debug      <M: LogMsg>(&self, _:M){}
+    fn info       <M: LogMsg>(&self, _:M){}
+    fn warning    <M: LogMsg>(&self, _:M){}
+    fn error      <M: LogMsg>(&self, _:M){}
```
mwu-tow, line 51: We want these two to log (as if enabled) -- the disabled parser after all should be "disabled for anything below warning"

src/rust/lib/logger/src/lib.rs
```diff
@@ -35,112 +39,61 @@ impl<F: Fn() -> S, S:Str> LogMsg for F {
...
 
-    pub fn sub<T:Str>(&self, path: T) -> Self {
-        if self.path.is_empty() {
-            Self::new(path)
-        } else {
-            Self::new(format!("{}.{}", self.path, path.as_ref()))
-        }
+    /// Creates a new logger with this logger as a parent.
+    fn sub(logger:&impl AnyLogger, path:impl Str) -> Self where Self:Sized {
+        Self::new(format!("{}.{}", logger.path(), path.as_ref()))
```
mwu-tow, line 57: This loses the piece of logic that if previous logger was empty (e.g. was `default`) its sublogger behaves as if root, rather than having a leading dot in path

src/rust/ensogl/src/display/scene.rs
```diff
@@ -779,20 +779,21 @@ impl SceneData {
 
         let display_object = display::object::Instance::new(&logger);
         let context        = web::get_webgl2_context(&dom.layers.canvas);
-        let sub_logger     = logger.sub("shape_dirty");
+        let sub_logger     = Logger::sub(&logger,"shape_dirty").into();
         let shape_dirty    = ShapeDirty::new(sub_logger,Box::new(on_mut.clone()));
-        let sub_logger     = logger.sub("symbols_dirty");
+        let sub_logger     = Logger::sub(&logger,"symbols_dirty").into();
```
mwu-tow, line 784: We should get rid of Logger::sub(…).`into` calls in this function and rather have the `new` function accept `Into<Logger>`

PR #509, opened 2020-06-01
Fixing the blur on zoom bug

wdanilo:
### Pull Request Description
- Implements [Mouse gets blurry after zoom out.](https://github.com/luna/ide/issues/467)

### Checklist


src/rust/ensogl/src/display/symbol/gpu.rs
```diff
@@ -435,12 +433,9 @@ impl Symbol {
...
                 match binding.scope.as_ref() {
-                    Some(ScopeType::Mesh(s)) =>
-                        this.init_attribute_binding(program,binding,*s),
-                    Some(_) =>
-                        this.init_uniform_binding(program,binding,&mut texture_unit_iter),
-                    None =>
-                        {}
+                    Some(ScopeType::Mesh(s)) => this.init_attribute_binding(program,binding,*s),
+                    Some(_) => this.init_uniform_binding(program,binding,&mut texture_unit_iter),
+                    None    => {}
```
MichaelMauderer, line 438: Alignment.

src/rust/ensogl/src/display/symbol/gpu/registry.rs
```diff
@@ -38,6 +38,7 @@ pub struct SymbolRegistry {
     symbol_dirty    : SymbolDirty,
     logger          : Logger,
     view_projection : Uniform<Matrix4<f32>>,
+    z_zoom_1        : Uniform<f32>,
```
MichaelMauderer, line 41: It might be useful to re-iterate part of the doc from the `Camera2D` about what the `z_zoom_1` value does and why it is called that: ("The `z_zoom_1` value describes the z-axis distance at which the `zoom` value is `1.0`"). My first impression was that it should just be called `z_zoom`, but with the description, the name makes more sense.

src/rust/ensogl/src/display/symbol/gpu.rs
wdanilo: The aligment was broken only not to use multiline expressions here.

src/rust/ensogl/src/display/symbol/gpu/registry.rs
wdanilo: Added comment about it. I think after the release I would need to spend few days refactoring this part as its getting messy slowly

PR #525, opened 2020-06-03
Implement segmented output ports.

MichaelMauderer:
### Pull Request Description
Adds multiple output ports to nodes that are highlighted on hover.

Peek 2020-06-15 10-26

### Important notes
* hack: there is not yet a good way to manage the order of shapes. To compensate for this a method that initializes the shapes is provided and called at the right place to ensure the correct order of shapes. The correct management needs to be still implemented once shape order management is available.

### Checklist

### DIFF
--- a/src/rust/ensogl/src/display/shape/primitive/glsl/shape.glsl
+++ b/src/rust/ensogl/src/display/shape/primitive/glsl/shape.glsl
@@ -84,7 +84,9 @@ Sdf difference (Sdf a, Sdf b) {
     return intersection(a,inverse(b));
 }
 
-
+Sdf grow (Sdf a, float size) {
+    return Sdf(a.distance - size);
+}
 
 // ================
 // === BoundSdf ===
@@ -133,6 +135,11 @@ BoundSdf resample (BoundSdf a, float multiplier) {
 
 BoundSdf pixel_snap (BoundSdf a) {
     a.distance = floor(a.distance) + 0.5;
+    return a;
+}
+
+BoundSdf grow (BoundSdf a, float size) {
+    a.distance = a.distance - size;
     return a;
 }
 
@@ -239,6 +246,14 @@ Shape resample (Shape s, float multiplier) {
 Shape pixel_snap (Shape s) {
     Id       id    = s.id;
     BoundSdf sdf   = pixel_snap(s.sdf);
+    Srgba    color = unpremultiply(s.color);
+    color.raw.a /= s.alpha;
+    return shape(id,sdf,color);
+}
+
+Shape grow (Shape s, float value) {
+    Id       id    = s.id;
+    BoundSdf sdf   = grow(s.sdf,value);
     Srgba    color = unpremultiply(s.color);
     color.raw.a /= s.alpha;
     return shape(id,sdf,color);
--- a/src/rust/ensogl/src/display/shape/primitive/glsl/shape.glsl
+++ b/src/rust/ensogl/src/display/shape/primitive/glsl/shape.glsl
@@ -140,6 +140,7 @@ BoundSdf pixel_snap (BoundSdf a) {
 
 BoundSdf grow (BoundSdf a, float size) {
     a.distance = a.distance - size;
+    a.bounds   = grow(a.bounds,size);
     return a;
 }
 
--- a/src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
+++ b/src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
@@ -292,6 +292,18 @@ impl Canvas {
             shape
         })
     }
+
+    /// Shrink the shape by the given value.
+    pub fn shrink<T:Into<Var<f32>>>
+    (&mut self, num:usize, s:Shape, value:T) -> Shape {
+        self.if_not_defined(num, |this| {
+            let value:Glsl = value.into().glsl();
+            let expr = iformat!("return grow({s.getter()},-{value});");
+            let mut shape = this.new_shape_from_expr(num,&expr);
+            shape.add_ids(&s.ids);
+            shape
+        })
+    }
 }
 
 
--- a/src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
+++ b/src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
@@ -280,6 +280,18 @@ impl Canvas {
             shape
         })
     }
+
+    /// Grow the shape by the given value.
+    pub fn grow<T:Into<Var<f32>>>
+    (&mut self, num:usize, s:Shape, value:T) -> Shape {
+        self.if_not_defined(num, |this| {
+            let value:Glsl = value.into().glsl();
+            let expr = iformat!("return grow({s.getter()},{value});");
+            let mut shape = this.new_shape_from_expr(num,&expr);
+            shape.add_ids(&s.ids);
+            shape
+        })
+    }
 }
 
 
--- a/src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
+++ b/src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
@@ -263,7 +263,7 @@ impl Canvas {
             let color:Glsl = color.into().glsl();
             this.add_current_function_code_line(iformat!("Shape shape = {s.getter()};"));
             this.add_current_function_code_line(iformat!("Srgba color = srgba({color});"));
-            let expr = iformat!("return set_color(shape,color);");
+            let expr      = iformat!("return set_color(shape,color);");
             let mut shape = this.new_shape_from_expr(num,&expr);
             shape.add_ids(&s.ids);
             shape
@@ -274,7 +274,7 @@ impl Canvas {
     pub fn pixel_snap
     (&mut self, num:usize, s:Shape) -> Shape {
         self.if_not_defined(num, |this| {
-            let expr = iformat!("return pixel_snap({s.getter()});");
+            let expr      = iformat!("return pixel_snap({s.getter()});");
             let mut shape = this.new_shape_from_expr(num,&expr);
             shape.add_ids(&s.ids);
             shape
@@ -286,8 +286,8 @@ impl Canvas {
     (&mut self, num:usize, s:Shape, value:T) -> Shape {
         self.if_not_defined(num, |this| {
             let value:Glsl = value.into().glsl();
-            let expr = iformat!("return grow({s.getter()},{value});");
-            let mut shape = this.new_shape_from_expr(num,&expr);
+            let expr       = iformat!("return grow({s.getter()},{value});");
+            let mut shape  = this.new_shape_from_expr(num,&expr);
             shape.add_ids(&s.ids);
             shape
         })
@@ -296,13 +296,8 @@ impl Canvas {
     /// Shrink the shape by the given value.
     pub fn shrink<T:Into<Var<f32>>>
     (&mut self, num:usize, s:Shape, value:T) -> Shape {
-        self.if_not_defined(num, |this| {
-            let value:Glsl = value.into().glsl();
-            let expr = iformat!("return grow({s.getter()},-{value});");
-            let mut shape = this.new_shape_from_expr(num,&expr);
-            shape.add_ids(&s.ids);
-            shape
-        })
+        let value = value.into();
+        self.grow(num, s, -value)
     }
 }
 
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -24,7 +24,7 @@ use ensogl::gui::component;
 
 use super::edge;
 use crate::component::visualization;
-use crate::component::node::port::output::OutPutPorts;
+use crate::component::node::port::output::OutputPorts;
 
 
 pub const NODE_SHAPE_PADDING : f32 = 40.0;
@@ -215,7 +215,7 @@ impl NodeModel {
         let logger  = Logger::new("node");
         edge::sort_hack_1(scene);
 
-        OutPutPorts::order_hack(&scene);
+        OutputPorts::order_hack(&scene);
         let main_logger = Logger::sub(&logger,"main_area");
         let drag_logger = Logger::sub(&logger,"drag_area");
         let main_area   = component::ShapeView::<shape      ::Shape>::new(&main_logger  ,scene);
@@ -283,8 +283,8 @@ impl NodeModel {
         let height = self.height();
 
         let size = Vector2::new(width+NODE_SHAPE_PADDING*2.0, height+NODE_SHAPE_PADDING*2.0);
-        self.main_area.shape.sprite.size().set(size);
-        self.drag_area.shape.sprite.size().set(size);
+        self.main_area.shape.sprite.size.set(size);
+        self.drag_area.shape.sprite.size.set(size);
         self.main_area.mod_position(|t| t.x = width/2.0);
         self.main_area.mod_position(|t| t.y = height/2.0);
         self.drag_area.mod_position(|t| t.x = width/2.0);
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -288,7 +288,7 @@ impl NodeModel {
         self.drag_area.mod_position(|t| t.x = width/2.0);
         self.drag_area.mod_position(|t| t.y = height/2.0);
 
-        self.output_ports.frp.set_size.emit(Some(size));
+        self.output_ports.frp.set_size.emit(V2::from(size));
         self.output_ports.mod_position(|t| t.x = width/2.0);
         self.output_ports.mod_position(|t| t.y = height/2.0);
     }
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -297,7 +297,7 @@ impl NodeModel {
         self.drag_area.mod_position(|t| t.x = width/2.0);
         self.drag_area.mod_position(|t| t.y = height/2.0);
 
-        self.output_ports.frp.set_size.emit(V2::from(size));
+        self.output_ports.frp.set_size.emit(size);
         self.output_ports.mod_position(|t| t.x = width/2.0);
         self.output_ports.mod_position(|t| t.y = height/2.0);
     }
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -94,7 +94,7 @@ pub mod shape {
             let select         = select2 - select;
             let select         = select.fill(color::Rgba::from(selection_color));
 
-            let out = select + shape + shadow;
+            let out = select + shadow + shape;
             out.into()
         }
     }
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -215,7 +215,7 @@ impl NodeModel {
         let logger  = Logger::new("node");
         edge::sort_hack_1(scene);
 
-        OutPutPorts::init_shape_order_hack(&scene);
+        OutPutPorts::order_hack(&scene);
         let main_logger = Logger::sub(&logger,"main_area");
         let drag_logger = Logger::sub(&logger,"drag_area");
         let main_area   = component::ShapeView::<shape      ::Shape>::new(&main_logger  ,scene);
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -27,8 +27,8 @@ use crate::component::visualization;
 use crate::component::node::port::output::OutPutPorts;
 
 
-pub(crate) const NODE_SHAPE_PADDING : f32 = 40.0;
-pub(crate) const NODE_SHAPE_RADIUS  : f32 = 14.0;
+pub const NODE_SHAPE_PADDING : f32 = 40.0;
+pub const NODE_SHAPE_RADIUS  : f32 = 14.0;
 
 
 // ============
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -87,7 +87,7 @@ pub mod shape {
             let select         = select2 - select;
             let select         = select.fill(color::Rgba::from(selection_color));
 
-            let out = select + shape;
+            let out = select + shape + shadow;
             out.into()
         }
     }
@@ -254,7 +254,7 @@ impl NodeModel {
 
 
         // TODO: Determine number of output ports based on node semantics.
-        let output_ports = OutputPorts::new(&scene, 4);
+        let output_ports = OutputPorts::new(&scene, 1);
         display_object.add_child(&output_ports);
 
 
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -9,8 +9,8 @@ pub use port::Expression;
 
 use crate::prelude::*;
 
-use enso_frp;
 use enso_frp as frp;
+use enso_frp;
 use ensogl::data::color;
 use ensogl::display::Attribute;
 use ensogl::display::Buffer;
@@ -21,14 +21,21 @@ use ensogl::display::traits::*;
 use ensogl::display;
 use ensogl::gui::component::Animation;
 use ensogl::gui::component;
+use std::num::NonZeroU32;
 
 use super::edge;
 use crate::component::visualization;
 use crate::component::node::port::output::OutputPorts;
 
 
+
+// =================
+// === Constants ===
+// =================
+
 pub const NODE_SHAPE_PADDING : f32 = 40.0;
 pub const NODE_SHAPE_RADIUS  : f32 = 14.0;
+
 
 
 // ============
@@ -254,7 +261,7 @@ impl NodeModel {
 
 
         // TODO: Determine number of output ports based on node semantics.
-        let output_ports = OutputPorts::new(&scene, 20);
+        let output_ports = OutputPorts::new(&scene, NonZeroU32::new(10).unwrap());
         display_object.add_child(&output_ports);
 
 
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -254,7 +254,7 @@ impl NodeModel {
 
 
         // TODO: Determine number of output ports based on node semantics.
-        let output_ports = OutputPorts::new(&scene, 1);
+        let output_ports = OutputPorts::new(&scene, 20);
         display_object.add_child(&output_ports);
 
 
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -254,7 +254,7 @@ impl NodeModel {
 
 
         // TODO: Determine number of output ports based on node semantics.
-        let output_ports = OutputPorts::new(&scene, 10);
+        let output_ports = OutputPorts::new(&scene, 4);
         display_object.add_child(&output_ports);
 
 
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -28,6 +28,7 @@ use crate::component::node::port::output::OutPutPorts;
 
 
 pub(crate) const NODE_SHAPE_PADDING : f32 = 40.0;
+pub(crate) const NODE_SHAPE_RADIUS  : f32 = 14.0;
 
 
 // ============
@@ -50,7 +51,7 @@ pub mod shape {
             let height : Var<Distance<Pixels>> = "input_size.y".into();
             let width  = width  - NODE_SHAPE_PADDING.px() * 2.0;
             let height = height - NODE_SHAPE_PADDING.px() * 2.0;
-            let radius = 14.px();
+            let radius = NODE_SHAPE_RADIUS.px();
             let shape  = Rect((&width,&height)).corners_radius(radius);
             let shape  = shape.fill(color::Rgba::from(bg_color));
 
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -253,6 +253,7 @@ impl NodeModel {
         let frp = Frp{input};
 
 
+        // TODO: Determine number of output ports based on node semantics.
         let output_ports = OutputPorts::new(&scene, 5);
         display_object.add_child(&output_ports);
 
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -87,7 +87,7 @@ pub mod shape {
             let select         = select2 - select;
             let select         = select.fill(color::Rgba::from(selection_color));
 
-            let out = select + shadow + shape;
+            let out = select + shape;
             out.into()
         }
     }
--- a/src/rust/lib/graph-editor/src/component/node.rs
+++ b/src/rust/lib/graph-editor/src/component/node.rs
@@ -254,7 +254,7 @@ impl NodeModel {
 
 
         // TODO: Determine number of output ports based on node semantics.
-        let output_ports = OutputPorts::new(&scene, 5);
+        let output_ports = OutputPorts::new(&scene, 10);
         display_object.add_child(&output_ports);
 
 
--- a/src/rust/lib/graph-editor/src/component/node/port/output.rs
+++ b/src/rust/lib/graph-editor/src/component/node/port/output.rs
@@ -410,6 +410,24 @@ impl display::Object for ShapeView {
 
 
 
+// ===============
+// === PortId  ===
+// ===============
+
+/// Id of a specific port inside of `OutPutPortsData`.
+#[derive(Clone,Copy,Default,Debug,Eq,PartialEq)]
+pub struct PortId {
+    index: usize,
+}
+
+impl PortId {
+    fn new(index:usize) -> Self {
+        Self{index}
+    }
+}
+
+
+
 // =================
 // === Port Frp  ===
 // =================
@@ -484,18 +502,6 @@ fn init_port_frp<Shape: PortShape +CloneRef+'static>
 // ===========
 // === Frp ===
 // ===========
-
-/// Id of a specific port inside of `OutPutPortsData`.
-#[derive(Clone,Copy,Default,Debug,Eq,PartialEq)]
-pub struct PortId {
-    index: usize,
-}
-
-impl PortId {
-    fn new(index:usize) -> Self {
-        Self{index}
-    }
-}
 
 /// Frp API of the `OutPutPorts`.
 #[derive(Clone,CloneRef,Debug)]
--- a/src/rust/lib/graph-editor/src/component/node/port/output.rs
+++ b/src/rust/lib/graph-editor/src/component/node/port/output.rs
@@ -15,6 +15,7 @@ use ensogl::gui::component::Animation;
 use ensogl::gui::component;
 
 use crate::node::NODE_SHAPE_PADDING;
+use crate::node::NODE_SHAPE_RADIUS;
 
 
 
@@ -53,7 +54,7 @@ pub mod port_area {
             let hover_area        = hover_area.fill(color::Rgba::new(0.0,0.0,0.0,0.000_001));
 
             let shrink           = 1.px() - 1.px() * &grow;
-            let radius           = 14.px();
+            let radius           = NODE_SHAPE_RADIUS.px();
             let port_area_size   = 4.0.px() * &grow;
             let port_area_width  = &width  + (&port_area_size - &shrink) * 2.0;
             let port_area_height = &height + (&port_area_size - &shrink) * 2.0;
--- a/src/rust/lib/graph-editor/src/component/node/port/output.rs
+++ b/src/rust/lib/graph-editor/src/component/node/port/output.rs
@@ -19,6 +19,7 @@ use ensogl::math::algebra::Clamp;
 use crate::node;
 
 
+
 // =================
 // === Constants ===
 // =================
@@ -31,15 +32,37 @@ const SEGMENT_GAP_WIDTH   : f32 = 2.0;
 const SHOW_DELAY_DURATION : f32 = 150.0;
 const HIDE_DELAY_DURATION : f32 = 150.0;
 
-const PORT_AREA_WIDTH     : f32  = 4.0;
+
 
 // ==============
 // === Shapes ===
 // ==============
 
-/// The port area shape is based on a single shape that gets offset to show a different slice for
-/// each segment. Each shapes represents a window of the underlying shape.
-pub mod port_area {
+/// Helper trait that allows us to abstract the API of the `multi_port_area::Shape` and the
+/// `single_port_area::Shape`. This is needed to avoid code duplication for functionality that can
+/// work with either shape.
+#[allow(missing_docs)]
+trait PortShapeApi {
+    fn set_grow(&self, grow_value:f32);
+    fn set_opacity(&self, opacity:f32);
+}
+
+/// The port area shape is based on a single shape that gets cropped to show the appropriate
+/// segment.
+///
+/// The base shape looks roughly like this:
+/// ```text
+///      .                            .
+///      .                            .
+///       .                          .
+///          .                    .
+///              """""""""""""
+///     |  r    |              |   r   |
+///     |            width             |
+/// ```
+/// where r is the radius of the left and right quarter circle shapes.
+///
+pub mod multi_port_area {
     use super::*;
     use ensogl::display::shape::*;
     use std::f32::consts::PI;
@@ -54,59 +77,63 @@ pub mod port_area {
         Var::<f32>::from(format!("(step(float({1}),float({0})) * step(float({0}),float({2})))", value, lower_bound, upper_bound))
     }
 
+    /// Compute the rotation of the plane along the shape border. The plane needs to be rotated
+    /// to be perpendicular with the outer shape border. That means, it needs to be rotate in the
+    /// segments that have a curved path, and needs to be perpendicular in the inner segment.
     fn compute_crop_plane_angle(full_shape_border_length:&Var<f32>, corner_segment_length:&Var<f32>, position:&Var<f32>) ->Var<f32> {
-        // TODO implement proper abstraction for non-branching "if/then/else" or "case" ins shaderland
+        // TODO implement proper abstraction for non-branching "if/then/else" or "case" in shaderland
+        // Here we use a trick to use a pseudo-boolean float that is either 0 or 1 to multiply a
+        // value that should be returned, iff it's case is true. That way we can add return values
+        // of different "branches" of which exactly one will be non-zero.
+
         let start                      = 0.0.into();
         let middle_segment_start_point = corner_segment_length;
         let middle_segment_end_point   = full_shape_border_length - corner_segment_length;
         let end                        = full_shape_border_length;
 
-        let base_rotation = Var::<f32>::from(90.0_f32.to_radians());
+        let default_rotation           = Var::<f32>::from(90.0_f32.to_radians());
 
-        // Case 1:
-        let case_1            = in_range(position, &start, &middle_segment_start_point);
-        let case_1_value_base = Var::<f32>::from(1.0) - (position / corner_segment_length).clamp(0.0.into(), 1.0.into());
-        let case_1_scale      = 90.0_f32.to_radians();
-        let case_1_value      = case_1 * (case_1_value_base * case_1_scale + &base_rotation);
+        let case_1_pseudo_bool      = in_range(position, &start, &middle_segment_start_point);
+        let case_1_value_normalised = Var::<f32>::from(1.0) - (position / corner_segment_length).clamp(0.0.into(), 1.0.into());
+        let case_1_scale            = 90.0_f32.to_radians();
+        let case_1_value            = case_1_pseudo_bool * (case_1_value_normalised * case_1_scale + &default_rotation);
 
-        // Case 2
-        let case_2            = in_range_inclusive(position, &middle_segment_start_point, &middle_segment_end_point);
-        let case_2_value_base = &base_rotation;
-        let case_2_value      = case_2 * case_2_value_base;
+        let case_2_pseudo_bool = in_range_inclusive(position, &middle_segment_start_point, &middle_segment_end_point);
+        let case_2_value_base  = &default_rotation;
+        let case_2_value       = case_2_pseudo_bool * case_2_value_base;
 
         // Case 3
-        let case_3            = in_range_inclusive(position, &middle_segment_end_point, &end);
-        let case_3_value_base = ((position - middle_segment_end_point) / corner_segment_length).clamp(0.0.into(), 1.0.into());
-        let case_3_scale      = -90.0_f32.to_radians();
-        let case_3_value      = case_3 * (case_3_value_base * case_3_scale + &base_rotation);
+        let case_3_pseudo_bool      = in_range_inclusive(position, &middle_segment_end_point, &end);
+        let case_3_value_normalised = ((position - middle_segment_end_point) / corner_segment_length).clamp(0.0.into(), 1.0.into());
+        let case_3_scale            = (-90.0_f32).to_radians();
+        let case_3_value            = case_3_pseudo_bool * (case_3_value_normalised * case_3_scale + &default_rotation);
 
-        (case_1_value + case_2_value + case_3_value).into()
+        case_1_value + case_2_value + case_3_value
     }
 
     /// Returns a value between 0 and 1 that indicates the position along the straight center segment.
     fn calculate_crop_plane_position_relative_to_center_segment(full_shape_border_length:&Var<f32>, corner_segment_length:&Var<f32>, position:&Var<f32>) -> Var<f32> {
-        // let start                      = 0.0.into();
+        // TODO implement proper abstraction for non-branching "if/then/else" or "case" in shaderland
+        // See function above for explanation of the branching.
+
         let middle_segment_start_point = corner_segment_length;
         let middle_segment_end_point   = full_shape_border_length - corner_segment_length;
         let end                        = full_shape_border_length;
 
-        let one  = Var::<f32>::from(1.0);
-
         // Case 1: always zero can be ignored
 
-        // Case 2
-        let case_2            = in_range_inclusive(position, &middle_segment_start_point, &middle_segment_end_point);
-        let case_2_value_base = (position - middle_segment_start_point) / (&middle_segment_end_point - middle_segment_start_point);
-        let case_2_value      = case_2 * case_2_value_base;
+        let case_2_pseudo_bool      = in_range_inclusive(position, &middle_segment_start_point, &middle_segment_end_point);
+        let case_2_value_normalised = (position - middle_segment_start_point) / (&middle_segment_end_point - middle_segment_start_point);
+        let case_2_value            = case_2_pseudo_bool * case_2_value_normalised;
 
-        // Case 3: always 1.0
-        let case_3            = in_range_inclusive(position, &middle_segment_end_point, &end);
-        let case_3_value      = case_3 * one;
+        let case_3_pseudo_bool = in_range_inclusive(position, &middle_segment_end_point, &end);
+        let case_3_value       = case_3_pseudo_bool * 1.0;
 
-        (case_2_value + case_3_value).into()
+        case_2_value + case_3_value
 
     }
 
+    /// Compute the plane at the location of the given port index.
     fn compute_crop_plane(index:&Var<f32>, port_num: &Var<f32>, width: &Var<f32>, corner_radius:&Var<f32>, position_offset:&Var<f32>) -> AnyShape {
         let corner_circumference     = corner_radius * 2.0 * PI;
         let corner_segment_length    = &corner_circumference * 0.25;
@@ -114,15 +141,15 @@ pub mod port_area {
         let full_shape_border_length = &center_segment_length + &corner_segment_length * 2.0;
 
 
-        let position_relative    = (index / port_num);
-        let crop_segment_pos = &position_relative * &full_shape_border_length + position_offset;
+        let position_relative = index / port_num;
+        let crop_segment_pos  = &position_relative * &full_shape_border_length + position_offset;
 
         let crop_plane_pos_relative = calculate_crop_plane_position_relative_to_center_segment(&full_shape_border_length, &corner_segment_length, &crop_segment_pos);
         let crop_plane_pos          = crop_plane_pos_relative * &center_segment_length + corner_radius;
 
-        let plane_rotation_angle  = compute_crop_plane_angle(&full_shape_border_length, &corner_segment_length, &crop_segment_pos);
-        let plane_shape_offset     = Var::<Distance<Pixels>>::from(&crop_plane_pos - width * 0.5);
-        let crop_shape       = HalfPlane().rotate(plane_rotation_angle).translate_x(plane_shape_offset);
+        let plane_rotation_angle = compute_crop_plane_angle(&full_shape_border_length, &corner_segment_length, &crop_segment_pos);
+        let plane_shape_offset   = Var::<Distance<Pixels>>::from(&crop_plane_pos - width * 0.5);
+        let crop_shape           = HalfPlane().rotate(plane_rotation_angle).translate_x(plane_shape_offset);
         crop_shape.into()
     }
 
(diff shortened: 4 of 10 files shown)


wdanilo:
As a fast first review,  few things to fix based on the gif:
1. When the ports appear, you should already be able to click . So as soon as they appear one of the ports should be selected. There is no way in showing output ports if they are not clicable
2. There is no animation regarding showing all the ports. On master now, whe nmouse apporaches the node, the ports are nicely going from under the node. On your gif they appear instantly

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -327,13 +275,14 @@ impl NodeModel {
...
         self.drag_area.shape.sprite.size().set(size);
-        self.output_area.shape.sprite.size().set(size);
         self.main_area.mod_position(|t| t.x = width/2.0);
         self.main_area.mod_position(|t| t.y = height/2.0);
         self.drag_area.mod_position(|t| t.x = width/2.0);
         self.drag_area.mod_position(|t| t.y = height/2.0);
-        self.output_area.mod_position(|t| t.x = width/2.0);
-        self.output_area.mod_position(|t| t.y = height/2.0);
+
+        self.output_ports.frp.set_size.emit(Some(size));
```
wdanilo, line 283: why Some(size) and not just size?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+
+
+
+// =================
+// === Constants ===
+// =================
+
+const BASE_SIZE:f32         = 0.5;
+const HIGHLIGHT_SIZE:f32    = 1.0;
+const SEGMENT_GAP_WIDTH:f32 = 2.0;
```
wdanilo, line 26: please align also `:`  here :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+
+            let hover_area_size   = 20.0.px();
+            let hover_area_width  = &width  + &hover_area_size * 2.0;
+            let hover_area_height = &height / 2.0 + &hover_area_size;
+            let hover_area        = Rect((&hover_area_width,&hover_area_height));
+            let hover_area        = hover_area.translate_y(-hover_area_height/2.0);
+            let hover_area        = hover_area.fill(color::Rgba::new(0.0,0.0,0.0,0.000_001));
+
+            let shrink           = 1.px() - 1.px() * &grow;
+            let radius           = 14.px();
```
wdanilo, line 55: this is node radius. Should use the right constants here.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+            let port_area_cropped = crop_window.intersection(port_area_aligned);
+
+            let color             = Var::<color::Rgba>::from("srgba(0.25,0.58,0.91,input_opacity)");
+            let port_area_colored = port_area_cropped.fill(color);
+
+            (port_area_colored + hover_area).into()
+        }
+    }
+}
+
```
wdanilo, line 90: 2 more nlns :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+// ===========
+// === Frp ===
+// ===========
+
+
```
wdanilo, line 95: rm nl

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+type PortId = usize;
+
+#[derive(Clone,CloneRef,Debug)]
+pub struct Frp {
+    pub set_size        : frp::Source<Option<Vector2<f32>>>,
```
wdanilo, line 100: use `V2` instead of `Vector2`. We are moving to `V2` as it has default impl defined. It is defined in math/algebra.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+                    });
+
+                    eval view.events.mouse_out ([set_port_sizes_all,set_port_opacity_all](_) {
+                         set_port_sizes_all.emit(0.0);
+                         set_port_opacity_all.emit(0.0);
+                    });
+
+                    eval view.events.mouse_down ([frp](_) {
+                        frp.on_port_mouse_down.emit(index);
+                    });
```
wdanilo, line 282: `eval` uses `f!`. `f!` works this way that if you skip the `[..]` part, then the first word before dot will be cloned. Here you are cloning `frp`, which is the first word before dot. Just skip the `[...]` part :) Moreover, use `eval_` and skip the `(_)` part as well.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+                        port_size.set_target_value(*size)
+                    }));
+
+                    is_opacity_target <- set_port_opacity.map(move |(id, _)| *id==index);
+                    opacity_change    <- set_port_opacity.gate(&is_opacity_target);
+                    _change_size      <- opacity_change.map(f!(((_, opacity))  {
+                        port_opacity.set_target_value(*opacity)
+                    }));
+
+                    eval view.events.mouse_over ([port_size,set_port_sizes_all,port_opacity,set_port_opacity_all](_) {
```
wdanilo, line 268: too long

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+
+                    is_resize_target <- set_port_size.map(move |(id, _)| *id==index);
+                    size_change      <- set_port_size.gate(&is_resize_target);
+                    _change_size     <- size_change.map(f!(((_, size))  {
+                        port_size.set_target_value(*size)
+                    }));
+
+                    is_opacity_target <- set_port_opacity.map(move |(id, _)| *id==index);
+                    opacity_change    <- set_port_opacity.gate(&is_opacity_target);
+                    _change_size      <- opacity_change.map(f!(((_, opacity))  {
```
wdanilo, line 264: eval

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+                let shape          = &view.shape;
+                let port_size    = Animation::<f32>::new(&network);
+                let port_opacity = Animation::<f32>::new(&network);
+                frp::extend! { network
+                     eval port_size.value ((size) shape.grow.set(*size));
+                     eval port_opacity.value   ((size) shape.opacity.set(*size));
+
+                    is_resize_target <- set_port_size.map(move |(id, _)| *id==index);
+                    size_change      <- set_port_size.gate(&is_resize_target);
+                    _change_size     <- size_change.map(f!(((_, size))  {
```
wdanilo, line 258: eval

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+
+        // Init ports
+        {
+            for (index,view) in data.borrow().ports.iter().enumerate() {
+                let shape          = &view.shape;
+                let port_size    = Animation::<f32>::new(&network);
+                let port_opacity = Animation::<f32>::new(&network);
+                frp::extend! { network
+                     eval port_size.value ((size) shape.grow.set(*size));
+                     eval port_opacity.value   ((size) shape.opacity.set(*size));
```
wdanilo, line 254: alignment

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+            def _set_port_opacities = set_port_opacity_all.map(f!([set_port_opacity,data](size) {
+                let port_num = data.borrow().ports.len();
+                for index in 0..port_num {
+                    set_port_opacity.emit((index,*size));
+                };
+            }));
+        }
+
+        // Init ports
+        {
```
wdanilo, line 247: create a separate function for that instead of such brackets. This is hacky and its hard to understand whats happening here. Moreover, you are borrowing value in the next line and it is borrowed for the next 50 lines. It is really hard to track down. You are in the initialization stage. This should be possible to be expressed without any borrow.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+
+// ===================
+// === OutPutPorts ===
+// ===================
+
+#[derive(Debug,Clone,CloneRef)]
+pub struct OutputPorts {
+    pub frp            : Frp,
+        network        : frp::Network,
+        data           : Rc<RefCell<OutPutPortsData>>,
```
wdanilo, line 185: instead of putting so many variables in RefCell, but only the needed ones. A lot of borrow ping pong below is needed because of this design. In `OutPutPortsData` you have a lot of fields which are already internally mutable, like `scene`, `logger`, or `display_object`. Putting them in another `RefCell` is bad design. Moreover, why do you have multiple `display_object` for the same thing? Again because of this design. Use `RefCells` on the smallest parts possible.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+        display_object : display::object::Instance,
+}
+
+impl OutputPorts {
+    pub fn new(scene:&Scene, number_of_ports:u8) -> Self {
+        let logger         = Logger::new("bubble");
```
wdanilo, line 191: this is not bubble!

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+#[derive(Debug,Clone,CloneRef)]
+pub struct OutputPorts {
+    pub frp            : Frp,
+        network        : frp::Network,
+        data           : Rc<RefCell<OutPutPortsData>>,
+        display_object : display::object::Instance,
+}
+
+impl OutputPorts {
+    pub fn new(scene:&Scene, number_of_ports:u8) -> Self {
```
wdanilo, line 190: why u8 here? Why not just `usize`?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+        let gap_width      = SEGMENT_GAP_WIDTH;
+
+        let mut ports      = Vec::default();
+        ports.resize_with(number_of_ports as usize,|| component::ShapeView::new(&logger,&scene));
+        let ports = ports;
+
+        let data =  OutPutPortsData {display_object:display_object.clone_ref(),scene,logger,size,
+                                     ports,gap_width}.init();
+        let data = Rc::new(RefCell::new(data));
+
```
wdanilo, line 208: A lot of things are badly aligned here. If you need to use such patterns as here, it is always much better to create a `OutPutPortsData::new` method and use it here instead.

src/rust/lib/graph-editor/src/component/node/port/output.rs
wdanilo, line 190: no docs. Please enable docs requirements on this module.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+                };
+            }));
+        }
+
+        // Init ports
+        {
+            for (index,view) in data.borrow().ports.iter().enumerate() {
+                let shape          = &view.shape;
+                let port_size    = Animation::<f32>::new(&network);
+                let port_opacity = Animation::<f32>::new(&network);
```
wdanilo, line 251: alignment

src/rust/lib/graph-editor/src/component/node/port/output.rs
wdanilo, line 264: there should be no spaces in cases like `(move |(id,_)| *id == index)`

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,300 @@
...
+                    });
+                }
+            }
+        }
+    }
+
+    // TODO: Implement proper sorting and remove.
+    /// Hack function used to register the elements for the sorting purposes. To be removed.
+    pub(crate) fn init_shape_order_hack(scene:&Scene) {
+        let logger = Logger::new("hack");
```
wdanilo, line 291: rename it pls to `order_hack`. I knwo we have it like this in other parts, but its needs to be fixed there as well

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -796,7 +796,7 @@ impl GraphEditorModelWithNetwork {
         frp::new_bridge_network! { [self.network, node.view.main_area.events.network]
             eval_ node.view.drag_area.events.mouse_down(touch.nodes.down.emit(node_id));
             eval  node.view.ports.frp.cursor_style ((style) cursor_style.emit(style));
-            eval_ node.view.frp.output_ports.mouse_down (output_press.emit(node_id));
+            eval  node.view.output_ports.frp.port_mouse_down ((_ix) output_press.emit(node_id));
```
wdanilo, line 799: use `eval_`

src/rust/lib/graph-editor/src/component/node.rs
MichaelMauderer: Because the FRP requires `Default` to be implemented and it didn't seem useful to implement a wrapper for this.

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: It doesn't really represent an platform-specific index here, so it should be one of the normal integer types. I've increased it to a u32.

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -24,10 +24,11 @@ use ensogl::gui::component;
 
 use super::edge;
 use crate::component::visualization;
+use crate::component::node::port::output::OutputPorts;
 
 
-
-const NODE_SHAPE_PADDING : f32 = 40.0;
+pub(crate) const NODE_SHAPE_PADDING : f32 = 40.0;
+pub(crate) const NODE_SHAPE_RADIUS  : f32 = 14.0;
```
wdanilo, line 31: just do it `pub`. There is no reason to make it `pub(crate)`.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+use ensogl::display::Sprite;
+use ensogl::display::scene::Scene;
+use ensogl::display::shape::*;
+use ensogl::display::traits::*;
+use ensogl::display;
+use ensogl::gui::component::Animation;
+use ensogl::gui::component;
+
+use crate::node::NODE_SHAPE_PADDING;
+use crate::node::NODE_SHAPE_RADIUS;
```
wdanilo, line 18: lets use as few imports as possible. These are good candidates to be used as `node::SHAPE_PADDING` instead :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+use crate::prelude::*;
+
+use enso_frp as frp;
+use enso_frp;
+use ensogl::data::color;
+use ensogl::display::Attribute;
+use ensogl::display::Buffer;
+use ensogl::display::Sprite;
+use ensogl::display::scene::Scene;
+use ensogl::display::shape::*;
```
wdanilo, line 11: This is a very wide import. Let's use `*` imports only for prelude, traits, and local modules.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+
+/// The port area shape is based on a single shape that gets offset to show a different slice for
+/// each segment. Each shapes represents a window of the underlying shape.
+pub mod port_area {
+    use super::*;
+
+    ensogl::define_shape_system! {
+        (style:Style, grow:f32, shape_width:f32, offset_x:f32, padding:f32, opacity:f32) {
+            let width  : Var<Distance<Pixels>> = shape_width.into();
+            let height : Var<Distance<Pixels>> = "input_size.y".into();
```
wdanilo, line 45: Why width is not computed the same way height is ?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+            let height = height - NODE_SHAPE_PADDING.px() * 2.0;
+
+            let hover_area_size   = 20.0.px();
+            let hover_area_width  = &width  + &hover_area_size * 2.0;
+            let hover_area_height = &height / 2.0 + &hover_area_size;
+            let hover_area        = Rect((&hover_area_width,&hover_area_height));
+            let hover_area        = hover_area.translate_y(-hover_area_height/2.0);
+            let hover_area        = hover_area.fill(color::Rgba::new(0.0,0.0,0.0,0.000_001));
+
+            let shrink           = 1.px() - 1.px() * &grow;
```
wdanilo, line 56: this var needs better name

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+            let radius           = NODE_SHAPE_RADIUS.px();
+            let port_area_size   = 4.0.px() * &grow;
```
wdanilo, line 58: what is this 4.0?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+
+
+// ===========
+// === Frp ===
+// ===========
+
+/// Id of a specific port inside of `OutPutPortsData`.
+type PortId = usize;
```
wdanilo, line 101: Lets use newtypes for that! :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+
+
+
+// =======================
+// === OutPutPortsData ===
+// =======================
+
+/// Internal data of the `OutPutPorts`.
+#[derive(Debug)]
+pub struct OutPutPortsData {
```
wdanilo, line 134: "Output" is a single word in English.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+///  * when a different port is hovered, it is highlighted immediately.
+///  * when none of the ports is hovered all of the `OutputPorts` disappear. Note: there is a very
+///    small delay for disappearing to allow for smooth switching between ports.
+///
+#[derive(Debug,Clone,CloneRef)]
+pub struct OutputPorts {
+    /// The FRP api of the `OutPutPorts`.
+    pub frp            : Frp,
+        network        : frp::Network,
+        data           : Rc<OutPutPortsData>,
```
wdanilo, line 215: alignment

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+        let size           = Cell::new(Vector2::zero());
+        let gap_width      = Cell::new(SEGMENT_GAP_WIDTH);
+
+        let mut ports      = Vec::default();
+        ports.resize_with(number_of_ports as usize,|| component::ShapeView::new(&logger,&scene));
+        let ports          = RefCell::new(ports);
+
+        OutPutPortsData {display_object,logger,size,ports,gap_width}.init()
+    }
+    fn init(self) -> Self {
```
wdanilo, line 156: nl

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+        self
+    }
+
+    fn update_shape_layout_based_on_size(&self) {
+        let port_num    = self.ports.borrow().len() as f32;
+        let width       = self.size.get().x;
+        let height      = self.size.get().y;
+        let port_width  = (width - NODE_SHAPE_PADDING) / port_num;
+        let port_size   = Vector2::new(port_width, height);
+        let gap_width   = self.gap_width.get();
```
wdanilo, line 167: alignment

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+    /// The FRP api of the `OutPutPorts`.
+    pub frp            : Frp,
+        network        : frp::Network,
+        data           : Rc<OutPutPortsData>,
+}
+
+impl OutputPorts {
+    /// Constructor.
+    pub fn new(scene:&Scene, number_of_ports:u32) -> Self {
+
```
wdanilo, line 221: rm nl :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+        let network        = default();
+        let frp            = Frp::new(&network);
+        let data           = OutPutPortsData::new(scene.clone_ref(),number_of_ports);
+        let data           = Rc::new(data);
+
```
wdanilo, line 226: rm nl :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+        data           : Rc<OutPutPortsData>,
+}
+
+impl OutputPorts {
+    /// Constructor.
+    pub fn new(scene:&Scene, number_of_ports:u32) -> Self {
+
+        let network        = default();
+        let frp            = Frp::new(&network);
+        let data           = OutPutPortsData::new(scene.clone_ref(),number_of_ports);
```
wdanilo, line 224: If `OutPutPortsData::new` requires the user of its API to clone the argument which implements `CloneRef`, then it is badly designed. Instead, `OutPutPortsData::new` should consume reference and clone it for itself.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    fn init_frp(&mut self) {
+        let network = &self.network;
+        let frp     = &self.frp;
+        let data    = &self.data;
+
+        frp::extend! { network
+            eval  frp.set_size ((size) data.set_size(size.into()));
+
+            def set_port_size      = source::<(PortId,f32)>();
```
wdanilo, line 243: use the new frp syntax . Just `<-` instead of def

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+            def set_port_sizes_all = source::<f32>();
+
+            def _set_port_sizes = set_port_sizes_all.map(f!([set_port_size,data](size) {
```
wdanilo, line 246: eval

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,311 @@
...
+                let port_num = data.ports.borrow().len();
+                for index in 0..port_num {
+                    set_port_size.emit((index,*size));
```
wdanilo, line 249: This is antipattern. Do not use map to emit events to frp. This is very error prone. Instead use the `<=` operator to iterate over frp values.

wdanilo:
Plus, it doesnt work for bigger amount of ports. It should properly divide the output port to chunks of the same length and the division should also be always "normal" to the node border.

image

wdanilo:
Plus the animation on show is slower than on hide. Plus it doesn't work well. Sometimes ports show even if they should be hidden. Videos posted on discord.

wdanilo:
Plus, `SEGMENT_GAP_WIDTH` is set to `2`, while it renders as `4` px.

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: I think in general that is the preferred way to do this in Rust. The caller might have a spare owned version that they don't need and can themselve decide how and when to create something new.

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer:
```
eval set_port_opacity_all ([set_port_opacity,data](opacity) {
    let port_num = data.ports.borrow().len();
    for index in 0..port_num {
        set_port_opacity.emit((index,*opacity));
    }
});
```
versus
```
port_ix_to_change <= set_port_opacity_all.map(f_!([data] {
    Vec::<usize>::from_iter(0..data.ports.borrow().len())
}));
set_port_opacity_for_port <- all(&port_ix_to_change,&set_port_opacity_all);
eval set_port_opacity_for_port ([set_port_opacity]((index, opacity)) {
   set_port_opacity.emit((*index,*opacity));
});
```
The later seems a lot more complex. Why is the emitting of events in the loop error prone?

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: There are two widths here: the whole shape and the visible slice.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,365 @@
...
+                    port_size.set_target_value(BASE_SIZE);
+                );
+            }
+        }
+
+        // FIXME this is a hack to ensure the ports are invisible at startup.
+        // Right now we get some of FRP mouse events on startup that leave the
+        // ports visible by default.
+        // Once that is fixed, remove this line.
+        delay_hide.finish();
```
wdanilo, line 350: I was asking aobut it in previous PR - tell me more about it. When it happens. It needs to be fixed before those or lands. What events you get and where?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,365 @@
...
+            delay_hide_finished    <- delay_hide.value.map(|t| *t>=TWEEN_END_VALUE );
+            on_delay_show_finished <- delay_show_finished.gate(&delay_show_finished).constant(());
+            on_delay_hide_finished <- delay_hide_finished.gate(&delay_hide_finished).constant(());
+
+            visible                <- delay_show_finished.map(|v| *v);
+
+            mouse_over_while_inactive  <- port_mouse_over.gate_not(&visible).constant(());
+            mouse_over_while_active    <- port_mouse_over.gate(&visible).constant(());
+
+            eval mouse_over_while_inactive ([delay_show,delay_hide](_){
```
wdanilo, line 280: use eval_

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,365 @@
...
+                delay_hide.stop();
+                delay_show.rewind();
+                delay_show.set_end_value(TWEEN_END_VALUE);
```
wdanilo, line 283: why this is needed?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,365 @@
...
+            let shape        = &view.shape;
+            let port_size    = Animation::<f32>::new(&network);
+            let port_opacity = Animation::<f32>::new(&network);
+
+            frp::extend! { network
+
+
+                // === Mouse Event Handling == ///
+
+                eval_ view.events.mouse_over(port_mouse_over.emit(PortId{index}));
```
wdanilo, line 311: please create smart consturcotr for it. Either in a form of `PortId(index)` or `PortId::new(index)` . you can also use `into`. But using raw constructors everywhere is something we try not to do.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,365 @@
...
+
+                // === Visibility and Highlight Handling == ///
+
+                 def _hide_all = hide_all.map(f_!([port_size,port_opacity]{
+                     port_size.set_target_value(0.0);
+                     port_opacity.set_target_value(0.0);
+                 }));
+
+                // Through the provided ID we can infer whether this port should be highlighted.
+                is_selected      <- activate_ports_with_selected.map(move |id| id.index == index);
```
wdanilo, line 330: why this `move` is needed?

MichaelMauderer: To move `index` (laversion: `port_id`) into the closure.

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: We need to cancel racing timer to avoid both timers ending.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+            let overall_width  : Var<Distance<Pixels>> = "input_size.x".into();
+            let overall_height : Var<Distance<Pixels>> = "input_size.y".into();
+
+            let base_shape_data = port_base_shape_data(&overall_width, &overall_height, &grow);
+            let BaseShapeData{ port_area,hover_area,radius,width } = base_shape_data;
+
+            let left_shape_crop  = compute_crop_plane(&index,
+                                                      &port_num,
+                                                      &width.clone().into(),
+                                                      &radius.clone().into(),
```
wdanilo, line 255: can we implement the code without all the explicit `clone().into()` everywhere, please? It would be much nicer to read this way.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+                                                      &padding_right);
+
+            let port_area  = port_area.difference(&left_shape_crop);
+            let port_area  = port_area.intersection(&right_shape_crop);
+
+            let left_hover_crop  = compute_crop_plane(&index, &port_num, &width.clone().into(),
+                                                      &radius.clone().into(), &0.0.into());
+            let right_hover_crop = compute_crop_plane(&(Var::<f32>::from(1.0) + &index), &port_num,
+                                                      &width.into(), &radius.into(), &0.0.into());
+
```
wdanilo, line 270: What are `left_shape_crop`, `right_shape_crop`, left_hover_crop`, and `right_hover_crop`? This shader looks very, very heavy and I believe it could be simplified a lot. Lets have a chat about it tomorrow.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+pub mod single_port_area {
+    use super::*;
+    use ensogl::display::shape::*;
+
+    ensogl::define_shape_system! {
+        (style:Style, grow:f32, opacity:f32) {
+            let overall_width  : Var<Distance<Pixels>> = "input_size.x".into();
+            let overall_height : Var<Distance<Pixels>> = "input_size.y".into();
+
+            let base_shape_data = port_base_shape_data(&overall_width, &overall_height, &grow);
```
wdanilo, line 305: According to the guide, there sohuld be no spaces around `,` - please fix here and in other places.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+///      .                            .
+///       .                          .
+///          .                    .
+///              """""""""""""
+///     |  r    |              |   r   |
+///     |            width             |
+/// ```
+/// where r is the radius of the left and right quarter circle shapes.
+fn port_base_shape_data
+(width:&Var<Distance<Pixels>>, height:&Var<Distance<Pixels>>, grow:&Var<f32>) -> BaseShapeData {
```
wdanilo, line 77: why this is not `::new` function of `BaseShapeData`?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+            let left_hover_crop  = compute_crop_plane(&index, &port_num, &width.clone().into(),
+                                                      &radius.clone().into(), &0.0.into());
+            let right_hover_crop = compute_crop_plane(&(Var::<f32>::from(1.0) + &index), &port_num,
+                                                      &width.into(), &radius.into(), &0.0.into());
+
+            let hover_area = hover_area.difference(&left_hover_crop);
+            let hover_area = hover_area.intersection(&right_hover_crop);
+            let hover_area = hover_area.fill(color::Rgba::new(0.0,0.0,0.0,0.000_001));
+
+            // FIXME: Use colour from style and apply transparency there.
```
wdanilo, line 275: haha, a side note, "colour" is British spelling :D

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+            self.grow.set(grow_value)
+        }
+
+        fn set_opacity(&self, opacity:f32) {
+            self.opacity.set(opacity)
+        }
+    }
+}
+
+/// Implements a simplified version of the multi_port_area::Shape shape for the case where there is
```
wdanilo, line 294: Lets create a big section here to divide these codes. They are pretty big and it would be cool to discover it while fast scrolling :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+                })
+            }
+        }
+    }
+
+    /// Resize all the port output shapes to fit the new layout requirements for thr given
+    /// parameters.
+    fn update_shape_layout_based_on_size_and_gap(&self, size:Vector2<f32>, gap_width:f32) {
+        match self {
+            ShapeView::Single { view }   => {
```
wdanilo, line 365: alignment of `=>` - just do not align it to anything, as the line above or below do not contain `=>`

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+                let shape = &view.shape;
+                shape.sprite.size.set(size);
+            }
+            ShapeView::Multi { views }   => {
+                let port_num  = views.len() as f32;
```
wdanilo, line 370: 2 spaces before `=`

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+                for (index, view) in views.iter().enumerate(){
+                    let shape = &view.shape;
+                    shape.sprite.size.set(size);
+                    shape.index.set(index as f32);
+                    shape.port_num.set(port_num);
+                    shape.padding_left.set(gap_width * 0.5);
+                    shape.padding_right.set(-gap_width * 0.5);
+                }
+                views[0]              .shape.padding_left .set(0.0);
+                views[views.len() - 1].shape.padding_right.set(0.0);
```
wdanilo, line 380: after we change the shapes to be drawn always with (0,0) in the center of the shapes, these should not be needed anymore (there's should be other simplifications as well).

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+            }
+            ShapeView::Multi { views }   => {
+                let port_num  = views.len() as f32;
+                for (index, view) in views.iter().enumerate(){
+                    let shape = &view.shape;
+                    shape.sprite.size.set(size);
+                    shape.index.set(index as f32);
+                    shape.port_num.set(port_num);
+                    shape.padding_left.set(gap_width * 0.5);
+                    shape.padding_right.set(-gap_width * 0.5);
```
wdanilo, line 377: Do you really need to set `padding` as parameter to the shader? This makes shaders more complex than they should be. Lets talk about it tomorrow.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+    fn set_parent<T:display::object::Object>(&self, parent:&T) {
+        match self {
+            ShapeView::Single { view }   => {
+                view.display_object().set_parent(parent);
+            }
+            ShapeView::Multi { views }   => {
+                views.iter().for_each(|view|  view.display_object().set_parent(parent))
+            }
+        }
+    }
```
wdanilo, line 394: This is a hack, please do not hack into display objects :P Instead, in the multi scenario, create a `display::object::Instance`, add all shapes to it, and implement `impl display::Object for ShapeView` instead :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+    let PortFrp { network,port_mouse_over,port_mouse_out,port_mouse_down,
+        hide_all,activate_ports_with_selected } = frp;
+
+    let shape        = &view.shape;
+    let port_size    = Animation::<f32>::new(&network);
+    let port_opacity = Animation::<f32>::new(&network);
+
+    frp::extend! { network
+
+            // === Mouse Event Handling == ///
```
wdanilo, line 431: indent

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+
+
+
+// ===========
+// === Frp ===
+// ===========
+
+/// Id of a specific port inside of `OutPutPortsData`.
+#[derive(Clone,Copy,Default,Debug,Eq,PartialEq)]
+pub struct PortId {
```
wdanilo, line 476: PortId should not be defined in `FRP` section.

wdanilo, line 476: I think a better name would be `PortIndex` instead. Correct me if Im wrong, but these are indexes of ports starting from left right?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+            activate_ports <- any(mouse_over_while_active,on_delay_show_finished);
+            eval_ activate_ports (delay_hide.stop_and_rewind());
+
+            activate_ports_with_selected <- port_mouse_over.sample(&activate_ports);
+
+            hide_all <- on_delay_hide_finished.map(f_!(delay_show.stop_and_rewind()));
+
+        }
+
+        let port_frp = PortFrp {network         : network.clone_ref(),
```
wdanilo, line 656: never, ever do it, pls. Netwrok has to be owned by 1 thing, not multiple.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+                                port_mouse_down : frp.on_port_mouse_down.clone_ref(),
+                                port_mouse_over,port_mouse_out,hide_all,
+                                activate_ports_with_selected};
```
wdanilo, line 659: please use one style. Either fields with `:` or just field names with commas. Just create more vars instead.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+            }
+        }
+    }
+}
+
+
+
+// =============================
+// === Port Frp Setup Helper ===
+// =============================
```
wdanilo, line 401: Lets just name it `PortFrp`

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+
+/// Helper struct to pass the required FRP endpoints to set up the FRP of a port shape view.
```
wdanilo, line 403: Let's not write everywhere that something is a "helper" structure. In fact, telling that it is "helper" doesn't tell much and just make docs longer. In fact, I believe this is not a helper structure at all! This is FRP definition of a single port, this is a very important structure! :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+#[derive(Clone,CloneRef,Debug)]
+struct PortFrp {
+    network                      : frp::Network,
+
+    port_mouse_over              : frp::Source<PortId>,
+    port_mouse_out               : frp::Source<PortId>,
+    port_mouse_down              : frp::Source<PortId>,
```
wdanilo, line 410: this is FRP for a single port, right? If so, the prefix `port_` is not needed here

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+
+    hide_all                     : frp::Stream<()>,
+    activate_ports_with_selected : frp::Stream<PortId>
```
wdanilo, line 413: I do not understand the name `activate_ports_with_selected`. I don't know how to improve it cause I don't understand it. Lets chat about it pls :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+             eval port_size.value    ((size) shape.set_grow(*size));
+             eval port_opacity.value ((size) shape.set_opacity(*size));
+
+
+            // === Visibility and Highlight Handling == ///
+
+             def _hide_all = hide_all.map(f_!([port_size,port_opacity]{
+                 port_size.set_target_value(0.0);
+                 port_opacity.set_target_value(0.0);
+             }));
```
wdanilo, line 449: eval

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+
+            // Through the provided ID we can infer whether this port should be highlighted.
+            is_selected      <- activate_ports_with_selected.map(move |id| *id == port_id);
```
wdanilo, line 452: why move?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+
+            eval_ show_highlighted ([port_opacity,port_size]{
+                port_opacity.set_target_value(1.0);
+                port_size.set_target_value(HIGHLIGHT_SIZE);
+            });
+
+            eval_ show_normal ([port_opacity,port_size]
+                port_opacity.set_target_value(0.5);
+                port_size.set_target_value(BASE_SIZE);
+            );
```
wdanilo, line 464: Here you are not using `{ ... }` and above you are using those. I just realized this macro has a bug. multiline syntax without `{ ... } ` should not be allowed!

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+        network : frp::Network,
+        data    : Rc<OutputPortsData>,
+}
+
+impl OutputPorts {
+    /// Constructor.
+    pub fn new(scene:&Scene, number_of_ports:NonZeroU32) -> Self {
+        let network = default();
+        let frp     = Frp::new(&network);
+        let data    = OutputPortsData::new(scene.clone_ref(), number_of_ports);
```
wdanilo, line 587: Lets not design an API which requires cloning on usage side. Pass `&scene` instead. If you want to make it generic, you can always use `scene: impl Into<Scene>` because there is `impl From<&Scene> for Scene` defined.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+        let data    = Rc::new(data);
+        OutputPorts {data,network,frp}.init()
+    }
+
+    fn init(mut self) -> Self {
+        self.init_frp();
+        self
+    }
+
+    fn init_frp(&mut self) {
```
wdanilo, line 597: If we have single init function, lets just leave it named `init` without making 2 separate ones calling each other :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,682 @@
...
+        // // Right now we get some of FRP mouse events on startup that leave the
+        // // ports visible by default.
+        // // Once that is fixed, remove this line.
+        delay_hide.from_now_to(TWEEN_END_VALUE);
+    }
+
+    // TODO: Implement proper sorting and remove.
+    /// Hack function used to register the elements for the sorting purposes. To be removed.
+    pub(crate) fn order_hack(scene:&Scene) {
+        let logger = Logger::new("hack");
```
wdanilo, line 673: Oh sorry, I meant, rename the name of the logger, not the function :D

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -948,7 +948,7 @@ impl GraphEditorModelWithNetwork {
         frp::new_bridge_network! { [self.network, node.main_area.events.network]
             eval_ node.drag_area.events.mouse_down(touch.nodes.down.emit(node_id));
             eval  node.ports.frp.cursor_style ((style) cursor_style.emit(style));
-            eval_ node.frp.output_ports.mouse_down (output_press.emit(node_id));
+            eval_ node.view.output_ports.frp.port_mouse_down (output_press.emit(node_id));
```
wdanilo, line 951: This is unsafe now. What happens if someone adds output port to a node? If I understand correctly, we are then regenerating the whole output port structure, right? So this binding will stop working. Isntead, we should expose frp endpoint in node frp for that and in case we change output t ports, we should rebind frp there.

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: Internally yes, but that should not matter for the API user. They should just assume that this struct identifies a Port in an OutputPort struct. So I'd rather keep the  `Id` instead of `Index` suffix.

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: I've changed it. But we should also consider changing the ownership semantics of `Network` then to disallow cloning general to prevent this from ever happening.

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: To move `port_id` into the closure.

src/rust/lib/graph-editor/src/component/node.rs
```diff
@@ -301,14 +257,15 @@ impl NodeModel {
         });
         display_object.add_child(&ports);
 
+        let frp = Frp{input};
 
-        let output_ports = OutputPortsEvents { shape_view_events:output_area.events.clone_ref() };
 
-        let frp = Frp{input,output_ports};
+        // TODO: Determine number of output ports based on node semantics.
+        let output_ports = OutputPorts::new(&scene, NonZeroU32::new(10).unwrap());
```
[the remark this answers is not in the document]
MichaelMauderer: I think instead of implicitly fixing it. it might be better to enable whoever will provide the value to do some error handling instead. That would allow us to, for example, show an error message at the source of the problem.

src/rust/lib/graph-editor/src/lib.rs
MichaelMauderer: Hm. You are right if we regenerate the whole output area, then this would not work. But this endpoint could still work if we implement the changing of the ports internally to the `OutputPort` struct. Might be a question of how we want to handle the FRP design for the node in general: do we want to reroute, or give access to internal structs?

src/rust/lib/graph-editor/src/component/node/port/output.rs
MichaelMauderer: I've updated the name and added a doc. Let me know if this is clearer.

src/rust/ensogl/src/display/shape/primitive/glsl/shape.glsl
```diff
@@ -136,6 +138,11 @@ BoundSdf pixel_snap (BoundSdf a) {
     return a;
 }
 
+BoundSdf grow (BoundSdf a, float size) {
+    a.distance = a.distance - size;
+    return a;
+}
```
wdanilo, line 144: this is bad. it doestn update the bounds.

src/rust/ensogl/src/display/shape/primitive/glsl/shape.glsl
```diff
@@ -244,6 +251,14 @@ Shape pixel_snap (Shape s) {
     return shape(id,sdf,color);
 }
 
+Shape grow (Shape s, float value) {
+    Id       id    = s.id;
+    BoundSdf sdf   = grow(s.sdf,value);
+    Srgba    color = unpremultiply(s.color);
+    color.raw.a /= s.alpha;
+    return shape(id,sdf,color);
+}
```
wdanilo, line 260: its 100% ok. just FYI - in most cases the colors will not work ok. Basically we should always apply colors after shapes are done.

src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
```diff
@@ -280,6 +280,30 @@ impl Canvas {
...
+
+    /// Grow the shape by the given value.
+    pub fn grow<T:Into<Var<f32>>>
+    (&mut self, num:usize, s:Shape, value:T) -> Shape {
+        self.if_not_defined(num, |this| {
+            let value:Glsl = value.into().glsl();
+            let expr = iformat!("return grow({s.getter()},{value});");
+            let mut shape = this.new_shape_from_expr(num,&expr);
+            shape.add_ids(&s.ids);
+            shape
```
wdanilo, line 292: I see that my code is broken as well (maybe because of some refactoring), but please, keep the proper alignment in new code :)

src/rust/ensogl/src/display/shape/primitive/shader/canvas.rs
```diff
@@ -280,6 +280,30 @@ impl Canvas {
...
+    (&mut self, num:usize, s:Shape, value:T) -> Shape {
+        self.if_not_defined(num, |this| {
+            let value:Glsl = value.into().glsl();
+            let expr = iformat!("return grow({s.getter()},-{value});");
+            let mut shape = this.new_shape_from_expr(num,&expr);
+            shape.add_ids(&s.ids);
+            shape
+        })
+    }
 }
```
wdanilo, line 307: why not just use the `grow` function above? `Var<f32>` implements `Neg`

src/rust/lib/graph-editor/src/lib.rs
wdanilo: I like the current design if we keep `output_ports.frp` as something which will automatically rebind internally when needed.

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+/// Implements the shape for a segment of the OutputPort with multiple output ports.
+pub mod multi_port_area {
+    use super::*;
+    use ensogl::display::shape::*;
+    use std::f32::consts::PI;
+
+    /// Compute the angle perpendicular to the shape border.
+    fn compute_border_perpendicular_angle
+    (full_shape_border_length:&Var<f32>, corner_segment_length:&Var<f32>, position:&Var<f32>)
+        -> Var<f32> {
```
wdanilo, line 135: remove indent pls - its part of the signaturee, so indent should be the same as arg list

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+        // TODO implement proper abstraction for non-branching "if/then/else" or "case" in
+        // shaderland
+        // Here we use a trick to use a pseudo-boolean float that is either 0 or 1 to multiply a
+        // value that should be returned, iff it's case is true. That way we can add return values
+        // of different "branches" of which exactly one will be non-zero.
+
+        // Transform coordinate system to be centered in the shape.
+        // The `distance`s here describe the distance along the shape border, so not straight line
+        // x coordinate, but the length of the path along the shape.
+        let center_distance          = position - full_shape_border_length  / 2.0;
```
wdanilo, line 145: Why you have to "Transform coordinate system to be centered in the shape."? All the functions that are drawing shapes should accept arguments with coords already centered. Please lets keep this assumption so our life will be much easier :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+    /// segment and apply the appropriate x-offset.
+    ///
+    /// * `full_shape_border_length` should be the length of the shapes border path.
+    /// * `corner_segment_length`    should be the quarter circumference of the circles on the
+    ///                              sides of base shape.
+    /// * `position_on_path`         should be the position along the shape border
+    ///                              (not the pure x-coordinate).
+    fn calculate_crop_plane_position_relative_to_center_segment
+    (full_shape_border_length:&Var<f32>, corner_segment_length:&Var<f32>, position_on_path:&Var<f32>)
+     -> Var<f32> {
```
wdanilo, line 175: aligment

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+        let middle_segment_start_point = corner_segment_length;
+        let middle_segment_end_point   = full_shape_border_length - corner_segment_length;
+
+        // Case 1: The left circle, always 0, achieved through clamping.
+        // Case 2: The middle segment.
+        let middle_segment_plane_position_x = (position_on_path - middle_segment_start_point)
+            / (&middle_segment_end_point - middle_segment_start_point);
+        // Case 2: The right circle, always 1, achieved through clamping.
+
```
wdanilo, line 184: You have 2 "Case 2" here

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+        middle_segment_plane_position_x.clamp(0.0.into(), 1.0.into())
+    }
+
+    /// Compute the crop plane at the location of the given port index. Also takes into account an
+    /// `position_offset` that is given as an offset along the shape boundary.
+    ///
+    /// The crop plane is a HalfPlane that is perpendicular to the border of the shape and can be
```
wdanilo, line 191: please write all names that refer to code names using backticks :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+        let crop_plane_pos          = crop_plane_pos_relative * &center_segment_length;
+        let crop_plane_pos          = crop_plane_pos + corner_radius;
+
+        let plane_rotation_angle = compute_border_perpendicular_angle
+            (&full_shape_border_length,&corner_segment_length,&crop_segment_pos);
+        let plane_shape_offset  = Var::<Distance<Pixels>>::from(&crop_plane_pos - width * 0.5);
+
+        let crop_shape          = HalfPlane();
+        let crop_shape          = crop_shape.rotate(plane_rotation_angle);
+        let crop_shape          = crop_shape.translate_x(plane_shape_offset);
```
wdanilo, line 215: strange indentation  :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+    fn new(number_of_ports:NonZeroU32, logger:&Logger, scene:&Scene) -> Self {
+        let number_of_ports = number_of_ports.get();
+        if number_of_ports == 1 {
+            ShapeView::Single { view: component::ShapeView::new(&logger,&scene) }
+        } else {
+            let display_object = display::object::Instance::new(logger);
+            let mut views = Vec::default();
+            let number_of_ports = number_of_ports as usize;
+            views.resize_with(number_of_ports,|| component::ShapeView::new(&logger,&scene));
+            views.iter().for_each(|view|  view.display_object().set_parent(&display_object));
```
wdanilo, line 338: double space after `|view|`

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+            ShapeView::Multi {display_object,views}
+        }
+    }
+
+    /// Set up the frp for all ports.
+    fn init_frp(&self, network:&frp::Network, port_frp:PortFrp) {
+        match self {
+            ShapeView::Single {view}  => init_port_frp(&view,PortId::new(0),port_frp,network),
```
wdanilo, line 346: double space

src/rust/lib/graph-editor/src/component/node/port/output.rs
wdanilo: What is the status of this comment? :)

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+
+/// Struct that contains all required FRP endpoints to set up the FRP of a port shape view.
+#[derive(Clone,CloneRef,Debug)]
+struct PortFrp {
+    mouse_over                            : frp::Source<PortId>,
+    mouse_out                             : frp::Source<PortId>,
+    mouse_down                            : frp::Source<PortId>,
+
+    hide_all                              : frp::Stream<()>,
+    /// Activate the port and if it has the given PortId, show it highlighted.
```
wdanilo, line 421: a very stragne indentation

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+// =================
+
+/// Struct that contains all required FRP endpoints to set up the FRP of a port shape view.
+#[derive(Clone,CloneRef,Debug)]
+struct PortFrp {
+    mouse_over                            : frp::Source<PortId>,
+    mouse_out                             : frp::Source<PortId>,
+    mouse_down                            : frp::Source<PortId>,
+
+    hide_all                              : frp::Stream<()>,
```
wdanilo, line 420: What does "hide_all" mean in a sin gel port FRP ?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+        eval_ view.events.mouse_over(mouse_over.emit(port_id));
+        eval_ view.events.mouse_out(mouse_out.emit(port_id));
+        eval_ view.events.mouse_down(mouse_down.emit(port_id));
+
+
+         // === Animation Handling == ///
+
+         eval port_size.value    ((size) shape.set_grow(*size));
+         eval port_opacity.value ((size) shape.set_opacity(*size));
+
```
wdanilo, line 451: wrong indent

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+        // === Visibility and Highlight Handling == ///
+
+         eval_ hide_all ([port_size,port_opacity]{
+             port_size.set_target_value(0.0);
+             port_opacity.set_target_value(0.0);
+         });
+
+        // Through the provided ID we can infer whether this port should be highlighted.
+        is_selected      <- activate_and_highlight_selected.map(move |id| *id == port_id);
```
wdanilo, line 461: what is this move for ?

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+        let size           = Cell::new(Vector2::zero());
+        let gap_width      = Cell::new(SEGMENT_GAP_WIDTH);
+        let ports          = ShapeView::new(number_of_ports,&logger,scene);
+        let ports          = RefCell::new(ports);
+
+        OutputPortsData {display_object,logger,size,ports,gap_width}.init()
+    }
+
+    fn init(self) -> Self {
+        self.ports.borrow().display_object().set_parent(&self.display_object);
```
wdanilo, line 551: do not use `display_obeject()` explicitelly. just skip it

`self.ports.borrow().set_parent(&self);`

src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -0,0 +1,692 @@
...
+            activate_ports <- any(mouse_over_while_active,on_delay_show_finished);
+            eval_ activate_ports (delay_hide.stop_and_rewind());
+
+            activate_and_highlight_selected <- mouse_over.sample(&activate_ports);
+
+            hide_all <- on_delay_hide_finished.map(f_!(delay_show.stop_and_rewind()));
+
+        }
+
+        let network         = network.clone_ref();
```
wdanilo, line 662: alignment

wdanilo: Plus why do you need it ?

PR #543, opened 2020-06-05
 Delayed appearance of multi ports

MichaelMauderer:
### Pull Request Description
Adds a delay to the visibility onset on ports. This avoids the blinking of ports during mouse movements over nodes that are not intended to interact with the node.

Peek 2020-06-05 14-01

### Important Notes
* hack: there are some invalid mouse events on startup that would leave the ports visible by default. This is compensated for by manually hiding the ports, even though this would otherwise b the default state.

### Checklist

### DIFF
--- a/src/rust/lib/graph-editor/src/component/node/port/output.rs
+++ b/src/rust/lib/graph-editor/src/component/node/port/output.rs
@@ -26,6 +26,9 @@ const BASE_SIZE           : f32 = 0.5;
 const HIGHLIGHT_SIZE      : f32 = 1.0;
 const SEGMENT_GAP_WIDTH   : f32 = 2.0;
 
+const SHOW_DELAY_DURATION : f32 = 150.0;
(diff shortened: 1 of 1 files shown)


src/rust/lib/graph-editor/src/component/node/port/output.rs
```diff
@@ -237,63 +241,131 @@ impl OutputPorts {
...
+                    port_size.set_target_value(BASE_SIZE);
+                ));
             }
         }
+
+        // FIXME this is a hack to ensure the ports are invisible at startup.
+        // Right noe we get some of FRP mouse events on startup that leave the
+        // ports visible by default.
+        // Once that is fixed, remove this line.
+        on_hide_delay_finish.emit(());
```
wdanilo, line 368: We need to fix it. IT sounds strange and dangerous. Please inspect it and write to me what exactly mouse events are propagated during startup.

PR #552, opened 2020-06-09
Fixed textfield key masking on Mac

notdanilo:
### Pull Request Description
This PR fixes an issue on MacOS which kept text field typing when using `command` modifier key.

### Important Notes
* Platform::query() was used to deal with OS intricacies.

### Checklist

### DIFF
--- a/src/rust/ensogl/src/display/shape/text/text_field/frp/keyboard.rs
+++ b/src/rust/ensogl/src/display/shape/text/text_field/frp/keyboard.rs
@@ -133,7 +133,7 @@ impl TextFieldKeyboardFrp {
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/rust/ensogl/src/display/shape/text/text_field/frp/keyboard.rs
```diff
@@ -133,16 +133,24 @@ impl TextFieldKeyboardFrp {
         }
     }
 
+    fn enable_writing(mask:&keyboard::KeyMask) -> bool {
```
mwu-tow, line 136: Should be `enables_writing` (as this is check, not the action of enabling).

PR #564, opened 2020-06-16
Routing type information to the graph editor integration layer

mwu-tow:
### Pull Request Description
This PR provides:
* registry for type information;
* receiving type updates from LS and routing appropriate updates;
* feeding this information up to the graph editor.

### Checklist


src/rust/ide/src/controller/project.rs
```diff
@@ -153,14 +174,58 @@ impl Handle {
...
+    /// Returns a handling function capable of processing updates from the json-rpc protocol.
+    /// Such function will be then typically used to process events stream from the json-rpc
+    /// connection handler.
+    pub fn json_event_handler
+    (&self) -> impl Fn(enso_protocol::language_server::Event) -> futures::future::Ready<()> {
+    // TODO [mwu]
+    //  This handler for JSON-RPC notifications is very similar to the function above that handles
+    //  binary protocol notifications. However, it is not practical to generalize them, as the
+    //  underlying RPC handlers and their types are separate.
+    //  This generalization should be reconsidered once the old JSON-RPC handler is phased out.
```
farmaazon, line 199: We have already an issue for unifying handlers, don't we?
If we have, its description should also contain note to refactor this functions according to this TODO.

src/rust/ide/src/executor/test_utils.rs
```diff
@@ -61,12 +63,29 @@ impl TestWithLocalPoolExecutor {
             self.run_task(task);
         }
     }
+
+    /// Runs all tasks in the pool and returns if no more progress can be made on any task.
+    pub fn run_until_stalled(&mut self) {
+        self.executor.run_until_stalled();
+    }
```
farmaazon, line 70: If we exposed API this way, perhaps we don't need `when_stalled` any more? Instead of:
```rust
let mut test = TestWithLocalPoolExecutor::set_up()
...
test.when_stalled(|| { ..something...});
```
We can have just
```rust
let mut test = TestWithLocalPoolExecutor::set_up()
...
test.run_until_stalled();
 ..something...
```

And if somehow it's still needed, it should use self.run_until_stalled instead of  self.executor.run_until_stalled();

src/rust/ide/src/executor/test_utils.rs
```diff
@@ -61,12 +63,29 @@ impl TestWithLocalPoolExecutor {
...
+        self.executor.run_until_stalled();
+        assert_eq!(0,self.running_task_count.get(),"The tasks are not complete!");
+    }
+
+    /// Runs all tasks until stalled and tries retrieving value from the future.
+    /// If the future cannot complete, panics.
+    pub fn expect_completion<R>(&mut self, fut:impl Future<Output=R>) -> R {
+        self.run_until_stalled();
+        fut.boxed_local().expect_ready()
+    }
```
farmaazon, line 83: It's not clear why we need this function. Can you add a note to docs?

src/rust/ide/src/model/execution_context.rs
```diff
@@ -2,16 +2,21 @@
...
+use enso_protocol::language_server::ExpressionValuesComputed;
+use enso_protocol::language_server::MethodPointer;
 use enso_protocol::language_server::VisualisationConfiguration;
+use flo_stream::MessagePublisher;
+use flo_stream::Subscriber;
 use std::collections::HashMap;
 use uuid::Uuid;
 
 
-
```
farmaazon, line 14: I think this line should still be here.

src/rust/ide/src/model/execution_context.rs
```diff
@@ -173,26 +267,31 @@ pub struct ExecutionContext {
     stack:RefCell<Vec<LocalCall>>,
     /// Set of active visualizations.
     visualizations: RefCell<HashMap<VisualizationId,AttachedVisualization>>,
+    /// Set of active visualizations.
```
farmaazon, line 270: Copied doc.

src/rust/ide/src/model/execution_context.rs
```diff
@@ -24,6 +29,95 @@ pub type ExpressionId = ast::Id;
...
+        });
+        self.emit(update);
+    }
+
+    /// Subscribe to notifications about changes in the registry.
+    pub fn subscribe(&self) -> Subscriber<ComputedValueExpressions> {
+        self.updates.borrow_mut().subscribe()
+    }
+
+    /// Borrows the immutable cache reference.
```
farmaazon, line 113: outdated doc

src/rust/ide/src/model/execution_context.rs
```diff
@@ -24,6 +29,95 @@ pub type ExpressionId = ast::Id;
...
+        let removed_keys = self.map.borrow().keys().copied().collect_vec();
+        self.map.borrow_mut().clear();
+        if !removed_keys.is_empty() {
+            self.emit(removed_keys);
+        }
+    }
+
+    /// Store the information from the given update received from the Language Server.
+    pub fn apply_update(&self, values_computed:ExpressionValuesComputed) {
+        let update = values_computed.updates.iter().map(|update| update.id).collect();
```
farmaazon, line 97: rename it to updated_expressions, or something.

src/rust/ide/src/model/execution_context.rs
```diff
@@ -173,26 +267,31 @@ pub struct ExecutionContext {
...
     pub fn new(logger:impl Into<Logger>, entry_point:DefinitionName) -> Self {
-        let logger         = logger.into();
-        let stack          = default();
-        let visualizations = default();
-        Self {logger,entry_point,stack,visualizations}
+        let logger             = logger.into();
+        let stack              = default();
+        let visualizations     = default();
+        let expression_info    = default();
+        Self {logger,entry_point,stack,visualizations, computed_value_info_registry: expression_info }
```
farmaazon, line 281:
```suggestion
        let logger                       = logger.into();
        let stack                        = default();
        let visualizations               = default();
        let computed_value_info_registry = default();
        Self {logger,entry_point,stack,visualizations, computed_value_info_registry }
```

src/rust/ide/src/executor/test_utils.rs
mwu-tow: I think both APIs should stay in place. Well, the latter is more general, so we could rewrite all our tests on top of it but there's no value.
Basically `when_stalled` allows you to setup responses and then run the asynchronous tasks with `await` sugar and whatnot.
Without this, you need to to manually poll the futures. It is sometimes necessary (e.g. as you want to test that something completes only after receiving some external stimulus) but I'd rather do that only when justified by the scenario we want to test.

The latter comment will be applied.

src/rust/ide/json-rpc/src/macros.rs
```diff
@@ -202,6 +220,10 @@ macro_rules! make_rpc_methods {
 /// ```
 #[macro_export]
 macro_rules! expect_call {
+    ($mock:ident.$method:ident() => $result:expr) => {
+        let result = $result;
+        $mock.expect.$method(move || result)
+    };
```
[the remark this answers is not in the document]
mwu-tow: The next branch would lead to infinite recursion in this case. 
The next branch purpose is to replace `(arg,…)` with `(arg=val,…)`. However, when there are no arguments, it becomes self-entrant no-op and there is no means to enter the third branch.

PR #574, opened 2020-06-18
Disconnect / reconnect edge source and target by clicking on edges. (Downwards edges)

MichaelMauderer:
### Pull Request Description
Enables edge disconnect on edge click. Adds hover indicator about which end is going to be disconnected.

### Important Notes
Only works for downwards edges.

### Checklist


wdanilo:
It also doesnt work with connections going upwards
image

wdanilo:
Moreover, there are 2 additional problems that we know about, but they still need to be fixed before this PR gets merged (so on master we have correct visual behavior). Consider the following picture:

t1

Especially this fragment:

t2

There are two artifacts visible, green and red:

- **The green artifact**  
   We know it is caused by overlapping transparent shapes. The only solution that doesn't break the complex edge ordering here is to just use different color instead of transparency for the faded-out edges. Let's do it this way.

- **The red artifact**
   We've been talking about it shortly a week ago. There were different ideas on how to fix it, including making the whole mouse cursor blue to hide it, but this is a bad solution as colored cursor always means that you are in an active drag state, not that you are hovering something. Instead, we should discover here the intersection point of the mouse and the edge and draw there a small circle that would make the line nicely rounded (see image below). Please note that computing the intersection is simple as the edge consists of a horizontal line, part of the circle, and vertical line (let's start with the downwards edge and then implement it for upwards as well). Please also note that keeping the division line at wrong angle (pink) will give us bad results and that the division angle needs to be always normal (perpendicular) to the border of the edge. However, computing this should be simple having the intersection point already computed. 

   t2

   Regarding the division lines, please just compute them in the way shown below. If the intersection fragment is on placed on the horizontal line, just dim shapes according to the green lines. If the intersection point is placed on the rounded corner, dim shapes according to the purple lines. If the intersection point is placed on the vertical line, dim lines according to the orange lines:
   
   t3

   As I mentioned earlier - let's start with edges going downwards, then let's verify it, and let's create a separate PR for edges going upwards. The latter case would use just the same algorithm - we have there just 2 more corners, so the algorithm to compute the intersection point and the division lines should be just the same. There is one thing to remember here - we do not want the arrow on the lines to be half-dimmed. It should be either fully dimmed or not at all:

   image

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -328,28 +530,63 @@ pub fn sort_hack_2(scene:&Scene) {
 // === FRP ===
 // ===========
 
+/// FRP endpoints for aggregated mouse events.
+#[derive(Clone,CloneRef,Debug)]
+#[allow(missing_docs)]
+pub struct ShapeViewEventsProxy {
```
wdanilo, line 536: Looking at the docs and the name of this struct I do not understand what it is used for.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -328,28 +530,63 @@ pub fn sort_hack_2(scene:&Scene) {
...
+    pub mouse_down : frp::Stream<()>,
```
wdanilo, line 537: () is the default parametrization for Stream, just skip it

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -328,28 +530,63 @@ pub fn sort_hack_2(scene:&Scene) {
...
+    on_mouse_out  : frp::Source<()>,
+}
+
+#[allow(missing_docs)]
+impl ShapeViewEventsProxy {
+    pub fn new(network:&frp::Network) -> Self {
+        frp::extend! { network
+            def on_mouse_over  = source();
+            def on_mouse_out   = source();
+            def on_mouse_down  = source();
```
wdanilo, line 552: Please fix alignments in the whole new code

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -101,43 +245,68 @@ macro_rules! define_corner_end {($($color:tt)*) => {
 
                 let shadow_size = 10.px() + 1.px();
                 let n_radius = &shadow_size + 1.px() * dim.y();
-                let n_shape  = Rect((&shadow_size*2.0 + 2.px() * dim.x(),&n_radius*2.0)).corners_radius(n_radius);
+                let n_shape  = Rect(
+                (&shadow_size*2.0 + 2.px() * dim.x(),&n_radius*2.0)).corners_radius(n_radius);
```
wdanilo, line 249: If you want to divide a line to multi-line expression, always use indentation. Now its hard to visually read it :)

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -500,9 +749,60 @@ impl EdgeModelData {
...
+
+    fn is_in_upper_half(&self, point:Vector2<f32>) -> bool {
+        let start = nalgebra::Point2::from(self.display_object.position().xy());
+        let end   = nalgebra::Point2::from(self.target_position.get());
+        let point = nalgebra::Point2::from(point);
+        let mid_y = (start.y + end.y) / 2.0;
+        point.y > mid_y
+    }
+
+    /// Returns whether the given hover position belongs to the Input or Output part of the
```
wdanilo, line 766: Input and Output are written using big letters, so I assume they should be connected to some structs in the code. If so, use backticks to indicate it.

MichaelMauderer:
@wdanilo 
I've made the changes to the computations as suggested: splitting orthogonal, using a round "joint" to hide the split (and some refactoring that came with that. So some of the visual glitches are gone now. 

Peek 2020-06-30 17-29

The code is not committed here yet, as I'm still hunting down one issue with the last shape directly at the port. 

Peek 2020-06-30 17-30

If you have any thoughts about why that shape could behave differently than the others, let me know.
I'll continue on this later today to figure out why that is happening.

wdanilo:
@MichaelMauderer this looks beautiful! And no, I don't have a clue why it's behaving differently. I believe these shapes should behave the same way. Btw, this is really beautiful!!! ❤️

wdanilo:
@MichaelMauderer One more question - how hard it would be to add animation to dimming? If that would take more than 1h, lets skip it for now, but if that's around 1h, lets add it - then we will have a final, very polished view of that! <3

MichaelMauderer:
> @MichaelMauderer One more question - how hard it would be to add animation to dimming? If that would take more than 2h, lets skip it for now, but if that's around 1-2h, lets add it - then we will have a final, very polished view of that! <3
 I think right now it would take longer, but after the refactoring for the colored edges, that should be way more easy. So let's file an issue and do it then!

MichaelMauderer:
So this PR is updated and (according to my testing), works glitch-free for downwards edges. The splitting id deactivated for upwards edges. 

Downwards edges on the right side: 
Peek 2020-07-01 12-06
Downwards edges on the left side:
Peek 2020-07-01 12-09
Upwards edges:
Peek 2020-07-01 12-10

@wdanilo I also addressed the review comments. Please, have another look, play with the PR, and let me know if you find any other issues here. 

In the meantime, I'll work on the changes for upwards edges and loopy bits.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+}
+
+impl<'a> SnapData<'a> {
+    fn new(position: Vector2<f32>, target_shape: &'a dyn EdgeShape) -> Self {
+        SnapData{position,target_shape}
+    }
+}
+
+/// The MultiShape trait allows operations on a collection of `EdgeShape`.
+trait MultiShape {
```
MichaelMauderer, line 186: This needs a better more specific name. Maybe also the splitting logic should live somewhere else than this trait.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+    /// Return the `ShapeViewEvents` of all sub-shapes.
+    fn events(&self) -> Vec<ShapeViewEvents>;
+    /// Return references to all `EdgeShape`s in this MultiShape.
+    fn edge_shape_views(&self) -> Vec<&dyn EdgeShape>;
```
MichaelMauderer, line 190: Instead of returning references here, it might be better to use `Id`s to avoid lifetime issues.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+                // these.
+                let snap_distance = (snap_position - point).norm();
+                if snap_distance < SNAP_THRESHOLD && shape.contains(point){
+                    return Some(SnapData::new(snap_position,*shape))
+                }
+            }
+            None
+        });
+
+        // Return the snap that is closest to the hover.
```
MichaelMauderer, line 223: We should not try to snap to all shapes. Instead, we should discover the correct shape from the hover event directly and only snap to that one. This will avoid also needing the `SNAP_THRESHOLD` in the first place.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+
+/// Holds the data required to split a shape into two parts.
+///
+/// The `area` indicates which side of the split will receive special coloring.
+#[derive(Clone,Copy,Debug)]
+struct SplitData {
+    split_position : Vector2<f32>,
+    area           : Area,
+    cut_angle      : f32
+}
```
wdanilo, line 163: I have a minor naming issue here. I know we are using suffix "Data" here and there (and I have problems with these as well), but basically, everything is just Data. when we use this suffix in other place in codes its basically used to indicate internal Data of some Rc'd struct. Here, we are talking about a split information and I just believe that there is no logical difference between names `SplitData` and just `Split` - as reader, you don't get more information seeing suffix "Data", because everything is just data :D

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+// ========================
+// === Edge Shape Trait ===
+// ========================
+
+/// Indicates which area should be recolored.
+#[derive(Clone,Copy,Debug,Eq,PartialEq)]
+enum Area {
+    Above,
+    Below,
+}
```
wdanilo, line 153: This needs better name. Moreover, I'm not sure if above / below are good names, especially in context of upwards edges. I would name it more like `split::Side::Source` and `split::Side::Target`, what do you think about it? Of course, I'm open to better names! If you don't want to create separate module for split.yet, then `SplitSide::Source` etc

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+    fn contains(&self, point:Vector2<f32>) -> bool {
+        self.contains_local(self.to_shape_coordinate_system(point))
+    }
+}
+
+
+
+// ========================
+// === Edge Shape Trait ===
+// ========================
```
wdanilo, line 146: This section contains too many non-related things.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -13,7 +13,10 @@ use ensogl::display::scene::Scene;
 use ensogl::display::shape::*;
 use ensogl::display::traits::*;
 use ensogl::display;
+use ensogl::gui::component::ShapeViewEvents;
 use ensogl::gui::component;
+use nalgebra::UnitComplex;
```
wdanilo, line 18: Lets use `rotate_90(v:Vector2<f32>) -> Vector2<f32> { Vector2(v.y,v.x) }`, or something like that instead :)

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+        let local          = self.to_shape_coordinate_system(point);
+        let local_snapped  = self.snap_to_shape_local(local)?;
+        let global_snapped = self.from_shape_to_global_coordinate_system(local_snapped);
+        Some(global_snapped)
+    }
+
+    /// Set the hover split for this shape. The `split_data` indicates where the shape should be
+    /// split and which side should be recolored.
+    fn enable_hover_split(&self, split_data:SplitData) {
+        let SplitData {split_position,area,cut_angle} = split_data;
```
wdanilo, line 92: Not a big thing, but lets talk. Why do you use this form? I don't like using it because if in the future I add more info to that struct, this line will stop compiling? Sometimes this is exactly our intention when we want to always be sure we use all fields that are there, but here, this code should work no matter what new fields you add there. Moreover, if we use that in object notation, we can use shorter fields names, like `split.position`.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+        let base_rotation   = self.display_object().rotation().z;
+        let local_unrotated = point - self.display_object().global_position().xy();
+        UnitComplex::new(-base_rotation) * local_unrotated
+    }
+
+    /// Convert the given shape local coordinate to the global coordinates system.
+    fn from_shape_to_global_coordinate_system(&self, point:Vector2<f32>) -> Vector2<f32> {
+        let base_rotation   = self.display_object().rotation().z;
+        let local_unrotated = UnitComplex::new(base_rotation) * point;
+        local_unrotated + self.display_object().global_position().xy()
```
wdanilo, line 124: Please, do not use explicit `.display_object()` ever. All functions are exported there, so `self.global_position()` should work. Fix it everywhere and in case you will find use case where you need something more, lets talk. Explicit call like here should always be avoided.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+            },
+        }
+        // Compute position in shape local coordinate system.
+        let offset = self.to_shape_coordinate_system(split_position);
+        self.set_hover_split_offset(offset)
+    }
+
+    /// Disable the hover split on this shape.
+    fn disable_hover_split(&self) {
+        self.set_hover_split_offset(Vector2::new(INFINITE, INFINITE));
```
wdanilo, line 109: fyi, since few weeks, we have exported functions for easier construction of vectors. You can do now this: `Vector2(INFINITE,INFINITE)` instead of `Vector2::new(INFINITE,INFINITE)`

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+/// The `area` indicates which side of the split will receive special coloring.
+#[derive(Clone,Copy,Debug)]
+struct SplitData {
+    split_position : Vector2<f32>,
+    area           : Area,
+    cut_angle      : f32
+}
+
+impl SplitData {
+    fn new(split_position:Vector2<f32>, area: Area, cut_angle:f32) -> Self {
```
wdanilo, line 166: Minor formatting issue - in few places in the code you have space after `:` in arg list. Just search the code for ": " and you should instantly discover all places

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+        SplitData {split_position,area,cut_angle}
+    }
+}
+
+/// `SnapData` is used as the result value of snapping operations on `MultiShapes`. It holds the
+/// position and shape that a hover position was snapped to, that is the shape and the position
+/// on the shape that was closest to the hover
+struct SnapData<'a> {
```
wdanilo, line 174: 1. I think better name would be `SnapTarget` (see the comment above about "everything is data". Moreover, the docs do not explain clear enough what is the position fold. Would you mind rephrasing it, please?

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
 
 
 
+// ========================
+// === Edge Shape Trait ===
+// ========================
+
+/// Edge shape defines the common behaviour of the sub-shapes used to create a Edge.
+trait EdgeShape: ensogl::display::Object {
+    fn set_hover_split_offset(&self, offset:Vector2<f32>);
```
wdanilo, line 62: This needs either documentation or slightly better naming. Is it `set_split_origin`?

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+    }
+
+    /// Check whether the given shape local coordinate is within the bounds of the shape.
+    fn contains_local(&self, point:Vector2<f32>) -> bool {
+        let size = self.sprite().size.get();
+        if point.x < -size.x / 2.0 || point.y < -size.y / 2.0 {
+            return false
+        }
+        point.x < size.x / 2.0 && point.y < size.y / 2.0
+    }
```
wdanilo, line 134: This should not be needed after using on hover events

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+        }
+    }
+
+    /// Snap the given position to our sub-shapes. Returns `None` if the given position cannot be
+    /// snapped to any sub-shape (e.g., because it is too far away).
+    fn snap_position_to_shape(&self, point:Vector2<f32>) -> Option<SnapData> {
+        let shapes    = self.edge_shape_views();
+        let snap_data = shapes.iter().filter_map(|shape|{
+            let maybe_snap_position     = shape.snap_to_shape(point);
+            if let Some(snap_position) = maybe_snap_position {
```
wdanilo, line 210: alignment

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+            .map(|(_, snap_data)| snap_data)
+    }
+
+    /// Try to snap the given position to the shape and compute the perpendicular cut angle. If we
+    /// can find valid values, we return a `SplitData` struct. This can fail, if the hover is too
+    /// far from any of our sub-shapes.
+    fn get_split_data_for_hover(&self,position:Vector2<f32>,area:Area) -> Option<SplitData> {
+        let maybe_snap_data = self.snap_position_to_shape(position);
+        if let Some(snap_data) = maybe_snap_data {
+            // TODO refactor into shape
```
wdanilo, line 238: do it before the merge pls :)

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+        SplitData {split_position,area,cut_angle}
+    }
+}
+
+/// `SnapData` is used as the result value of snapping operations on `MultiShapes`. It holds the
+/// position and shape that a hover position was snapped to, that is the shape and the position
+/// on the shape that was closest to the hover
+struct SnapData<'a> {
+    position     : Vector2<f32>,
+    target_shape : &'a dyn EdgeShape,
```
wdanilo, line 176: I don't like using dyn traits because they are slow, especially when rendering a lot of shapes. Maybe its necessary here, but I would leave here a request - think about it and think if there is any other design that would not require it without making the code MUCH bigger. I'd love to hear your thoughts here. In the past I was able to remove multiple dyn traits from your PRs (visualizations handling), so I'm just wondering if we can just simplify some assumptions and some generalizations to also make this more performant. Of course if that would increase the boilerplate 2x or 3x then its not worth, but lets talk about it.

src/rust/lib/graph-editor/src/component/edge.rs
wdanilo: Our naming convention here would be `AnyEdgeShape`

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+            .min_by(|(distance_a, _), (distance_b, _)| {
+                distance_a.partial_cmp(&distance_b).unwrap_or(Ordering::Equal)
+            })
+            .map(|(_, snap_data)| snap_data)
+    }
+
+    /// Try to snap the given position to the shape and compute the perpendicular cut angle. If we
+    /// can find valid values, we return a `SplitData` struct. This can fail, if the hover is too
+    /// far from any of our sub-shapes.
+    fn get_split_data_for_hover(&self,position:Vector2<f32>,area:Area) -> Option<SplitData> {
```
wdanilo, line 235: minor - add spaces after `,`

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+        let maybe_snap_data = self.snap_position_to_shape(position);
+        if let Some(snap_data) = maybe_snap_data {
```
wdanilo, line 237: use `.map` instead

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+
+    /// Split the shape at the given `position` and highlight the given `area`. This might fail if
+    /// the given position is too far from the shape. In that case, splitting is disabled.
+    fn enable_hover_split(&self, position:Vector2<f32>, area:Area) {
+        if let Some(split_data) = self.get_split_data_for_hover(position,area) {
+            self.edge_shape_views().iter().for_each(|shape| shape.enable_hover_split(split_data));
+        } else {
+            self.disable_hover_split()
+        }
+    }
```
wdanilo, line 255: the naming of these function is bad 0 looking at its body. Function named `ENABLE_hover_split` cannot call `DISABLE_hover_split` inside.

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
 // === Shape Definitions ===
 // =========================
 
-macro_rules! define_corner_start {($($color:tt)*) => {
+/// SplitShape allows a shape to be split along a line and each sub-shape to be colored separately.
+struct SplitShape {
+    part_a : AnyShape,
+    part_b : AnyShape,
+    joint  : AnyShape,
+}
```
wdanilo, line 275: the names are hard to understand here. What is `part_a` and `part_b`? What is `joint`? Cover that in docs pls :)

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+
+impl SplitShape {
+    /// Splits the shape in two at the line given by the offset and rotation. Will render a
+    /// circular "joint" at the given `offset`, if `joint_radius` > 0.0.
+    fn new
+    (base_shape:AnyShape, offset:&Var<Vector2<f32>>,rotation:&Var<f32>,joint_radius:&Var<Distance<Pixels>>) -> Self {
```
wdanilo, line 281: add spaces after `,` here and in other places

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+            .translate_y(&offset_y);
+
+        let part_a = base_shape.intersection(&split_plane).into();
+        let part_b = base_shape.difference(&split_plane).into();
+
+        let joint_radius = Var::<Distance<Pixels>>::from(joint_radius);
+        let joint        = Circle(joint_radius)
+                            .translate_x(&offset_x)
+                            .translate_y(&offset_y)
+                            .into();
```
wdanilo, line 297: just use 2 lets instead please :)

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -44,23 +47,280 @@ const NODE_PADDING       : f32 = node::SHADOW_SIZE;
...
+    /// circular "joint" at the given `offset`, if `joint_radius` > 0.0.
+    fn new
+    (base_shape:AnyShape, offset:&Var<Vector2<f32>>,rotation:&Var<f32>,joint_radius:&Var<Distance<Pixels>>) -> Self {
+        let offset_x    = Var::<Distance<Pixels>>::from(offset.x());
+        let offset_y    = Var::<Distance<Pixels>>::from(offset.y());
+        let rotation    = Var::<Angle<Radians>>::from(rotation.clone());
+        let split_plane = HalfPlane()
+            .rotate(&rotation)
+            .translate_x(&offset_x)
+            .translate_y(&offset_y);
```
wdanilo, line 288: just use 2 lets instead, please :)

src/rust/lib/graph-editor/src/component/edge.rs
```diff
@@ -69,30 +329,54 @@ macro_rules! define_corner_start {($($color:tt)*) => {
...
+
+            fn sprite(&self) -> &Sprite{
+                &self.shape.sprite
+            }
+
+            fn snap_to_shape_local(&self, _point:Vector2<f32>) -> Option<Vector2<f32>> {
+                // FIXME this needs to be implemented for upwards edges
+                None
+            }
+
```
wdanilo, line 365: rm nl

src/rust/lib/graph-editor/src/component/edge.rs
MichaelMauderer: This was only used as a shorthand. I've removed it.

src/rust/lib/graph-editor/src/component/edge.rs
MichaelMauderer: This looks like introducing other languages constructors syntax. Is this not violating some of our own style guidelines, too? (Uppercase functions?).

src/rust/lib/graph-editor/src/component/edge.rs
MichaelMauderer: Implementing `ObjectOps` and avoiding the call to `.display_object()` seems to mess with my ability to return a reference to the `EdgeShape`. For now, I would prefer to leave it as is, instead of tracking down why `ObjectOps` requires a static lifetime only to avoid this method call.

src/rust/lib/graph-editor/src/component/edge.rs
MichaelMauderer: I would like to do both of these things as part of the next PR that has up and down edges working if that is fine with you.  It avoids some backporting/merging work

src/rust/lib/graph-editor/src/component/edge.rs
MichaelMauderer: The next iteration of this will use ShapeIDs in some places where we use references to shapes right now. But I think in some places we cannot avoid it here, as we need to abstract over all the base shapes that make up the whole edge shape. But I will keep this in mind.

src/rust/lib/graph-editor/src/component/edge.rs
MichaelMauderer: I moves to use `Rotate2` which is more appropriate. I would like to keep  the genral rotation to keep the code more robust against new shapes that might have a diffeent rotation.

wdanilo:
Closing in favor of https://github.com/enso-org/ide/pull/640

PR #575, opened 2020-06-18
Use correct position to calculate edge dragging.

MichaelMauderer:
### Pull Request Description
Fixed edges getting desynchronized when panning / zooming scene during edge connection.

### Checklist

### DIFF
--- a/src/rust/ensogl/src/display/scene.rs
+++ b/src/rust/ensogl/src/display/scene.rs
@@ -934,8 +934,10 @@ impl SceneData {
         });
     }
 
-    pub fn transform_screen_to_scene_coordinates(&self, pos:Vector2<f32>) -> Vector2<f32> {
-       self.camera().position().xy() +  (pos.xy() / self.camera().zoom())
+    pub fn transform_screen_to_scene_coordinates(&self, position:Vector3<f32>) -> Vector3<f32> {
+        let position = position / self.camera().zoom();
+        let position = Vector4::new(position.x, position.y, position.z, 1.0);
+        (self.camera().inversed_view_matrix() * position).xyz()
     }
 }
 
--- a/src/rust/ensogl/src/display/scene.rs
+++ b/src/rust/ensogl/src/display/scene.rs
@@ -933,6 +933,10 @@ impl SceneData {
             self.context.viewport(0,0,canvas.width as i32, canvas.height as i32);
         });
     }
+
+    pub fn transform_screen_to_scene_coordinates(&self, pos:Vector2<f32>) -> Vector2<f32> {
+       self.camera().position().xy() +  (pos.xy() / self.camera().zoom())
+    }
 }
 
 impl display::Object for SceneData {
--- a/src/rust/ensogl/src/display/scene.rs
+++ b/src/rust/ensogl/src/display/scene.rs
@@ -934,7 +934,7 @@ impl SceneData {
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -1841,8 +1845,8 @@ fn new_graph_editor(world:&World) -> GraphEditor {
 
     // === Move Edges ===
     detached_edge           <- any(&inputs.some_edge_targets_detached,&inputs.some_edge_sources_detached);
-    cursor_pos_on_detach    <- cursor.frp.position.sample(&detached_edge);
-    edge_refresh_cursor_pos <- any (cursor_pos_on_detach,cursor.frp.position);
+    cursor_pos_on_detach    <- mouse_pos_in_scene.sample(&detached_edge);
+    edge_refresh_cursor_pos <- any (cursor_pos_on_detach,mouse_pos_in_scene);
```
wdanilo, line 1849: Please be very careful about changing the logic! We were using cursor position, now you are using mouse position. So it will not work when snapping ports (cursor moves then, mouse not).

src/rust/ensogl/src/display/scene.rs
```diff
@@ -933,6 +933,10 @@ impl SceneData {
             self.context.viewport(0,0,canvas.width as i32, canvas.height as i32);
         });
     }
+
+    pub fn transform_screen_to_scene_coordinates(&self, pos:Vector2<f32>) -> Vector2<f32> {
+       self.camera().position().xy() +  (pos.xy() / self.camera().zoom())
+    }
```
wdanilo, line 939: this should be done using proper matrix computations instead. See computations in `cursor.rs` as an example.

src/rust/lib/graph-editor/src/lib.rs
```diff
@@ -1501,6 +1501,12 @@ fn new_graph_editor(world:&World) -> GraphEditor {
     let outputs        = UnsealedFrpOutputs::new();
     let sealed_outputs = outputs.seal(); // Done here to keep right eval order.
 
+    // === Mouse Cursor Transform ===
+    frp::extend! { network
+        mouse_pos_in_scene <- cursor.frp.position.map(f!((position) {
```
wdanilo, line 1506: we have two names. Mosue and cursor. In my previous PR I was telling that you are using wrong name to implement the logic. Now you are using name "mouse" to refer to "cursor". Please, be very careful regarding names!

src/rust/ensogl/src/display/scene.rs
```diff
@@ -933,6 +933,12 @@ impl SceneData {
             self.context.viewport(0,0,canvas.width as i32, canvas.height as i32);
         });
     }
+
+    pub fn transform_screen_to_scene_coordinates(&self, position:Vector3<f32>) -> Vector3<f32> {
```
wdanilo, line 937: this can be done shorter with the same logic as `screen_to_scene_coordinates` :)

src/rust/ensogl/src/display/scene.rs
```diff
@@ -933,6 +933,12 @@ impl SceneData {
...
+        let position = position / self.camera().zoom();
+        let position = Vector4::new(position.x, position.y, position.z, 1.0);
+        (self.camera().inversed_view_matrix() * position).xyz()
```
wdanilo, line 940: Are you 100% sure it works when scene is both zoomed and panned?

MichaelMauderer: Peek 2020-06-19 11-19

PR #578, opened 2020-06-18
Open project with CLI arg

notdanilo:
### Pull Request Description
This PR enables IDE to open a specific project by getting its name from CLI argument `--project`.

### Important Notes
* `ide/lib.rs` was reorganized and split into `constants.rs`, `tests.rs` and `ide.rs` for better maintainance.
* No other logic was changed besides adding a conditional branching when `--project` argument is present.
* Internal uses of `constants::DEFAULT_PROJECT_NAME` was removed.

### Checklist


src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+        Ok(controller::Project::new(logger,connection_json,connection_binary,project_name))
+    }
+
+    /// Creates a new project and returns its metadata, so the newly connected project can be
+    /// opened.
+    pub async fn create_project
+    ( logger          : &Logger
+    , project_manager : &impl project_manager::API
+    , name            : &String) -> FallibleResult<ProjectMetadata> {
+        info!(logger, "Creating a new project named `{name}`.");
```
wdanilo, line 96: Backticks should be used for code. For names use quotes. So replace it with:
`"Creating a new project '{name}'."` instead.

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+        let id = project_manager.create_project(name).await?.project_id;
+        Ok(ProjectMetadata {
+            id,
+            name        : ProjectName {name:name.clone()},
+            last_opened : None,
+        })
```
wdanilo, line 102: I would be really thankful if you change this 5 lines expr to 3 lines one :)

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+        let mut projects = response.projects.iter();
+        let project      = projects.find(|project_metadata| {
+            project_metadata.name.name == *project_name
+        });
+        let project_metadata = if let Some(project) = project {
+            project.clone()
+        } else {
+            println!("Attempting to create {}", project_name);
+            Self::create_project(logger,project_manager,project_name).await?
+        };
```
wdanilo, line 121: use unwrap_or_else instead

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub async fn open_project_or_create_new
+    ( logger : &Logger
+    , project_manager : &impl project_manager::API
+    , project_name    : &String) -> FallibleResult<controller::Project> {
+        let projects_to_list = constants::MAXIMUM_LISTABLE_PROJECTS;
+        let response     = project_manager.list_recent_projects(&projects_to_list).await?;
+        let mut projects = response.projects.iter();
+        let project      = projects.find(|project_metadata| {
+            project_metadata.name.name == *project_name
+        });
```
wdanilo, line 115: 1. braces are not needed here - you can make 1 line expr from it instead.
2. `project_metadata.name.name` - this clearly indicates bad naming. Should be fixed.

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+            endpoints.language_server_binary_address,&project_metadata.name.name).await
+    }
+
+    /// Open most recent project or create a new project if none exists.
+    pub async fn open_most_recent_project_or_create_new
+    ( logger          : &Logger
+    , project_manager : &impl project_manager::API) -> FallibleResult<controller::Project> {
+        let projects_to_list = 1;
+        let mut response     = project_manager.list_recent_projects(&projects_to_list).await?;
+        let project_metadata = if let Some(project) = response.projects.pop() {
```
wdanilo, line 133: In all such cases use `unwrap_or_else` instead,

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+            project
+        } else {
+            let project_name = constants::DEFAULT_PROJECT_NAME.to_string();
+            Self::create_project(logger,project_manager,&project_name).await?
+        };
+        let endpoints = project_manager.open_project(&project_metadata.id).await?;
+        Self::open_project(logger,endpoints.language_server_json_address,
+                     endpoints.language_server_binary_address,&project_metadata.name.name).await
```
wdanilo, line 141: too big indent

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub async fn open_most_recent_project_or_create_new
+    ( logger          : &Logger
+    , project_manager : &impl project_manager::API) -> FallibleResult<controller::Project> {
+        let projects_to_list = 1;
+        let mut response     = project_manager.list_recent_projects(&projects_to_list).await?;
+        let project_metadata = if let Some(project) = response.projects.pop() {
+            project
+        } else {
+            let project_name = constants::DEFAULT_PROJECT_NAME.to_string();
+            Self::create_project(logger,project_manager,&project_name).await?
```
wdanilo, line 137: I completely do not understand this logic. It takes `&str`, converts it to `String` only to convert it back to `&str` in the next line

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+    async fn initialize_project_manager(&mut self) -> FallibleResult<()> {
+        let transport        = self.connect_to_project_manager().await?;
+        self.project_manager = Some(Self::setup_project_manager(transport));
+        Ok(())
+    }
+
+    /// Sets up the project view, including the controller it uses.
+    pub async fn setup_project_view(&self) -> Result<ProjectView,failure::Error> {
+        let logger       = &self.logger;
+        let pm           = self.project_manager.as_ref().expect("Couldn't get Project Manager.");
```
wdanilo, line 153: This error is not self explanatory. What does it mean that you couldn't "get" project manager?

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+        let arguments    = ensogl::system::web::Arguments::new();
+        let project = if let Some(project_name) = arguments.get("project") {
```
wdanilo, line 155: alignment of =

src/rust/ide/src/view/node_editor.rs
```diff
@@ -459,14 +468,13 @@ impl GraphEditorIntegratedWithControllerModel {
...
         //   Currently it is not possible to:
-        //    * use other project name than the default;
         //    * enter other module than the initial (namely, "Main")
         //    * describe that visualization's expression wishes to be evaluated in any other
         //      context.
         //   Because of that for now we will just hardcode the `visualization_module` using
         //   fixed defaults. In future this will be changed, then the editor will also get access
         //   to the customised values.
-        let project_name         = crate::DEFAULT_PROJECT_NAME;
+        let project_name         = self.project.project_name.as_ref();
```
wdanilo, line 477: this shows wrong naming here as well. If you have a field `project`, then it should not have a field `project_name` because it already is a project, so it should not have another names. It should be `self.project.name` instead.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -503,7 +503,35 @@ pub async fn sleep(duration:Duration) {
 #[cfg(not(target_arch = "wasm32"))]
 pub use async_std::task::sleep;
 
+/// Stores arguments extracted from Location search. e.g. http://localhost/?arg0=value0&arg1=value1.
```
wdanilo, line 506: Location is written using upper name but it is neither in backtics nor it is a popular name of something. What is this location referring to ?

src/rust/lib/system/web/src/lib.rs
```diff
@@ -503,7 +503,35 @@ pub async fn sleep(duration:Duration) {
...
+#[derive(Shrinkwrap,Debug)]
+pub struct Arguments {
+    pub hash_map : HashMap<String,String>
+}
+
+impl Arguments {
+    fn args_from_search(search:&str) -> HashMap<String,String> {
+        if search.chars().nth(0) == Some('?') {
+            search[1..].split("&").filter_map(|arg| {
```
wdanilo, line 515: what is this `1` ?

src/rust/lib/system/web/src/lib.rs
```diff
@@ -503,7 +503,35 @@ pub async fn sleep(duration:Duration) {
...
+                let x : Vec<&str> = arg.split("=").collect();
+                match x.len() {
+                    1 => Some((x[0].to_string(), "".to_string())),
+                    2 => Some((x[0].to_string(), x[1].to_string())),
+                    _ => None
```
wdanilo, line 520: this code can be made safer. It uses `x.len()` to guarantee that `x[0]` or `x[1]` do not panic, but instead you could use `head` and `tail` and Options to do it the same way but safer regarding future code refactoring.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -503,7 +503,35 @@ pub async fn sleep(duration:Duration) {
...
+                }
+            }).collect()
+        } else {
+            default()
+        }
+    }
 
+    /// Creates a new arguments map from location search.
+    pub fn new() -> Self {
+        let search   = window().location().search().unwrap_or_else(|_| String::from(""));
```
wdanilo, line 530: please read docs to `Option`. There are many places in the code where you can use better methods than these. For example here, `unwrap_or_default()`.

src/rust/ide/src/constants.rs
```diff
@@ -0,0 +1,28 @@
...
+pub const SOURCE_DIRECTORY:&str = "src";
+
+/// An invocable language expression that serialize given input into JSON.
+pub const SERIALIZE_TO_JSON_EXPRESSION:&str = "x -> x.json_serialize";
+
+/// Endpoint used by default by a locally run Project Manager.
+pub const PROJECT_MANAGER_ENDPOINT:&str = "ws://127.0.0.1:30535";
+
+/// Default project name used by IDE on startup.
+pub const DEFAULT_PROJECT_NAME:&str = "Project";
```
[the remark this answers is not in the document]
mwu-tow: >  what is default project name? It is the name of Unnamed Project? if so, it should be named UnnamedProject instead.

As of now I don't think there's a notion of an unnamed project in the IDE.

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+        })
+    }
+
+    /// Open the named project or create a new project if it doesn't exist.
+    pub async fn open_project_or_create_new
+    ( logger : &Logger
+    , project_manager : &impl project_manager::API
+    , project_name    : &String) -> FallibleResult<controller::Project> {
+        let projects_to_list = constants::MAXIMUM_LISTABLE_PROJECTS;
+        let response     = project_manager.list_recent_projects(&projects_to_list).await?;
```
mwu-tow, line 111: This should be noted with a TODO note that it should be changed once Engine adjusts their API as planned. 
See https://github.com/luna/enso/issues/819

src/rust/lib/system/web/src/lib.rs
```diff
@@ -503,7 +503,35 @@ pub async fn sleep(duration:Duration) {
...
+pub struct Arguments {
+    pub hash_map : HashMap<String,String>
+}
+
+impl Arguments {
+    fn args_from_search(search:&str) -> HashMap<String,String> {
+        if search.chars().nth(0) == Some('?') {
+            search[1..].split("&").filter_map(|arg| {
+                let x : Vec<&str> = arg.split("=").collect();
+                match x.len() {
```
mwu-tow, line 517: I expect that slice patterns could be used to have it both clean and safe

src/rust/ide/src/ide.rs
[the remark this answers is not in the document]
mwu-tow: Also, that'd enable testing the new code — now it is not easy to test that IDE actually opens the project from the passed name.

src/rust/ide/src/constants.rs
```diff
@@ -0,0 +1,28 @@
+//! Global constants used across whole application.
+
+/// A name of language this IDE supports
+pub const LANGUAGE_NAME:&str = "Enso";
+
+/// A file extension of modules of language this IDE supports without leading dot.
+pub const LANGUAGE_FILE_EXTENSION:&str = "enso";
+
+/// A file extension of modules of language this IDE supports with leading dot.
+pub const LANGUAGE_FILE_DOT_EXTENSION:&str = ".enso";
```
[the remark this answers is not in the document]
wdanilo: I prefer to allocate than do somethingl ike what is currently done there.

src/rust/ide/src/ide.rs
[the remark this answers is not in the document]
wdanilo: Yes, do not panic if you can

wdanilo:
And please, always lint code before requesting PR. Its really important as otherwise we need one more ping-pong in review.

src/rust/lib/system/web/src/lib.rs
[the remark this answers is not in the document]
farmaazon: > I'm wondering if we want/need to deal in any way with the URI component encoding (all these `&` <-> `%26` etc)

Perhaps we should use some library for reading URIs?

src/rust/ide/src/constants.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// The directory in the project that contains all the source files.
+pub const SOURCE_DIRECTORY:&str = "src";
+
+/// An invocable language expression that serialize given input into JSON.
+pub const SERIALIZE_TO_JSON_EXPRESSION:&str = "x -> x.json_serialize";
```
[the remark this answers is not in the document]
wdanilo: @notdanilo talk with Engine about it then. This has to work.

src/rust/ide/src/constants.rs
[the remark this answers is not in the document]
wdanilo: @notdanilo then add comment about it please :)

src/rust/ide/src/constants.rs
```diff
@@ -0,0 +1,28 @@
...
+pub const PROJECT_MANAGER_ENDPOINT:&str = "ws://127.0.0.1:30535";
+
+/// Default project name used by IDE on startup.
+pub const DEFAULT_PROJECT_NAME:&str = "Project";
+
+/// Visualization folder where IDE can look for user-defined visualizations per project.
+pub const PROJECT_VISUALIZATION_FOLDER:&str = "visualization";
+
+/// This is the maximum number of listable projects, used in `ProjectManager::list_recent_projects`.
+pub const MAXIMUM_LISTABLE_PROJECTS : u32 = std::i32::MAX as u32 - 1;
```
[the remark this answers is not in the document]
wdanilo: but why? Why maximum count of anything is expressed in `i32` instead of `usize`? When negative numbers are needed there?

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+impl IDE {
+    /// Creates a new IDE instance.
+    pub fn new() -> Self {
+        let config          = SetupConfig::new_local();
+        let logger          = Logger::new("IDE");
+        let _executor       = Self::setup_global_executor();
+        let project_view    = None;
+        let project_manager = None;
```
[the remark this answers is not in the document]
wdanilo: I understand that default doesnt really mean anything. Should we introduce `empty` then which would be used as part of monoid definition? This thought is in my head for months already and more and more. I believe we should do it.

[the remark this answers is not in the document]
mwu-tow: @farmaazon 
> But here we do care what is the value we're assigning.

I somehow can't imagine not relying on `default` for `Option` being `None` and for `Vec` being an empty vector (though in terms of trait docstring itself, the implementation that returned default element repeated 7 times would be valid).

While the trait itself does not give any guarantees about how the default value shall be chosen, implementors of Default are allowed to.

@wdanilo 
To be honst, I don't really like having a custom solution for something that is already covered by an established idiom in Rust (i.e. using default to denote empty container). 
Also, I don't think that Monoid's `empty` is the exact semantics we need here — `Option<T>` is not monoid if `T` is not monoid, whereas "empty" should always be well-defined for `Option` no matter the type stored.

src/rust/ide/src/config.rs
```diff
@@ -1,4 +1,24 @@
...
+
+/// Configuration data necessary to initialize IDE.
+///
+/// We will eventually want to load it from a configuration file.
+#[derive(Clone,Debug)]
+pub struct Startup {
+    /// WebSocket endpoint of the project manager service.
+    pub project_manager_endpoint    : String,
+    /// The project name we want to open on startup passed from the optional `--project` argument
+    pub project_name_from_arguments : Option<String>
```
mwu-tow, line 13: Consider rename.

mwu-tow: Consider using `Option<ProjectName>`

src/rust/ide/src/ide.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub fn setup_project_manager
+    (transport:impl json_rpc::Transport + 'static) -> project_manager::Client {
+        let project_manager = project_manager::Client::new(transport);
+        executor::global::spawn(project_manager.runner());
+        project_manager
+    }
+
+    /// Connect to language server.
+    pub async fn open_project
+    ( logger          : &Logger
```
mwu-tow, line 83: Consider `&self` for the first arg instead of logger

PR #599, opened 2020-06-23
Expression id in span tree

farmaazon:
### Pull Request Description
This is a part of #545: The optional Expression Id was added to SpanTree, to allow reading type information by views.

### Important Notes
Due to nature of opr::Chain and prefix::Chain, I decided to not put ast reference to SpanTree node yet. But the current design needs discussion, because it is sometimes troublesome.

### Checklist
Please include the following checklist in your PR:

- [x] The documentation has been updated if necessary.
- [x] All code conforms to the [Rust](https://github.com/luna/enso/blob/main/docs/style-guide/rust.md).
- [x] All code has been tested where possible.
- [x] All code has been profiled where possible.

PR #602, opened 2020-06-24
Passing Type Information To Graph Editor View

mwu-tow:
# Pull Request Description
As discussed with @wdanilo , this PR adds passing to graph editor events that set types for given expression ids.

### Important Notes
* this PR is only usable for graph editor after #599 is merged — otherwise graph editor wouldn't be able to associate types with ports.

### Checklist

### DIFF
--- a/src/rust/ide/src/view/node_editor.rs
+++ b/src/rust/ide/src/view/node_editor.rs
@@ -258,28 +258,6 @@ impl GraphEditorIntegratedWithControllerModel {
         Ok(())
     }
 
-    /// Set given type (or lack of such) on the given sub-expression.
-    fn set_type(&self, id:ExpressionId, typename:graph_editor::OptionalType) {
-        let event = (id,typename);
-        self.editor.frp.inputs.set_expression_type.emit_event(&event);
-    }
-
-    /// Look up the typename for the given expression in the execution controller's registry and
-    /// pass the data to the editor view.
-    fn update_type_on(&self, id:ExpressionId) {
-        let typename = self.lookup_typename(&id);
-        self.set_type(id,typename)
-    }
-
-    /// Like `update_type_on` but for multiple expressions.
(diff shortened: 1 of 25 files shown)


src/rust/ide/src/view/node_editor.rs
```diff
@@ -339,20 +355,27 @@ impl GraphEditorIntegratedWithControllerModel {
...
-            self.expression_views.borrow_mut().insert(node, expression);
+            self.editor.frp.inputs.set_node_expression.emit_event(&(id,code_and_trees));
+            self.expression_views.borrow_mut().insert(id, expression);
+        }
+
+        // Set initially available type information on ports (identifiable expression's sub-parts).
+        for expression_part in node.info.expression().iter_recursive() {
+            if let Some(id) = expression_part.id {
+                self.update_type_on(id)
+            }
```
farmaazon, line 378: In case of span tree we remember expression and update view only when it changed. But here we notify them each time we got any update from the controllers.

I think we can move this loop inside the if above; we do the full refresh only when expression changes (or node was just put on scene), and do only partial updates when we get notification about new type.

src/rust/ide/src/view/node_editor.rs
```diff
@@ -258,9 +258,25 @@ impl GraphEditorIntegratedWithControllerModel {
...
+    }
+
+    /// Like `update_type_on` but for multiple expressions.
+    fn update_types_on(&self, expressions_to_update:&[ExpressionId]) -> FallibleResult<()> {
+        debug!(self.logger, "Updating type information for IDs: {expressions_to_update:?}.");
+        for id in expressions_to_update {
+            self.update_type_on(*id)
+        }
         Ok(())
     }
```
farmaazon, line 281: (Not crucial): In this section I started with the topmost function and then put functions used by it, etc. So please move those functions lower, or rearrange the whole section.

Issue #585, opened 2020-06-19
Editable project name in view.


notdanilo:
### Summary
This task is to add control displaying project name, which will be docked to the top-center point of the window (and thus over the node view). 

User can click on this name and start to edit it. When user press enter, the name is commited; text field loses focus and FRP signal about project name change is emitted - the actual project rename will be implemented in #584.

The name should be dark gray and highlighted when mouse hovers over it, or when it is edited.

### Value
This is a user-friendly way to renaming projects, far better than popping-up dialog. 

### Acceptance Criteria & Test Cases
Tests and a working example

wdanilo:
Danilo, based on the video that you have provided, there are several things that look bad right now:
1. There is no indication when the field is active. Part of the task was to make it dark when not in edit mode, light when in edit mode. This is not done.
2. When you are doing renaming (the text is light) and you press escape, you should "cancel the renaming", while hitting "enter" should apply the renaming. After hitting enter, the name should get back to grey state to show the editing is done.
3. After name is changed, the text should not be in editable form anymore, so cursor should not be there.
4. The text needs to be centered, also while typing.

Moreover, on the posted video I have noticed that the text size on nodes and the text size of the project name are different. What is causing it?

sylwiabr:
Completely not working on MacOS but closing in favor of https://github.com/enso-org/ide/issues/693

PR #617, opened 2020-06-26
Create temporary Untitled project and renaming projects

notdanilo:
### Pull Request Description
This PR introduces a project rename method for the project manager with tests and a drafty project name view.

### Important Notes
* A temporary ProjectName view structure was created to display the project name and also enable renaming. This will be refined on task https://github.com/enso-org/ide/issues/585.
* Many things regarding TextField, such as the centered text alignment, was done in a drafty manner because it will be replaced soon with the new text field.

### Checklist


src/rust/ide/src/controller/project.rs
```diff
@@ -98,9 +101,12 @@ impl ExecutionContextsRegistry {
 
 /// Project controller's state.
 #[allow(missing_docs)]
-#[derive(Clone,CloneRef,Debug)]
+#[derive(Clone,CloneRef,Derivative)]
+#[derivative(Debug)]
 pub struct Handle {
-    pub project_name        : Rc<String>,
+    pub project_metadata    : Rc<RefCell<ProjectMetadata>>,
```
mwu-tow, line 107: I think that `ProjectMetadata` is not the type we want to store here — it represents the project entry in the list. In particular, `lastOpened` will be most certainly outdated after we open the project. 

***

Also, this might be a good opportunity to split off the project model from the project controller, to achieve same model as we have for e.g. execution context. 
The model should contain mutable state that corresponds to some remote entity, while controller should be immutable with regards to remote operations, it is just a set of data that we can use to call the remote.
The refcell itself should also be the implementation detail of that model type.

@farmaazon will likely be able to explain this better and give advice whether this is something we want pursue soon-ish for project.

src/rust/ide/src/controller/project.rs
```diff
@@ -114,25 +120,27 @@ impl Handle {
...
     ( parent                     : impl AnyLogger
+    , project_manager            : impl project_manager::API + 'static
     , language_server_client     : language_server::Connection
     , mut language_server_binary : binary::Connection
-    , project_name               : impl Str
+    , project_metadata           : ProjectMetadata
     ) -> Self {
         let logger = Logger::sub(parent,"Project Controller");
-        info!(logger,"Creating a project controller for project {project_name.as_ref()}");
+        info!(logger,"Creating a project controller for project {project_metadata.name.deref()}");
```
mwu-tow, line 129: I'd implement `Display` for the `ProjectName`, even if just for avoiding this.

src/rust/ide/src/controller/project.rs
```diff
@@ -304,6 +312,20 @@ impl Handle {
...
+    pub fn project_name(&self) -> ProjectName {
+        self.project_metadata.borrow().name.clone()
+    }
+
+    /// Rename project.
+    pub async fn rename_project(&self, name:impl Str) -> FallibleResult<()> {
+        let name                 = name.into();
+        let mut project_metadata = self.project_metadata.borrow_mut();
+        self.project_manager.rename_project(&project_metadata.id,&name).await?;
+        *project_metadata.name = ProjectName::new(name).to_string();
```
mwu-tow, line 326: This looks *very scary*. `.await` yields the control while we have ongoing `borrow_mut`.
If we request a rename, then try reading the name while waiting for engine's answer, this will likely panic.

We must always try to keep refcell's borrows as localized as possible and avoid allowing non-local code execution while we have lock in place.

src/rust/ide/src/view/project_name.rs
```diff
@@ -0,0 +1,100 @@
...
+        });
+        self.text_field.set_position(offset);
+    }
+
+    /// Get the project name.
+    pub fn name(&self) -> String {
+        self.project_controller.project_name().to_string()
+    }
+
+    /// Rename the project name.
```
mwu-tow, line 79:
```suggestion
    /// Change the project name.
```

mwu-tow: or perhaps even "apply the project name change" — and I'd always consider adding a word like "apply" to the method name as well, as now it is not clear whether this refers to "project name as in this structure" or "project name as in the state of project model in controller".

src/rust/ide/src/view/project_name.rs
```diff
@@ -0,0 +1,100 @@
...
+        Self {logger,display_object,text_field,project_controller}.initialize()
+    }
+
+    fn initialize(self) -> Self {
+        let project_name = self.clone_ref();
+        self.text_field.set_text_edit_callback(move |change| {
+            // If the text edit callback is called, the TextEdit must be still alive.
+            let field_content = project_name.text_field.get_content();
+            let new_name      = field_content.replace("\n","");
+            if change.inserted == "\n" {
```
mwu-tow, line 54: If we have something to confirm edit, do we have also something to abort edit? 
e.g. if we press Escape or move focus away I'd expect that we should restore the name into the text field

src/rust/ide/src/view/project_name.rs
```diff
@@ -0,0 +1,100 @@
...
+            }
+            // Keep only one line.
+            project_name.text_field.set_content(&new_name);
+            project_name.setup_center_alignment();
+        });
+        self.setup_center_alignment();
+        self
+    }
+
+    fn setup_center_alignment(&self) {
```
mwu-tow, line 65: I'm not sure how much trouble would that be — but centering is a logic that shouldn't be specific to the project name. Ideally this should be part of text field itself.

src/rust/ensogl/src/display/shape/text/text_field.rs
```diff
@@ -126,6 +126,18 @@ shared! { TextField
...
+            let mut callback = callback;
+            callback(&self.content)
+        }
+
+        /// Run `callback` with the underlying `&mut TextFieldContent`.
+        pub fn with_mut_content<F:FnMut(&mut TextFieldContent)>(&mut self, callback:F) {
+            let mut callback = callback;
+            callback(&mut self.content)
+        }
+
```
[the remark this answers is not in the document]
wdanilo: Yup, it looks like overly complex design to me. Btw remember the text field will be replaced soon.

src/rust/ide/src/controller/project.rs
wdanilo: Yep, we should never have such code. Borrow must should be always limited to a narrow scope.

src/rust/ide/src/view/project_name.rs
wdanilo: This is one of my comments in the issue: https://github.com/enso-org/ide/issues/585. But I'll move the comments to the PR to make them more visible.

src/rust/ide/src/view/project_name.rs
wdanilo: @mwu-tow this will be part of new text field that I'm implementing now.

src/rust/ensogl/src/display/shape/text/text_field.rs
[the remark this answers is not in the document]
notdanilo: Yes, I kept in mind that the text field will be replaced for many things in this PR.

src/rust/ide/src/controller/project.rs
notdanilo: Good catch! Thanks for pointing it out.

src/rust/ide/src/view/project_name.rs
notdanilo: Yes, this logic should belong to the text field. But I didn't want to put it in there because it was going to be replaced. But it's good to know it's a WIP. :)

src/rust/ide/src/controller/project.rs
notdanilo: As discussed on Discord, we will leave this refactoring for later since it needs to be planned as a separated task.

notdanilo:
@wdanilo your comments are appliable to the view tasks which I am currently working. So let's progress with this PR disregarding them.

Here is an animation of how the implementation looks. Keep in mind a bug on the project manager service still exists, which is visible on this animation, and it is considered a blocker. A bug report was filed to keep track of it: https://github.com/enso-org/enso/issues/959

project-renaming

src/rust/ensogl/src/display/shape/text/text_field.rs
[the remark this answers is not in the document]
farmaazon: in the old TextFieldFrp after each change of content you should update the displayed stuff, see `set_content` for reference

src/rust/ide/src/controller/project.rs
notdanilo: I introduced a new struct called `ProjectData` inside `controller/project.rs`. The MVC refactoring is ready in the project name view PR.

src/rust/ensogl/src/display/shape/text/text_field.rs
```diff
@@ -126,6 +126,22 @@ shared! { TextField
     }
 
     impl {
+        /// Run `callback` with the underlying `mut TextFieldContent`.
+        pub fn with_content<F:FnMut(&TextFieldContent)>(&self, callback:F) {
+            let mut callback = callback;
+            callback(&self.content)
+        }
```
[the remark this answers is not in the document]
notdanilo: This is not used and I removed it.

src/rust/ide/src/controller/project.rs
```diff
@@ -92,15 +94,31 @@ impl ExecutionContextsRegistry {
...
 // ==========================
 
 /// Project controller's state.
 #[allow(missing_docs)]
-#[derive(Clone,CloneRef,Debug)]
+#[derive(Clone,CloneRef,Derivative)]
+#[derivative(Debug)]
 pub struct Handle {
-    pub project_name        : Rc<String>,
+    pub project_data        : Rc<RefCell<ProjectData>>,
```
farmaazon, line 119: We can't change project id after opening, can we? If not, maybe it will be sufficient to keep only Rc<RefCell<ProjectName>>, and id as a separate immutable field?

src/rust/ensogl/src/display/camera/camera2d.rs
```diff
@@ -79,7 +79,7 @@ pub struct Clipping {
 impl Default for Clipping {
     fn default() -> Self {
         let near = 1.0;
-        let far  = 1000.0;
+        let far  = 10000.0;
```
wdanilo, line 82: add todo comment that this is not a final solution (will fail on bigger screens). Instead, when resizing a screen this should be updated - write about ti in the todo pls

src/rust/ide/src/ide.rs
```diff
@@ -81,14 +81,13 @@ impl IdeInitializer {
     /// Connect to language server.
     pub async fn open_project
     ( logger           : &Logger
-    , project_manager  : &impl project_manager::API
-    , project_metadata : &ProjectMetadata
+    , project_manager  : impl project_manager::API + 'static
```
wdanilo, line 84: where this 'static comes from? I'm very careful about such constraints because they populate easily trough the codebase and its hard to prune them later. We need to understand exactly where such things come from. In case of functions that are boxed and stored somewhere its obvious, here its not and needs description

src/rust/ide/src/view/project_name.rs
```diff
@@ -0,0 +1,102 @@
+//! This module provides a view for project's name which can be used to edit it.
```
wdanilo, line 1: This project name view is wrong. It contains controller in view. Tehre is another PR which implements that in a correct way. This should not be part of this PR. I was writing about it earlier. This code bundles view and controller and should not land on master.

src/rust/ide/src/ide.rs
notdanilo: I refactored this code to leave 'static out of it.

src/rust/ide/src/view/project_name.rs
notdanilo: @wdanilo this file is completely removed in https://github.com/enso-org/ide/pull/639. My suggestion is:

1. Once https://github.com/enso-org/ide/pull/639 is approved, I can merge it into this PR's branch.
2. Then merge everything into `main`.

src/rust/ide/src/tests.rs
```diff
@@ -22,6 +22,7 @@ async fn failure_to_open_most_recent_project_is_reported() {
         let name             = crate::constants::DEFAULT_PROJECT_NAME;
         let project_metadata = IdeInitializer::get_most_recent_project_or_create_new
             (&logger,&client,name).await.expect("Couldn't get most recent or create new project.");
+        let client           = Rc::new(client);
```
farmaazon, line 25:
```suggestion
        let client  = Rc::new(client);
```

src/rust/ide/src/view/project_name.rs
wdanilo: Lets do it this way :)

PR #621, opened 2020-06-30
The IDE team process specification

farmaazon:
### Pull Request Description
The main goal of this change is to add a specification about our workflow, task management etc. into our docs. The specified process differs (I hope improves) our current process of estimating and describing tasks. Please read it carefully and share you comments.

For the few next sprints we probably won't follow exactly the described process (I need to speed up specifying epics), but this is what we want to eventually achieve.

Additionally it refactors somewhat docs structure according to the requirement of pushing it to dev.enso.org. The docs still should be improved in the near future.

### Checklist


docs/contributing/README.md
```diff
@@ -0,0 +1,7 @@
+# Enso IDE Contributing Documentation
+
+This directory contains helpful resources for anyone who wants to start Enso IDE development. The
+ main guideline is available [here](../CONTRIBUTING.md).
```
wdanilo, line 4: indentation

wdanilo:
I love it!

docs/CONTRIBUTING.md
```diff
@@ -76,8 +76,7 @@ need the following setup:
 Please note that you should not use a code auto-formatter in this codebase. Please read the
 following documents to learn more about reasons behind this decision and the recommended code style
 guide. Be sure to carefully read the
-[Rust style guide 1](https://github.com/luna/ide/blob/main/docs/style-guide.md)
-and the [Rust style guide 2](https://github.com/luna/enso/blob/main/doc/style-guide/rust.md)
```
mwu-tow, line 80: We shouldn't remove reference to the other document without having them unified.
Currently they both contain partially separate and complementing, partially contradictory guides.

docs/contributing/process.md
```diff
@@ -0,0 +1,15 @@
+# Enso Team Process
```
mwu-tow, line 1: Should we discern between IDE team and Enso Team? (this process seems to apply only to the IDE team)

What is the audience of this document? Since it describes our internal process (and references data available only through zenhub) I'd consider putting this in the internal repo.
Nothing secret here but I don't think this information is  useful to community.

wdanilo: Michal, thi is meant for the IDE team now, maybe it will be applied to other teams in the future if it will work well here. Regarding moving it to internal repo - I this its worth keeping it here. As soon as we will have more contributors its worth including them in the process (at least those who are very active). Its also nice to be very transparent regarding our process - it just shows people how we are working and I always appreciate companies which are so transparent regarding their processes. Does it make sense to you? :)

farmaazon: Actually, I meant to IDE team process. When it will be extended to the whole Enso, then we should change the title.

docs/CONTRIBUTING.md
farmaazon: The link is dead, so I thought that the rust style guide was removed from enso repository. But I'll fix that anyway.

docs/contributing/process.md
```diff
@@ -0,0 +1,15 @@
...
+This document specify our core team workflow, described as a lifecycle of a task:
+
+* A newly created task should be appropriately described, but not estimated yet - it will be during
+ Backlog Refinement meeting. The task is put in "New Tasks" column. 
+* At the beginning of the sprint team leads put to the "To Refine" column all tasks which will be 
+ refined during the next Backlog refinement.
+* Each team member should read the descriptions of tasks in "To Refine" column and ask questions
+ and raise all concerns. All the conversation should be recorded in the issue's comments.  
+* During Backlog Refinement we confirm that the task description is clear and estimate it with
+ "Scrum poker" methodology. The task is then moved to the "Backlog" column. If it turns out that
```
MichaelMauderer, line 12: Just to clarify: my understanding of our current process is that we actually do a time estimate ("number of days needed to complete a task"), instead of "[Scrum Poker](https://en.wikipedia.org/wiki/Planning_poker)" which tends to estimate abstract story points. Is this meant as a change to the current way we estimate? If not we might want to clarify this here.

farmaazon:
Pushed requested changes. I removed the "Scrum poker" term, I thought that this is mainly about cards, but I seems to be more than that.

docs/contributing/process.md
mwu-tow: @wdanilo Yes. I mostly wonder if this is the right place, since looking at path "contributing/process.md" it can suggest to the community that this is a process that needs to be followed to contribute. 
Though I'm not sure how it should be placed.

README.md
```diff
@@ -95,7 +95,7 @@ highly appreciate every contribution. If you love the vision behind Enso and you
 want to redefine the data processing world, join us and help us track down bugs,
 implement new features, improve the documentation or spread the word! Join our
 community on a [Discord chat](http://chat.luna-lang.org) and read the
```
notdanilo, line 97: Do we have an updated enso.org link for our Discord chat?

docs/README.md
```diff
@@ -0,0 +1,5 @@
+# Enso IDE documentation
+ 
+* [**Contributing guidelines**](./contributing/README.md) - helpful instructions for anyone who
+ wants to contribute. 
+* [**Product specification**](./product/README.md) - a specification from the user perspective.
```
notdanilo, line 5: :+1:

PR #626, opened 2020-07-01
Remove shared libraries from the IDE repo

iamrecursion:
### Pull Request Description
1.  Removes `prelude`, and renames it to `enso-prelude`.
2.  Removes `shapely`, and renames it to `enso-shapely`.
3.  Removes `macro-utils` and renames it to `enso-macro-utils`.

Closes enso-org/enso#938.

### Important Notes
All tests appear to be passing on my end.

### Checklist


iamrecursion:
> One question I have is how the workflow for updating these libraries will look now: so far we were able to just edit them and immediately use changes, now it seems that this needs to go the crates.io.
> How the releases our shared libraries on crates.io work — do we have them somehow automated?

If you want to depend on them properly they'd have to be released to crates.io, yes. I know this makes things more difficult, but it doesn't make sense for me to effectively fork `prelude` and co for use while doing the parser rewrite. @wdanilo said he's going to look into workflows for updating them locally before pushing to crates.

As for the release workflow, just ask me and I'll add you as an owner for the crate on crates.io. From there it's as simple as `cargo publish`.

PR #639, opened 2020-07-04
Editable project name in view

notdanilo:
### Pull Request Description
This PR implements editable project name views with proper visuals, frp networks and animations.

### Important Notes
* Text `focus` and `blur` aren't implemented because the current implementation of `TextField` always keep at least one `TextField` focused and doesn't notify about the blur event as it should. Implementing this would require some refactoring on `TextField`, so I decided to leave it as it is and support `focus`/`blur` events when the new `TextField 2.0` arrives.

### Demo

project-renaming-2

### Checklist


src/rust/ide/src/view/project_name.rs
```diff
@@ -4,19 +4,60 @@ use crate::prelude::*;
...
 use nalgebra::Vector2;
 
+
+
+// =================
+// === Constants ===
+// =================
+
+const HIGHLIGHTED_TEXT_COLOR : color::Rgba = color::Rgba::new(1.0,1.0,1.0,1.0);
+const DARK_GRAY_TEXT_COLOR   : color::Rgba = color::Rgba::new(0.6,0.6,0.6,1.0);
```
wdanilo, line 33: `(0.6,0.6,0.6,1.0)` doesnt play nice with bg. Bg is a little yellowish. Shoulduse alpha instead.

src/rust/ide/src/view/project_name.rs
```diff
@@ -4,19 +4,60 @@ use crate::prelude::*;
...
+mod shape {
+    use super::*;
+
+    ensogl::define_shape_system! {
+            (style:Style, selection:f32) {
+                let bg_color = color::Rgba::new(0.0,0.0,0.0,0.000001);
+                let width  : Var<Distance<Pixels>> = "input_size.x".into();
+                let height : Var<Distance<Pixels>> = "input_size.y".into();
+                let shape  = Rect((&width,&height));
+                let shape  = shape.fill(bg_color);
```
wdanilo, line 50: you dont need all of these. Just use Plane

src/rust/ide/src/view/project_name.rs
```diff
@@ -4,19 +4,60 @@ use crate::prelude::*;
...
+                shape.into()
+            }
+        }
+}
+
 /// The project name's view used for visualizing the project name and renaming it.
 #[derive(Debug,Clone,CloneRef)]
 pub struct ProjectName {
+    network            : frp::Network,
+    main_area          : component::ShapeView<shape::Shape>,
```
wdanilo, line 60: View should not contain controller

src/rust/ide/src/view/project_name.rs
```diff
@@ -58,18 +104,44 @@ impl ProjectName {
             project_name.text_field.set_content(&new_name);
             project_name.setup_center_alignment();
         });
+
+        let network      = &self.network;
+        let project_name = self.clone_ref();
+        frp::extend! { network
+            def _foo = self.main_area.events.mouse_over.map(f_!(project_name.highlight_text()));
+            def _foo = self.main_area.events.mouse_out .map(f_!(project_name.darken_text()));
```
wdanilo, line 112: `_foo` is bad name. In such case use eval. And use animations - they will change this frp net

src/rust/ide/src/view/project_name.rs
```diff
@@ -31,21 +72,26 @@ impl ProjectName {
...
+        let project_controller    = project_controller.clone();
+        let main_logger           = Logger::sub(&logger,"main_area");
+        let main_area             = component::ShapeView::<shape::Shape>::new(&main_logger,scene);
+        Self{main_area,network,logger,display_object,text_field,project_controller}.initialize()
     }
 
     fn initialize(self) -> Self {
+        self.text_field.set_content(&self.project_controller.project_name());
+        self.display_object.add_child(&self.text_field.display_object());
+        self.display_object.add_child(&self.main_area);
```
wdanilo, line 94: never use `display_object` explicitly. `self.display_object. ...` is bad. The same `(&self.text_field.display_object());` All things like that should ook just as `self.add_child(&self.text_field);` always.

src/rust/ide/src/view/project_name.rs
```diff
@@ -58,18 +104,44 @@ impl ProjectName {
             project_name.text_field.set_content(&new_name);
             project_name.setup_center_alignment();
         });
+
+        let network      = &self.network;
+        let project_name = self.clone_ref();
```
wdanilo, line 109: You are leaking memory here

src/rust/ide/src/view/project_name.rs
```diff
@@ -58,18 +104,44 @@ impl ProjectName {
...
+        let mut width = 0.0;
         self.text_field.with_mut_content(|content| {
             let mut line = content.line(0);
             if line.len() > 0 {
-                offset.x = -line.get_char_x_position(line.len() - 1) / 2.0;
+                width = line.get_char_x_position(line.len() - 1);
             }
         });
+        let offset = Vector3::new(-width/2.0,0.0,0.0);
+        let height = 16.0;
```
wdanilo, line 142: hardcoded literals

src/rust/ide/src/view/project_name.rs
notdanilo: @wdanilo oh, it works for `self` but not for `TextField`. `TextField`'s `DisplayObject' is inside a `RefCell`, so it can't implement `display::Object` to get its reference.

src/rust/lib/graph-editor/src/component/project_name.rs
```diff
@@ -0,0 +1,300 @@
...
+
+// =================
+// === Constants ===
+// =================
+
+const HIGHLIGHTED_TEXT_COLOR : color::Rgba = color::Rgba::new(1.0,1.0,1.0,1.0);
+const DARK_GRAY_TEXT_COLOR   : color::Rgba = color::Rgba::new(1.0,1.0,1.0,0.6);
+
+/// Default project name used by IDE on startup.
+pub const DEFAULT_PROJECT_NAME:&str = "Unnamed";
```
notdanilo, line 33: I don't like to keep it here.

src/rust/ide/src/constants.rs
```diff
@@ -16,7 +16,7 @@ pub const SERIALIZE_TO_JSON_EXPRESSION:&str = "x -> x.json_serialize";
 pub const PROJECT_MANAGER_ENDPOINT:&str = "ws://127.0.0.1:30535";
 
 /// Default project name used by IDE on startup.
-pub const DEFAULT_PROJECT_NAME:&str = "Untitled";
+pub const DEFAULT_PROJECT_NAME:&str = graph_editor::component::project_name::DEFAULT_PROJECT_NAME;
```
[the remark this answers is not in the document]
farmaazon: In my opinion, those two constants have different roles
1. In ide crate this is the name we create/open UnnamedUntitled project.
2. Here it is just the default string in "ProjectName" widget. It can be even empty, the controllers should set proper name by themself, not rely on the widget defaults.

Therefore I opt for keeping two constants separately. They can be even different - that's better, we will see when controllers forget to set the proper project name.

src/rust/lib/graph-editor/src/component/project_name.rs
```diff
@@ -0,0 +1,313 @@
...
+        self.setup_text_field_content();
+        self
+    }
+
+    fn reset_name(&self) {
+        info!(self.logger, "Resetting project name.");
+        self.setup_text_field_content();
+    }
+
+    fn setup_text_field_content(&self) {
```
mwu-tow, line 232: `setup` not the best word here

src/rust/lib/graph-editor/src/component/project_name.rs
mwu-tow: Likely should be replaced with some other placeholder string like `[UNKNOWN NAME]` or sth.

src/rust/ide/src/view/node_editor.rs
```diff
@@ -260,6 +275,24 @@ impl GraphEditorIntegratedWithControllerModel {
...
+            executor::global::spawn(async move {
+                if let Err(e) = project_controller.rename_project(&name).await {
+                    info!(logger, "Couldn't rename project: {e}");
+                    project_name.frp.reset_name.emit(());
+                }
+            });
+        }
+    }
+}
+
```
mwu-tow, line 295: missing nl

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -894,9 +896,9 @@ impl Deref for GraphEditorModelWithNetwork {
 }
 
 impl GraphEditorModelWithNetwork {
-    pub fn new<S:Into<Scene>>(scene:S, cursor:cursor::Cursor) -> Self {
+    pub fn new(world:&World, cursor:cursor::Cursor) -> Self {
```
wdanilo, line 899: Please, do not do this change. Scene was here on purpose. Passing world is way too generic.

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -970,16 +973,26 @@ pub struct GraphEditorModel {
...
         let logger         = Logger::new("GraphEditor");
         let display_object = display::object::Instance::new(&logger);
         let nodes          = Nodes::new(&logger);
 //        let visualizations = Stage::new(scene.clone_ref(), Logger::new("VisualisationCollection"));
         let edges          = default();
         let frp            = FrpInputs::new(network);
         let touch_state    = TouchState::new(network,&scene.mouse.frp);
-        Self {logger,display_object,scene,cursor,nodes,edges,touch_state,frp}//visualizations }
+        let project_name   = component::ProjectName::new(world);
+        scene.add_child(&display_object);
```
wdanilo, line 990: why this change? Its not responsibility of graph editor model to add itself to scene!

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -970,16 +973,26 @@ pub struct GraphEditorModel {
...
+        scene.add_child(&project_name);
```
wdanilo, line 991: this is also bad - a component, like graph editor model, should not add things to scene, which is parent of the components. You can add project name as child to graph editor, but not this way

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1451,6 +1463,14 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
     let outputs        = UnsealedFrpOutputs::new();
     let sealed_outputs = outputs.seal(); // Done here to keep right eval order.
 
+    // === Cancel project name editing ===
+
+    frp::extend! { network
+        eval inputs.cancel_project_name_editing((_) {
```
wdanilo, line 1469: use eval_ instead and write it one line without curly braces

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+use ensogl::animation::linear_interpolation;
+
+
+
+// =================
+// === Constants ===
+// =================
+
+const HIGHLIGHTED_TEXT_COLOR : color::Rgba = color::Rgba::new(1.0,1.0,1.0,1.0);
+const DARK_GRAY_TEXT_COLOR   : color::Rgba = color::Rgba::new(1.0,1.0,1.0,0.6);
```
wdanilo, line 31: this is not "dark gray" anymore. The name shouldreflect semantically what it is used for. Like non-active color, or something.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// Project name used as a placeholder in ProjectName view.
+pub const UNKNOWN_PROJECT_NAME:&str = "[UNKNOWN NAME]";
```
wdanilo, line 34: docs should tell when it is used in the gui and in which situations it can appear.

wdanilo: Moreover, it looks hacky, but I cannot tell anything how to fix it until thedocs tell when it is used.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+
+
+// =============
+// === Shape ===
+// =============
+
+mod shape {
```
wdanilo, line 42: what is this shape? this mod should be named like background or soemthing.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+// === FrpInputs ===
+// =================
+
+#[derive(Debug,Clone,CloneRef)]
+#[allow(missing_docs)]
+pub struct FrpInputs {
+    /// Rename the project.
+    pub rename     : frp::Source<String>,
+    /// Reset the project name to the one before editing.
+    pub reset_name : frp::Source<()>
```
wdanilo, line 65: `()` is thedefaultparametrization

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+// =================
+// === FrpInputs ===
+// =================
+
+#[derive(Debug,Clone,CloneRef)]
+#[allow(missing_docs)]
+pub struct FrpInputs {
+    /// Rename the project.
+    pub rename     : frp::Source<String>,
```
wdanilo, line 63: this should be named "name" , as this is input stream of names of this project.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo, line 65: this should be named "cancel_editing" instead.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+// ==================
+// === FrpOutputs ===
+// ==================
+
+#[derive(Debug,Clone,CloneRef)]
+#[allow(missing_docs)]
+pub struct FrpOutputs {
+    /// Emits the new project name when it's renamed.
+    pub renamed : frp::Source<String>
```
wdanilo, line 89: this should be named "name" . It doesn't matter from frp perspective fit changed because of renaming or other action - this is the updated name of the project.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// Create new FrpOutputs.
+    pub fn new(network:&frp::Network) -> Self {
+        frp::extend! {network
+            def renamed = source();
+        }
+        Self{renamed}
+    }
+}
+
+// ===========
```
wdanilo, line 102: spacing here and in other places

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+// === Frp ===
+// ===========
+
+#[derive(Debug,Clone,CloneRef,Shrinkwrap)]
+#[allow(missing_docs)]
+pub struct Frp {
+    #[shrinkwrap(main_field)]
```
wdanilo, line 109: for frp API dont use shjrinkwrap - create anual derefs so intellij will show hints. frps apis are often complex or nested so its cool to have hints for them.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+        Self{network,inputs,outputs}
+    }
+}
+
+impl Frp {
+    /// Create new Frp.
+    pub fn new() -> Self {
+        default()
+    }
+}
```
wdanilo, line 129: just minor thing - lets always implement `new` and then implement Default as `new()`. This way if you have more methods they are in one block with constructor.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+// ==================
+// === Animations ===
+// ==================
+
+/// ProjectName's animations handlers.
+#[derive(Debug,Clone,CloneRef)]
+pub struct Animations {
+    highlight   : Animation<f32>,
+    positioning : Animation<Vector3<f32>>
```
wdanilo, line 141: rename to position

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+
+// ==================
+// === Animations ===
+// ==================
+
+/// ProjectName's animations handlers.
+#[derive(Debug,Clone,CloneRef)]
+pub struct Animations {
+    highlight   : Animation<f32>,
```
wdanilo, line 140: rename to opacity, because it is not really highlight

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+    positioning : Animation<Vector3<f32>>
+}
+
+impl Animations {
+    /// Create new animations handlers.
+    pub fn new(network:&frp::Network) -> Self {
+        let highlight   = Animation::<_>::new(&network);
```
wdanilo, line 147: you dont need ::<_> here

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+#[derive(Debug,Clone,CloneRef)]
+#[allow(missing_docs)]
+pub struct ProjectNameModel {
+    logger         : Logger,
+    animations     : Animations,
+    view           : component::ShapeView<shape::Shape>,
+    display_object : display::object::Instance,
+    text_field     : TextField,
+    project_name   : Rc<RefCell<String>>,
```
wdanilo, line 167: display object should be before view and text_field - just as a cleaning step, because it is their parent. Doesnt really matter fro the logicalperspective, but I'd prefer to keep cleaned drops here

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl ProjectNameModel {
+    /// Create new ProjectNameModel.
+    pub fn new(world:&World,frp:&Frp) -> Self {
+        let scene                 = world.scene();
+        let logger                = Logger::new("ProjectName");
+        let display_object        = display::object::Instance::new(&logger);
+        let font                  = scene.fonts.get_or_load_embedded_font("DejaVuSansMono").unwrap();
+        let size                  = Vector2::new(600.0,100.0);
```
wdanilo, line 178: hardocded constants here and there - moveall ofthem to section constants on top of the file

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+        let text_size             = 16.0;
+        let text_field_properties = TextFieldProperties{base_color,font,size,text_size};
+        let text_field            = TextField::new(&world,text_field_properties);
+        let view_logger           = Logger::sub(&logger,"view_logger");
+        let view                  = component::ShapeView::<shape::Shape>::new(&view_logger,scene);
+        let project_name          = Rc::new(RefCell::new(UNKNOWN_PROJECT_NAME.to_string()));
+        let renamed_output        = frp.outputs.renamed.clone();
+        let animations            = Animations::new(&frp.network);
+        Self{logger,view,display_object,text_field,project_name,renamed_output
+            ,animations}.initialize()
```
wdanilo, line 189: we name it `init` everythere, so lets use it here as well

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+        let offset = Vector3::new(-width/2.0,0.0,0.0);
+        let mut height = default();
+            self.text_field.with_content(|content| height = content.line_height);
+        self.view.shape.sprite.size.set(Vector2::new(width,height));
+        self.view.set_position(Vector3::new(0.0,-height/2.0,0.0));
+        self.animations.positioning.set_target_value(offset);
+    }
+
+    fn initialize(self) -> Self {
+        self.add_child(&self.text_field.display_object());
```
wdanilo, line 209: never, ever use `.disaply_object()` explicitelly - I add this comment to almost all PRs here.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+            // If the text edit callback is called, the TextEdit must be still alive.
+            let field_content = project_name.text_field.get_content();
+            let new_name      = field_content.replace("\n","");
+            if change.inserted == "\n" {
+                project_name.rename(&new_name);
+            }
+            // Keep only one line.
+            project_name.text_field.set_content(&new_name);
+            project_name.setup_center_alignment();
+        });
```
wdanilo, line 222: ehh, a lot of logic that is a hack - good its here, but should be removed after new text field is there. Add comment about it pls

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.setup_text_field_content();
+        self
+    }
+
+    fn reset_name(&self) {
+        info!(self.logger, "Resetting project name.");
+        self.setup_text_field_content();
+    }
+
+    fn setup_text_field_content(&self) {
```
wdanilo, line 233: this name doesnt tell whatthisfunction does.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.text_field.set_content(&self.project_name.borrow());
+        self.setup_center_alignment();
+    }
+
+    fn highlight_text(&self) {
+        self.animations.highlight.set_target_value(1.0);
+    }
+
+    fn darken_text(&self) {
```
wdanilo, line 242: this function does not darket the text - it fades it out. Moreover, it doesso only if not selected, so the name is misleading

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    fn setup_text_field_content(&self) {
+        self.text_field.set_content(&self.project_name.borrow());
+        self.setup_center_alignment();
+    }
+
+    fn highlight_text(&self) {
+        self.animations.highlight.set_target_value(1.0);
+    }
```
wdanilo, line 240: imo such functions should be just part of the frp definition - when reading frp block it would be easier to understand what it does instead of jumping to "model.something" one-liners. Its good tokeep in model more complex things, but things like just controlling animations, should be embedded directly in frp block so it would be clear what is happening there.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.text_field.set_base_color(base_color);
+    }
+
+    fn set_position(&self, value:Vector3<f32>) {
+        self.text_field.set_position(value);
+    }
+
+    fn rename(&self, name:impl Str) {
+        let name = name.into();
+        *self.project_name.borrow_mut() = name.clone();
```
wdanilo, line 259: why `name.clone()`?  Whydo you want to allocate if for example the argument was a string passed as ownership? Use `name.into()` - will be moreefficient

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl ProjectName {
+    /// Create a new ProjectName view.
+    pub fn new(world:&World) -> Self {
+        let frp     = Frp::new();
+        let model   = Rc::new(ProjectNameModel::new(world,&frp));
+        let network = &frp.network;
+        frp::extend! { network
+            eval model.view.events.mouse_over((_) {model.highlight_text()});
```
wdanilo, line 287: eval_ and do not usecurly pbraces in oneliners

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+            eval model.view.events.mouse_out ((_) {model.darken_text()});
+            eval frp.inputs.reset_name((_) {model.reset_name()});
+            eval frp.inputs.rename((name) {model.rename(name)});
+        }
+
+        // Animations
```
wdanilo, line 293: Make it sub-section stylecomment please

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        Self{frp,model}
+    }
+}
+
+impl display::Object for ProjectNameModel {
+    fn display_object(&self) -> &display::object::Instance {
+        &self.display_object
+    }
+}
```
wdanilo, line 308: wrong section

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: @wdanilo I have explained this before. `TextField` doesn't implement `display::Object` because of how it was designed.

src/rust/ide/lib/view/src/graph_editor.rs
notdanilo: @wdanilo `ProjectName` uses `TextField` which needs `World` to be created. This can be refactored out when we get the new `TextField` implementation. Sounds good?

src/rust/ide/lib/view/src/graph_editor.rs
notdanilo: I was testing something (`TextField` was displayed - with some errors - even if it wasn't added as a child of a display object) and I forgot to remove this line.

src/rust/ide/lib/view/src/graph_editor.rs
notdanilo: I had one main reason to do it: It was a workaround to display the `ProjectName` in the same position both for debug shapes and IDE.

As an action for your comment, I added it to `display_object` instead, but I fixed the position of `GraphEditor` on `IDE`. Which also fixes the connection lines not following the cursor.

I will need to ask @farmaazon and @mwu-tow to review this PR again.

connections

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: The doc now says:
`/// Project name used as a placeholder in ProjectName view when it's initialized.`

Please check this discussion: https://github.com/enso-org/ide/pull/639#discussion_r451362250

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: I renamed it to `update_text_field_content` as Michal suggested.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: Yeah, I thought about doing that, but keeping it like `eval_ model.view.events.mouse_over(model.fade_in_text());` is more descriptive.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: Because I need to move it in the 4th line:
```rust
        let name = name.into();
        *self.project_name.borrow_mut() = name.clone();
        self.update_text_field_content();
        self.name_output.emit(name);
```

src/rust/ide/lib/view/src/graph_editor.rs
wdanilo: Why cant we change the impl of old text field to get `Into<Scence>` instead?

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+mod background {
+    use super::*;
+
+    ensogl::define_shape_system! {
+            (style:Style, selection:f32) {
+                let bg_color = color::Rgba::new(0.0,0.0,0.0,0.000_001);
+                Plane().fill(bg_color).into()
+            }
+        }
+}
```
wdanilo, line 52: indentations

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+
+// =================
+// === FrpInputs ===
+// =================
+
+#[derive(Debug,Clone,CloneRef)]
+#[allow(missing_docs)]
+pub struct FrpInputs {
+    /// Rename the project.
+    pub name : frp::Source<String>,
```
wdanilo, line 64: please update docs

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+
+
+// ==================
+// === FrpOutputs ===
+// ==================
+
+#[derive(Debug,Clone,CloneRef)]
+#[allow(missing_docs)]
+pub struct FrpOutputs {
+    /// Emits the new project name when it's renamed.
```
wdanilo, line 89: these docs are not correct. It just emits name whenever it changes, but it isdefinition of frp. I believe thisdocs is notneeded.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+
+impl Frp {
+    /// Create new Frp.
+    pub fn new() -> Self {
+        let network = frp::Network::new();
+        let inputs  = FrpInputs::new(&network);
+        let outputs = FrpOutputs::new(&network);
+        Self{network,inputs,outputs}
+    }
+}
```
wdanilo, line 138: Ok, this needs to be written in guide, but I would bethankful if we will be using the following layout:
- struct definition
- explicit Deref impl if any
- methods definitions
- all other impls

The reason is that Deref is important while inspectingthe layout of structure, so youshould know the structimplements it immediately. Methods are your public API(with deref), while all other impls impleent additional utilsand there could be a ton of impls andyoudontwant to have methods 500 lines below struct

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+    display_object : display::object::Instance,
+    view           : component::ShapeView<background::Shape>,
+    text_field     : TextField,
+    project_name   : Rc<RefCell<String>>,
+    name_output    : frp::Source<String>,
+}
+
+impl ProjectNameModel {
+    /// Create new ProjectNameModel.
+    pub fn new(world:&World,frp:&Frp) -> Self {
```
wdanilo, line 182: use Into<Scene> instead

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+        let text_field            = TextField::new(&world,text_field_properties);
+        let view_logger           = Logger::sub(&logger,"view_logger");
+        let view                  = component::ShapeView::<background::Shape>::new(&view_logger, scene);
+        let project_name          = Rc::new(RefCell::new(UNKNOWN_PROJECT_NAME.to_string()));
+        let renamed_output        = frp.outputs.name.clone();
+        let animations            = Animations::new(&frp.network);
+        Self{logger,view,display_object,text_field,project_name,
+            name_output: renamed_output
+            ,animations}.init()
+    }
```
wdanilo, line 200: this is soooo ugly. Please, rename the var instead.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: I don't think this is the right explanation. It would be OK for me if in every place you would create TODOs, but there are no todos. Moreover, text field should define it. Add it there instead of such code here.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+    logger         : Logger,
+    animations     : Animations,
+    display_object : display::object::Instance,
+    view           : component::ShapeView<background::Shape>,
+    text_field     : TextField,
+    project_name   : Rc<RefCell<String>>,
+    name_output    : frp::Source<String>,
+}
+
+impl ProjectNameModel {
```
wdanilo, line 180: Soehow I feel your implementation is leaking memory. This is the reason I marked it now as "request changes". Please test what happens if you remove it from graph editor after some key is pressed. Will it be disposed?

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: That's not true at all. `emit` accepts reference

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,327 @@
...
+        let network = &frp.network;
+        frp::extend! { network
+            eval_ model.view.events.mouse_over(model.fade_in_text());
+            eval_ model.view.events.mouse_out (model.fade_out_text());
+            eval_ frp.inputs.cancel_editing(model.reset_name());
+            eval frp.inputs.name((name) {model.rename(name)});
+        }
+
+
+        // == Animations
```
wdanilo, line 312: this is not right section comment

src/rust/ide/src/view/node_editor.rs
```diff
@@ -167,6 +173,21 @@ impl GraphEditorIntegratedWithController {
...
             }));
         }
+
+
+        // === Project Renaming ===
+
+        let project_name = &model.editor.project_name;
+        frp::extend! {network
+            eval project_name.frp.outputs.name((name) {
+                model.rename_project(name.to_string());
```
wdanilo, line 183: why this api requires allocation?

src/rust/ide/src/view/node_editor.rs
```diff
@@ -271,6 +292,25 @@ impl GraphEditorIntegratedWithControllerModel {
...
+
+impl GraphEditorIntegratedWithControllerModel {
+    fn rename_project(&self, name:String) {
+        if self.project_controller.project_name().to_string() != name {
+            let project_controller = self.project_controller.clone_ref();
+            let project_name       = self.editor.project_name.clone_ref();
+            let logger             = self.logger.clone_ref();
+            executor::global::spawn(async move {
+                if let Err(e) = project_controller.rename_project(&name).await {
+                    info!(logger, "Couldn't rename project: {e}");
```
wdanilo, line 305: this is not a sentence

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
farmaazon: Well, I think it is. The only way to implement it AFAIR is to completely refactor TextField - and you are doing it anyway.

Look at commented code https://github.com/enso-org/ide/blob/main/src/rust/ensogl/src/display/shape/text/text_field.rs#L511 To implement this "From" properly the TextField cannot be under the shared! macro. But as I said that involves much refactoring, which is worthless if you replace that textfield anyway.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
farmaazon: As a side note: this is a drawback of the philosophy of getting "AsRef/Into" everywhere - it is not clear when function can take reference, or need to take the whole value. If the function would just take a reference instead of impl ToRef, the compiler would point here, that functions needs reference only and it would be clear that clone is not needed here.

src/rust/ensogl/src/display/shape/text/text_field.rs
```diff
@@ -126,6 +126,12 @@ shared! { TextField
     }
 
     impl {
+        /// Run `callback` with the underlying `&TextFieldContent`.
+        pub fn with_content<F:FnMut(&TextFieldContent)>(&self, callback:F) {
+            let mut callback = callback;
+            callback(&self.content);
+        }
+
```
farmaazon, line 134: This function seems to be written to avoid restrictions of shared! macro... which restrictions were put there for reason. Now it is unsafe, because nobody is warned that callback cannot use any function of text_field, or we get double borrow panic.

I see you need it only for getting line height - please add function for getting height instead.

The same is for `with_mut_content`, however it has no simple replacement. It should at least has a comment which will warn about potential dangers.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: @farmaazon very interesting observation indeed.

src/rust/ide/lib/view/src/graph_editor.rs
wdanilo: Always when testing add empty `// fixme` comments not to forget about it!

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: Ok, I get this rerasoning. And although I really like it, I believe that writing `alpha.set_target_value(FADE_OUT_ALPHA)` is not much worse. I really think that in case of emitting events from within eval, we should keep it as part of frp for clearness purposes.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: Ok, it looks hacky, just like internal details showing to the user. Please change it to "Unknown". Is it ok @mwu-tow ?

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: @farmaazon I'm not sure if you have read my whole comment or only the first sentence. I told I'm ok with leaving it IFF in every usage place there would be fixme comment in order not to forget updating it after new text field kicks in

notdanilo: FIXME notes everywhear

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: My problem is with its counterpart:

```rust
    fn fade_out_text(&self) {
        if !self.text_field.is_focused() {
            self.animations.opacity.set_target_value(0.0);
        }
    }
```

`fn fade_out_text` has more logic to it. The current `TextField` doesn't implement focus events and I have to poll for its state. I will keep both functions in there with a `TODO` note to make use of the newest `TextField`'s frp.

src/rust/ide/lib/view/src/graph_editor.rs
notdanilo: @wdanilo because the old `TextField` needs to get `FocusManager` from `World`.

notdanilo: I changed the implementation of TextField to get `Into<Scene>`. 1h later :)

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: The original issue was fixed.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: omg idk what happened here

src/rust/ide/src/view/node_editor.rs
notdanilo: @wdanilo because `name` is moved to an async closure inside of `rename_project`.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: @wdanilo sorry, maybe I am too sleepy. But what's the expected action here?

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: Yes, I just tested it and it's dropped. It's tested on the branch `wip/dg/project_renaming_view_drop_test`.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: Relayouting the code according ot the poitns there - write to me on Discord if that's still not clear :)

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
wdanilo: Still, I believe it leaks memory. Look at `let project_name = self.clone_ref();` which you then pass to a closure managed by `self` - there is a loop.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    fn update_text_field_content(&self) {
+        self.text_field.set_content(&self.project_name.borrow());
+        self.update_center_alignment();
+    }
+
+    fn fade_in_text(&self) {
+        self.animations.opacity.set_target_value(1.0);
+    }
```
wdanilo, line 243: We've been talking about that in reviews earlier - could you just use these calls directly in FRP instead of delegating them this way to model?

src/rust/ide/src/view/node_editor.rs
wdanilo: Hmm, I still do not understand it. No matter what happens inside `rename_project`, why its signature is designed to force passing the ownership ?

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: Oh, done. I need to relayout https://github.com/enso-org/ide/pull/662 to follow this guideline.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
notdanilo: @wdanilo you seem to be right. But I have two notes about it:

1. I don't know how `CloneRef` aren't implemented. But when I run the test on the `wip/dg/project_renaming_view_drop_test` branch, the drop function is called. But disregard this note, the following one is more important.
2. This workaround for keeping the text field in a single line and checking if the change was committed with a newline is the one responsible for leaking memory. There is no easy fix for it without refactoring `TextField`. My suggestion is to leave a `FIXME` note to implement it correctly when the new `TextField` component arrives.

And one last thing: It would be cool to have two different components like `MultiLineText` and `SingleLineText`, with one being a specialization of the other. I see many places where `SingleLineText` could be useful.

src/rust/ide/lib/view/src/graph_editor/component/project_name.rs
```diff
@@ -212,6 +212,10 @@ impl ProjectNameModel {
         let project_name = self.clone_ref();
         //FIXME[dg]: This section to check newline and keep TextField in a single line is hacky
         // and should be removed once the new TextField is implemented.
+        // Also: Taking TextField reference inside a closure managed by itself is a reference
+        // loop, which means we are leaking memory. This cannot be fixed without refactoring
+        // TextField. This code will not suffer of memory leak when using the events emitted from
+        // TextField 2.0` implementation.
```
farmaazon, line 218: But you can just pass the project_name as a weak reference to the closure and avoid leaking memory, cannot you?

Issue #570, opened 2020-06-16
Edge lines create loop when dragging from input port


MichaelMauderer:

farmaazon:
During Backlog Refinement we agreed, that:
1. Refactoring should be moved to another task.
2. Those bugs are not essential to be done in this release.

PR #640, opened 2020-07-05
Disconnect / reconnect edge source and target by clicking on edges. (Full) 

MichaelMauderer:
### Pull Request Description
Implement full edge splitting.

### Important Notes
Note comments and discussions from the previous partial PR: https://github.com/enso-org/ide/pull/574

### Checklist


wdanilo:
Hi, amazing progress! It works smoothly and a lot of issues are gone. Still, there are few:

1. Clicking on edge disconnects the wrong part:
    
2. Clicking on output port and then input port creates a new edge while it should not.
    
3. You can ... connect edge to edge now, and even create edge loop!
    
4. There is edge precedence error, which is strange. When dragging an edge, all of its parts should e drawn in front layer (at least this is behavior before this PR). Now, its not the case anymore. It works for output-only attached edges, but not for input-only attached ones:
    
    
5. There are some strange "dots" appearing around arrow when hovering the edge:
    
6. The arrow top has strange artifacts when hovering:
    
7. The division between top and bottom edge part is not in the middle of the nodes:
    
8. Plus, create even wider hover area for edges please, so this situation would also hover the edge:
    image

wdanilo:
I was checking the status of today implementation and here are I think new cases I found:

9. Desynchronised position on hover when zooed in
    

10. This glitch:
    

11. This actually should just not be possible (and I think we overlooked it) - when dragging edge with only output port connected, the mouse should not get snapped to inputs because it is impossible to create input <-> input connection and it is deeply confusing.
    

12. When starting a connection when zoomed, the connection is drawn not to the mouse position if mouse was not yet moved.
    

13. After starting a connection it is impossible to click in the background to drop it, probably because of the glitches that are seen there. I suspect this might be issue introduced by my connection implementation and handling corner cases of very short connections - so Im documenting it here, but if this is my issue (please check on aster if it behaves the same way for short connections), I'll fix it as another ticket.
    

14. This glitch

wdanilo:
Plus, I was browsing the code fast (not real review yet), and I have a few comments - please make sure that:
1. You do not use `.display_object()` explicitly.
2. Spacing is correct (between sections, function args, etc)
3. I'd be thankful if you make sure that all docs of the abstractions you introduce would contain clear descriptions regarding **what they are used for**. So for example, `SemanticSplit` - I see what it does, but I was unable to really answer the question of why this abstraction has to exist and what it really is used for. Such info in the documentation would drastically help understand what's going.

MichaelMauderer:
@wdanilo 
* 1 fixed
* 2 fixed, but we should also set up an issue to do the correct edge handling for the multi-output port shapes.
* 3 fixed
* 4 fixed
* 5 fixed
* 6 fixed 
* 7 fixed 
* 8 fixed 
* 9 This is also on `main` and I will add this to a new issue about zooming breaking interactions (#635). 
* 10 fixed
* 11 This is partially fixed but comes at the trade-off that either we have no snapping when not doing edge dragging or the behavior you describe. A full fix should go into a bugfix issue because it will need some additional work on handling the cursor style from the port manager (#659). 
* 12 fixed
* 13 The dropping of edges seems by clicking on the background seems to work for me. I will open a new issue for the shape glitch.  (#658)
* 14 fixed
* 15 fixed

wdanilo:
Hi! I'm happy it's getting polished, but unfortunately, a set of new bugs appeared (these are regressions as were not happening earlier as far as I know):

16. It is impossible now to create multiple output edges - clicking on output port when an output edge exists, does nothing.
17. There is a new glitch (was not present before):
    
18. The edge highlighting is inverted - it just should work other way around. When mouse is close to the input port and you click on the edge, the part of the edge close to the port should stay active, not the second part. I realized that after seeing issue 19 (below).
19. When input is connected, clicking on it does nothing. Instead, when you click on input port, it should just disconnect existing edge (it should be consistent with the edge highlight) - when you move over the edge, and then you hover the port, the input port should behave just like "the last part of the edge".
20. This happens now (this makes me especially worried as it looks like you are keeping a mutable-like state on the edges, which you should not):
    
21. I cannot connect edges anymore in this configuration (moreover, a lot of "jumping" is seen here, as in the issue 20):
    
22. This glitch (happens with 2 nodes connected together too):
    
22. This glitch (I bet it was reported already, but cannot find the number for it):
    

**Moreover, few issues reported earlier are not yet fixed. Please be sure to fix and TEST all reported issues before making a PR - this would save us a lot of time. In particular:**

4. Still seeing the glitch:
    
7. The division point is still not in the middle of the distance between nodes, see the attached image - the mouse is placed at the division point, and the rectangles have the same height:
    rects
9. This still happens - but it's ok, as you described, let's make another issue from that.
11. This issue still exists - ok, we can make it a separate issue, but the solution for that will be simple - I believe a simple FRP gate would do the trick. But, let's keep it separate.
15. It's not fixed:

MichaelMauderer:
@wdanilo 

**5** Fixed. According to my measurements, this is halfway between the edges of the nodes now. (Which means one line is longer, as it overlaps the node)
**16** Fixed.
**4 + 8 + 17 + 20**: This is all related to my initial fix for (4) and shows that there is more work to be done on changing the shape of the edge to have the correct ordering behavior in all those cases. Since this is not really related to edge disconnecting I have opened issue #672 to deal with those. I've added what I found out about how to resolve the issues there, too.
**18 + 19** This should be a new issue, as it is new behavior unrelated to the split/highlight.disconnect of this PR (and I want to avoid growing this PR even more).
**21 + 8r:** Fixed. (The original 8 was a different underlying issue) There is a trade-off here: the arrow must only appear if the mainline is as high as the arrow. Otherwise we end up in a bad shape configuration where there is a discontinuity in the line.
**22** Fixed by limiting the snapping to the visible circle quadrant.
**23** Fixed. Was caused by hovering information being used when detached. Fixed by not allowing the hover split when edge is detached.

MichaelMauderer:
One more thing that needs fixing before re-review though

Peek 2020-07-17 12-48

MichaelMauderer:
@wdanilo This is ready for another look.

wdanilo:
Michael, I'm happy to see it ats this stage, good work! Edges are fun :) 

Replying to above message:

- **4 + 17 + 20** Ok, lets do it as separate issue. I have linked it to https://github.com/enso-org/ide/issues/570.
- **18** I do not agree this is a separate issue. The current highlighting is wrong - it is inverted. Just invert it - highlight the other part of the edge.
- **19** I agree this could be a separate issue – please create it and link it to https://github.com/enso-org/ide/issues/570


Moreover, I found new things:

24. The height of the edge jumps when hovering input port - it was not behaving like that:
    

25. This glitch is similar to 15 (blinking of the edge) - to reproduce it - place the left node just SLIGHTLY below the right node:
    

16. This is pretty bad - if nodes are on the same Y-axis, its impossible to disconnect one of them. HOWEVER, let's make it part of https://github.com/enso-org/ide/issues/570 as well.



Moreover, I found a new thing that is not implemented but should be part of this task. However, as I discovered it really late, let's make it part of https://github.com/enso-org/ide/issues/570. The thing is that we should support 2 modes of disconnecting - clicking, and press-nad-drag mode, just like on input / output ports. Currently, edges support only first mode.


**So to sum this up, please fix 18, and 24, 25, and then let's merge it :)**

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1830,29 +1966,38 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
-    edge_refresh_cursor_pos_no_hover <- edge_refresh_cursor_pos.gate_not(&is_hovering_output);
+    edge_refresh_on_node_hover        <- all(edge_refresh_cursor_pos,hover_node).gate(&is_hovering_output);
+    edge_refresh_cursor_pos_no_hover  <- edge_refresh_cursor_pos.gate_not(&is_hovering_output);
+    edge_refresh_cursor_pos_on_hover  <- edge_refresh_on_node_hover.map(|(pos,_)| *pos);
 
-    eval edge_refresh_cursor_pos ((position) {
-        edges.detached_target.for_each(|id| {
+    refresh_target      <- any(&edge_refresh_cursor_pos_on_hover,&edge_refresh_cursor_pos_no_hover);
+    refresh_source      <- edge_refresh_cursor_pos_no_hover.map(|pos| *pos);
+    snap_source_to_node <- edge_refresh_on_node_hover.map(|(_,target)| target.clone());
```
wdanilo, line 1989: you can use `._1()` instead

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1830,29 +1966,38 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
-    edge_refresh_on_node_hover       <- all(edge_refresh_cursor_pos,hover_node);
-    edge_refresh_cursor_pos_no_hover <- edge_refresh_cursor_pos.gate_not(&is_hovering_output);
+    edge_refresh_on_node_hover        <- all(edge_refresh_cursor_pos,hover_node).gate(&is_hovering_output);
+    edge_refresh_cursor_pos_no_hover  <- edge_refresh_cursor_pos.gate_not(&is_hovering_output);
+    edge_refresh_cursor_pos_on_hover  <- edge_refresh_on_node_hover.map(|(pos,_)| *pos);
 
-    eval edge_refresh_cursor_pos ((position) {
-        edges.detached_target.for_each(|id| {
+    refresh_target      <- any(&edge_refresh_cursor_pos_on_hover,&edge_refresh_cursor_pos_no_hover);
+    refresh_source      <- edge_refresh_cursor_pos_no_hover.map(|pos| *pos);
```
wdanilo, line 1988: what does this line do? If im correct, this does not do anything.

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1868,7 +2013,7 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
         });
     });
 
-    eval edge_refresh_on_node_hover ([nodes,edges,model]((_,target)) {
+    eval snap_source_to_node ([nodes,edges,model]((target)) {
```
wdanilo, line 2016: unnecesary parenthesis - this should be catched by linter btw

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+
+
+fn up() -> Vector2<f32> {
+    Vector2(1.0,0.0)
+}
+
+fn point_rotation(point:Vector2<f32>) -> nalgebra::Rotation2<f32> {
+    nalgebra::Rotation2::rotation_between(&point,&up())
+}
+
```
wdanilo, line 57: I know it was my change, but I did not clean my code. This should be in a section imo

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+
+// =======================
+// === Hover Extension ===
+// =======================
+
+/// Add an invisible hover area to the provided shape. The base shape should already be colored
+/// otherwise coloring it later will also color the hover area.
+fn hover_area(base_shape:AnyShape, size:Var<Pixels>) -> AnyShape {
+    let hover_area = base_shape.grow(size).fill(color::Rgba::new(0.0,0.0,0.0,0.000_001));
```
wdanilo, line 190: this sohuld use the color constant defined somewhere above

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
 
-macro_rules! define_corner_start {($($color:tt)*) => {
+/// Joint definition
+pub mod joint {
+    use super::*;
+
+    ensogl::define_shape_system! {
+            (
+            ) {
+                let radius : Var<Pixels> = "min(input_size.y,input_size.x)".into();
```
wdanilo, line 284: wouldn't these be always the same ? Is there any place where we keep the input_size not rectangular? If no, let's not waste gpu computing power ;)

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+                let width  = &LINE_WIDTH.px();
+                let shape  = corner_base_shape(&radius,width,&angle,&start_angle);
 
                 let shadow_size = 10.px();
-                let n_radius = &shadow_size + 1.px() * dim.y();
-                let n_shape  = Rect((&shadow_size*2.0 + 2.px() * dim.x(),&n_radius*2.0)).corners_radius(n_radius);
-                let n_shape  = n_shape.fill(color::Rgba::new(1.0,0.0,0.0,1.0));
+                let node_radius = &shadow_size + 1.px() * dim.y();
+                let node_shape  = Rect(
+                    (&shadow_size*2.0 + 2.px() * dim.x(),&node_radius*2.0)).corners_radius(node_radius);
```
wdanilo, line 329: just make vars `node_width` and `node_height` :)

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+                let node_shape  = node_shape.fill(color::Rgba::new(1.0,0.0,0.0,1.0));
```
wdanilo, line 330: again, this is refactored somewhere to constants. Lets replace all such colors by the constant.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+                let end_angle     =  start_angle + self.shape.angle.get();
+                let upper_bound   = start_angle.max(end_angle);
+                let lower_bound   = start_angle.min(end_angle);
+
+                let correct_quadrant = lower_bound < vector_angle && upper_bound > vector_angle;
+                if correct_quadrant {
+                     Some(Vector2::new(closest_point.x, closest_point.y))
+                } else {
+                    None
+                }
```
wdanilo, line 386: As we've been discussing - this is too complex and this is just a hotfix. There should be TODO note about it with link to an issue about it before this PR is merged. (the TODO should be in every such place)

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+
+            fn snap_local(&self, point:Vector2<f32>) -> Option<Vector2<f32>> {
+                let radius = self.shape.radius.get();
+                let center = Vector2::zero();
+                let point_to_center = point.xy() - center;
+
+                let closest_point = center + point_to_center / point_to_center.magnitude() * radius;
+                let vector_angle  = -nalgebra::Rotation2::rotation_between(&Vector2::new(0.0, 1.0),&closest_point).angle();
+                let start_angle   =  self.shape.start_angle.get();
+                let end_angle     =  start_angle + self.shape.angle.get();
```
wdanilo, line 459: 2 spaces after `=`

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+            fn events(&self) -> &ShapeViewEvents{
+                &self.events
+            }
+
+            fn normal_local(&self, point:Vector2<f32>) -> nalgebra::Rotation2<f32> {
+                point_rotation(point)
+            }
+
+            fn snap_local(&self, point:Vector2<f32>) -> Option<Vector2<f32>> {
+                let radius = self.shape.radius.get();
```
wdanilo, line 452: why this does not need to have allow(depraecated)?

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+            fn events(&self) -> &ShapeViewEvents{
+                &self.events
+            }
+
+            fn normal_local(&self, _:Vector2<f32>) -> nalgebra::Rotation2<f32> {
+                nalgebra::Rotation2::new(0.0)
+            }
+
+            fn snap_local(&self, point:Vector2<f32>) -> Option<Vector2<f32>> {
+                #[allow(deprecated)]
```
wdanilo, line 513: all such places needs to be propelry documented. No one will remember to fix that in 1 month from now without proper documentation and link to an issue.

src/rust/ensogl/src/display/symbol/gpu/geometry/compound/sprite.rs
```diff
@@ -119,6 +119,18 @@ impl Size {
...
+// === Getters ===
+
+#[allow(missing_docs)]
+impl Size {
+    #[deprecated(note = "This will be removed without replacement soon.")]
+    pub fn get(&self) -> Vector2<f32>{
+        if self.hidden.get() { self.value.get() }
+        else                 { self.attr.get()  }
+    }
+}
```
wdanilo, line 131: Let's add more context here. Someone reading such things should understand WHY it is deprecated. We want to keep here `Var`s instead of normal values, so `get` will not result in value soon.

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -960,10 +960,69 @@ impl GraphEditorModelWithNetwork {
...
+            self.frp.some_edge_targets_detached.emit(());
+        }
+        edge_id
+    }
+
+    fn new_edge_from_input
+     (&self,edge_click:&frp::Source<EdgeId>,edge_over:&frp::Source<EdgeId>,
+      edge_out:&frp::Source<EdgeId>) -> EdgeId {
+        let edge_id = self.create_edge(edge_click,edge_over,edge_out);
+        let first_detached = self.edges.detached_source.is_empty();
```
wdanilo, line 1018: alignment -- in other places as well, please make a fast self-review on github and check for all such places.

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -960,10 +960,69 @@ impl GraphEditorModelWithNetwork {
...
+        self.edges.detached_target.insert(edge_id);
+        if first_detached {
+            self.frp.some_edge_targets_detached.emit(());
+        }
+        edge_id
+    }
+
+    fn new_edge_from_input
+     (&self,edge_click:&frp::Source<EdgeId>,edge_over:&frp::Source<EdgeId>,
+      edge_out:&frp::Source<EdgeId>) -> EdgeId {
```
wdanilo, line 1016: in such case please use param by line

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
+                let end_angle     =  start_angle + self.shape.angle.get();
+                let upper_bound   = start_angle.max(end_angle);
+                let lower_bound   = start_angle.min(end_angle);
+
+                let correct_quadrant = lower_bound < vector_angle && upper_bound > vector_angle;
+                if correct_quadrant {
+                     Some(Vector2::new(closest_point.x, closest_point.y))
+                } else {
+                    None
+                }
```
wdanilo, line 468: in such cases use always `as_some` from boolinator exported in proelude

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
-                let height     = height - (2.0 * PADDING).px();
-                let triangle   = Triangle(width,height);
-                let offset     = (LINE_WIDTH/2.0).px();
-                let triangle_l = triangle.translate_x(-&offset);
-                let triangle_r = triangle.translate_x(&offset);
-                let shape      = triangle_l + triangle_r;
-                let shape      = shape.fill(color::Rgba::from($($color)*));
+                let focus_split_angle  = focus_split_angle.into();
+                let focus_split_center = focus_split_center.px();
+                let shape = Triangle(width-1.0.px(),height-1.0.px());
```
wdanilo, line 532: what are these `-1.px()` ? Btw you can write `1.px()` instead of `1.0.px()`. They should be named variable telling what they are for.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -251,23 +672,45 @@ pub mod back {
...
             /// Constructor.
             pub fn new(logger:Logger, scene:&Scene) -> Self {
                 let display_object = display::object::Instance::new(&logger);
-                $(let $field = component::ShapeView::new(Logger::sub(&logger,stringify!($field)),scene);)*
-                $(display_object.add_child(&$field);)*
-                Self {logger,display_object,$($field),*}
+                $(let $field = component::ShapeView::new(
+                    Logger::sub(&logger,stringify!($field)),scene);)*
+                    $(display_object.add_child(&$field);)*
+                let mut shape_view_events:Vec<ShapeViewEvents> = Vec::default();
```
wdanilo, line 694: a very very minor thing. Just ocnidering our alignment rules, it's always better to write such things as `= Vec::<ShapeViewEvents>::default()` - then left side of `=` is shorter. Doesnt matter here but I thought I will mention it.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
+            LayoutState::TopCenterLeftLoop  => false,
+            LayoutState::DownLeft           => true,
+            LayoutState::DownRight          => true,
+        }
+    }
+
+    fn is_input_above_output(self) -> bool {
+        !self.is_output_above_input()
+    }
+
```
wdanilo, line 847: rm nl

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
+/// The semantic split, splits the sub-shapes according to their relative position from `OutputPort`
+/// to `InputPort` and allows access to the three different groups of shapes:
+///   - shapes that are input side of the split;
+///   - shapes that are at the split;
+///   - shapes that are output side of the split.
+///
+/// Note that "at the split" also includes the shapes adjacent to the actual split because they
+/// need to be treated as if they were at the split location to avoid glitches at the shape
+/// boundaries.
+/// // FIXE: please, tell more about these glitches - I do not understand why they happen after this description
```
wdanilo, line 865: please fix this fixme :) OH I MADE TYPO HERE! Sorry :)

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
+                split_index = Some(index);
+                break
+            }
+        }
+        let split_index = split_index?;
+
+        Some(SemanticSplit {ordered_part_ids,split_index})
+    }
+
+    /// Return a ordered vector that contains the ids of the shapes in the order they appear in the
```
wdanilo, line 908: a -> an

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
+    /// edge. Shapes that are to be handled as in the same place, are binned into a sub-vector.
```
wdanilo, line 909: Are you sure you wanted to use the "binned" verb here? My diet tells it means that something is put in a bin, but this could also be my weak english

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
+    /// focus on or all focus off.
+    fn semantically_binned_edges(edge_data:&EdgeModelData) -> Vec<Vec<display::object::Id>> {
+        let front  = &edge_data.front;
+        let back   = &edge_data.back;
+        let layout = edge_data.layout_state.get();
+        match layout {
+            LayoutState::DownLeft | LayoutState::DownRight => {
+                vec![
+                    vec![EdgeShape::id(&front.side_line),  EdgeShape::id(&back.side_line)  ],
+                    vec![EdgeShape::id(&front.corner),     EdgeShape::id(&back.corner)     ],
```
wdanilo, line 928: I was writing in my previous PR that this is incorrect. `front.corner` and `back.corner` do not "occupy the same space", as docs tell. Front corner is the part that is visible next to node, while back corner is a very short part which fills the hole between the shape and `front.corner`. They are visible at the same time, always. - see the definition of the shapes - they differ . One is `ring - node_shape` and one is `ring * node_shape`.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
+}
+
+#[allow(missing_docs)]
+impl ShapeViewEventsProxy {
+    pub fn new(network:&frp::Network) -> Self {
+        frp::extend! { network
+            def on_mouse_over = source();
+            def on_mouse_out  = source();
+            def on_mouse_down = source();
+
```
wdanilo, line 1009: don't use defs pls . Use <- instead

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
+            mouse_down <- on_mouse_down.constant(());
+            mouse_over <- on_mouse_over.constant(());
+            mouse_out  <- on_mouse_out.constant(());
```
wdanilo, line 1012: the type of `on_mouse_down` is `()` so this constant doesn't make sense here. If you want to convert from `Source` to `Stream`, use `into` instead.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,37 +779,274 @@ pub fn sort_hack_2(scene:&Scene) {
...
-    pub target_attached           : frp::Source<bool>,
-    pub redraw                    : frp::Source<()>,
+    pub source_width    : frp::Source<f32>,
+    pub source_height   : frp::Source<f32>,
+    pub target_position : frp::Source<Vector2>,
+    pub target_attached : frp::Source<bool>,
+    pub source_attached : frp::Source<bool>,
+    pub redraw          : frp::Source,
+
+    pub hover_position  : frp::Source<Option<Vector2<f32>>>,
```
wdanilo, line 1031: Few lines above we are using `Vector2` and here we are using `Vector2<f32>` - lets make it consistent and lets skip the parametrization

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -432,16 +1125,39 @@ impl Edge {
...
+            // eval input.source_attached ((t) source_attached.set(*t));
+            eval input.target_attached ((t) target_attached.set(*t));
+            eval input.source_width    ((t) source_width.set(*t));
+            eval input.source_height   ((t) source_height.set(*t));
+            eval input.hover_position  ((t) hover_position.set(*t));
+
+            eval shape_events.on_mouse_over ((id) {
+                hover_target.set(Some(*id));
+            });
+            eval shape_events.on_mouse_out  ((_)  {
```
wdanilo, line 1154: use eval_

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -432,16 +1125,39 @@ impl Edge {
...
+            // FIXME This should be enabled for #672
+            // eval input.source_attached ((t) source_attached.set(*t));
+            eval input.target_attached ((t) target_attached.set(*t));
+            eval input.source_width    ((t) source_width.set(*t));
+            eval input.source_height   ((t) source_height.set(*t));
+            eval input.hover_position  ((t) hover_position.set(*t));
+
+            eval shape_events.on_mouse_over ((id) {
+                hover_target.set(Some(*id));
+            });
```
wdanilo, line 1153: make such things one-liners and do not use curly braces here

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -513,11 +1253,33 @@ impl EdgeModelData {
...
+        // let fully_attached   = self.target_attached.get() && self.source_attached.get();
+        let fully_attached          = self.target_attached.get();
+        let node_half_width         = self.source_width.get() / 2.0;
+        let target_node_half_height = NODE_HEIGHT / 2.0;
+        let source_node_half_height = self.source_height.get() / 2.0;
+        let source_node_circle      = Vector2(node_half_width- source_node_half_height, 0.0);
+        let source_node_radius      = source_node_half_height;
+
+        // === Update Highlights ===
+
```
wdanilo, line 1266: such things should have 2 enters above and one below according to our style guide

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -573,6 +1335,16 @@ impl EdgeModelData {
         let downward_far   = -target.y > min_len_for_non_curved_line || target_is_below_node;
         let is_down        = if is_flat_side {downward_flat} else {downward_far};
 
+        // === Layout State ===
+        // Initial guess at our layout. Will be refined for some edge cases when we have more
+        // layout informaiton.
```
wdanilo, line 1340: What does it mean "when we have more layout information"? I mean, this should be explained better here - what is to be refined and what info is missing at this point? Just add few words about it - with an example info that is missing here that will be known later

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -573,6 +1335,16 @@ impl EdgeModelData {
...
+        let state = match (is_down,(side < 0.0)) {
+            (true,true) => LayoutState::DownLeft,
+            (true,false) => LayoutState::DownRight,
+            (false,true) => LayoutState::UpLeft,
+            (false,false) => LayoutState::UpRight,
+        };
```
wdanilo, line 1346: alignment

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -884,7 +1665,206 @@ impl EdgeModelData {
...
+    /// Indicates whether the height difference between input and output is too small to  use the
+    /// y value to assign the `EndDesignation` for a given point.
+    fn input_and_output_y_too_close(&self) -> bool {
+        let target_y = self.position().y;
+        let source_y = self.target_position.get().y;
+        let delta_y  = target_y - source_y;
+        delta_y > 0.0 && delta_y < MIN_SOURCE_TARGET_DIFFERENCE_FOR_Y_VALUE_DISCRIMINATION
+    }
+
+    // FIXE: several lines are too long.
```
wdanilo, line 1727: please fix this "FIXE" (my keyboard doesnt like the key "m" :( )

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -884,7 +1665,206 @@ impl EdgeModelData {
...
+            (true, PortType::InputPort)   => 0.0,
+            (true, PortType::OutputPort)  => 2.0 * RIGHT_ANGLE,
+        }
+    }
+
+    /// These corrections are needed as sometimes shapes are in places that lead to inconsistent
+    /// results, e.g., the side line leaving the node from left/right or right/left. The shape
+    /// itself does not have enough information about its own placement to determine which end
+    /// is pointed towards the `Target` or `Source` part of the whole edge. So we need to account
+    /// for these here based on the specific layout state we are in.
```
wdanilo, line 1759: Could you please describe an example case preferably with ascii-art here? It would be sooooo helpful to understand it. Now I understand what is happening here, but I bet in couple of months this will be hard to follow.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -884,7 +1665,206 @@ impl EdgeModelData {
...
+    fn get_cut_angle_correction(&self, shape_role:ShapeRole) -> f32 {
+        let layout_state = self.layout_state.get();
+
+        let flip = 2.0 * RIGHT_ANGLE;
+
+        // These rules are derived from the algorithm in `redraw`. In some layout configurations
+        // shapes are inverted top/down or left/right and we need to apply the appropriate
+        // corrections here. Sometimes these are just the side-effect of some layouting mechanics
+        // without visual justification (e.g., the `PortLine` sometimes ends up with a negative
+        // height and is thus flipped upside down.
```
wdanilo, line 1769: super cool

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -884,7 +1665,206 @@ impl EdgeModelData {
...
+    /// FocusSplit the shape at the given `position` and highlight the given `EndDesignation`. This
+    /// might fail if the given position is too far from the shape.
+    fn try_enable_focus_split
+    (&self, position:Vector2<f32>, focus_shape_id:display::object::Id, part: PortType)
+    -> Result<SnapTarget, ()>{
+        let snap_data      = self.try_point_snap(position,focus_shape_id).ok_or(())?;
+        let semantic_split = SemanticSplit::new(&self,snap_data.target_shape_id).ok_or(())?;
+        let angle          = self.cut_angle_for_shape(snap_data.target_shape_id,position,part).ok_or(())?;
+
+        // Completely disable/enable focus for shapes that are not close the split base don their
```
wdanilo, line 1840: there are some mistakes in words here

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1617,25 +1682,101 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
-    new_output_edge <- node_output_touch.down.map(f_!(model.new_edge_from_output()));
-    new_input_edge  <- node_input_touch.down.map(f_!(model.new_edge_from_input()));
+    new_output_edge <- create_edge_from_output.map(f_!([model,edge_mouse_down,edge_over,edge_out] {
+        Some(model.new_edge_from_output(&edge_mouse_down,&edge_over,&edge_out))
+    })).unwrap();
+    new_input_edge <- create_edge_from_input.map(f!([model,edge_mouse_down,edge_over,edge_out]((target)){
+        if model.is_node_connected_at_input(target.node_id,target.port.to_vec()) {
+            return None
+        };
+        Some(model.new_edge_from_input(&edge_mouse_down,&edge_over,&edge_out))
```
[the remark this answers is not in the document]
MichaelMauderer: Doesn't seem to be implemented? `method not found in 'graph_editor::EdgeId'`

src/rust/ide/lib/view/src/graph_editor.rs
MichaelMauderer: This is mostly for naming ourpooses to make clear how this is used next.

src/rust/ide/lib/view/src/graph_editor.rs
MichaelMauderer: Linter does not complain but it is fixed.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
 const LINE_WIDTH         : f32 = 4.0;
+const ARROW_SIZE_X       : f32 = 20.0;
+const ARROW_SIZE_Y       : f32 = 20.0;
+
+const HOVER_EXTENSION    : f32 = 10.0;
+
 const MOUSE_OFFSET       : f32 = 2.0;
 const NODE_PADDING       : f32 = node::SHADOW_SIZE;
-const PADDING            : f32 = 4.0;
+const PADDING            : f32 = 4.0 + HOVER_EXTENSION;
```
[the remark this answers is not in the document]
MichaelMauderer: If I remember correctly this was to void the hover area beeing clipped at the edges of the sprite.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -39,118 +28,550 @@ const LINE_SHAPE_WIDTH   : f32 = LINE_WIDTH + 2.0 * PADDING;
...
 const RIGHT_ANGLE        : f32 = std::f32::consts::PI / 2.0;
 
+// TODO: Maybe find a better name.
+const MIN_SOURCE_TARGET_DIFFERENCE_FOR_Y_VALUE_DISCRIMINATION : f32 = 45.0;
```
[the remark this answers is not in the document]
MichaelMauderer: This is related to the issue, that we want to split edges based on the y-value. But if they are next to each other that doesn't work. So we need to switch between "use y-value" and "use euclidian distance" when the y-distance is too small. This value is the threshold for that switch.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
MichaelMauderer: Because we always were able to get the values from shapes. Just not sprites. Should this be different?

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
MichaelMauderer: What is the advantage of that over the explicit `Some(...)`?
Besides I am always getting a not-implemented error for `as_some`. Need to investigate why.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
MichaelMauderer: They are put in a bin :) This is derived from [data binning/bucketing](https://en.wikipedia.org/wiki/Data_binning)

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
MichaelMauderer: `on_mouse_down` here is an `frp::Source<display::object::Id>`.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
MichaelMauderer: I've added a note about this. It doesn't really matter as there is no way a focus split can create a self-intersection on those shapes. We can just handle them as part of the same  shape group.

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
wdanilo: Interesting. I don't know. I will think about it. Thanks for pointing it out!

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
wdanilo: Because the code is much nicer!

```rust
if correct_quadrant {
                     Some(Vector2::new(closest_point.x, closest_point.y))
                } else {
                    None
                }
```
vs
```rust
correct_quadrant.as_some(Vector2::new(closest_point.x,closest_point.y))
```

build/run.js
```diff
@@ -127,7 +127,7 @@ commands.build.rust = async function(argv) {
 
         console.log('Checking the resulting WASM size.')
         let stats = fss.statSync(paths.dist.wasm.mainOptGz)
-        let limit = 3.54
+        let limit = 3.58
```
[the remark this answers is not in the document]
MichaelMauderer: After re-basing the increase is from  3.59 (main) to 3.62 for this PR. No libraries added.

PR #642, opened 2020-07-06
Searcher uses node editor's controller to add nodes

mwu-tow:
### Pull Request Description
Rather than passing node editor and graph controller separately to the node searcher, the searcher will be able to access node editor's controller directly.
Thanks to that, when node editor changes the displayed graph (e.g. due to entering a node), the searcher's copy of controller won't end up dangling.
This addresses #634

### Checklist

### DIFF
--- a/src/rust/ide/src/view/node_editor.rs
+++ b/src/rust/ide/src/view/node_editor.rs
@@ -129,8 +129,8 @@ impl GraphEditorIntegratedWithController {
     }
 
     /// Get the controller associated with this graph editor.
-    pub fn controller(&self) -> controller::ExecutedGraph {
-        self.model.controller.clone_ref()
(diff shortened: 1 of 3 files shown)


src/rust/ide/src/view/node_editor.rs
```diff
@@ -127,6 +127,11 @@ impl GraphEditorIntegratedWithController {
     pub fn graph_editor(&self) -> GraphEditor {
         self.model.editor.clone_ref()
     }
+
+    /// Get the controller associated with this graph editor.
+    pub fn controller(&self) -> controller::ExecutedGraph {
+        self.model.controller.clone_ref()
+    }
```
farmaazon, line 134: I wonder if we could return reference here? Are there any downsides of doing that?

PR #647, opened 2020-07-08
Pass the method pointers information to the graph editor.

mwu-tow:
### Pull Request Description
Similarly to what we do with type information, this PR also adds passing the method pointer information to the graph editor view.

### Checklist

### DIFF
--- a/src/rust/ide/lib/view/src/graph_editor.rs
+++ b/src/rust/ide/lib/view/src/graph_editor.rs
@@ -373,6 +373,7 @@ pub struct FrpInputs {
     pub set_node_expression          : frp::Source<(NodeId,node::Expression)>,
     pub set_node_position            : frp::Source<(NodeId,Vector2)>,
     pub set_expression_type          : frp::Source<(ast::Id, OptionalType)>,
+    pub set_method_pointer           : frp::Source<(ast::Id, OptionalMethodPointer)>,
     pub cycle_visualization          : frp::Source<NodeId>,
     pub set_visualization            : frp::Source<(NodeId,Option<visualization::Path>)>,
     pub register_visualization       : frp::Source<Option<visualization::Definition>>,
@@ -410,6 +411,7 @@ impl FrpInputs {
             set_node_expression          <- source();
             set_node_position            <- source();
             set_expression_type          <- source();
+            set_method_pointer           <- source();
             set_visualization_data       <- source();
             cycle_visualization          <- source();
             set_visualization            <- source();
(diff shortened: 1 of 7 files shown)


src/rust/ide/src/view/node_editor.rs
```diff
@@ -375,27 +376,33 @@ impl GraphEditorIntegratedWithControllerModel {
...
+    /// Look up the computed information for a given expression and pass the information to the
+    /// graph editor view.
+    ///
+    /// The computed value information includes the expression type and the target method pointer.
+    fn refresh_computed_info(&self, id:ExpressionId) {
+        let info     = self.lookup_computed_info(&id);
+        let info     = info.as_ref();
+        let typename = info.and_then(|info| info.typename.clone());
+        self.set_type(id,graph_editor::OptionalType(typename));
+        let method_pointer = info.and_then(|info| info.method_pointer.clone());
```
farmaazon, line 409:
```suggestion
        let typename = info.and_then(|info| info.typename.clone_ref());
        self.set_type(id,graph_editor::OptionalType(typename));
        let method_pointer = info.and_then(|info| info.method_pointer.clone_ref());
```

mwu-tow: Actually Option<Rc<blah>> isn't CloneRef.

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -373,6 +373,7 @@ pub struct FrpInputs {
     pub set_node_expression          : frp::Source<(NodeId,node::Expression)>,
     pub set_node_position            : frp::Source<(NodeId,Vector2)>,
     pub set_expression_type          : frp::Source<(ast::Id, OptionalType)>,
+    pub set_method_pointer           : frp::Source<(ast::Id, OptionalMethodPointer)>,
```
wdanilo, line 376: no space after ,

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -655,6 +658,22 @@ pub struct OptionalType(pub Option<ImString>);
...
+
+/// Information about target definition for node entering.
+///
+/// `None` means that the node cannot be entered.
+// TODO [mwu]
+//  As currently there is no good place to wrap Rc into a newtype that can be easily depended on
+//  both by `ide-view` and `ide` crates, we put this as-is. Refactoring should be considered in the
+//  future, once code organization and emerging patterns are more clear.
+#[derive(Clone,Debug,Default,Shrinkwrap)]
+pub struct OptionalMethodPointer(pub Option<Rc<enso_protocol::language_server::MethodPointer>>);
```
wdanilo, line 673: Why Option is inside? `Rc` is for me implementation detail, but `Option` is a structural pattern telling that the field can be missing or not. I would definitely prefer to get `Option<MethodPointerSomething>` in frp rather than this, especially as there are special functions for handling options in frp, like for example `foo <= bar` - `foo` will be emitted only if `bar` was Some.

PR #651, opened 2020-07-09
Disable yargs to make IDE work on macOS Big Sur (#650)

BinarySoftware:
### Pull Request Description
The IDE would crash when trying to open it on macOS Big Sur, beta 1 & 2, so i've found the problem, first thinking it was electron-related, but then created simple electron app with the same configs and it worked, so i ran through code bit by bit , and disabled yargs, and it made the app boot up and connect to the engine.

### Important Notes
As far as i can tell yargs is pretty necessary to the IDE, so we should take a look why it doesnt work at the first place and maybe file an issue on [their git]|(https://github.com/yargs/yargs)

### Checklist

### DIFF
--- a/build/run.js
+++ b/build/run.js
@@ -127,7 +127,7 @@ commands.build.rust = async function(argv) {
 
         console.log('Checking the resulting WASM size.')
         let stats = fss.statSync(paths.dist.wasm.mainOptGz)
-        let limit = 3.51
+        let limit = 3.54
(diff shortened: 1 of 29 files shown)


wdanilo:
Maciek, awesome job in finding the source of the problem 👏. However, this is not a proper solution to disable command line switches on macOS! This is evil and terrible. This PR will not be merged. Could you investigate it further, please?

PR #654, opened 2020-07-09
Searcher Controller

farmaazon:
### Pull Request Description
This PR introduces a new controller for the Searcher. Currently it loads and provides an initial list of suggestions. Additionally there is a raw integration with view: then the current Searcher is put into the screen, it logs all the suggestions retrieved from the Engine. 

### Important Notes
Currently the logs always show empty list, because Engine does not have implementation of the API yet.

### Checklist

### DIFF
--- a/docs/product/searcher.md
+++ b/docs/product/searcher.md
@@ -24,7 +24,7 @@ TODO: Add more description when visual part will be implemented.
 ### Suggestion of Node Expressions
 
 The suggestion list is obtained from the Engine using `search/completion` method of Language Server.
-The parameters of the call depends on the Current searcher input, picked suggestions so far, and
+The parameters of the call depend on the Current searcher input, suggestions picked suggestions so far, and
 if we add a node connected to selection. The current implementation can be found in 
 `ide::controller::searcher` module.
 
--- a/docs/product/searcher.md
+++ b/docs/product/searcher.md
@@ -25,7 +25,7 @@ TODO: Add more description when visual part will be implemented.
 
 The suggestion list is obtained from the Engine using `search/completion` method of Language Server.
 The parameters of the call depend on the Current searcher input, suggestions picked suggestions so far, and
-if we add a node connected to selection. The current implementation can be found in 
+if we are adding a node connected to selection. The current implementation can be found in 
 `ide::controller::searcher` module.
 
 ### Suggestion Database
--- a/docs/product/searcher.md
+++ b/docs/product/searcher.md
@@ -10,26 +10,26 @@ tags: [product]
 ### Behaviour
 
 The Searcher Panel can be brought to the screen in two different ways:
-* when user starts to edit node - the node became Searcher input and panel appears below, 
-* when user press tab with mouse over Graph Editor Panel - the new Searcher input appears with
-  Searcher panel below. Additionally, if there is exactly one node selected, the connection is
-  displayed between the selected node and the Searcher input.
-  
-When the new searcher is displayed by pressing Tab and there is no selected node, and the user
-didn't type anything in searcher input yet, the panel below input shows the main menu. In other
-cases the list of suggestions is displayed instead.
+* when user starts to edit node's expression - the node becomes Searcher input
+  and panel appears below, 
+* when user press tab with mouse over Graph Editor Panel - the new Searcher
+  input appears with Searcher panel below. Additionally, if there is exactly
+  one node selected, the connection is displayed between the selected node
+  and the Searcher input.
 
 TODO: Add more description when visual part will be implemented.
 
 ### Suggestion of Node Expressions
 
-The suggestion list is obtained from the Engine using `search/completion` method of Language Server.
-The parameters of the call depend on the Current searcher input, suggestions picked suggestions so far, and
-if we are adding a node connected to selection. The current implementation can be found in 
+The suggestion list is obtained from the Engine using `search/completion` method
+of Language Server. The parameters of the call depend on the Current searcher
+input, suggestions picked suggestions so far, and if we are adding a node
+connected to selection. The current implementation can be found in 
 `ide::controller::searcher` module.
 
 ### Suggestion Database
 
-The `search/completion` method of Language Server returns a list of keys to the Suggestion
-Database instead of the whole entries. The database is retrieved by IDE on Project opening, and 
-keeps it up to date. The Database is implemented in `ide::model::suggestion_database` module. 
+The `search/completion` method of Language Server returns a list of keys to
+the Suggestion Database instead of the whole entries. The database is
+retrieved by IDE on Project opening, and keeps it up to date. The Database
+is implemented in `ide::model::suggestion_database` module. 
--- a/docs/product/searcher.md
+++ b/docs/product/searcher.md
@@ -1,0 +1,14 @@
+---
+layout: developer-doc
+title: Searcher Panel In Graph Editor
+category: product
+tags: [product]
+---
+
+# Searcher Panel In Graph Editor
+
+## Behaviour
+
+## Suggestion Database
+
+## Searcher Controller
--- a/docs/product/searcher.md
+++ b/docs/product/searcher.md
@@ -30,6 +30,6 @@ if we are adding a node connected to selection. The current implementation can b
 
 ### Suggestion Database
 
-The`search/completion` method of Language Server returns a list of keys to the Suggestion
+The `search/completion` method of Language Server returns a list of keys to the Suggestion
 Database instead of the whole entries. The database is retrieved by IDE on Project opening, and 
 keeps it up to date. The Database is implemented in `ide::model::suggestion_database` module. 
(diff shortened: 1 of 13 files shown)


src/rust/ide/src/model/module.rs
```diff
@@ -236,6 +236,12 @@ impl QualifiedName {
     }
 }
 
+impl<'a> Into<&'a String> for &'a QualifiedName {
+    fn into(self) -> &'a String {
+        &self.0
+    }
+}
+
```
mwu-tow, line 244: can't we have rather something like AsRef<str> ?

src/rust/ide/src/view/node_searcher.rs
```diff
@@ -80,6 +86,28 @@ impl NodeSearcher {
         self.display_object.add_child(&self.text_field.display_object());
         self.text_field.clear_content();
         self.text_field.set_focus();
+        let module     = self.node_editor.displayed_module();
+        //TODO[ao]: Now we use some predefined location, until this task will be done:
+        // https://github.com/enso-org/ide/issues/653 . This code should be replaced with
+        // the proper Searcher view integration anyway.
+        let position   = TextLocation { line:2, column:4 };
```
mwu-tow, line 93: Doesn't it bother server if the location is out of bounds?
Perhaps (0,0) would be a safer default?

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -1,17 +1,95 @@
...
+    /// Argument lists of suggested object (atom or function). If the object does not take any
+    /// arguments, the list is empty.
+    pub arguments : Vec<Argument>,
+    /// A type returned by the suggested object.
+    pub return_type : String,
+    /// A documentation associated with object. If there is no documentation, the field is an empty
+    /// string.
+    pub documentation : String,
+    /// A type of the "self" argument. This field is `None` for non-method suggestions.
+    pub self_type : Option<String>,
```
mwu-tow, line 44: Why once we use empty string for no data and once `None`?

src/rust/ide/src/view/node_searcher.rs
```diff
@@ -16,12 +16,15 @@ use ensogl::display::shape::text::text_field::TextFieldProperties;
 use ensogl::display::world::World;
 use ensogl::display;
 use ensogl::traits::*;
+use data::text::TextLocation;
```
mwu-tow, line 19: use decls order

src/rust/ide/src/view/node_searcher.rs
```diff
@@ -80,6 +86,28 @@ impl NodeSearcher {
...
+                    match notification {
+                        controller::searcher::Notification::NewList => {
+                            let list = controller.suggestions();
+                            info!(logger,"New list in Searcher: {list:?}");
+                        }
+                    }
+                }
+            }
+            futures::future::ready(())
+        }));
```
mwu-tow, line 109: So we add new subscription each time when `show` is called.
Do we unsubscribe when hiding the searcher?

src/rust/ide/src/view/node_searcher.rs
```diff
@@ -80,6 +86,28 @@ impl NodeSearcher {
         self.display_object.add_child(&self.text_field.display_object());
```
mwu-tow, line 86: docstring says `/// Show NodeSearcher if it is invisible.`
but I don't see in this function any "isn't this already visible?" checks

docs/product/searcher.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: developer-doc
```
wdanilo, line 2: I did not notice it earlier - please note that all of our docs should have 80 chars limit. We made change toallow 100 chars limit for Rust code, but docs across the wholeproject should be formatted according to the rulesin main enso repo, which specifies80 chars limit

src/rust/ide/src/controller/searcher.rs
```diff
@@ -0,0 +1,265 @@
...
+    Error(Rc<failure::Error>)
+}
+
+impl Suggestions {
+    /// Check if suggestion list is still loading.
+    pub fn is_loading(&self) -> bool {
+        match self {
+            Self::Loading => true,
+            _             => false,
+        }
```
wdanilo, line 45: Why not just `self == Self::Loading` ?

src/rust/ide/src/controller/searcher.rs
```diff
@@ -0,0 +1,265 @@
...
+
+// =====================
+// === Notifications ===
+// =====================
+
+/// The notification emitted by Searcher Controller
+#[derive(Copy,Clone,Debug,Eq,PartialEq)]
+pub enum Notification {
+    /// A new Suggestion list is available.
+    NewList
```
wdanilo, line 81: Notification::NewList does not tell anything about the factthat this is a __suggestion__ list

src/rust/ide/src/controller/searcher.rs
```diff
@@ -0,0 +1,265 @@
...
+        let position    = self.position.deref().into();
+        let request     = self.language_server.completion(module,&position,&self_type,&return_type,&tags);
+        let data        = self.data.clone_ref();
+        let database    = self.database.clone_ref();
+        let logger      = self.logger.clone_ref();
+        let notifier    = self.notifier.clone_ref();
+
+        self.data.borrow_mut().current_list = Suggestions::Loading;
+        executor::global::spawn(async move {
+            info!(logger,"Requesting new suggestion list");
```
wdanilo, line 172: all logs should be sentences.

src/rust/ide/src/model/module.rs
wdanilo: yep, AsRef would be better

src/rust/ide/src/view/layout.rs
```diff
@@ -106,10 +106,10 @@ impl ViewLayout {
         let logger        = Logger::sub(logger,"ViewLayout");
         let world         = &application.display;
         let text_editor   = TextEditor::new(&logger,world,text_controller,kb_actions,fonts);
-        let node_editor   = NodeEditor::new(&logger,application,graph_controller,project,
-            visualization_controller);
+        let node_editor   = NodeEditor::new(&logger,application,graph_controller,
+            project.clone_ref(),visualization_controller);
```
wdanilo, line 110: Why do we have an API which requires clone ref on the call side? This sohuld accept &project instead.

src/rust/ide/src/view/node_searcher.rs
farmaazon: Thanks for pointing out, that I forget to remove controller on hide. When removed, the notifications publisher should perish and the executed task should finish.

src/rust/ide/src/controller/searcher.rs
farmaazon: I cannot implement Eq for this structure, because it contains `failure::Error` which does not implement Eq.

src/rust/ide/src/view/layout.rs
farmaazon: No, because it takes project. If it would take reference, it would have to clone_ref it anyway, and now it sometimes can omit cloning if the user does not need the `project` anymore.

docs/product/searcher.md
```diff
@@ -0,0 +1,35 @@
...
+### Behaviour
+
+The Searcher Panel can be brought to the screen in two different ways:
+* when user starts to edit node - the node became Searcher input and panel appears below, 
+* when user press tab with mouse over Graph Editor Panel - the new Searcher input appears with
+  Searcher panel below. Additionally, if there is exactly one node selected, the connection is
+  displayed between the selected node and the Searcher input.
+  
+When the new searcher is displayed by pressing Tab and there is no selected node, and the user
+didn't type anything in searcher input yer, the panel below input shows the main menu. In other
```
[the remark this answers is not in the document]
farmaazon: What main menu is - this is still to be turned out. Maybe I remove this paragraph now.

src/rust/ide/src/view/layout.rs
wdanilo: Then why it doesnt accept `Into<..>` ? Then you can pass reference instead of requiring the user to clone it on call side. If this is CloneRefable, then `Into` is implemented for its reference by default.

src/rust/ide/src/controller/searcher.rs
wdanilo: Oh, I did not notice it. Btw, thy we are using `fialure::Error` in the first place? Why we do not keep all possible errors just as enums?

src/rust/ide/src/view/layout.rs
farmaazon: Hmm, I could, however for me it's a minimal difference if I call &project, and project.clone_ref() - I need to put some info anyway, and now you more clearly see what is happening during call. I usually follow the Rust philosophy telling that "explicit is better than implicit".

Another argument: Why converting anything should be responsibility of "new" function, if it's clear responsibility is just to create the `NodeEditor`?

src/rust/ide/src/controller/searcher.rs
farmaazon: We do keep all errors as enums, but failure helps a lot if you have two error's kind and want to return both of them. And those error kinds are often just an implementation detail, which failure::Error elegantly hides.

However, this crate is deprecated and we have issue to replace it: https://github.com/enso-org/ide/issues/443

PR #660, opened 2020-07-14
Obtaining child AST text span

mwu-tow:
### Pull Request Description
Adds utilities to check the position of AST nodes in the text relative to their parents.
Also, fixed a number of byte/codepoint indexing mismatches and added some tests for this (though still coverage could be improved).

`impl std::ops::Index<Span> for str ` still needs more refactoring, will finish this tomorrow.

### Checklist

### DIFF
--- a/src/rust/ide/lib/ast/impl/src/lib.rs
+++ b/src/rust/ide/lib/ast/impl/src/lib.rs
@@ -311,8 +311,8 @@ impl Ast {
         self.iter_recursive().find(|ast| ast.repr() == repr)
     }
 
-    /// Get the index (relative to self) for given child node.
-    pub fn index_of_child(&self, child:&Ast) -> FallibleResult<Index> {
+    /// Get the offset relative to self for a given child node.
+    pub fn child_offset(&self, child:&Ast) -> FallibleResult<Index> {
         let searched_token  = Token::Ast(&child);
         let mut found_child = false;
         let mut position    = 0;
@@ -333,7 +333,7 @@ impl Ast {
     /// Get the span (relative to self) for a child node identified by given crumb.
     pub fn span_of_child_at(&self, crumb:&Crumb) -> FallibleResult<Span> {
         let child = self.get(crumb)?;
-        let index = self.index_of_child(child)?;
+        let index = self.child_offset(child)?;
         Ok(Span::new(index, Size::new(child.len())))
     }
 }
@@ -890,7 +890,7 @@ impl TokenConsumer for IdMapBuilder {
     fn feed(&mut self, token:Token) {
         match token {
             Token::Off(val) => self.offset += val,
-            Token::Chr(val) => self.offset += val.len_utf8(),
+            Token::Chr( _ ) => self.offset += 1,
             Token::Str(val) => self.offset += val.chars().count(),
             Token::Ast(val) => {
                 let begin = self.offset;
@@ -934,7 +934,7 @@ impl TokenConsumer for LengthBuilder {
     fn feed(&mut self, token:Token) {
         match token {
             Token::Off(val) => self.length += val,
-            Token::Chr(val) => self.length += val.len_utf8(),
+            Token::Chr( _ ) => self.length += 1,
             Token::Str(val) => self.length += val.chars().count(),
             Token::Ast(val) => val.shape().feed_to(self),
         }
@@ -953,7 +953,7 @@ impl HasLength for Token<'_> {
     fn len(&self) -> usize {
         match self {
             Token::Off(val) => *val,
-            Token::Chr(val) => val.len_utf8(),
+            Token::Chr( _ ) => 1,
             Token::Str(val) => val.chars().count(),
             // The below is different and cannot be unified with `LengthBuilder` because below will
             // use the cached length, while `LengthBuilder` will traverse subtree.
@@ -1055,7 +1055,7 @@ where F:FnMut(Index,&Ast) {
     fn feed(&mut self, token:Token) {
         match token {
             Token::Off(val) => self.index += val,
-            Token::Chr(val) => self.index += val.len_utf8(),
+            Token::Chr( _ ) => self.index += 1,
             Token::Str(val) => self.index += val.chars().count(),
             Token::Ast(val) => {
                 (self.callback)(Index::new(self.index), val);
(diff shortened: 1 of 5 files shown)


src/rust/ide/lib/ast/impl/src/lib.rs
```diff
@@ -303,6 +310,32 @@ impl Ast {
...
+            } else if !found_child {
+                position += token.len()
+            }
+        });
+        if found_child {
+            Ok(Index::new(position))
+        } else {
+            Err(NoSuchChild.into())
+        }
+    }
```
farmaazon, line 331: `Index` term in this context is really misleading for me. It's not natural that this means "index of character". Usually we use rather "offset" term.

src/rust/ide/lib/ast/impl/src/lib.rs
```diff
@@ -852,8 +890,8 @@ impl TokenConsumer for IdMapBuilder {
     fn feed(&mut self, token:Token) {
         match token {
             Token::Off(val) => self.offset += val,
-            Token::Chr( _ ) => self.offset += 1,
-            Token::Str(val) => self.offset += val.len(),
+            Token::Chr(val) => self.offset += val.len_utf8(),
+            Token::Str(val) => self.offset += val.chars().count(),
```
farmaazon, line 894: Once we add the number of bytes, and then chars count?

src/rust/ide/lib/ast/impl/src/lib.rs
```diff
@@ -896,8 +934,8 @@ impl TokenConsumer for LengthBuilder {
     fn feed(&mut self, token:Token) {
         match token {
             Token::Off(val) => self.length += val,
-            Token::Chr( _ ) => self.length += 1,
-            Token::Str(val) => self.length += val.len(),
+            Token::Chr(val) => self.length += val.len_utf8(),
+            Token::Str(val) => self.length += val.chars().count(),
```
farmaazon, line 938: Ditto

src/rust/ide/lib/ast/impl/src/lib.rs
```diff
@@ -911,6 +949,19 @@ impl<T:HasTokens> HasLength for T {
     }
 }
 
+impl HasLength for Token<'_> {
+    fn len(&self) -> usize {
+        match self {
+            Token::Off(val) => *val,
+            Token::Chr(val) => val.len_utf8(),
+            Token::Str(val) => val.chars().count(),
```
farmaazon, line 957: Ditto

src/rust/ide/lib/ast/impl/src/lib.rs
```diff
@@ -1004,8 +1055,8 @@ where F:FnMut(Index,&Ast) {
     fn feed(&mut self, token:Token) {
         match token {
             Token::Off(val) => self.index += val,
-            Token::Chr( _ ) => self.index += 1,
-            Token::Str(val) => self.index += val.len(),
+            Token::Chr(val) => self.index += val.len_utf8(),
+            Token::Str(val) => self.index += val.chars().count(),
```
farmaazon, line 1059: Ditto

src/rust/lib/data/src/text.rs
```diff
@@ -113,7 +113,7 @@ impl Display for Size {
 
 impl From<&str> for Size {
     fn from(text:&str) -> Self {
-        Size::new(text.len())
+        Size::new(text.chars().count())
     }
 }
```
farmaazon, line 118: This is not a today change, but I have distaste for this conversion. What does it mean that string is _converted_ to its length? Those are completely different things.

src/rust/ide/lib/ast/impl/src/lib.rs
mwu-tow: -> child_offset

src/rust/ide/lib/ast/impl/src/lib.rs
mwu-tow: Stupid mistake. One was wrong and one was correct but I just mindlessly flipped both. Fixed.

src/rust/lib/data/src/text.rs
mwu-tow: Agreed as discussed today. The impl has been removed in favor of `from_str` constructor.

src/rust/ide/lib/ast/impl/src/lib.rs
```diff
@@ -311,8 +311,8 @@ impl Ast {
         self.iter_recursive().find(|ast| ast.repr() == repr)
     }
 
-    /// Get the index (relative to self) for given child node.
-    pub fn index_of_child(&self, child:&Ast) -> FallibleResult<Index> {
+    /// Get the offset relative to self for a given child node.
```
farmaazon, line 314: I would still extend this docs. To be more clear that this is an offest of "begins of text representation" in number of codepoints (although the last can be deduced from the returned type).

src/rust/lib/data/src/text.rs
```diff
@@ -152,6 +167,17 @@ impl Span {
         Span {index:Index::new(0), size}
     }
 
+    /// Get the index of the last character in the span.
+    ///
+    /// If the span is empty or starts at 0, returns `None`.
```
farmaazon, line 172:
```suggestion
    /// If the span is empty or ends at 0, returns `None`.
```

src/rust/lib/data/src/text.rs
```diff
@@ -152,6 +167,17 @@ impl Span {
...
+    pub fn last(&self) -> Option<Index> {
+        if self.is_empty() {
+            None
+        } else {
+            self.end().checked_sub(Size::new(1))
+        }
+    }
```
farmaazon, line 179: Technically, it is impossible now to have span which is not empty, but it end is at 0. But maybe we should leave it as it is, and fix when we will handle panics correctly.

src/rust/lib/data/src/text.rs
mwu-tow: I've removed that part, as you mentioned that this situation is impossible.

src/rust/lib/data/src/text.rs
mwu-tow: True.

src/rust/ide/lib/ast/impl/src/lib.rs
mwu-tow: I've extended somehow, though didn't feel like mentioning codepoints. The type expresses it, we should not need to repeat it on every usage.

src/rust/ide/lib/ast/impl/src/lib.rs
```diff
@@ -767,6 +803,11 @@ pub trait TokenConsumer {
     fn feed(&mut self, val:Token);
 }
 
+impl<F:FnMut(Token)> TokenConsumer for F {
+    fn feed(&mut self, val: Token) {
```
wdanilo, line 807: rm space after :

PR #661, opened 2020-07-14
Applying suggestion in Searcher

farmaazon:
### Pull Request Description
This PR implements the function for applying suggestion by user in the Searcher Controller (the action on "Tab"). It also contains a bit of the controller refactoring.

### Checklist

### DIFF
--- a/src/rust/ide/src/controller/searcher.rs
+++ b/src/rust/ide/src/controller/searcher.rs
@@ -14,13 +14,14 @@ use parser::Parser;
 // === Suggestion List ===
 // =======================
 
-pub type Completion = Rc<model::suggestion_database::Entry>;
+/// Suggestion for input completion: possible functions, arguments, etc.
+pub type CompletionSuggestion = Rc<model::suggestion_database::Entry>;
 
 /// A single suggestion on the Searcher suggestion list.
 #[derive(Clone,CloneRef,Debug,Eq,PartialEq)]
 pub enum Suggestion {
     /// Suggestion for input completion: possible functions, arguments, etc.
-    Completion(Completion)
+    Completion(CompletionSuggestion)
     // In future, other suggestion types will be added (like suggestions of actions, etc.).
 }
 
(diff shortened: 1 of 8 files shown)


src/rust/ide/src/controller/searcher.rs
```diff
@@ -208,59 +389,196 @@ mod test {
...
         assert_eq!(searcher.suggestions().list(), Some(&expected_list));
         let notification = subscriber.next().boxed_local().expect_ready();
         assert_eq!(notification, Some(Notification::NewSuggestionList));
     }
+
+    #[wasm_bindgen_test]
+    fn parsed_input() {
+        let parser = Parser::new_or_panic();
+
+        fn args_repr(prefix:&ast::prefix::Chain) -> Vec<String> {
```
mwu-tow, line 473: as it returns a sequence of reprs, I'd name this `arg_reprs`

src/rust/ide/src/controller/searcher.rs
```diff
@@ -146,33 +267,93 @@ impl Searcher {
...
+        }
+        Ok(())
+    }
+
+    /// User picked a completion suggestion.
+    ///
+    /// This function should be called when user chooses some completion suggestion. It still will
+    /// be editing the input. The picked suggestion will be remembered,
+    pub fn pick_completion(&self, picked_suggestion:CompletionSuggestion) -> String {
+        let added_ast         = ast::Ast::var(&picked_suggestion.name);
```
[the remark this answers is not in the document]
farmaazon: So, shall I parse the name string instead? Otherwise I don't have distinction between method name being a var and operator being an operator.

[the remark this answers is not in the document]
mwu-tow: perhaps... I don't think we support operators now (and that'd require likely some other syntax) but parsing will allow telling apart Cons and vars — and these two I think we need to support now.
If we store user input in form of AST, then yes — I think we need to parse whatever gets appended from suggestion. Or suggestions themselves should contain AST with name, alternatively.

src/rust/ide/src/controller/searcher.rs
```diff
@@ -208,59 +392,196 @@ mod test {
...
+            client_setup(&mut client);
+            let searcher = Searcher {
+                logger          : default(),
+                data            : default(),
+                notifier        : default(),
+                module          : Rc::new(module_path.qualified_module_name("Test")),
+                position        : Immutable(TextLocation::at_document_begin()),
+                database        : default(),
+                language_server : language_server::Connection::new_mock_rc(client),
+                parser          : Parser::new_or_panic(),
```
wdanilo, line 415: why you dont use `..default()` pattern  ? Can Searcher impl Default?

src/rust/ide/src/controller/searcher.rs
```diff
@@ -208,59 +392,196 @@ mod test {
...
+                self_type     : None
+            };
+            let entry2 = model::suggestion_database::Entry {
+                name : "TestVar1".to_string(),
+                ..entry1.clone()
+            };
+            let entry9 = model::suggestion_database::Entry {
+                name : "TestFunction2".to_string(),
+                ..entry1.clone()
+            };
```
wdanilo, line 433: Could we instead have a method like `entry9 = entry1.with_name("TestFunction2")` instead of these 4 lines of code?

src/rust/ide/src/controller/searcher.rs
farmaazon: I don't think Searcher can impl Default, because there is no such thing as "Default Language Server connection".

PR #663, opened 2020-07-15
 Ability to set edge color based on type.

MichaelMauderer:
### Pull Request Description
Add facilities to compute colors from types (basic hashing and mapping algorithm) as well as facilities to color edges, ports and the cursor dependent on their respective types. 

### Important Notes
Does not enforce correct connections. Of two different colored ports are connected, one color will be used for the edge.
Known Issue: `text::Area` changes the cursor on hover internally, and thus cancels the color of the cursor if it was dragging a typed edge. The ``text::Area` should be refactored to emit an event instead, that can be used in the `GraphEditor` to implement the correct behavior. The correct behavior is most likely to not allow changing the cursor while it is dragging an edge. 

### Checklist

### DIFF
--- a/src/rust/ide/lib/view/src/debug_scenes/shapes.rs
+++ b/src/rust/ide/lib/view/src/debug_scenes/shapes.rs
@@ -72,19 +72,6 @@ impl DummyTypeGenerator {
     }
 }
 
-/// Allows modifying every node in a span tree through a closure. Can be used to initialise the
-/// span tree with type information for the debug scene.
-fn visit_span_tree_nodes_mut<F:FnMut(&mut span_tree::Node)>(span_tree:&mut span_tree::SpanTree, mut f:F) {
-    visit_node_and_children_mut(&mut span_tree.root, &mut f);
-}
-
-fn visit_node_and_children_mut<F:FnMut(&mut span_tree::Node)>(node: &mut span_tree::Node, f: &mut F) {
-    f(node);
-    node.children.iter_mut().for_each(|child| {
-        visit_node_and_children_mut(&mut child.node, f);
-    })
-}
-
 
 
 // ========================
@@ -239,16 +226,18 @@ use span_tree::traits::*;
 
 pub fn expression_mock() -> Expression {
     let code             = "open \"data.csv\"".into();
-    let mut output_span_tree = span_tree::SpanTree::default();
-    let mut input_span_tree  = span_tree::builder::TreeBuilder::new(15)
-        .add_leaf(0,4,span_tree::node::Kind::Operation,PrefixCrumb::Func)
+    let output_span_tree = span_tree::SpanTree::default();
+    let input_span_tree  = span_tree::builder::TreeBuilder::new(15)
+        .add_child(0,4,span_tree::node::Kind::Operation,PrefixCrumb::Func)
+        .set_expression_id(Uuid::new_v4())
+        .done()
         .add_empty_child(5,span_tree::node::InsertType::BeforeTarget)
-        .add_leaf(5,10,span_tree::node::Kind::Target{is_removable:false},PrefixCrumb::Arg)
+        .add_child(5,10,span_tree::node::Kind::Target{is_removable:false},PrefixCrumb::Arg)
+        .set_expression_id(Uuid::new_v4())
+        .done()
         .add_empty_child(15,span_tree::node::InsertType::Append)
         .set_expression_id(Uuid::new_v4())
         .build();
-    visit_span_tree_nodes_mut(&mut input_span_tree, |node| node.expression_id = Some(Uuid::new_v4()));
-    visit_span_tree_nodes_mut(&mut output_span_tree, |node| node.expression_id = Some(Uuid::new_v4()));
     Expression {code,input_span_tree,output_span_tree}
 }
 
@@ -257,26 +246,35 @@ pub fn expression_mock2() -> Expression {
     let val              = ast::crumbs::SegmentMatchCrumb::Body {val:pattern_cr};
     let parens_cr        = ast::crumbs::MatchCrumb::Segs {val,index:0};
     let code             = "make_maps size (distribution normal)".into();
-    let mut output_span_tree = span_tree::SpanTree::default();
-    let mut input_span_tree  = span_tree::builder::TreeBuilder::new(36)
+    let output_span_tree = span_tree::SpanTree::default();
+    let input_span_tree  = span_tree::builder::TreeBuilder::new(36)
         .add_child(0,14,span_tree::node::Kind::Chained,PrefixCrumb::Func)
-        .add_leaf(0,9,span_tree::node::Kind::Operation,PrefixCrumb::Func)
-        .add_empty_child(10,span_tree::node::InsertType::BeforeTarget)
-        .add_leaf(10,4,span_tree::node::Kind::Target {is_removable:true},PrefixCrumb::Arg)
-        .add_empty_child(14,span_tree::node::InsertType::Append)
-        .done()
+            .add_child(0,9,span_tree::node::Kind::Operation,PrefixCrumb::Func)
+                .set_expression_id(Uuid::new_v4())
+                .done()
+            .add_empty_child(10,span_tree::node::InsertType::BeforeTarget)
+            .add_child(10,4,span_tree::node::Kind::Target {is_removable:true},PrefixCrumb::Arg)
+                .set_expression_id(Uuid::new_v4())
+                .done()
+            .add_empty_child(14,span_tree::node::InsertType::Append)
+            .set_expression_id(Uuid::new_v4())
+            .done()
         .add_child(15,21,span_tree::node::Kind::Argument {is_removable:true},PrefixCrumb::Arg)
-        .add_child(1,19,span_tree::node::Kind::Argument {is_removable:false},parens_cr)
-        .add_leaf(0,12,span_tree::node::Kind::Operation,PrefixCrumb::Func)
-        .add_empty_child(13,span_tree::node::InsertType::BeforeTarget)
-        .add_leaf(13,6,span_tree::node::Kind::Target {is_removable:false},PrefixCrumb::Arg)
-        .add_empty_child(19,span_tree::node::InsertType::Append)
-        .done()
-        .done()
+            .set_expression_id(Uuid::new_v4())
+            .add_child(1,19,span_tree::node::Kind::Argument {is_removable:false},parens_cr)
+                .set_expression_id(Uuid::new_v4())
+                .add_child(0,12,span_tree::node::Kind::Operation,PrefixCrumb::Func)
+                    .set_expression_id(Uuid::new_v4())
+                    .done()
+                .add_empty_child(13,span_tree::node::InsertType::BeforeTarget)
+                .add_child(13,6,span_tree::node::Kind::Target {is_removable:false},PrefixCrumb::Arg)
+                    .set_expression_id(Uuid::new_v4())
+                    .done()
+                .add_empty_child(19,span_tree::node::InsertType::Append)
+                .done()
+            .done()
         .add_empty_child(36,span_tree::node::InsertType::Append)
         .build();
-    visit_span_tree_nodes_mut(&mut input_span_tree, |node| node.expression_id = Some(Uuid::new_v4()));
-    visit_span_tree_nodes_mut(&mut output_span_tree, |node| node.expression_id = Some(Uuid::new_v4()));
     Expression {code,input_span_tree,output_span_tree}
 }
 


src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -582,43 +611,88 @@ impl OutputPortsData {
...
+
+        // === Update FRP bindings ===
         let data    = &self.data;
+        let frp     = &self.frp;
 
         // Used to set and detect the end of the tweens. The actual value is irrelevant, only the
         // duration of the tween matters and that this value is reached after that time.
         const TWEEN_END_VALUE:f32 = 1.0;
 
+        // TODO memory leak from tween?
```
MichaelMauderer, line 682: Is it okay to create a new one here, or does this leak memory? If so we need to instantiate them somewhere else.

MichaelMauderer: For example in the `network_internal`.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -582,43 +611,88 @@ impl OutputPortsData {
...
 pub struct OutputPorts {
     /// The FRP api of the `OutPutPorts`.
-    pub frp     : Frp,
-        network : frp::Network,
-        data    : Rc<OutputPortsData>,
+    pub frp              : Frp,
+        network          : frp::Network,
+        network_internal : Rc<RefCell<frp::Network>>,
+        data             : Rc<OutputPortsData>,
+        type_map         : TypeColorMap,
```
wdanilo, line 618: Hmm, I see this type_map already in second place in the code. Are we keeping the same information twice in different code places or is this a different information?

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -582,43 +611,88 @@ impl OutputPortsData {
 #[derive(Debug,Clone,CloneRef)]
 pub struct OutputPorts {
     /// The FRP api of the `OutPutPorts`.
-    pub frp     : Frp,
-        network : frp::Network,
-        data    : Rc<OutputPortsData>,
+    pub frp              : Frp,
+        network          : frp::Network,
+        network_internal : Rc<RefCell<frp::Network>>,
```
wdanilo, line 616: This name is mysterious. What is an "internal frp network" ?

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -582,43 +611,88 @@ impl OutputPortsData {
...
+        data             : Rc<OutputPortsData>,
+        type_map         : TypeColorMap,
+        expression       : Rc<RefCell<Expression>>,
```
wdanilo, line 619: Why output ports keep expression information? This is not their responsibility. Is this information duplicated now? If so, it's bad, as we should not duplicate information across strucutres.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -582,43 +611,88 @@ impl OutputPortsData {
...
+
+        // We want to be able to match each `PortId` to a `Crumb` from the expression, including
+        // the root.
+        let number_of_ports = 1 + expression.output_span_tree.root_ref().leaf_iter().count() as u32;
+        let mut id_map      = HashMap::<PortId,span_tree::Crumbs>::from_iter(expression.output_span_tree.root_ref()
+            .leaf_iter()
+            .enumerate()
+            .map(|(index, node)| {
+                (PortId::new(index + 1), node.crumbs)
+        }));
```
wdanilo, line 665: can we use one or two more variables to explain what this code does, please?

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
wdanilo: Why you think it will leak memory? You may be right, but I need explanation to justify it.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -659,26 +733,43 @@ impl OutputPorts {
...
+    pub fn get_port_color(&self, crumbs:&[span_tree::Crumb]) -> Option<color::Lcha> {
+        let ast_id = self.expression.borrow().get_id_for_crumbs(crumbs)?;
+        self.type_map.type_color(ast_id)
+    }
+
+    /// Set the type information for the given `ast::Id`.
+    pub fn set_expression_type(&self, id:ast::Id, maybe_type:Option<Type>) {
+        // TODO should we ignore this if it is not part f our expression?
+        match maybe_type {
+            Some(r#type) => self.type_map.insert(id,r#type),
```
wdanilo, line 769: I don't think this is really the var name that we are looking. Even `tp` would be better than this strange monster ;)

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -659,26 +733,43 @@ impl OutputPorts {
...
+    }
+
+    /// Return the color of the port indicated by the given `Crumb`.
+    pub fn get_port_color(&self, crumbs:&[span_tree::Crumb]) -> Option<color::Lcha> {
+        let ast_id = self.expression.borrow().get_id_for_crumbs(crumbs)?;
+        self.type_map.type_color(ast_id)
+    }
+
+    /// Set the type information for the given `ast::Id`.
+    pub fn set_expression_type(&self, id:ast::Id, maybe_type:Option<Type>) {
```
wdanilo, line 766: I still dont understand why outut ports base on some expressions, so Im not really checking this code till we talk about it.

src/rust/ide/lib/view/src/graph_editor/component/type_coloring.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::graph_editor::SharedHashMap;
+use crate::graph_editor::Type;
+
+use ensogl::data::color;
+use std::collections::hash_map::DefaultHasher;
+use std::hash::Hasher;
+
+
+/// Color that should be used if no type information is available.
+pub const DEFAULT_TYPE_COLOR : color::Lcha = color::Lcha::new(0.5, 0.0, 0.0, 1.0);
```
wdanilo, line 15: no type information is not DEFAULT type color. The name is misleading. It should be named MISSING_TYPE_INFO_COLOR instead

src/rust/ide/lib/view/src/graph_editor/component/type_coloring.rs
```diff
@@ -0,0 +1,55 @@
...
+
+// ================================
+// === Type to Color Conversion ===
+// ================================
+
+fn type_to_hash(type_information:Type) -> u64 {
+    let mut hasher = DefaultHasher::new();
+    type_information.0.hash(&mut hasher);
+    hasher.finish()
+}
```
wdanilo, line 27: why `Type` jsut does not implement hash insteaD?

wdanilo: Moreover, what is the relation of this function and the section name it is placed in ?

src/rust/ide/lib/view/src/graph_editor/component/type_coloring.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// Return the color that corresponds to the given type. Can be used to color edges and ports.
+pub fn color_for_type(type_information:Type) -> color::Lch {
+    let hue =  (type_to_hash(type_information) % 360) as f32 / 360.0;
```
wdanilo, line 31: spacing

wdanilo, line 31: I'm not sure we are using 360 degrees for hue. Arent we using 0-1 there? To be checked and confirmed

src/rust/ide/lib/view/src/graph_editor/component/type_coloring.rs
```diff
@@ -0,0 +1,55 @@
...
+    color::Lch::new(0.5, 0.8, hue)
```
wdanilo, line 32: spacing

wdanilo, line 32: literals - should be constants

src/rust/ide/lib/view/src/graph_editor/component/type_coloring.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+
+
+// ================
+// === Type Map ===
+// ================
+
+/// `TypeMap` allows to keep track of the type and color of a `ast::Id`. It allows to store the
+/// `ast::Id` -> `Type` mapping and infer the colour for the given `ast::Id` from that.
```
wdanilo, line 42: It is a good rule to not start documentation with the name of what you describe. Jsut start with a sentence telling what it does. Btw, the doc tells about `TypeMap` but struct is named differently.

wdanilo: Moreover the docs tell something aobut colors, the name as well, but the implementation does not use colors anywhere. This needs to be better explained.

src/rust/ide/lib/view/src/debug_scenes/shapes.rs
```diff
@@ -52,6 +53,43 @@ where T:frp::HasOutput<Output=Out>, T:Into<frp::Stream<Out>>, Out:frp::Data {
...
 
+#[derive(Clone,Debug,Default)]
+struct DummyTypeGenerator {
+    type_counter : u32
+}
+
+impl DummyTypeGenerator {
+    fn get_dummy_type(&mut self) -> Type {
+        self.type_counter += 1;
+        Type(ImString::new(format!("dummy_type_{}",self.type_counter)))
```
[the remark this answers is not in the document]
farmaazon: > Do we have to explicitly call `ImString` constructor? We should be able to write `Type::from(...)` instead.

I suspect we could also call just `format!("dummy_type_{}",self.type_counter)),into()` if you like

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -19,6 +19,7 @@ pub mod builtin;
 pub mod data;
 
 use crate::graph_editor::component::node;
+use crate::graph_editor::component::type_coloring::DEFAULT_TYPE_COLOR;
```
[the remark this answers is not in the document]
MichaelMauderer: Yes. This is now `MISSING_TYPE_COLOR`.

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1779,7 +1818,7 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
     })).unwrap();
 
     outputs.edge_added <+ new_output_edge;
-    new_edge_source <- new_output_edge.map2(&node_output_touch.down, move |id,node_id| (*id,EdgeTarget::new(node_id,default())));
+    new_edge_source <- new_output_edge.map2(&node_output_touch.down, move |id,target| (*id,target.clone()));
```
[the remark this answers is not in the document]
MichaelMauderer: No, it contains an `Rc<span_tree::Crumbs>`, which is not copy.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
MichaelMauderer: We keep it both in the `OutputPorts` and the `port::Manager` as both of these need to color their ports. I guess we could move this functionality into the node, or alternatively create/refactor a unified `Manager` for both input and output ports.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
MichaelMauderer: We are using this here so we can look up the `ast::Id` for a given `&[span_tree::Crumb]`, which we need to determine the color of our ports. See also https://github.com/enso-org/ide/pull/663#discussion_r459478405
This is similar to what happens in the `port::Manager`. Maybe we should unify that?

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
MichaelMauderer: My understanding is, that the output ports are based on the `Expression::output_span_tree`, which determines the actual number of output ports and also the type of the output ports.

src/rust/ide/lib/view/src/graph_editor/component/type_coloring.rs
MichaelMauderer: I've added a note. But essentially, if the hash is random, the 360 here is irrelevant. We normalise to 0..1 in the end.

build/run.js
```diff
@@ -127,7 +127,7 @@ commands.build.rust = async function(argv) {
 
         console.log('Checking the resulting WASM size.')
         let stats = fss.statSync(paths.dist.wasm.mainOptGz)
-        let limit = 3.55
+        let limit = 3.58
```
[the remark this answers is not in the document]
MichaelMauderer: I believe this is indicated wrongly and also includes the changes from the edge-disconnect branch. Let's see what happens after rebasing here.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
MichaelMauderer: Whenever we change the number of ports, we need to re-create the frp bindings of the `port_network` (formerly `network_internal`). Right now we drop the `port_network` and set up new bindings to the `network`. But the `Tween` is created in the `network` which persists. 

So the question is: will it create a new tween every time or is it fine to create a new one?

Writing it out like this I heavily suspect that it's bad and needs to be another way.

src/rust/ide/lib/view/src/debug_scenes/shapes.rs
```diff
@@ -53,6 +54,60 @@ where T:frp::HasOutput<Output=Out>, T:Into<frp::Stream<Out>>, Out:frp::Data {
...
+fn visit_span_tree_nodes_mut<F:FnMut(&mut span_tree::Node)>(span_tree:&mut span_tree::SpanTree, mut f:F) {
+    visit_node_and_children_mut(&mut span_tree.root, &mut f);
+}
+
+fn visit_node_and_children_mut<F:FnMut(&mut span_tree::Node)>(node: &mut span_tree::Node, f: &mut F) {
+    f(node);
+    node.children.iter_mut().for_each(|child| {
+        visit_node_and_children_mut(&mut child.node, f);
+    })
+}
```
wdanilo, line 103: These are span_tree utilities - move them to span_tree implementation please. Chat with @farmaazon about the. correct placement AND if there are no ready utils for that as Adam was working a lot on different iterator types.

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -654,10 +655,14 @@ impl display::Object for Edge {
 /// Typename information that may be associated with the given Port.
 ///
 /// `None` means that type for the port is unknown.
-#[derive(Clone,Debug,Shrinkwrap)]
+#[derive(Clone,Debug,Hash,Shrinkwrap)]
 pub struct Type(pub ImString);
 
-
+impl From<String> for Type {
+    fn from(s: String) -> Self {
```
wdanilo, line 662: spacing

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -825,6 +830,15 @@ impl Edges {
     pub fn insert(&self, edge:Edge) {
         self.all.insert(edge.id(),edge);
     }
+
+    pub fn detached_edges_iter(&self) -> impl Iterator<Item=EdgeId> {
+        let detached_target = self.detached_target.raw.borrow();
+        let detached_source = self.detached_source.raw.borrow();
+        let mut detached = detached_target.iter().copied().collect_vec();
+        let detached_source_iter = detached_source.iter().copied();
```
wdanilo, line 838: alignment

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1389,6 +1422,42 @@ impl GraphEditorModel {
             }
         };
     }
+
+    /// Return the color of an edge target. Returns `None` if no type information is associated
+    ///with the target port.
+    fn try_get_edge_target_color(&self, edge_target:EdgeTarget) -> Option<color::Lcha> {
+        let node              = self.nodes.get_cloned_ref(&edge_target.node_id)?;
+        // We don't know whether this is a input or output port, so we need to check both.
```
wdanilo, line 1430: This comment is not necessary IMO. The code is very clear about it

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1389,6 +1422,42 @@ impl GraphEditorModel {
             }
         };
     }
+
+    /// Return the color of an edge target. Returns `None` if no type information is associated
+    ///with the target port.
```
wdanilo, line 1427: space before "with"

src/rust/ide/lib/view/src/graph_editor.rs
```diff
@@ -1948,24 +1999,11 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
-    //   treated as unknown type by default);
-    // * integration layer will not emit "unset" signal for ports that were removed by changing the
-    //   node expression (`set_node_expression` event) or were removed altogether (`remove_node`);
-    // * integration layer reserves right to emit spurious signals (i.e. unsetting type that was
-    //   not previously set or setting same type multiple times). This should improve, as we get
-    //   better updates from engine services (see https://github.com/enso-org/enso/issues/441 ).
-    trace inputs.set_expression_type;
-
+    frp::extend! { network
+    eval inputs.set_expression_type (((node_id,ast_id,maybe_type)) model.set_node_expression_type(*node_id,*ast_id,maybe_type.clone()));
```
wdanilo, line 2006: too long line (I know there are too long lines in other places too, Im trying to fix them slowly)

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -319,9 +321,14 @@ macro_rules! define_corner_start {($color:expr, $highlight_color:expr) => {
...
             , focus_split_angle:f32
+            , color_rgba:Vector4<f32>
+            , focus_color_rgba:Vector4<f32>
             ) {
-                let width  = &LINE_WIDTH.px();
-                let shape  = corner_base_shape(&radius,width,&angle,&start_angle);
+                let width           = &LINE_WIDTH.px();
+                let shape           = corner_base_shape(&radius,width,&angle,&start_angle);
+                let color           = Var::<color::Rgba>::from(color_rgba);
+                let focus_color = Var::<color::Rgba>::from(focus_color_rgba);
```
wdanilo, line 330: alignment

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -401,9 +418,13 @@ macro_rules! define_corner_end {($color:expr, $highlight_color:expr) => {
...
```
wdanilo, line 427: alignment

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -471,19 +500,22 @@ macro_rules! define_corner_end {($color:expr, $highlight_color:expr) => {
...
-            (focus_split_center:Vector2<f32>, focus_split_angle:f32) {
-                let width  = LINE_WIDTH.px();
+            (focus_split_center:Vector2<f32>, focus_split_angle:f32, color_rgba:Vector4<f32>,
+             focus_color_rgba:Vector4<f32>) {
+                let width           = LINE_WIDTH.px();
                 let height : Var<Pixels> = "input_size.y".into();
-                let shape  = Rect((width.clone(),height));
+                let shape           = Rect((width.clone(),height));
+                let color           = Var::<color::Rgba>::from(color_rgba);
+                let focus_color = Var::<color::Rgba>::from(focus_color_rgba);
```
wdanilo, line 514: alignment

src/rust/ide/lib/view/src/graph_editor/component/edge.rs
```diff
@@ -519,19 +559,22 @@ macro_rules! define_line {($color:expr, $highlight_color:expr) => {
...
     pub mod arrow {
         use super::*;
         ensogl::define_shape_system! {
-            (focus_split_center:Vector2<f32>, focus_split_angle:f32) {
+            (focus_split_center:Vector2<f32>, focus_split_angle:f32, color_rgba:Vector4<f32>,
+             focus_color_rgba:Vector4<f32>) {
                 let width  : Var<Pixels> = "input_size.x".into();
                 let height : Var<Pixels> = "input_size.y".into();
+                let color              = Var::<color::Rgba>::from(color_rgba);
+                let focus_color    = Var::<color::Rgba>::from(focus_color_rgba);
```
wdanilo, line 572: alignemnt

src/rust/ide/lib/view/src/graph_editor/component/node/port.rs
```diff
@@ -140,6 +143,21 @@ impl Expression {
         let output_span_tree = default();
         Self {code,input_span_tree,output_span_tree}
     }
+
+    fn get_id_for_crumbs(&self, crumbs:&[span_tree::Crumb]) -> Option<ast::Id> {
+        // We don't know whether this is from the input or output span tree, so we need to
+        // check both.
+        let span_trees = [&self.input_span_tree, &self.output_span_tree];
```
wdanilo, line 150: IMO the sae effecet would be when naming the var `span_trees_to_check`. You can leave comment if you want, but I'm trying to remove form code all non-essential comments (comments which cannot be explained by good var names), like comments about some system depended choices (e.g. "On Windows we need this because this and that").

src/rust/ide/lib/view/src/graph_editor/component/node/port.rs
```diff
@@ -285,9 +314,21 @@ impl Manager {
...
+        self.type_color_map.type_color(ast_id)
+    }
+
     pub fn width(&self) -> f32 {
         self.width.get()
     }
+
+    pub fn set_expression_type(&self, id:ast::Id, maybe_type:Option<Type>) {
+        match maybe_type {
+            Some(r#type) => self.type_color_map.insert(id,r#type),
```
wdanilo, line 328: Please fix all such naming in the PR . `r#type` is really not the best name for the var. In case you don't agree,l sets chat about it! But I was writing about it previously already :)

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -22,9 +22,12 @@ use ensogl::display;
 use ensogl::gui::component::Animation;
 use ensogl::gui::component::Tween;
 use ensogl::gui::component;
-use std::num::NonZeroU32;
+use span_tree;
 
 use crate::graph_editor::node;
+use crate::graph_editor::Type;
+use crate::graph_editor::component::node::Expression;
+use crate::graph_editor::component::type_coloring::{TypeColorMap, MISSING_TYPE_COLOR};
```
wdanilo, line 30: split this use pls

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -324,9 +325,8 @@ enum ShapeView {
 
 impl ShapeView {
     /// Constructor.
-    fn new(number_of_ports:NonZeroU32, logger:&Logger, scene:&Scene) -> Self {
-        let number_of_ports = number_of_ports.get();
-        if number_of_ports == 1 {
+    fn new(number_of_ports:u32, logger:&Logger, scene:&Scene) -> Self {
+        if number_of_ports <= 1 {
             ShapeView::Single { view: component::ShapeView::new(&logger,&scene) }
```
wdanilo, line 330: This needs a description in doc comment to this function now (that you output Single) if port count was 0

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -373,13 +373,32 @@ impl ShapeView {
...
+    }
 }
 
 impl display::Object for ShapeView {
     fn display_object(&self) -> &display::object::Instance {
         match self {
-            ShapeView::Single { view }               => view.display_object(),
-            ShapeView::Multi  { display_object, .. } => display_object,
+            ShapeView::Single{ view }               => view.display_object(),
+            ShapeView::Multi{ display_object, .. } => display_object,
```
wdanilo, line 401: alignment

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
wdanilo: As I've written in discord:

You have 2 functions set_expression_type which do the same, they should be merged together. You have duplicated logic of this type_map – this needs to be refactored in one place. I'm ok with keeping references to a single, internally mutable type_map if you want. Or even keeping immutable type-map (see the im libarary - this could sometimes be useful - a node can create type map and can distribute it to input and output port manager. The im hashmaps are immutable and have fast clone (based on Rc)). Anyway, whatever solution we choose, we should never duplicate the logic of creating a strucutre, so lets create and manage type map in a single place.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
wdanilo: As discussed on discord, lets rename it to `expression` and `pattern` in the appropriate places.

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
```diff
@@ -582,43 +611,91 @@ impl OutputPortsData {
...
+
+        // We want to be able to match each `PortId` to a `Crumb` from the expression, including
+        // the root.
+        let number_of_ports     = 1 + expression.output_span_tree.root_ref().leaf_iter().count() as u32;
+        let expression_nodes    = expression.output_span_tree.root_ref().leaf_iter();
+        // Create a `PortId` for every leaf and store them in tuples.
+        let port_ids_for_crumbs = expression_nodes.enumerate().map(|(index, node)| {
+            (PortId::new(index + 1), node.crumbs)
+        });
+        let mut id_map      = HashMap::<PortId,span_tree::Crumbs>::from_iter(port_ids_for_crumbs);
```
wdanilo, line 666: rm alignment

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
wdanilo: You mean that you create animations `delay_show` and `delay_hide` ON EVERY `set_expression` right? This is of course bad - you should create them on init. There is nothing bad in the creation of animations on init and THEN connecting to them multiple times from new networks. When this network will be removed, these animations will correctly handle the dropped network (will clean unused connections). But of course, creating new nodes in the same network constantly is a memory leak

src/rust/ide/lib/view/src/graph_editor/component/node/port/output.rs
wdanilo: Ok, these are patterns!

PR #664, opened 2020-07-16
Deserialization of the `executionContext/executionFailed` notification.

mwu-tow:
### Pull Request Description
Previously the `executionContext/executionFailed` notification wasn't correctly serialized, leading to deserialization errors in logs.
Now it is deserialized and lead to logging proper errors, as reported by the language server.

### Checklist

### DIFF
--- a/build/run.js
+++ b/build/run.js
@@ -127,7 +127,7 @@ commands.build.rust = async function(argv) {
 
         console.log('Checking the resulting WASM size.')
         let stats = fss.statSync(paths.dist.wasm.mainOptGz)
-        let limit = 3.54
+        let limit = 3.56
         let size = Math.round(100 * stats.size / 1024 / 1024) / 100
         if (size > limit) {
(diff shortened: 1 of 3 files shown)


build/run.js
```diff
@@ -127,7 +127,7 @@ commands.build.rust = async function(argv) {
 
         console.log('Checking the resulting WASM size.')
         let stats = fss.statSync(paths.dist.wasm.mainOptGz)
-        let limit = 3.54
+        let limit = 3.56
```
wdanilo, line 130: where this increase comes from? I'd understand 0.01 MB, but 0.02 MB seems big for these changes.

farmaazon: IDK, but maybe it's autogenerated code from serde macros?

wdanilo: I'm wondering it Michal did not increase it more than needed :P Because it seems like a lot.

PR #669, opened 2020-07-16
Update CONTRIBUTING.md

wdanilo:
### Pull Request Description
filling missing info about code sizes.


docs/CONTRIBUTING.md
```diff
@@ -172,7 +172,21 @@ have prepared several scripts which maximally automate the process:
...
+  2. If the PR does include new libraries, you are allowed to increase the
+     limit by 10KB, but you should also consider if it is possible to get the
+     same results without using a new library (even by implementing few lines 
+     of code from the library in sources of the project).
+  3. If the PR does include new libraries, and the limit is exceeded by more
+     than 10KB, check which part of the code contributet to it, and talk about 
+     it with the code owner.
+  4. If the PR does include new libraries, and the limit is exceeded by more 
+     than 50KB, it would probably not be merged. Research possible alternatives
+     before talking with code owner about this case.
```
farmaazon, line 189: So the first solution is... to split PR to two lesser parts of code. That's good, but I feel that this is not the intended result. 

However I don't have in mind better solution, so actually this could be the best we could do to keep library reasonably small.

wdanilo: It's not the intention - if its more than 10KB - talk with code owner to check if it can be made smaller. It doesn't mean it would not be merged.

docs/CONTRIBUTING.md
```diff
@@ -172,7 +172,21 @@ have prepared several scripts which maximally automate the process:
   not grew too much in comparison to the previous release. Watching the
   resulting binary size is one of the most important responsibility of each
   contributor in order to keep the project small and suitable for web-based
-  usage.
+  usage. In case the size will exceed the limits:
+  1. If the PR does not include any new libraries, you are allowed to increase 
+     the limit by 10KB. In case the limit will be exceeded by more than 10KB,
+     check which part of the code contributet to it, and talk about it with the
```
iamrecursion, line 178:
```suggestion
     check which part of the code contributed to it, and talk about it with the 
```

docs/CONTRIBUTING.md
```diff
@@ -172,7 +172,21 @@ have prepared several scripts which maximally automate the process:
   not grew too much in comparison to the previous release. Watching the
   resulting binary size is one of the most important responsibility of each
   contributor in order to keep the project small and suitable for web-based
-  usage.
+  usage. In case the size will exceed the limits:
+  1. If the PR does not include any new libraries, you are allowed to increase
```
iamrecursion, line 176: These should be a bulleted list not a numbered one. The numbers imply ordering.

docs/CONTRIBUTING.md
```diff
@@ -172,7 +172,21 @@ have prepared several scripts which maximally automate the process:
...
+  1. If the PR does not include any new libraries, you are allowed to increase 
+     the limit by 10KB. In case the limit will be exceeded by more than 10KB,
+     check which part of the code contributet to it, and talk about it with the 
+     code owner.
+  2. If the PR does include new libraries, you are allowed to increase the
+     limit by 10KB, but you should also consider if it is possible to get the
+     same results without using a new library (even by implementing few lines 
+     of code from the library in sources of the project).
+  3. If the PR does include new libraries, and the limit is exceeded by more
+     than 10KB, check which part of the code contributet to it, and talk about
```
iamrecursion, line 185:
```suggestion
     than 10KB, check which part of the code contributed to it, and talk about 
```

PR #685, opened 2020-07-22
Suggestions fixes

farmaazon:
### Pull Request Description
This PR fixes all discrepancies between IDE and Engine since Engine delivered implementation of suggestions.

### Checklist


src/rust/ide/src/double_representation/graph.rs
```diff
@@ -430,7 +430,7 @@ main =
         println!("zz");
```
notdanilo, line 430: Why do we have this print? I know it isn't part of this PR, but it just caught my attention.

farmaazon: I remove those in my next PR.

PR #686, opened 2020-07-22
Generate HTML Documentation in Suggestion Entries

### LINKED ISSUE #682: Html documentation in Suggestion Entries.
### Summary
Currently, when we retrieve the suggestion entries from the Engine, we store the documentation as is. Instead, it should be converted to HTML using the docparser.

### Specification
The conversion should be performed when creating `model::suggestion_database::Entry` structure.

### Acceptance Criteria & Test Cases
Unit tests.

BinarySoftware:
### Pull Request Description
closes #682 
Finally, the Documentation Parser and Generator is used somewhere (Yayyy 🎉 🎉 🎉 )
Modifies docstring from suggestion database to HTML code, enabling us to present documentation for nodes in searcher

### Checklist

### DIFF
--- a/src/rust/ide/src/model/suggestion_database.rs
+++ b/src/rust/ide/src/model/suggestion_database.rs
@@ -11,6 +11,7 @@ use language_server::types::SuggestionDatabaseUpdateEvent;
 pub use language_server::types::SuggestionEntryArgument as Argument;
 pub use language_server::types::SuggestionEntryId as EntryId;
 pub use language_server::types::SuggestionsDatabaseUpdate as Update;
+pub use parser::DocParser as DocParser;
 
 
 // =============
@@ -51,16 +52,18 @@ impl Entry {
         let this = match entry {
             SuggestionEntryAtom {name,module,arguments,return_type,documentation} =>
                 Self {
-                    name,arguments,return_type,documentation,
+                    name,arguments,return_type,
                     module    : module.try_into()?,
                     self_type : None,
+                    documentation: Entry::gen_doc(documentation),
                     kind      : EntryKind::Atom,
                 },
             SuggestionEntryMethod {name,module,arguments,self_type,return_type,documentation} =>
                 Self {
-                    name,arguments,return_type,documentation,
+                    name,arguments,return_type,
                     module    : module.try_into()?,
                     self_type : Some(self_type),
+                    documentation: Entry::gen_doc(documentation),
                     kind      : EntryKind::Method,
                 },
             SuggestionEntryFunction {name,module,arguments,return_type,..} =>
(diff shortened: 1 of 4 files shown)


src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -11,6 +11,7 @@ use language_server::types::SuggestionDatabaseUpdateEvent;
 pub use language_server::types::SuggestionEntryArgument as Argument;
 pub use language_server::types::SuggestionEntryId as EntryId;
 pub use language_server::types::SuggestionsDatabaseUpdate as Update;
+pub use parser::DocParser as DocParser;
```
farmaazon, line 14:
```suggestion
use parser::DocParser;
```
And it should be put into right place according to our Rust Style Guide

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -51,16 +52,18 @@ impl Entry {
         let this = match entry {
             SuggestionEntryAtom {name,module,arguments,return_type,documentation} =>
                 Self {
-                    name,arguments,return_type,documentation,
+                    name,arguments,return_type,
                     module    : module.try_into()?,
                     self_type : None,
+                    documentation: Entry::gen_doc(documentation),
```
farmaazon, line 58: Align `:`

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -98,6 +101,21 @@ impl Entry {
     pub fn with_name(self, name:impl Into<String>) -> Self {
         Self {name:name.into(),..self}
     }
+
+    /// Generates HTML documentation for documented suggestion
+    fn gen_doc(doc: Option<String>) -> Option<String> {
```
farmaazon, line 106: The None actually indicates an error. So this method should return FallibleResult<String>

Then, during conversion we should also log an error when we are unable to parse documentation (and leave `documentation` field as None - as it is now).

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -98,6 +101,21 @@ impl Entry {
...
+        match doc {
+            Some(d) => {
+                let parser = DocParser::new_or_panic();
+                let output = parser.generate_html_doc_pure(d);
+                match output {
+                    Ok(result) => Some(result),
+                    Err(_)     => None,
+                }
+            },
+            None    => None,
```
farmaazon, line 116:
```suggestion
            None => None,
```
If lines are not consecutive, then we don't align anytihng.

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -46,36 +47,46 @@ pub struct Entry {
 
 impl Entry {
     /// Create entry from the structure deserialized from the Language Server responses.
-    pub fn from_ls_entry(entry:language_server::types::SuggestionEntry) -> FallibleResult<Self> {
+    pub fn from_ls_entry(entry:language_server::types::SuggestionEntry, logger:impl AnyLogger)
+        -> FallibleResult<Self> {
         use language_server::types::SuggestionEntry::*;
+        let convert_doc = |doc: Option<String>| doc.and_then(|doc| match Entry::gen_doc(doc) {
+            Ok(d)  => Some(d),
+            Err(err) => {
```
wdanilo, line 55: alignment

wdanilo, line 55: there are functions to convert Result to Option, please use them instead, it will be shorter

BinarySoftware: Yes there are - like `.ok()` or `.err()`, but they're unusable in our case, as we want to log the error, which would be impossible with `.ok` and `.err` would lose the value.
Here - all of available transformations : https://docs.google.com/drawings/d/1EOPs0YTONo_FygWbuJGPfikO9Myt5HwtiFUHRuE1JVM/preview

PR #706, opened 2020-07-31
Argument Completion Suggestions

mwu-tow:
### Pull Request Description
This one PR refers both issues regarding suggestion for the arguments (https://github.com/enso-org/ide/issues/629 and https://github.com/enso-org/ide/issues/612) and is a result of cooperation of @farmaazon and @mwu-tow 

### Checklist

- [] All code has been profiled where possible.


src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -230,6 +259,28 @@ impl SuggestionDatabase {
         self.entries.borrow().values().cloned().find(|entry| entry.method_id().contains(&id))
     }
 
+    pub fn lookup_by_name_and_location
+    (&self, name:impl Str, location:TextLocation) -> Vec<Rc<Entry>> {
+        self.entries.borrow().values().cloned().filter(|entry| {
```
mwu-tow, line 264: here and below only the elements returned should be cloned

src/rust/ide/src/controller/searcher.rs
```diff
@@ -578,57 +587,150 @@ impl Searcher {
...
+                self.database.get(*id).map_none(||
+                    error!(self.logger,"Missing entry {id} in Suggestion Database.")
+                ).map(Suggestion::Completion)
+            });
+            suggestions.extend(entries);
+        }
+        Ok(suggestions)
+    }
+
+    fn possible_function_calls(&self) -> Vec<CompletionSuggestion> {
```
mwu-tow, line 656: We could also use information about function arity to reject some overloads (e.g. if entry takes 2 parameters and we already provided 3 arguments).

src/rust/ide/src/controller/searcher.rs
```diff
@@ -578,57 +587,150 @@ impl Searcher {
...
+            } else {
+                let name     = &call.function_name;
+                let location = *self.position_in_code;
+                Some(self.database.lookup_by_name_and_location(name,location))
+            }
+        };
+        opt_result().unwrap_or(Vec::new())
+    }
+
+    fn module_method_called(&self, call:&SimpleFunctionCall) -> Option<QualifiedName> {
```
mwu-tow, line 672: This method really needs rename, it's really hard to grasp what it's doing.

src/rust/ide/src/controller/searcher.rs
```diff
@@ -578,57 +587,150 @@ impl Searcher {
...
+            .and_then(|expr| ast::opr::as_access_chain(&expr.func))
+            .is_some()
+    }
+}
+
+
+struct SimpleFunctionCall {
+    this_argument : Option<Ast>,
+    function_name : String,
+}
```
mwu-tow, line 717: Shouldn't this also include support for the non sugared universal invocation syntax `method this`?

src/rust/ide/src/controller/searcher.rs
```diff
@@ -957,6 +1187,34 @@ mod test {
         assert!(Rc::ptr_eq(&arg.picked_suggestion,&entry2));
     }
 
+    #[test]
```
mwu-tow, line 1190: this whole test is total TODO, I've just started it.
Basically I expect to finish commented out cases, add more checks and more expected method calls, as needed.

src/rust/ide/src/controller/searcher.rs
```diff
@@ -578,57 +587,150 @@ impl Searcher {
...
+            }
+        };
+        opt_result().unwrap_or(Vec::new())
+    }
+
+    fn module_method_called(&self, call:&SimpleFunctionCall) -> Option<QualifiedName> {
+        let module_ast      = self.graph.graph().module.ast();
+        let position        = *self.position_in_code;
+        let module          = double_representation::module::Info {ast:module_ast};
+        let this_name       = ast::identifier::name(call.this_argument.as_ref()?)?;
```
mwu-tow, line 676: To be considered what happens if "this_argument" is a qualified name, e.g. we have "Foo.Bar.method". Is this even legal?

src/rust/ide/src/controller/searcher.rs
farmaazon: In such case the `this_argument` is None.

O, or perhaps we should rename `this_argument` to the `dot_applied_argument`

src/rust/ide/src/controller/searcher.rs
farmaazon: Actually the only place in enso docs where I found qualified names is only import syntax. So as I understood, you cannot use qualified names in code, only in imports. However this may change in future.

src/rust/ide/src/controller/searcher.rs
farmaazon: Those are filtered out anyway in `return_types_for_argument_completion`

src/rust/ide/src/controller/searcher.rs
farmaazon: This is a little more complicated and not in the task scope, therefore I leave it as it is now.

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -230,6 +259,28 @@ impl SuggestionDatabase {
         self.entries.borrow().values().cloned().find(|entry| entry.method_id().contains(&id))
     }
 
+    pub fn lookup_by_name_and_location
```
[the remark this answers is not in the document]
farmaazon: now all those lookups takes module and location as argument.

PR #710, opened 2020-08-05
Add menu to select visualisation for given type + test switching visualizations

MichaelMauderer:
### Pull Request Description
Adds a menu to select visualizations for a node

Peek 2020-09-24 17-06

Peek 2020-09-24 17-23

### Known Issues

**Text Artifacts**
There seem to be some visual artifacts from the text area, which might be caused by the list of the label.

Peek 2020-09-22 11-01

At the end in the gif there are a bunch of letters at the scene origin all jumbled up. 

**`ListView` alows selection when invisible**
See https://github.com/enso-org/ide/issues/804
Update: No longer an issue as we resize the ListView now and avoid this issue that way.

**Visualisation containers don't have a nice background in Dark mode**
Not addressed in this PR as there seems to be some work in progress about the container. (It has been moved to an HTML entity that at the moment just has a hardcoded color)

**`ListView` resize performance**
In debug mode running an animation that resizes the `ListView` drops fps on my machine <30 fps. Seems to work fine in prod.

### Checklist


wdanilo:
The result presented on the screenshot does not exactly match the description in the task: https://github.com/enso-org/ide/issues/541. In particular:
1. The task tells:
    > after you HOVER visualization, on top of it (right upper corner with a few pixel distance from borders), there should appear a dark, transparent, rounded rectangle with the name of the current visualization selected and a triangle down.

    Which do not happen now. Now the triangle is there always, after you hover, no dark background appears, and after hover there is no name of the currently selected visualization visible. 

2. The task states:
    > After pressing this button, a drop down should appear, and its appearance should be similar to this <image>

    Which again doesn't happen right now in the style shown in the screenshot. The thing here is that the dropdown should not fill the whole visualization. It should be a "separate component". Sometimes visualization will be very narrow (like vertical slider with 20px width) - you want to have these controls still accessible there. So it was meant to look like this:

There are moreover few things to be fixed:
3. The visualization names should be much shorter. For example, instead of "Raw Text Visualization" it should be named "JSON". Then name "Visualization" doesn't make sense - every option is visualization. This will help us make that much nicer.

---------------------

So, basically, before we hover, we have this: 
    n2
     Then, after we hover some of the green area:
    n2
    We should see (the red boxes are other icons in the future) (the top of the dark area should be rounded, the bottom should not):
     n2
    After pressing this area (not like now that you need to be over the arrow, which makes it hard to select):
    n2
    A pop-up window should appear, in a style shown in the task (the colors need to still be adjusted in the final implementation) (all corners of popup should be rounded):
    image
    n2
    In the drop down list, there should not be the visualization that is currently chosen (of course).

Moreover, when we hover the list, lets make the first selection appear with transparency rather than growing it left to right - it looks bad now and just transparency animation on the beginning will do its job perfectly.

MichaelMauderer:
@wdanilo Updated to address the raised issues.

src/rust/ide/view/graph-editor/src/component/visualization/container/action_bar.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    ensogl::define_shape_system! {
+        () {
+            let width  : Var<Pixels> = "input_size.x".into();
+            let height : Var<Pixels> = "input_size.y".into();
+            let radius             = node::NODE_SHAPE_RADIUS.px() ;
+            let background_rounded = Rect((&width,&height)).corners_radius(&radius);
+            let background_sharp   = Rect((&width,&height/2.0)).translate_y(-&height/4.0);
+            let background         = background_rounded + background_sharp;
+            let fill_color         = color::Rgba::from(color::Lcha::new(0.1,0.013,0.18,0.6));
```
BinarySoftware, line 78: Please use theme manager for that, and show how it looks in light mode.

src/rust/ide/view/graph-editor/src/component/visualization/container/action_bar.rs
```diff
@@ -0,0 +1,289 @@
...
+use ensogl::gui::component;
+use ensogl_text as text;
+
+
+
+// =================
+// === Constants ===
+// =================
+
+const HOVER_COLOR    : color::Rgba = color::Rgba::new(1.0,0.0,0.0,0.000_001);
```
BinarySoftware, line 26: Please use theme manager for that, and show how it looks in light mode.

src/rust/ide/view/graph-editor/src/component/visualization/container/action_bar.rs
```diff
@@ -0,0 +1,289 @@
...
+            background,
+        }.init()
+    }
+
+    fn init(self) -> Self {
+        self.add_child(&self.hover_area);
+
+        self.set_label("None");
+
+        self.visualisation_chooser_label.frp.set_default_color.emit(color::Rgba::new(1.0,1.0,1.0,1.0));
```
BinarySoftware, line 151: Please use theme manager for that, and show how it looks in light mode.

src/rust/ide/view/graph-editor/src/component/visualization/container/visualization_chooser.rs
```diff
@@ -0,0 +1,309 @@
...
+use ensogl::gui::component;
+use ensogl_gui_list_view as list_view;
+
+
+
+// =================
+// === Constants ===
+// =================
+
+const HOVER_COLOR    : color::Rgba = color::Rgba::new(1.0,0.0,0.0,0.000_001);
```
BinarySoftware, line 24: Duplicate, also please use theme manager for that, and show how it looks in light mode.

src/rust/ide/view/graph-editor/src/component/visualization/container/visualization_chooser.rs
```diff
@@ -0,0 +1,309 @@
...
+    use super::*;
+
+    ensogl::define_shape_system! {
+        () {
+            let width            = Var::<Pixels>::from("input_size.x");
+            let height           = Var::<Pixels>::from("input_size.y");
+            let triangle         = Triangle(width,height);
+            let triangle_down    = triangle.rotate(Var::<f32>::from(std::f32::consts::PI));
+
+            let fill_color       = color::Rgba::from(color::Lcha::new(0.8,0.013,0.18,1.0));
```
BinarySoftware, line 43: Please use theme manager for that, and show how it looks in light mode.

src/rust/ide/view/graph-editor/src/component/visualization/container/visualization_chooser.rs
```diff
@@ -0,0 +1,309 @@
...
+        Self {frp,model}.init(app)
+    }
+
+    fn init(self, app:&Application) -> Self {
+        let network  = &self.frp.network;
+        let frp      = &self.frp;
+        let model    = &self.model;
+
+        let scene    = app.display.scene();
+        let mouse    = &scene.mouse.frp;
```
BinarySoftware, line 205:
```suggestion
        let network = &self.frp.network;
        let frp     = &self.frp;
        let model   = &self.model;

        let scene   = app.display.scene();
        let mouse   = &scene.mouse.frp;
```

src/rust/ide/view/graph-editor/src/component/visualization/container/action_bar.rs
```diff
@@ -0,0 +1,289 @@
...
+/// ```text
+///     / ---------------------------- \
+///    |            <label>         V   |
+///    |--------------------------------|
+///
+/// ```
+#[derive(Clone,CloneRef,Debug)]
+pub struct QuickActionBar {
+         model : Rc<Model>,
+    pub frp    : Frp
```
BinarySoftware, line 224:
```suggestion
    model   : Rc<Model>,
    pub frp : Frp
```

BinarySoftware (commented):
Leaving some notes, Please, could You default to using theme manager? If You need any help, please ask me and i'll help you to set it up. We want a possibility in the future to easily switch between themes, so we cannot have hardcoded color literals, and now that i've put the theme manager into use, it makes sense to use it.

src/rust/ide/view/graph-editor/src/component/visualization/container/action_bar.rs
MichaelMauderer: This is not a real visible, color. Just an almost transparent overlay to get mouse events.

MichaelMauderer:
Updated to conform with the new theming, use the new list widget and FRP macros. Also, some glitches squashed.

src/rust/ide/view/graph-editor/src/component/visualization/container/visualization_chooser.rs
```diff
@@ -0,0 +1,315 @@
...
+// ===============================
+// === Visualisation Path List ===
+// ===============================
+
+/// The `VisualisationPathList` allows us to show a `Vec<visualization::Path>` in the
+/// `list_view::ListView` by implementing `list_view::entry::ModelProvider`.
+#[derive(Clone,Debug,Default)]
+struct VisualisationPathList {
+    content: Vec<visualization::Path>
+}
```
farmaazon, line 80: Wrapping vector of displayable things seems to me like a very common pattern, thence I would consider making a batch-implementation of `list_view::entry::ModelProvider` for every `Vec<T:Into<Model>>` and `From<T:Display>` for `Model` - then you will not need such implementations.

Additionally, `ListView` structure could have helper method `set_entries(&self, entries:impl entry::ModelProvider)` which just cast entries and emit frp endpoint.

farmaazon (commented):
A small suggestion from my side

wdanilo:
Seems super cool! Fast check-questions:
1. Does it correctly work with dark theme as well?
2. Why the default selection is "None"? When openning a visualization, a default one should be selected - forcin user to make extra step deosnt make any sense. The most default visualization is the JSON one (all Enso structs will have json printing implemented by default), but for some types other defaults should be defined. Till we dont have the mechanissm for associating types with their defaults, we can defautl to JSON always. But keeping "None" is wrong and should not be like this.
3. I havent tested it yet, but on the gifs Ive seen youre always clicking next to arrow - will it work if you click more on the left - on the currently selected label as well?

MichaelMauderer:
> Seems super cool! Fast check-questions:
> 
>     1. Does it correctly work with dark theme as well?

Yes! I'll add screengrabs.

>     2. Why the default selection is "None"? When openning a visualization, a default one should be selected - forcin user to make extra step deosnt make any sense. The most default visualization is the JSON one (all Enso structs will have json printing implemented by default), but for some types other defaults should be defined. Till we dont have the mechanissm for associating types with their defaults, we can defautl to JSON always. But keeping "None" is wrong and should not be like this.

I think that is just the current default behaviour. But I agree that something more sensible would be better. I can add that behaviour in this PR.
 
>     3. I havent tested it yet, but on the gifs Ive seen youre always clicking next to arrow - will it work if you click more on the left - on the currently selected label as well?

Right now it is just the arrow that opens the menu. But I can expand the area to encompas the label, too.

wdanilo:
Some minor visual comments:
image

1. This white background hurts my eyes in dark scene. It was much darker earlier. This needs to be fixed, it looks very bad.
2. There is no offset between the top bar (name of currently selected vis + triangle button) and the drop down menu - can we add few pixel offset between these, please?

src/rust/ensogl/lib/drop-down-menu/src/component.rs
```diff
@@ -0,0 +1,352 @@
...
+
+        let scene   = app.display.scene();
+        let mouse   = &scene.mouse.frp;
+
+        frp::extend! { network
+
+
+            // === Input Processing ===
+
+            eval  frp.input.set_entries ([model](entries) {
```
wdanilo, line 217: double space after eval

src/rust/ensogl/lib/list-view/src/entry.rs
```diff
@@ -114,6 +120,99 @@ impl ModelProvider for EmptyProvider {
...
+
+    /// Return the index to the unmasked underlying data. Will only be valid to use after
+    /// calling `clear_mask`.
+    pub fn unmasked_index(&self, ix:Id) -> Id {
+        match self.mask.get() {
+            None                 => ix,
+            Some(id) if ix < id  => ix,
+            Some(_)              => ix+1,
+        }
+    }
```
wdanilo, line 183: This needs better docs. After reading it I didnt getwhat really is "unmasked underlying data". It may be explained somewhere, but then thisdoc should point me to the explanation place.

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -142,35 +148,38 @@ pub mod overlay {
...
+    pub select                         : frp::Source,
+    pub deselect                       : frp::Source,
+    pub set_size                       : frp::Source<Vector2>,
+    pub enable_fullscreen              : frp::Source,
+    pub disable_fullscreen             : frp::Source,
+    pub  preprocessor                  : frp::Stream<EnsoCode>,
+
+    scene_shape                        : frp::Sampler<scene::Shape>,
+    size                               : frp::Sampler<Vector2>,
+    preprocessor_select                : frp::Source<EnsoCode>,
```
wdanilo, line 164: this looks like accintal change

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -142,35 +148,38 @@ pub mod overlay {
...
+            set_visualization              <- source();
+            set_data                       <- source();
+            select                         <- source();
+            deselect                       <- source();
+
+            set_size                       <- source();
+            enable_fullscreen              <- source();
+            disable_fullscreen             <- source();
+            preprocessor_select            <- source();
+            size                           <- set_size.sampler();
```
wdanilo, line 181: accidental change again ?

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -413,8 +436,13 @@ impl ContainerModel {
             self.fullscreen_view.background_dom.dom().set_style_or_warn("width", "0", &self.logger);
             self.fullscreen_view.background_dom.dom().set_style_or_warn("height", "0", &self.logger);
             // self.fullscreen_view.background.shape.sprite.size.set(zero());
+
+            let quick_action_size = Vector2::new(size.x, QUICK_ACTION_BAR_HEIGHT);
+            self.action_bar.frp.set_size.emit(quick_action_size);
         }
 
+        self.action_bar.set_position_y((size.y - QUICK_ACTION_BAR_HEIGHT) / 2.0);
```
wdanilo, line 444: sometimes its named "action bar" and soemetimes "quick action bar" - this sohuld be unified

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -429,6 +457,12 @@ impl ContainerModel {
         // self.view.background.shape.roundness.set(value);
         // self.fullscreen_view.background.shape.roundness.set(value);
     }
+
+    fn show_visualisation(&self) {
+        if let Some(vis) = self.visualization.borrow().as_ref() {
+            vis.display_object().set_parent(&self.view)
```
wdanilo, line 463: do not use `.display_object()` explicitly here

wdanilo (commented):
Beautiful code <3 Almost no comments from my side. I would have accepted it if not the visual issues!

src/rust/ensogl/lib/list-view/src/entry.rs
```diff
@@ -70,6 +70,12 @@ impl Model {
     }
 }
 
+impl<T:Display> From<T> for Model {
+    fn from(item: T) -> Self {
+        Model::new(format!("{}", item))
```
farmaazon, line 75:
```suggestion
        Model::new(item.to_string()))
```
`ToString` trait is implemented for everything implementing Display.

src/rust/ensogl/lib/list-view/src/entry.rs
```diff
@@ -114,6 +120,99 @@ impl ModelProvider for EmptyProvider {
 }
 
 
+// === Model Provider for Vectors ===
+
+/// An Entry Model Provider that wraps a Vec<T:Into<Model>>.
+#[derive(Clone,Debug)]
+pub struct VectorProvider<T> {
+    content : Vec<T>,
+}
```
farmaazon, line 129: Why do we need wrapper? Cannot we just `impl<T:Into<Model> + Debug + Clone> ModelProvider for Vec<T>`?

src/rust/ensogl/lib/list-view/src/entry.rs
```diff
@@ -114,6 +120,99 @@ impl ModelProvider for EmptyProvider {
...
+        VectorProvider{content}
+    }
+}
+
+
+// === Masked Model Provider ===
+
+/// An Entry Model Provider that wraps a Vec<T:Into<Model>> and allows the masking of a single item.
+#[derive(Clone,Debug)]
+pub struct SingleMaskedProvider {
```
farmaazon, line 154: At this point I would consider putting all providers and trait into `provider` submodule, and rename everything here removing `Provider` suffix. Then we could reimport "ModelProvider" in entry.rs

src/rust/ensogl/lib/list-view/src/entry.rs
MichaelMauderer: Yes! Did that and removed the wrapper.

src/rust/ensogl/lib/list-view/src/entry.rs
MichaelMauderer: I would leave that for later to avoid conflicts with the searcher PR.

MichaelMauderer:
>     1. This white background hurts my eyes in dark scene. It was much darker earlier. This needs to be fixed, it looks very bad.
That is the current state of main, this seems to be a hardcoded HTML element at the moment. (It's listed in the known issues)

>     2. There is no offset between the top bar (name of currently selected vis + triangle button) and the drop down menu - can we 

Added! Can be changed as desired via FRP endpoint.

>     3. In the white theme the dropdown looks strange - it has the same bg as the window behind it. Definitely a shadow is missing there.
I think shadows are in the works in  #799

>     4. Can we make drop down menu appear / disappear animation? Justl iek in this PR: #799
Done!

BinarySoftware:
One thing i found : It won't show visualizations defined in enso project folder

MichaelMauderer:
> One thing i found : It won't show visualizations defined in enso project folder

@BinarySoftware That's worth reporting as a separate issue for the IDE part that does visualization discovery.

BinarySoftware:
> > One thing i found : It won't show visualizations defined in enso project folder
> 
> 
> 
> @BinarySoftware That's worth reporting as a separate issue for the IDE part that does visualization discovery.

Ok great 😄

src/rust/ide/view/graph-editor/src/component/visualization/container/visualization_chooser.rs
```diff
@@ -0,0 +1,142 @@
...
+            // === Output Processing ===
+
+            frp.source.mouse_over   <+ menu.icon_mouse_over;
+            frp.source.mouse_out    <+ menu.icon_mouse_out;
+            frp.source.menu_closed  <+ menu.menu_closed;
+            frp.source.menu_visible <+ menu.menu_visible;
+
+            selected_path <- model.selection_menu.frp.chosen_entry.map(f!([model](entry_id)({
+                entry_id.map(|entry_id| model.visualization_alternatives.borrow().get(entry_id).cloned()).flatten()
+
```
wdanilo, line 129: rm nl

src/rust/ide/view/graph-editor/src/component/visualization/path.rs
```diff
@@ -4,6 +4,7 @@ use crate::prelude::*;
 
 use crate::data;
 
+use fmt::Formatter;
```
wdanilo, line 7: I'd ask you to remove it and use `fmt::Formatter` instead in the use place. Less imports and there is just a single usage of that

wdanilo (approved):

src/rust/ensogl/lib/list-view/src/entry.rs
farmaazon: The searcher PR landed, so it can be refactored in this one.

src/rust/ensogl/lib/drop-down-menu/src/component.rs
```diff
@@ -0,0 +1,370 @@
...
+    selection_menu  : list_view::ListView,
+
+    // `SingleMaskedProvider` allows us to hide the selected element.
+    content         : RefCell<Option<list_view::entry::SingleMaskedProvider>>,
+}
+
+impl Model {
+    fn new(app:&Application) -> Self {
+        let logger         = Logger::new("visualization_chooser::Model");
+        let scene          = app.display.scene().clone_ref();
```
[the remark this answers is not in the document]
MichaelMauderer: The `clone_ref` actually isn't needed. Removed.

src/rust/ensogl/lib/drop-down-menu/src/component.rs
```diff
@@ -0,0 +1,370 @@
...
+            hide_menu <- source::<()>();
+            show_menu <- source::<()>();
+
+            eval_ hide_menu ([model,frp,menu_height]{
+                model.selection_menu.deselect_entries.emit(());
+                frp.source.menu_visible.emit(false);
+                frp.source.menu_closed.emit(());
+                /// FIXME: If we end at 0.0 the ListView will still display the first
+                /// content item. This avoids the slowdown close to 0.0, so we can
+                /// manually remove the LisTView from the scene at 0.0.
```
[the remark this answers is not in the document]
MichaelMauderer: Actually it's not really a `FIXME` but an explanation of the workaround for a bug. Left the note to explain what is going on here, but removed the fixme. I will file an issue for the bug though and mention that here.

PR #716, opened 2020-08-09
Ability to change text on nodes via GUI

wdanilo:
### Pull Request Description
- Implements [Ability to change text on nodes via GUI](https://github.com/enso-org/ide/issues/462)

### Important Notes
Please note that it is not working perfectly. For example, after getting into edit mode and selecting other node, the edit mode should be stopped. It is also possible to get into edit mode of two nodes at the same time. Also, the key binding `cmd + node press` is right now bundled to both node edit AND multi-node selection. All these issues will be resolved as part of the upcoming [Fixing shortcut manager](https://github.com/enso-org/ide/issues/713) issue.

### Checklist


src/rust/ensogl/lib/text/src/typeface/pen.rs
```diff
@@ -3,21 +3,40 @@
 //! [freetype documentation](https://www.freetype.org/freetype2/docs/glyphs/glyphs-3.html#section-1)
 
 use crate::prelude::*;
-
```
farmaazon, line 6: I believe this line should remain.

src/rust/ensogl/lib/text/src/typeface/pen.rs
```diff
@@ -83,9 +90,7 @@ impl Pen {
 #[cfg(test)]
 mod tests {
     use super::*;
-
     use crate::typeface::font::GlyphRenderInfo;
-
```
farmaazon, line 88: Those lines should remain.

PR #720, opened 2020-08-12
Update for Language Server API Changes

mwu-tow:
### Pull Request Description
This PR fixes IDE to work with engine after https://github.com/enso-org/enso/pull/1055

This also includes some refactoring for the module-related types that were affected by these changes.

### Checklist


src/rust/ide/src/double_representation/module.rs
```diff
@@ -320,8 +378,8 @@ pub fn locate
 /// desugaring implicit extensions methods for modules).
 pub fn lookup_method
 (ast:&known::Module, method:&language_server::MethodPointer) -> FallibleResult<definition::Id> {
-    let module_path = model::module::Path::from_file_path(method.file.clone())?;
-    let explicitly_extends_looked_type = method.defined_on_type == module_path.module_name();
+    let module_path = QualifiedName::try_from(method)?;
+    let explicitly_extends_looked_type = method.defined_on_type == module_path.name();
```
farmaazon, line 382:
```suggestion
    let module_path                    = QualifiedName::try_from(method)?;
    let explicitly_extends_looked_type = method.defined_on_type == module_path.name();
```

src/rust/ide/src/model/project.rs
```diff
@@ -60,9 +61,19 @@ pub trait API:Debug {
...
     }
+
+    /// Look up given id in the suggestion database and if it is a known method obtain a pointer to
+    /// it.
+    fn method_ptr_from_db_id
+    (&self, id:SuggestionId) -> FallibleResult<language_server::MethodPointer> {
+        let db    = self.suggestion_db();
+        let entry = db.lookup(id)?;
+        language_server::MethodPointer::try_from(entry.as_ref()).map_err(Into::into)
+    }
```
farmaazon, line 76: Can this function be implemented on SuggestionDatabase instead of Project Model?

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -230,6 +266,11 @@ impl SuggestionDatabase {
         self.entries.borrow().get(&id).cloned()
     }
 
+    /// Get suggestion entry by id. Returns `Result`.
+    pub fn lookup(&self, id:EntryId) -> Result<Rc<Entry>,NoSuchEntry> {
+        self.get(id).ok_or(NoSuchEntry(id))
+    }
+
```
farmaazon, line 273: Other "lookup" methods could return Result instead of Option as well.

mwu-tow: I've looked into this but come to a conclusion that this isn't as good idea.
Other lookup methods do not look for the `EntryId` but rather try matching by some arbitrary criterais. As such, not being able to match any entry in such cases is much less error than in this case. 
Also, all usages would just discard such errors. And each of them would require custom error types.
Too much work with no use.

src/rust/ide/src/double_representation.rs
```diff
@@ -24,3 +27,62 @@ pub mod test_utils;
...
+/// This value corresponds to contents of the `Cons` AST shape.
+pub struct ReferentName(String);
+
+impl ReferentName {
+    /// Check if the given text would be a valid referent name;
+    pub fn validate(name:impl AsRef<str>) -> Result<(), NotReferentName> {
+        let name = name.as_ref();
+        // TODO [mwu]
+        //  We should be able to call parser or sth to verify that other requirements for the
+        //  referent form identifiers are fulfilled.
```
farmaazon, line 62: When we'll do it? Add an info, that we could do it when rust parser will be available.

src/rust/ide/src/double_representation/module.rs
```diff
@@ -2,18 +2,81 @@
...
+
+        let segments:Vec<_> = Result::from_iter(names)?;
+        Self::new(segments).map_err(Into::into)
+    }
+
+    /// Build module's qualified name for a project with a given name.
+    pub fn qualified(&self, project_name:impl Into<String>) -> QualifiedName {
+        // TODO why safe?
+        QualifiedName::from_segments(project_name,&self.segments).unwrap()
+    }
```
farmaazon, line 62: In fact it is not safe, because user can give a name what is not a valid project name.

src/rust/ide/src/model/suggestion_database.rs
```diff
@@ -23,7 +23,7 @@ use enso_protocol::language_server::SuggestionId;
 // ==============
 
 #[allow(missing_docs)]
-#[derive(Debug,Fail,Clone,Copy)]
+#[derive(Debug,Fail,Clone,Copy,PartialEq)]
```
farmaazon, line 26:
```suggestion
#[derive(Debug,Clone,Copy,Eq,Fail,PartialEq)]
```

src/rust/lib/enso-prelude/src/result.rs
```diff
@@ -0,0 +1,19 @@
+//! This module defines utilities for working with the `Result` type.
+
+/// Adds utilities to the `Result` type.
+pub trait ResultOps {
+    type Item;
+    type Error;
+
+    /// Call the given handler if this is an error and promote `Result` to `Option`.
+    fn handle_err<F>(self, f:F) -> Option<Self::Item> where F : FnOnce(Self::Error);
```
farmaazon, line 9: This function is cool! As I remember, it could be used in many places in `node_editor.rs` file. If you want you can look through all `if let Err(` and replace them with this.

src/rust/ide/src/double_representation/module.rs
mwu-tow: Good catch, method is unused, I wanted to remove it but somehow missed.

src/rust/lib/enso-prelude/src/result.rs
mwu-tow: I don't think this really fits there, as these `if let Err`s never involve promotion to the option after error is handled.
If we just want to do something if there's an error, I consider `if let Err` idiomatic and reasonable way.

src/rust/ide/tests/language_server.rs
```diff
@@ -62,6 +64,9 @@ wasm_bindgen_test_configure!(run_in_browser);
 // #[wasm_bindgen_test::wasm_bindgen_test(async)]
 #[allow(dead_code)]
 async fn ls_text_protocol_test() {
+    // FIXME broken, as cannot know the project name
```
wdanilo, line 67: this should be marked as `FIXME[MWU]` and also the description should be better - right now I completely do not understand it.

src/rust/ide/src/double_representation.rs
```diff
@@ -24,3 +27,64 @@ pub mod test_utils;
...
 /// Link: https://github.com/luna/enso/blob/main/doc/syntax/encoding.md
 pub const INDENT : usize = 4;
+
+
+
+// ====================
+// === ReferentName ===
+// ====================
+
+
```
wdanilo, line 37: spacing - single enter here

Issue #464, opened 2020-05-26
Fix old visual elements state. After selected node is removed and new is created, it looks like it was already selected.


wdanilo:

sylwiabr:
please fill task description

PR #725, opened 2020-08-14
Fixing the bug: Fix old visual elements state. 

wdanilo:
### Pull Request Description
- Implements (Fix old visual elements state. After selected node is removed and new is created, it looks like it was already selected.)[https://github.com/enso-org/ide/issues/464]
- Refactors some doc comments 

### Checklist

### DIFF
--- a/src/rust/ensogl/lib/core/src/system/gpu/data/attribute.rs
+++ b/src/rust/ensogl/lib/core/src/system/gpu/data/attribute.rs
@@ -75,7 +75,7 @@ impl {
         })
     }
 
-    /// Adds a new named buffer to the scope.
+    /// Add a new named buffer to the scope.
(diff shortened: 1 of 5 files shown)


src/rust/ensogl/lib/core/src/system/gpu/data/attribute.rs
```diff
@@ -121,11 +121,14 @@ impl {
         })
     }
 
-    /// Disposes instance for reuse in the future. Please note that the disposed data still
+    /// Dispose instance for reuse in the future. Please note that the disposed data still
     /// exists in the buffer and will be used when rendering. It is yours responsibility to hide
-    /// id, fo example by degenerating vertices.
+    /// it, for example by degenerating the mesh (putting all vertices in the same place).
```
farmaazon: This docs should be adjusted: it not quite user responsibility to clear data, as this function set them to the default.

PR #772, opened 2020-08-28
Fix electron-related bugs

BinarySoftware:
### Pull Request Description
This PR fixes :
#247 - transparency of window.
#484 - moving window was really hard because of lack of frame.
#743 - window controls covered the project name. Now as the frame is reintroduced to the window, controls are part of it, and they don't overlap the scene.

### Checklist

### DIFF
--- a/src/js/lib/client/src/index.js
+++ b/src/js/lib/client/src/index.js
@@ -328,18 +328,26 @@ function createWindow() {
         windowPreferences.devTools = true
     }
 
-    if (args.frame) {
-        windowPreferences.frame         = true
-        windowPreferences.titleBarStyle = 'default'
-    }
+    // FIXME [MM]
+    // Currently enabled to improve usability and dragging of window as well
+    // as fixes #743 as well as #
+    // if (args.frame) {
+    //     windowPreferences.frame         = true
+    //     windowPreferences.titleBarStyle = 'default'
+    // }
+    windowPreferences.frame         = true
+    windowPreferences.titleBarStyle = 'default'
 
     if (args['background-throttling']) {
         windowPreferences.backgroundThrottling = true
     }
 
-    if (args.vibrancy == false) {
-        windowPreferences.vibrancy = false
-    }
+    // FIXME [MM]
+    // Currently disabled as of issue on Windows platform (#247)
+    // if (args.vibrancy == false) {
+    //     windowPreferences.vibrancy = false
+    // }
+    windowPreferences.vibrancy = false
 
     const window = new Electron.BrowserWindow(windowPreferences)
 
--- a/src/js/lib/client/src/index.js
+++ b/src/js/lib/client/src/index.js
@@ -315,41 +315,41 @@ function createWindow() {
         webPreferences       : webPreferences,
         width                : windowCfg.width,
         height               : windowCfg.height,
-        frame                : false,
+        frame                : true,
         devTools             : false,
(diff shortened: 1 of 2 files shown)


mwu-tow (changes requested):

src/js/lib/client/src/preload.js
```diff
@@ -34,19 +34,20 @@ window.showAnimation = animate_show(win)
...
+//     if(e.clientY < 36 && e.clientX < 80) {
+//         if(!windowButtonsVisible) {
+//             windowButtonsVisible = true
+//             win.setWindowButtonVisibility(windowButtonsVisible)
+//         }
+//     } else if (windowButtonsVisible) {
+//         windowButtonsVisible = false
+//         win.setWindowButtonVisibility(windowButtonsVisible)
+//     }
+// })
```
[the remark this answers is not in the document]
BinarySoftware: Actually i'd remove it completely, as it works only on macOS where buttons are on top-left side of the window, and it cannot simply work on Windows, it should check for platforms etc. I'd remove that magical piece of Code.

src/js/lib/client/src/preload.js
```diff
@@ -30,26 +30,6 @@ window.showAnimation = animate_show(win)
...
-            win.setWindowButtonVisibility(windowButtonsVisible)
-        }
-    } else if (windowButtonsVisible) {
-        windowButtonsVisible = false
-        win.setWindowButtonVisibility(windowButtonsVisible)
-    }
-})
-
-
-
```
mwu-tow: On non-macOS platforms this should be a no-op.

In case there is no frame, this should still run. Ideally this should be conditional on presence of the frame in the Window.

src/js/lib/client/src/index.js
```diff
@@ -315,34 +315,42 @@ function createWindow() {
...
+    // https://github.com/enso-org/ide/issues/743 as well as
+    // https://github.com/enso-org/ide/issues/484.
+    // If we don't want the default window frame,
+    // we can try to create our own solution in the future.
     if (args.frame) {
-        windowPreferences.frame         = true
-        windowPreferences.titleBarStyle = 'default'
+        windowPreferences.frame         = false
+        windowPreferences.titleBarStyle = 'hiddenInset'
     }
```
mwu-tow: Reverse the flag, use the yargs default feature.
Try the solution suggested https://github.com/yargs/yargs/issues/1590

src/js/lib/client/src/index.js
```diff
@@ -315,34 +315,42 @@ function createWindow() {
...
     if (args.dev) {
         windowPreferences.devTools = true
     }
 
+    // TODO [MM]
+    // Currently enabled to improve usability and dragging of window, fixes
+    // https://github.com/enso-org/ide/issues/743 as well as
+    // https://github.com/enso-org/ide/issues/484.
+    // If we don't want the default window frame,
+    // we can try to create our own solution in the future.
```
mwu-tow: No TODO comment needed here, create an issue instead for completing custom window frame and enabling it as default.

src/js/lib/client/src/preload.js
mwu-tow: Perhaps `args` from index.js can be accesed or window can be queried for options it was created with / frame presence?

mwu-tow (approved):

PR #784, opened 2020-09-09
Fix ability to toggle dev console in app.

BinarySoftware:
### Checklist


mwu-tow (approved):
Approved but please first:
1) if we support ability to toggle dev console, it should be documented somewhere (what is the shortcut and how to use it)
2) It does not work for me on Windows. If this change improves the behavior on other platforms, then fine, but please create an issue to fix this also on Windows.

PR #799, opened 2020-09-21
Documentation Panel in Searcher

farmaazon:
### Pull Request Description
This PR introduces panel next to suggestion list displaying the documentation of currently selected suggestion.

documentation

### Important Notes
Removed the static documentation panel visible on the right.

### Checklist


src/rust/ide/src/view/node_editor.rs
```diff
@@ -1087,3 +1081,26 @@ impl list_view::entry::ModelProvider for SuggestionProvider {
...
+
+fn doc_placeholder_for(suggestion:&controller::searcher::suggestion::Completion) -> String {
+    let title = match suggestion.kind {
+        EntryKind::Atom     => "Atom",
+        EntryKind::Function => "Function",
+        EntryKind::Local    => "Local variable",
+        EntryKind::Method   => "Method",
+    };
+    format!("{} `{}`\n\nNo documentation available", title,suggestion.code_to_insert(None))
+}
```
BinarySoftware, line 1106:
```suggestion
}

```

BinarySoftware (approved):
Great work, could you try to add drop shadow to list view?
```rust
ensogl::define_shape_system! {
        (radius:f32,roundness:f32) {
            let width  : Var<Pixels> = "input_size.x".into();
            let height : Var<Pixels> = "input_size.y".into();
            let radius        = 1.px() * &radius;
            let corner_radius = &radius * &roundness;
            let border_size_f = 16.0;
            let shadow        = Rect((&width,&height)).corners_radius(&corner_radius);
            let shadow_color  = color::LinearGradient::new()
                .add(0.0,color::Rgba::new(0.0,0.0,0.0,0.0).into_linear())
                .add(1.0,color::Rgba::new(0.0,0.0,0.0,0.20).into_linear());
            let shadow_color  = color::SdfSampler::new(shadow_color)
                .max_distance(border_size_f)
                .slope(color::Slope::Exponent(2.0));
            let shadow        = shadow.fill(shadow_color);
            shadow.into()
        }
```

farmaazon:
Pushed alternative version of documentation panel background:
image

wdanilo:
I've got a few visual comments. Consider the folliwing screenshot:
searcher

1. The background of the searcher and doc is not in the same color. I believe this shuld be fixed in this PR to keep the consistency. Of course it should be parametrized - changing one should affect the other (we will be adjusting colors soon, so its important).

2. The spacing in the documentation is bad (red rectangles). It may look OK on the right side of the window, but when displayed here, the spacing should be consistent. So there should be a doc "style" telling - hey, now youre next to searcher, dont use this huge top offset.

3. The slider on the right should not be visible all time - only when scrolling, and it should also look nicer (like in macos, but should look this way on every OS - our GUI should never be OS specific): 
image

4. The growing list highlight seems strange to me. Can we instead make it slide from top? I mean - we can treat the input text as the "currently selected field", so when we press button down, the selection should traverse from top to bottom.

5. When searcher is opened, we dont see doc to the currently written function in the text input - the doc should be there. Basically the idea is that when you start typing "fo", you see doc for "foo", so you select "foo", you close the searcher, and when you re-open it, you should be in the exact same place that you left, so you sohuld again see docs for "foo".

src/rust/ide/view/src/documentation.rs
```diff
@@ -222,36 +259,43 @@ impl View {
...
-                    frp.data_receive_error.emit(Some(e));
+    fn init(self) -> Self {
+        let network       = &self.frp.network;
+        let model         = &self.model;
+        let visualization = &self.visualization_frp;
+        let frp           = &self.frp;
+        frp::extend! { TRACE_ALL network
+
+
+            // === Displaying documentation ===
```
wdanilo, line 277: I know this is not formal, I will add it to guideline, but lets not use double enter if this is first item in the block - just use single, it looks much muhch better

src/rust/ide/view/src/project.rs
```diff
@@ -30,43 +28,37 @@ struct Model {
...
-        else          { self.display_object.add_child(&self.documentation)    }
+        Self{app,logger,display_object,graph_editor,searcher}
     }
 
     fn set_style(&self, is_light:bool) {
         if is_light { self.app.themes.set_enabled(&["dark"])  }
         else        { self.app.themes.set_enabled(&["light"]) }
     }
 
+    fn searcher_left_top_under_node_at(position:Vector2<f32>) -> Vector2<f32> {
```
wdanilo, line 53: This name is magic :D

src/rust/ide/view/src/project.rs
```diff
@@ -196,10 +178,18 @@ impl View {
...
-                }
-            ));
+                })
+            );
+            _eval <- graph.outputs.node_position_set.map2(&graph.outputs.edited_node,
+                f!([searcher_left_top]((node_id,position),edited_node_id) {
+                    if edited_node_id.contains(node_id) {
+                        let new_left_top = Model::searcher_left_top_under_node_at(*position);
+                        searcher_left_top.set_target_value(new_left_top);
+                    }
```
wdanilo, line 190: by reading the code I dont understand what it does. Could you refactor it to well named function instead please?

src/rust/ide/view/src/searcher.rs
```diff
@@ -36,16 +85,37 @@ struct Model {
...
+        display_object.add_child(&documentation);
         display_object.add_child(&list);
-        Self{logger,display_object,list}
+        list.set_position_x(SUGGESTION_LIST_X);
+        documentation.set_position_x(DOCUMENTATION_X);
+        Self{logger,display_object,list,documentation,doc_provider}
+    }
+
+    fn documentation_for_entry(&self, id:Option<list_view::entry::Id>) -> String {
+        if let Some(id) = id {
```
wdanilo, line 108: I believe that using `map` + `unwrap_or_default` will b much more clean here

src/rust/ensogl/lib/list-view/src/component.rs
```diff
@@ -61,12 +64,26 @@ mod background {
...
-            let height        = sprite_height - PADDING_PX.px() * 2.0;
+            let width         = sprite_width.clone()  - SHADOW_PX.px() * 2.0 - PADDING_PX.px() * 2.0;
+            let height        = sprite_height.clone() - SHADOW_PX.px() * 2.0 - PADDING_PX.px() * 2.0;
             let color         = style.get_color(ensogl_theme::vars::widget::list_view::background::color);
             let rect          = Rect((&width,&height)).corners_radius(CORNER_RADIUS_PX.px());
             let shape         = rect.fill(color::Rgba::from(color));
-            shape.into()
+
+            let border_size_f = 16.0;
+            let corner_radius = CORNER_RADIUS_PX.px() * 1.75;
```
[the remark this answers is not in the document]
farmaazon: I used the code from visualization container shape. However I see now that node.rs has better code, but still it doesn't use theme for colors and exponent.

src/rust/ide/src/view/node_editor.rs
```diff
@@ -1087,3 +1081,26 @@ impl list_view::entry::ModelProvider for SuggestionProvider {
         }
     }
 }
+
+impl ide_view::searcher::DocumentationProvider for controller::searcher::suggestion::List {
+    fn get_for_entry(&self, id:usize) -> Option<String> {
+        use controller::searcher::suggestion::Suggestion;
+        iprintln!("Getting documentation for {id}");
```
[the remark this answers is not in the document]
farmaazon: Old debugging print, I will remove.

farmaazon:
> 1. The background of the searcher and doc is not in the same color. I believe this shuld be fixed in this PR to keep the consistency. Of course it should be parametrized - changing one should affect the other (we will be adjusting colors soon, so its important).

have you read my comment?

> 2. The spacing in the documentation is bad (red rectangles). It may look OK on the right side of the window, but when displayed here, the spacing should be consistent. So there should be a doc "style" telling - hey, now youre next to searcher, dont use this huge top offset.

This is about spacing in docs, but what about the second red rectangle?

> 4. The growing list highlight seems strange to me. Can we instead make it slide from top? I mean - we can treat the input text as the "currently selected field", so when we press button down, the selection should traverse from top to bottom.

It can grow from up. Slide will also look strangely, because we don't have cropping.

> 5. When searcher is opened, we dont see doc to the currently written function in the text input - the doc should be there. Basically the idea is that when you start typing "fo", you see doc for "foo", so you select "foo", you close the searcher, and when you re-open it, you should be in the exact same place that you left, so you sohuld again see docs for "foo".

This was not expressed in any task requirement, and actually it is a bit of work, thus I would move it to a separate ask (or to the part2). 

> the idea is that when you start typing "fo", you see doc for "foo", so you select "foo"

So this is the same as "as user types, the first suggestion should be selected"?

src/rust/ensogl/lib/list-view/src/component.rs
```diff
@@ -61,12 +64,26 @@ mod background {
...
+            let width         = sprite_width  - PADDING_PX.px() * 2.0;
+            let height        = sprite_height - PADDING_PX.px() * 2.0;
+            let shadow        = Rect((&width,&height)).corners_radius(corner_radius);
+            let shadow_color  = color::LinearGradient::new()
+                .add(0.0,color::Rgba::new(0.0,0.0,0.0,0.0).into_linear())
+                .add(1.0,color::Rgba::new(0.0,0.0,0.0,0.20).into_linear());
```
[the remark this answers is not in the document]
farmaazon: But do we consider any other color than transparent in the first line?

src/rust/ide/view/src/searcher.rs
farmaazon: This won't be "or_default".

src/rust/ide/view/src/documentation/style.css
```diff
@@ -319,7 +319,7 @@
 }
 
 .Synopsis {
-    margin-top: 15px;
```
BinarySoftware, line 322: This works well with `pure` documentation, but will break `AST` doc, as far as i remember there is doc title which  then can have margin-bottom: 15px

src/rust/ensogl/lib/theme/src/lib.rs
```diff
@@ -183,6 +183,9 @@ define_theme! { dark
             }
             text {
                 color = color::Lcha::new(1.0,0.0,0.0,0.7);
+                highlight {
```
BinarySoftware, line 186: IDK but i feel like we should also add shadow color to dark-mode

wdanilo (approved):

PR #810, opened 2020-09-25
Scatterplot visualization

BinarySoftware:
### Pull Request Description
Creates scatterplot visualization according to task requirements : 
https://app.zenhub.com/workspaces/enso-ide-5e99c2847b8a3027b334f1b9/issues/enso-org/ide/797

Zrzut ekranu 2020-10-22 o 20 04 03

and dark mode:
Zrzut ekranu 2020-10-23 o 19 46 41

Example code to generate plot given above:
```ruby
    scatterplotTest = """
                    {"axis" : {"x" : { "label" : "x-axis label", "scale" : "linear" }, "y" : { "label" : "y-axis label", "scale" : "linear" }},
                    "focus" : { "x" : 1.7, "y" : 2.1, "zoom" : 3.0 },
                    "points" : {"labels" : "visible"},
                    "data" : [
                                { "x" : 0.1, "y" : 0.7, "label" : "foo", "color" : "ff0000", "shape" : "circle", "size" : 5.0 },
                                { "x" : 0.1, "y" : 0.7, "label" : "foo", "color" : "ffff00", "shape" : "circle", "size" : 2.0 },
                                { "x" : 0.2, "y" : 0.6, "label" : "bar", "color" : "00ff00", "shape" : "square", "size" : 0.5 },
                                { "x" : 0.3, "y" : 0.4, "label" : "baz", "color" : "0000ff", "shape" : "square", "size" : 0.1 },
                                { "x" : 0.5, "y" : 0.4, "label" : "bao", "color" : "f0000f", "shape" : "triangle", "size" : 1.0 }
                    ]}
```

### Checklist
Please include the following checklist in your PR:

- [x] The documentation has been updated if necessary.
- [x] All code conforms to the [Rust](https://github.com/luna/enso/blob/main/docs/style-guide/rust.md).
- [x] All code has been tested where possible.
- [ ] All code has been profiled where possible.

PR #811, opened 2020-09-26
Histogram visualization

BinarySoftware:
### Pull Request Description
Adds a d3.js based histogram visualization.
This PR also refactors common parts like loadScript to a single file, as it is used in many visualizations.

Test string : 
```ruby
    histogramTest = """
                    {"axis" : {"x" : { "label" : "x-axis label" }, "y" : { "label" : "y-axis label" }},
                    "data" : [
                        { "x" : 0.1 },{ "x" : 0.1 },{ "x" : 0.1 },{ "x" : 0.1 },{ "x" : 0.1 },{ "x" : 0.2 },{ "x" : 0.3 },{ "x" : 0.3 },{ "x" : 0.3 }
                    ]}
```

Zrzut ekranu 2020-09-26 o 11 19 53

### Checklist


wdanilo (changes requested):
Amazing <3

Notes to be fixed:
1. Please apply here all related comments from here: https://github.com/enso-org/ide/pull/810
2. Please implement all missing things from the task description (it is updated).

MichaelMauderer:
@wdanilo When doing the review, can you re-check the original spec? This is still the one that has been taken from scatter plot without any changes or updates. Some things in there do not make sense for a histogram (or are inconsistent like the data format which differs between the two examples).

MichaelMauderer:
@wdanilo One follow-up question: do we have a good way to test data updates to a visualization?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
+loadScript('https://d3js.org/d3.v4.min.js');
```
wdanilo, line 1: Please add file doc here.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+loadStyle('https://fontlibrary.org/face/dejavu-sans-mono')
```
wdanilo, line 2: When these styles / scripts are being loaded? The thing is that they should be loaded ON FIRST USAGE of the histogram, not earlier, as our GUI startup time is huge already. I believe that now they are loaded when gui starts, as we are parsing the JS file and we are caching the class definition, right? If I'm right, we need to fix it probably inside of constructor of the visualization. I would not think of any generic mechanism for it for now, let's see other visualizarions and then lets generalize it.

*EDIT*

In the future, I would allow for ES6 import syntax here and I would handle it in a special way (import on first usage);

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+   "color" :  "rgb(1.0,0.0,0.0)" },
+   "bins" : 10,
+   "data" : [
+     "values" : [0.1, 0.2, 0.1, 0.15, 0.7],
+   ]
+ }
+ */
+class Histogram extends Visualization {
+    static inputType = "Any"
+    static label     = "Histogram (JS)"
```
wdanilo, line 39: The labels should not really have info what language this is written in - users dont care aobut it

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+     * Update the internal data and plot settings with the ones from the new incoming data.
+     * If no new settings/data have been provided the old ones will be kept.
+     */
+    updateData(data, isUpdate) {
+        if (isUpdate) {
+            this.axisSpec    = data.axis         || this.axisSpec;
+            this.focus       = data.focus        || this.focus;
+            this.dataValues  = data.data.values  || this.data;
+            this.bins        = data.bins         || this.bins;
+        } else {
```
wdanilo, line 63: too big spacing before `||`

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+    /**
+     * Update the internal data and plot settings with the ones from the new incoming data.
+     * If no new settings/data have been provided the old ones will be kept.
+     */
+    updateData(data, isUpdate) {
+        if (isUpdate) {
+            this.axisSpec    = data.axis         || this.axisSpec;
+            this.focus       = data.focus        || this.focus;
+            this.dataValues  = data.data.values  || this.data;
+            this.bins        = data.bins         || this.bins;
```
wdanilo, line 62: I'm not sure eveything works here as intended. For exammple, if `data.bins` is `0` it will be discarded. Remember that in JS `0 || 5` = `5` ;)

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+            this.bins        = data.bins;
+        }
+    }
+
+    /**
+     * Return the focus area of the histogram. If none is set, returns undefined.
+     */
+    getFocus() {
+        return this.focus
+    }
```
wdanilo, line 76: We are trying to not prefix getters with `get` prefix. Moreover, remember that JS has getters / setters : https://www.w3schools.com/js/js_object_accessors.asp . I'm not telling to use them here, I prefer functions as well, but at least drop the get prefix pls

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+        const margin = this.getMargins(this.getAxisSpec());
+        return {
+            inner : {
+                width  : width - margin.left - margin.right,
+                height : height - margin.top - margin.bottom
+            },
+            outer : {
+                width  : width,
+                height : height,
+            },
```
wdanilo, line 116: in JS, this is the samme as `outer : {width,height}`

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+        };
+    }
+
+    /**
+     * Update the histogram with the current data and settings.
+     */
+    updateHistogram() {
+        const canvas  = this.getCanvasDimensions();
+        const divElem = this.createDivElem(canvas.outer.width,canvas.outer.height);
+        this.dom.appendChild(divElem);
```
wdanilo, line 127: Why during the update function we are modyfying the DOM? Sounds like a serious performance issue for me, but I might be wrong. Could you tell more about it ?

**EDIT** Oh, yo uare re-drawing the histogram from scratch here, right? If so, where is the function shich drops the old div appended to dom?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+            .attr("transform","translate(" + canvas.margin.left + "," + canvas.margin.top + ")");
+
+        const extremesAndDeltas = this.getExtremesAndDeltas(this.getData());
+        this.createLabels(this.getAxisSpec(),svg,canvas);
+        const [bars,scale,axis] = this.createHistogram(extremesAndDeltas,canvas,svg,this.getData(),this.getFocus());
+        const zoom = this.initPanAndZoom(canvas,svg,bars,scale,axis);
+
+        // TODO [MM]: In task specification buttons were on top of the visualization, but because
+        //            the visualization selector obfuscated them, they're now on the bottom.
+        //            This should be fixed in (#898).
```
wdanilo, line 143: <3

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+        this.createButtonFitAll(scale,axis,svg,bars,extremesAndDeltas,canvas,zoom);
```
wdanilo, line 144: Just a fast question - I really, really rarely see in OOP functions with so many arguments. Are you sure we dont want to store some of them as class fields?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+    /**
+     * Helper function for rescaling the data points with a new scale.
+     */
+    rescale(scale, axis, canvas, bars, with_animation) {
+        const animation_duration = (with_animation ? ANIMATION_DURATION : 0.0);
+        axis.x.transition().duration(animation_duration)
+            .call(d3.axisBottom(scale.x).ticks(canvas.inner.width / X_AXIS_LABEL_WIDTH));
+        bars
+            .transition().duration(animation_duration)
+            .attr('transform',d => "translate(" + scale.x(d.x0) + "," + scale.y(d.length) + ")scale(" + scale.zoom + ",1)")
```
wdanilo, line 272: loong lines! Please make all lines <100 chars

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+            if (focus.x !== undefined && focus.zoom !== undefined) {
+                let padding_x = extremesAndDeltas.dx * (1 / (2 * (focus.zoom)));
+                domain_x = [focus.x - padding_x,focus.x + padding_x];
+            }
+        }
+
+        let x = d3.scaleLinear()
+            .domain(domain_x)
+            .range([0,canvas.inner.width]);
+        let xAxis = svg.append("g")
```
wdanilo, line 292: You are sometimes adding nlns between lets, sometimes not. Make it consistent please. (Im ok with both)

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+            .attr("transform","translate(0," + canvas.inner.height + ")")
+            .call(d3.axisBottom(x));
+
+        let histogram = d3.histogram()
+            .value(d => d)
+            .domain(x.domain())
+            .thresholds(x.ticks(this.getBinCount())); // FIXME
```
wdanilo, line 299: whats this FIXME about ?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+
+        dataPoints.forEach(value => {
+            if (value < xMin) { xMin = value }
+            if (value > xMax) { xMax = value }
+        });
+
+        let dx        = xMax - xMin;
+        let padding_x = 0.1 * dx;
+
+        return {xMin:xMin,xMax:xMax,paddingX:padding_x,dx:dx};
```
wdanilo, line 396: you dont need this long form in JS, just enumerate fields, like `{xMin,xax,paddinxX,dx}` :)

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    /**
+     * Return margins for plots drawing area.
+     */
+    getMargins(axis) {
+        if (axis.x.label === undefined && axis.y.label === undefined) {
```
wdanilo, line 403: VERY mminor thing. Just create `let noXAxis = axis.x.label === undefined` and use it in ifs - it will be easier to read

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,533 @@
...
+        } else if (axis.x.label === undefined) {
+            return {top:20,right:10,bottom:20,left:45};
+        }
+        return {top:10,right:10,bottom:35,left:45};
+    }
+
+    /**
+     * Creates HTML div element as container for plot.
+     */
+    createDivElem(width,height) {
```
wdanilo, line 416: This function name is semantically incorrect. This function is not about Div creation, right? It does something more than that. So the name should reflect what it really does.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/loadingScripts.js
```diff
@@ -0,0 +1,30 @@
+/**
```
wdanilo, line 1: The name of the file is a little bit off to me. Every file in JS contains scripts, so I would drop the suffix "Scripts" here as well

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/loadingScripts.js
```diff
@@ -0,0 +1,30 @@
...
+    document.head.appendChild(script);
+}
+
+/**
+ * Helper function to load styles.
+ *
+ * It runs only once because of the visualization implementation : file is loaded once, then
+ * the onDataReceived() is called multiple times, so it won't load the same style on this or any
+ * other visualization show/hide.
+ */
```
wdanilo, line 21: I don't think this doc is correct here. Nothing in implementation of this function prevents multiple style loading, so this doc can bring confusion and false thinking that we will be safe to call it multiple times with the same style source. The same applies to doc comment above

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/scatterPlot.js
```diff
@@ -1,34 +1,3 @@
...
-    style.href  = url;
-    style.rel   = "stylesheet";
-    style.media = "screen";
-    style.type  = "text/css";
-
-    document.head.appendChild(style);
-}
-
 loadScript('https://d3js.org/d3.v4.min.js');
 loadStyle('https://fontlibrary.org/face/dejavu-sans-mono')
```
wdanilo, line 2: Please apply the same performance things here as well - load it on first usage

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
BinarySoftware: we have already ok() function for it in Scatterplot, ready to be copied ;)

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
MichaelMauderer: About removing the `FIXME`.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
MichaelMauderer: You are right. I did not completely refactor this before, but it is changed now.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
MichaelMauderer: It actually seems to be a bit more tricky because if we want to use it in the constructor, we need for the script to be loaded via `onload` on the script, which is probably the more correct way to do this anyway. But that requires us to start making things async/callback based, which makes this a bit more difficult. I would suggest moving this to another issue because it might require some more significant changes to all visualizations. (Unless there is an easier way that I don't know about)

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script.rs
```diff
@@ -20,14 +20,27 @@ pub fn table_view_visualization() -> visualization::java_script::FallibleDefinit
...
+    let source          = format!("{}{}",loading_scripts,source);
 
     visualization::java_script::Definition::new(data::builtin_library(),source)
 }
 
 /// Return a `JavaScript` Map visualization.
 pub fn map_view_visualization() -> visualization::java_script::FallibleDefinition {
-    let source = include_str!("java_script/mapView.js");
+    let loading_scripts = include_str!("java_script/loading.js");
+    let source          = include_str!("java_script/mapView.js");
```
wdanilo, line 42: I know its not part of this commit, but I dont like the name `mapView`. It should be `geoMap` instead. Every visualization is a "view", but we don't use the name "histogramView", nor other views.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
+/** Histogram Visualization. */
+loadScript("https://d3js.org/d3.v4.min.js");
+loadStyle("https://fontlibrary.org/face/dejavu-sans-mono");
```
wdanilo, line 3: please load the things only when visualization is used! Not during the start of GUI, as this slows down things drastically.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
...
+const X_AXIS_LABEL_WIDTH = 30;
+const ANIMATION_DURATION = 1000;
+const LINEAR_SCALE = "linear";
+const LIGHT_PLOT_COLOR = "#00E890";
+const DARK_PLOT_COLOR = "#E0A63B";
+const DEFAULT_NUMBER_OF_BINS = 10;
+const BUTTON_HEIGHT = 25;
+
+// Indicates whether script dependencies have been added to the DOM.
+let scriptsInitialised = false;
```
wdanilo, line 20: This variable is not used anywhere

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
...
+};
+
+const LABEL_STYLE = "font-family: DejaVuSansMonoBook; font-size: 10px;";
+const X_AXIS_LABEL_WIDTH = 30;
+const ANIMATION_DURATION = 1000;
+const LINEAR_SCALE = "linear";
+const LIGHT_PLOT_COLOR = "#00E890";
+const DARK_PLOT_COLOR = "#E0A63B";
+const DEFAULT_NUMBER_OF_BINS = 10;
+const BUTTON_HEIGHT = 25;
```
wdanilo, line 17: Some of these variables are not used anywhere. Please, check them and other functions / variables in this file. Moreover, please align all `=` here

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
...
+    if (isUpdate) {
+      this._axisSpec = ok(data.axis) ? data.axis : this._axisSpec;
+      this._focus = ok(data.focus) ? data.focus : this._focus;
+      this._dataValues = ok(data.data.values) ? data.data.values : this.data;
+      this._bins = ok(data.bins) ? data.bins : this._bins;
+    } else {
+      this._axisSpec = data.axis;
+      this._focus = data.focus;
+      this._dataValues = data.data.values;
+      this._bins = data.bins;
```
wdanilo, line 79: are you using some kind of formatter here? I was asking to use prettier during osme previosu comments with disabled semicolon option

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
...
+          "," +
+          scale.y(d.length) +
+          ")scale(" +
+          scale.zoom +
+          ",1)"
+      );
+  }
+
+  /**
+   * Update the d3 histogram with the current data.
```
wdanilo, line 353: Write few momre words how the update is performed pls. Is all graphical data thrown away and re-drawn, or is it done differently?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
...
+
+    if (noXAxis && noYAxis) {
+      return { top: 20, right: 20, bottom: 25, left: 35 };
+    } else if (noYAxis) {
+      return { top: 10, right: 20, bottom: 35, left: 35 };
+    } else if (noXAxis) {
+      return { top: 20, right: 10, bottom: 25, left: 55 };
+    }
+    return { top: 10, right: 10, bottom: 35, left: 55 };
+  }
```
wdanilo, line 530: a lot of hardcoded variables

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
...
+    const divElem = document.createElementNS(null, "div");
+    divElem.setAttributeNS(null, "class", "vis-histogram");
+    divElem.setAttributeNS(
+      null,
+      "viewBox",
+      0 + " " + 0 + " " + width + " " + height
+    );
+    divElem.setAttributeNS(null, "width", "100%");
+    divElem.setAttributeNS(null, "height", "100%");
+    divElem.setAttributeNS(null, "transform", "matrix(1 0 0 -1 0 0)");
```
wdanilo, line 545: magical mmatrix hardcoded. Its purpose would probably not be obvious to the reader

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,683 @@
...
+    const darkStrokeColor = `rgba(255,255,255,0.7)`;
+    const buttonLightColor = `#333`;
+    const darkBtnHoverColor = `rgba(255,255,255,0.5)`;
+    const darkSelectionFill = `#efefef`;
+
+    addStyleToElem(".selection", "rx: 4px;stroke: transparent;");
+    addStyleToElem(
+      "button",
+      `
+            margin-left: 5px;
```
wdanilo, line 563: You are using 2-space indentation everywhere, but here we've got 6-space one. Moreover, the closing quote is misaligned. In the next text as well

wdanilo (changes requested):

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
wdanilo: As we discussed, let's do it after this PR as the startup time in GUI is very important and right now its already ugly. Please discuss it with Sylwia as well.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
MichaelMauderer: This doesn't seem to work with prettier. So, I'm leaving this for the moment.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
MichaelMauderer: Extended the description. This has been reworked to no longer re-do the whole vis, but to do a proper update,

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
MichaelMauderer: Issue added: https://github.com/enso-org/ide/issues/985

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script.rs
MichaelMauderer: I'll do this as part of the map-initialisation PR.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/bubbleVisualization.js
```diff
@@ -1,38 +1,42 @@
 class BubbleVisualization extends Visualization {
```
BinarySoftware, line 1: Is this visualization really needed? Wasn't it just some kind of a nice demo 4 months ago using dummy code? Should it exist now? The same about ensogl-one.

docs/CONTRIBUTING.md
```diff
@@ -97,13 +97,18 @@ need the following setup:
...
 Please read the following documents to learn more about reasons behind this
 decision and the recommended code style guide. Be sure to carefully read the
 [Rust style guide 1](./contributing/style-guide.md)
 and the [Rust style guide 2](https://dev.enso.org/docs/style-guide/rust.html)
 before contributing to the codebase.
 
+We do, however, use [`prettier`](https://prettier.io/) for the JavaScript files in our code base. 
+If you have not installed it already you can do so via `npm install prettier`. To use it manually 
+via command line run `prettier --write` to all JavaScript files in the project. Alternatively, 
+there are plugins for many IDEs available to do this for you.
```
wdanilo, line 110: FYI, in my PR I've added `./run fmt` which auto formats all cargo toml files. We can extend it with auto formatting all js files too!

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/bubbleVisualization.js
wdanilo: No, these are perfect examples to users. They should not be enabled in GUI though.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,711 @@
...
+
+    /**
+     * Return vales to plot.
+     */
+    data() {
+        return this._dataValues || {}
+    }
+
+    /**
+     * Return the number of bins to use for the histogram..
```
wdanilo, line 112: double dot.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,711 @@
...
+     */
+    binCount() {
+        return this._bins || DEFAULT_NUMBER_OF_BINS
+    }
```
wdanilo, line 116: are you sure that if user sets 0 bins we should defautl to the default number of bins ?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -0,0 +1,711 @@
...
+            if (shortcuts.zoomIn(event)) {
+                zoomIn()
+                endBrushing()
+            }
+        }
+
+        /**
+         * Updates plot when brushing.
+         */
+        function updateChart() {
```
wdanilo, line 308: I believe that function name should contain something about brushing if these docs are valid

wdanilo (approved):

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/bubbleVisualization.js
BinarySoftware: So this vis can simply not be registered, and still stay in this place.

PR #814, opened 2020-09-30
Handling missing main method on startup and `here.main` syntax

mwu-tow:
### Pull Request Description
Currently IDE fails to load if `main` method definition in not present or if it uses the `here.main` syntax.
This PR fixes both:
1) method pointer lookup will correctly take into consideration the module name to resolve the `here` symbol (related engine issue: https://github.com/enso-org/enso/issues/1138 ) 
2) if the `main` method cannot be found during the startup, it will be created

Currently I am marking this PR as a draft, because there's one more missing piece — the standard `main` definition stub requires also explicit `Base` import.
To be discussed how to handle this.

### Checklist

### DIFF
--- a/src/rust/ide/src/double_representation/module.rs
+++ b/src/rust/ide/src/double_representation/module.rs
@@ -734,7 +734,7 @@ mod tests {
         info.expect_code("import Bar.Gar");
     }
 
-    #[test]
+    #[wasm_bindgen_test]
     fn implicit_method_resolution() {
         let parser         = parser::Parser::new_or_panic();
         let module_name    = ReferentName::new("Main").unwrap();
--- a/src/rust/ide/src/double_representation/module.rs
+++ b/src/rust/ide/src/double_representation/module.rs
@@ -2,11 +2,12 @@
 
 use crate::prelude::*;
 
+use crate::constants::keywords::HERE;
 use crate::double_representation::definition;
-use crate::double_representation::definition::DefinitionName;
 use crate::double_representation::definition::DefinitionProvider;
 use crate::double_representation::identifier;
 use crate::double_representation::identifier::Identifier;
+use crate::double_representation::identifier::NormalizedName;
 use crate::double_representation::identifier::LocatedName;
 use crate::double_representation::identifier::ReferentName;
 
@@ -587,7 +588,7 @@ pub fn locate
     // Not exactly regular - we need special case for the first crumb as it is not a definition nor
     // a children. After this we can go just from one definition to another.
     let first_crumb = crumbs_iter.next().ok_or(EmptyDefinitionId)?;
-    let mut child = ast.def_iter().find_by_name(&first_crumb)?;
+    let mut child   = ast.def_iter().find_by_name(&first_crumb)?;
     for crumb in crumbs_iter {
         child = definition::resolve_single_name(child,crumb)?;
     }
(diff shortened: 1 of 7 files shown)


src/rust/ide/src/double_representation/module.rs
```diff
@@ -598,18 +608,27 @@ pub fn locate
...
 /// desugaring implicit extensions methods for modules).
+///
+/// The `name` parameter is the name of the module that contains `ast`. It affects how the `here`
+/// keyword is resolved.
 pub fn lookup_method
-(ast:&known::Module, method:&language_server::MethodPointer) -> FallibleResult<definition::Id> {
-    let module_name                    = QualifiedName::try_from(method)?;
-    let explicitly_extends_looked_type = method.defined_on_type == module_name.name().as_ref();
-
+(name:&ReferentName, ast:&known::Module, method:&language_server::MethodPointer)
```
farmaazon, line 615: I would rename `name` parameter to `module_name`

src/rust/ide/src/double_representation/module.rs
```diff
@@ -598,18 +608,27 @@ pub fn locate
 ///
 /// The module is assumed to be in the file identified by the `method.file` (for the purpose of
 /// desugaring implicit extensions methods for modules).
+///
+/// The `name` parameter is the name of the module that contains `ast`. It affects how the `here`
+/// keyword is resolved.
 pub fn lookup_method
```
farmaazon, line 614: At this point when I read this module I wonder what decides when a function is a method of `Info` and when it isn't? Perhaps most of them could go inside Info, and even Info could easily keep module name next to ast?

That only under your consideration, it is not blocker for merging PR.

src/rust/ide/src/view/project.rs
```diff
@@ -91,20 +110,44 @@ pub async fn recreate_if_missing(project:&model::Project, path:&FilePath, defaul
...
+pub fn add_main_if_missing
+(module:&model::Module, main_ptr:&MethodPointer, parser:&Parser) -> FallibleResult<()> {
+    if module.lookup_method(main_ptr).is_err() {
+        let mut info  = module.info();
+        let main_code = default_main_method_code();
+        let main_ast  = parser.parse_line(main_code)?;
+        info.add_ast(main_ast,double_representation::module::Placement::End)?;
+        module.update_ast(info.ast);
+
+        // TODO [mwu] Add imports required by the default main definition (when needed).
```
farmaazon, line 125: Let's add it right now, or put here a reference to issue with managing new import syntax.

src/rust/ide/src/view/project.rs
```diff
@@ -91,20 +110,44 @@ pub async fn recreate_if_missing(project:&model::Project, path:&FilePath, defaul
...
+    }
+    Ok(())
+}
+
 impl ProjectView {
     /// Create a new ProjectView.
-    pub async fn new(logger:impl AnyLogger, model:model::Project)
-    -> FallibleResult<Self> {
+    pub async fn new(logger:impl AnyLogger, project:model::Project)
+                     -> FallibleResult<Self> {
```
farmaazon, line 133:
```suggestion
    -> FallibleResult<Self> {
```

farmaazon (approved):

src/rust/ide/src/double_representation/module.rs
mwu-tow: Leaving as-is, should be refactored in the future.

Functions being free and not in `Info` is mostly consequence of `Info` being introduced later than them. We don't want `Info` to contain name, as most of the methods don't need it.  
Another issue is the ergonomy of usage — having to wrap Ast into info every single time is not convenient. Some places also don't know name but want to process Ast.

The lookup function however should not really be a method should not be a method of plain info but something that contains both ast and name.  If other functions working on such name+ast pair appear, we should consider introducing it as a new structure (NamedInfo or sth similar). Info should not be structurally changed.

src/rust/ide/src/double_representation/module.rs
mwu-tow: done

src/rust/ide/src/view/project.rs
mwu-tow: actually this can fit in a single line

mwu-tow:
Had a call about imports issue. Decided to change the contents of the default generated main to just contain `"Hello world!" node with no dependency on the standard library, so the import won't be needed.

PR #834, opened 2020-10-06
Fast-action icons for nodes

MichaelMauderer:
### Pull Request Description
Implements #538 

Peek 2020-10-07 17-05
Peek 2020-10-07 16-56

### Important Notes
Issue #538 describes 4 icons, but only three actions. All icons are implemented, but only 3 are visible at the moment.
Exact colors and visuals are subject to discussion (can easily be configured).

### Checklist

Please include the following checklist in your PR:

- [x] The documentation has been updated if necessary.
- [x] All code conforms to the [Rust](https://github.com/luna/enso/blob/main/docs/style-guide/rust.md) style guide.
- [x] All code has been tested where possible.
- [x] All code has been profiled where possible.

PR #851, opened 2020-10-14
Refactor animations to better allow animations of color and other components.

MichaelMauderer:
### Pull Request Description
Changes the Animation interface to provide an abstraction that allows components to define an  "animation space vector". This vector needs to work like a cartesian coordinate and n-dimensional space and provide a way to be constructed from the given component (and construct a component from a given a nimation space vector.

For basic datatypes like `f64` `Vector4` the value itself is identical to the animation space vector, but e.g., for `color::Rgba` we want animations to happen in an aproriate color space and thus the animation space vector is a `Vector4` in the `color::Laba` space.

### Important Notes
Precondition for merging  #834 

### Checklist

### DIFF
--- a/src/rust/ensogl/lib/core/src/data/color.rs
+++ b/src/rust/ensogl/lib/core/src/data/color.rs
@@ -19,6 +19,7 @@ pub mod component;
 pub mod data;
 pub mod gradient;
 pub mod space;
+pub mod animations;
 
 pub use component::*;
 pub use self::data::*;
--- a/src/rust/ensogl/lib/core/src/data/color.rs
+++ b/src/rust/ensogl/lib/core/src/data/color.rs
@@ -19,7 +19,7 @@ pub mod component;
 pub mod data;
 pub mod gradient;
 pub mod space;
-pub mod animations;
+pub mod animation;
 
 pub use component::*;
 pub use self::data::*;
--- a/src/rust/ensogl/lib/core/src/gui/component.rs
+++ b/src/rust/ensogl/lib/core/src/gui/component.rs
@@ -291,10 +291,6 @@ impl<T:Animatable+Debug+Default+Clone+'static> Animation<T> {
         self.simulator.set_target_value(state.value);
     }
 
-    pub fn skip(&self) {
-        self.simulator.skip();
-    }
-
     pub fn target_value(&self) -> T {
         let value =  self.simulator.target_value();
         AnimationLinearSpace{value}.into()
--- a/src/rust/ensogl/lib/core/src/gui/component.rs
+++ b/src/rust/ensogl/lib/core/src/gui/component.rs
@@ -213,15 +213,6 @@ pub struct AnimationLinearSpace<T> {
 /// Shorthand for a  HasAnimationSpaceRepr::AnimationSpaceRepr wrapped in a `AnimationLinearSpace`.
 pub type AnimationSpaceRepr<T> = AnimationLinearSpace<<T as HasAnimationSpaceRepr>::AnimationSpaceRepr>;
 
-/// This is used to make type inference better on usage places - without it,
-/// Rust would force one to write the where part EVERYWHERE in use places.
-/// FIXME[MM] replace with version from prelude.
-pub trait FromAsInto<T> = Sized where T:Into<Self>;
-
-/// Can be transformed from and into.
-/// FIXME[MM] replace with version from prelude.
-pub trait BiInto<T> = Sized + Into<T> + FromAsInto<T>;
-
 pub trait Animatable = HasAnimationSpaceRepr + BiInto<AnimationSpaceRepr<Self>>;
 
 /// Convert the animation space value to the respective `Animatable`.
--- a/src/rust/ensogl/lib/core/src/gui/component.rs
+++ b/src/rust/ensogl/lib/core/src/gui/component.rs
@@ -290,7 +290,7 @@ impl<T:Animatable+frp::Data> Animation<T> {
     }
 
     pub fn target_value(&self) -> T {
-        let value =  self.simulator.target_value();
+        let value = self.simulator.target_value();
         from_animation_space(value)
     }
 }
--- a/src/rust/ensogl/lib/core/src/gui/component.rs
+++ b/src/rust/ensogl/lib/core/src/gui/component.rs
@@ -196,25 +196,31 @@ impl<T:Shape> display::Object for ShapeView<T> {
 // === Animatable ===
 // ==================
 
+/// Indicate what datatype to use in the animation space representation.
 pub trait HasAnimationSpaceRepr {
+    /// Representation in animation space. Needs to support linear interpolation and all
+    /// pre-requisites of `inertia::Value`.
     type AnimationSpaceRepr: inertia::Value;
 }
 
-pub struct AnimationLinearSpace<T>{
-    pub(crate) value: T
+/// Newtype that indicates that the wrapped value is valid to be used in animations.
+#[derive(Debug)]
+pub struct AnimationLinearSpace<T> {
+    /// Wrapped value representing the animation space value.
+    pub value: T
 }
 
+/// Shorthand for a  HasAnimationSpaceRepr::AnimationSpaceRepr wrapped in a `AnimationLinearSpace`.
 pub type AnimationSpaceRepr<T> = AnimationLinearSpace<<T as HasAnimationSpaceRepr>::AnimationSpaceRepr>;
 
-// This is used to make type inference better on usage places - without it,
-// Rust would force you to write the where part EVERYWHERE in use places.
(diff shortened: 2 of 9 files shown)


src/rust/ensogl/lib/core/src/gui/component.rs
```diff
@@ -189,31 +189,91 @@ impl<T> display::Object for ShapeView<T> {
...
 /// Smart animation handler. Contains of dynamic simulation and frp endpoint. Whenever a new value
 /// is computed, it is emitted via the endpoint.
-#[derive(CloneRef,Derivative,Debug,Shrinkwrap)]
+#[derive(CloneRef,Derivative,Debug)]
 #[derivative(Clone(bound=""))]
 #[allow(missing_docs)]
-pub struct Animation<T> {
-    #[shrinkwrap(main_field)]
-    pub simulator : DynSimulator<T>,
+pub struct Animation<T:Animatable<T>> {
```
wdanilo, line 239: Oh, this looks bad! `T:Animatable<T>`. You should never have traits which refer to itself this way. You defintely could have written it as `T:Animatable` instead - see my impl for reference

src/rust/ensogl/lib/core/src/gui/component.rs
```diff
@@ -189,31 +189,91 @@ impl<T> display::Object for ShapeView<T> {
...
 // =================
 
 /// Smart animation handler. Contains of dynamic simulation and frp endpoint. Whenever a new value
 /// is computed, it is emitted via the endpoint.
-#[derive(CloneRef,Derivative,Debug,Shrinkwrap)]
+#[derive(CloneRef,Derivative,Debug)]
 #[derivative(Clone(bound=""))]
 #[allow(missing_docs)]
-pub struct Animation<T> {
-    #[shrinkwrap(main_field)]
```
wdanilo, line 202: why this change? It makes usage ugly

src/rust/ensogl/lib/core/src/gui/cursor.rs
```diff
@@ -367,8 +367,8 @@ impl Cursor {
         radius.set_target_value(DEFAULT_RADIUS);
         size.set_target_value(DEFAULT_SIZE());
 
-        let fade_out_spring = inactive_fade.spring() * 0.2;
-        let fade_in_spring  = inactive_fade.spring();
+        let fade_out_spring = inactive_fade.simulator.spring() * 0.2;
+        let fade_in_spring  = inactive_fade.simulator.spring();
```
wdanilo, line 371: forcing user to know that "animation" uses "simulator" under the hood is leaking internal design to the api. its bad

wdanilo (changes requested):

src/rust/ensogl/lib/core/src/data/color.rs
```diff
@@ -19,6 +19,7 @@ pub mod component;
 pub mod data;
 pub mod gradient;
 pub mod space;
+pub mod animations;
```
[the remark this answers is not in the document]
wdanilo: Moreover, this is not responsibility of colors to think about animations, so this modul;e placemetn is bad. But read my other comments - they propose a better design which actually suits this place

src/rust/ensogl/lib/core/src/gui/component.rs
```diff
@@ -189,6 +189,61 @@ impl<T> display::Object for ShapeView<T> {
...
+
+/// Shorthand for a  HasAnimationSpaceRepr::AnimationSpaceRepr wrapped in a `AnimationLinearSpace`.
+pub type AnimationSpaceRepr<T> = AnimationLinearSpace<<T as HasAnimationSpaceRepr>::AnimationSpaceRepr>;
+
+/// This is used to make type inference better on usage places - without it,
+/// Rust would force one to write the where part EVERYWHERE in use places.
+pub trait FromAsInto<T> = Sized where T:Into<Self>;
+
+/// Can be transformed from and into.
+pub trait FromInto<T> = Sized + Into<T> + FromAsInto<T>;
```
wdanilo, line 218: Why not `Iso`? This is a standard form of isomorphism and imo this sohuld be named this way. Moreover, why not placing it in prelude like I proposed in my snippet?

src/rust/ensogl/lib/core/src/gui/component.rs
```diff
@@ -198,22 +253,45 @@ impl<T> display::Object for ShapeView<T> {
...
+pub struct Animation<T:Animatable> {
     #[shrinkwrap(main_field)]
-    pub simulator : DynSimulator<T>,
+    pub simulator : DynSimulator<T::AnimationSpaceRepr>,
     pub value     : frp::Stream<T>,
 }
 
-impl<T:inertia::Value> Animation<T> {
+#[allow(missing_docs)]
+impl<T:Animatable+Debug+Default+Clone+'static> Animation<T> {
```
wdanilo, line 263: Are these constraints all needed? I dont get where `'static'`, `Default`, and `Debug` comes from.

src/rust/ensogl/lib/core/src/gui/component.rs
```diff
@@ -198,22 +253,45 @@ impl<T> display::Object for ShapeView<T> {
...
     /// Constructor.
     pub fn new(network:&frp::Network) -> Self {
         frp::extend! { network
             def target = source::<T>();
         }
-        let simulator = DynSimulator::<T>::new(Box::new(f!((t) target.emit(t))));
+        let simulator = DynSimulator::<T::AnimationSpaceRepr>::new(Box::new(f!((t) {
+             target.emit(AnimationLinearSpace{value:t}.into())
```
wdanilo, line 270: Change it pls to `t.to_animation_linear_space()` - this can be a method generated automatically which uses `into` under the hood. Lets apply it to other places below as well :) Manual construction of wrappers everywhere adds a lot of boilerplate

src/rust/ensogl/lib/core/src/gui/component.rs
MichaelMauderer: We don't really define a mathematical isomorphism here. That is [a lot more specific](https://en.wikipedia.org/wiki/Isomorphism). All we do require is a `From<T>` and `Into<T>` implementation. From a Rust perspective, the name `FromInto` is much clearer about the requirements. 

I agree, needs to move to prelude though.

src/rust/ensogl/lib/core/src/gui/component.rs
MichaelMauderer: I think they come from the FRP. I will check that later.

src/rust/ensogl/lib/core/src/gui/component.rs
MichaelMauderer: I added a `from_animation_space` function that does that.

src/rust/ensogl/lib/core/src/gui/component.rs
MichaelMauderer: > I agree, needs to move to prelude though.

Just noticed prelude is in a different crate, though I would leave this here for now. If we feel strongly about the usefulness, we can open an issue to add it there and refactor here after it has landed.

src/rust/ensogl/lib/core/src/gui/component.rs
MichaelMauderer: Yeah, all of the contsraints are needed for the FRP. Otherwise `T` can't be used in `source::<T>();`

src/rust/ensogl/lib/core/src/gui/component.rs
wdanilo: The fact that prelude is in different crate should not prevent us from changing it. This should definitely be in prelude IMO. If youre not sure, lets talk about it. Also, lets talk more aobut the Iso name, as I believe this would be the correct name for that.

src/rust/ensogl/lib/core/src/gui/component.rs
wdanilo: If they come from FRP, please use their FRP type alias so we know they are from FRP! Expanding constraitns like this make it error prone in the future - they will change in frp and we will not know where thay came from. Instead, always try to use aliases prepared for it. So just use `T:frp::Data` instead!

src/rust/ensogl/lib/core/src/gui/component.rs
```diff
@@ -198,22 +251,45 @@ impl<T> display::Object for ShapeView<T> {
...
+        from_animation_space(value)
+    }
+
+    pub fn set_target_value(&self, target_value:T) {
+        let state:AnimationLinearSpace<_> = target_value.into();
+        self.simulator.set_target_value(state.value);
+    }
+
+    pub fn target_value(&self) -> T {
+        let value =  self.simulator.target_value();
```
wdanilo, line 290: double space

wdanilo (approved):

PR #856, opened 2020-10-20
Labels on nodes should reflect function argument names

wdanilo:
### Pull Request Description
- Implements [Labels on nodes should reflect function argument names, not only connected elements.](https://github.com/enso-org/ide/issues/528).
- This is the first implementation of arguments names display only. A much more sophisticated implementation has to be yet implemented, but it requires a lot of design talks, so it was postponed for now. The current implementation just display names of not connected arguments. It does not display names of already connected nor allow for folding expressions.

### Important Notes
- This PR contains also some SpanTree refactoring.
- :warning: This PR contains a LOT of TODO notes - they are meant to be solved before merging it! :warning:

### Checklist


src/rust/ide/lib/span-tree/example/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+                .kind(Kind::argument(true))
+                .crumbs(parens_cr)
+                .new_child(|t|t.size(12).kind(Kind::Operation).crumbs(PrefixCrumb::Func).new_id())
+                .new_child(|t|t.size(1))
+                .new_child(|t|t.size(6).kind(Kind::target(true)).crumbs(PrefixCrumb::Arg).new_id())
+                .new_child(|t|t.size(1))
+            ))
+        .new_child(|t|t.size(1));
+
+    println!("{:#?}",input_span_tree2);
```
farmaazon, line 82: As this does not display anything, I'd rather have it as a test instead of example. In tests you can also print span-tree using println!

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -48,8 +60,45 @@ impl Kind {
...
+        }
+    }
+
+    /// Match the value with `Kind::Empty(ExpectedArgument(_))`.
+    pub fn is_expected_argument(self) -> bool {
+        match self {
+            Self::Empty(InsertType::ExpectedArgument(_)) => true,
+            _                                            => false
+        }
+    }
```
farmaazon, line 87: Some of those function can be superseded by `matches!` macro.

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -58,15 +107,25 @@ pub enum InsertType {
     BeforeTarget,
     AfterTarget,
     Append,
+    // FIXME: When this insert type can be assigned to node without name?
```
farmaazon, line 110: what is "node without name?" you mean empty node (with 0-length span)?

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -58,15 +107,25 @@ pub enum InsertType {
...
     /// notation.
     ExpectedArgument(usize),
 }
 
+impl Default for InsertType {
+    fn default() -> Self {
+        Self::Append
+    }
+}
+
```
farmaazon, line 123: Why do we need this? Why Append is default?

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -92,43 +151,94 @@ pub fn parent_crumbs(crumbs:&[Crumb]) -> Option<&[Crumb]> {
...
+// ============
 // === Node ===
+// ============
 
 /// SpanTree Node.
 ///
 /// Each node in SpanTree is bound to some span of code, and potentially may have corresponding
 /// AST node.
-#[derive(Clone,Debug,Eq,PartialEq)]
+#[derive(Clone,Debug,Default,Eq,PartialEq)]
```
farmaazon, line 164: I don't like making span-tree nodes "Default". The span-tree node does not make any sense outside the SpanTree context, and it is very hard to express what in fact the "default nodes" are.

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -138,8 +248,99 @@ pub struct Child {
...
+    fn deref(&self) -> &Self::Target {
+        &self.node
+    }
+}
+
+impl DerefMut for Child {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.node
+    }
+}
```
farmaazon, line 262: I know we follow this pattern everythere, but I found this recently https://rust-lang.github.io/api-guidelines/predictability.html#only-smart-pointers-implement-deref-and-derefmut-c-deref

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -172,6 +373,7 @@ impl<'a> Ref<'a> {
         enso_data::text::Span::new(self.span_begin,self.node.size)
     }
 
+    // FIXME: These docs are invalid, this function is deeply magical
     /// Get the reference to child with given index. Returns None if index if out of bounds.
     pub fn child(mut self, index:usize) -> FallibleResult<Ref<'a>> {
```
farmaazon, line 378: I don't understand. Have you test which shows that docs are invalid?

The only magic thing is that we return our `Ref` structure instead of plain reference, but Ref structure have more useful information in it (like the crumbs to the current node).

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -48,8 +60,45 @@ impl Kind {
...
+        match self {
+            Self::Empty(InsertType::ExpectedArgument(_)) => true,
+            _                                            => false
+        }
+    }
 }
 
+impl Default for Kind {
+    fn default() -> Self {
+        Self::Empty(default())
```
farmaazon, line 92: `Empty` is not a good default for kind; empty nodes (and their insertion point) have sense only in context of their parent (so we may realize what kind of insertion it is). If default node does not have any parent, it should be "root" instead. But I don't see point in creating sole root node - instead you can just use `SpanTree::default()`

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -92,43 +151,94 @@ pub fn parent_crumbs(crumbs:&[Crumb]) -> Option<&[Crumb]> {
...
-            size           : Size::new(0),
-            children       : Vec::new(),
-            expression_id  : None,
-            parameter_info : None,
-        }
+        let kind           = Kind::Empty(insert_type);
+        let size           = default();
+        let children       = default();
+        let expression_id  = default();
+        let parameter_info = default();
```
farmaazon, line 187: What is the point in replacing obvious values with defaults here?

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -138,8 +248,99 @@ pub struct Child {
...
+// === ChildBuilder ===
+// ====================
+
+/// A builder pattern for `SpanTree`. A think wrapper for `Child` which adds useful methods for
+/// building properties of the current node.
+#[derive(Debug)]
+#[allow(missing_docs)]
+pub struct ChildBuilder {
+    pub child : Child,
+}
```
farmaazon, line 276: We already have "ChildGenerator" in generate.rs, and more appriopriate way for building span-tree in `span-tree::builder::Builder`

farmaazon: Add docs to both builders explaining why we keep them together

src/rust/ide/lib/span-tree/example/src/lib.rs
wdanilo: As agreed via phone, tests should not print a lot to console and Im adding doc comments about the purpose of this builder here

wdanilo: Also, Adam, I added second interface to the builder, so now you can either use magic function which computes automatically things for you, or an explict one, where you need to provide everything manually. Thus I believe this may replace the old builder in the future.

src/rust/ide/lib/span-tree/src/node.rs
wdanilo: As we were chatting via phone, default nodes can be just any node tha t you are adding to spantree while constructing it. A node can default to just insertion point, as this is completely valid default. It simplifies a lot of API. Again, Id love to chat more about it in case it, but in according to rust docs: `Sometimes, you want to fall back to some kind of default value, and don't particularly care what it is.`, which is completely valid here, and simplifies api a lot in different places where we really dont care about this value.

src/rust/ide/lib/span-tree/src/node.rs
wdanilo: Ok, so the function is not as magical as I tohught, but also it doesnt return none. Im fixing docs.

src/rust/ide/lib/span-tree/src/node.rs
wdanilo: As we agreed on the phone, lets leave it, as explained above

src/rust/ide/lib/span-tree/src/node.rs
wdanilo: I added docs. Moreover, I provided more explicit interface to the new builder, so it can be used for testing purposes as well! Lets clean this up when reworking the span tree API! :)

farmaazon (approved):

PR #866, opened 2020-10-22
Fix ExpectedArgument handling in SpanTree

farmaazon:
### Pull Request Description
There was a wrong behavior when connecting to "labeled" argument in expressions being an infix, for example call `Main.foo`. Additionally fixed some issues with AST id change after such operation.

### Checklist


src/rust/ide/lib/span-tree/src/action.rs
```diff
@@ -203,7 +222,7 @@ mod test {
 
     use crate::builder::TreeBuilder;
     use crate::generate::context;
-    use crate::node::Kind::Operation;
+    use crate::node::Kind::{Operation, Argument};
```
wdanilo, line 225: wrong import

wdanilo (approved):

PR #873, opened 2020-10-23
Spantree refactoring and extensions

wdanilo:
### Pull Request Description
- Some of these changes were required for (payload in SpanTree) [The ports & connections are not colored according to type information](https://github.com/enso-org/ide/issues/765)
- This contains refactoring of SpanTree according to some earlier discussions - the `span_tree::node::Kind` contains now much more partitioned information, so building improper SpanTree is prevented. 
- Moreover, the API was reworked to allow for a more modular workflow style.

### Checklist


src/rust/ide/lib/span-tree/src/generate.rs
```diff
@@ -207,17 +213,19 @@ impl SpanTreeGenerator for Ast {
...
         // Removing operands is possible only when chain has at least 3 of them
         // (target and two arguments).
-        let is_removable                                 = self.args.len() >= 2;
-        let node_and_offset:FallibleResult<(Node,usize)> = match &self.target {
+        let kind      = kind.into();
+        let removable = self.args.len() >= 2;
+        let node_and_offset:FallibleResult<(Node<T>,usize)> = match &self.target {
             Some(target) => {
-                let node = target.arg.generate_node(node::Kind::Target {is_removable},context)?;
+                let node = target.arg.generate_node(node::Kind::this().removable(),context)?;
```
farmaazon, line 225: Tests are failing, and here is the (one of) reason(s): before we mark infix arguments as removable only when there were more than 2 of them.

src/rust/ide/lib/span-tree/src/iter.rs
```diff
@@ -3,7 +3,7 @@ use crate::prelude::*;
 
 use crate::Node;
 use crate::node;
-
+use crate::node::Payload;
```
farmaazon, line 6:
```suggestion
use crate::node::Payload;

```

src/rust/ide/lib/span-tree/src/node/kind.rs
```diff
@@ -0,0 +1,303 @@
...
+
+
+// === Smart Kind Constructors ===
+
+#[allow(missing_docs)]
+impl Kind {
+    pub fn this            () -> This           { default() }
+    pub fn argument        () -> Argument       { default() }
+    pub fn insertion_point () -> InsertionPoint { default() }
+}
```
farmaazon, line 42: Is `Kind::this()` so better than `kind::This::default()`?

src/rust/ide/lib/span-tree/src/node/kind.rs
```diff
@@ -0,0 +1,303 @@
...
+            _                       => false
+        }
+    }
+}
+
+
+// === API ===
+
+impl Kind {
+    /// Smart name getter.
```
farmaazon, line 75: Smart getters in some libraries means "getter caching returned values". Let's name all those just "getter".

src/rust/ide/lib/span-tree/src/node/kind.rs
```diff
@@ -0,0 +1,303 @@
...
+}
+
+
+
+// ============
+// === This ===
+// ============
+
+/// Kind representing "this" node. For example, in the following expressions, `foo` is considered
+/// "this": `foo.bar`, `foo + bar`, `foo.+ bar`.
```
farmaazon, line 147: currently, in `bar foo` we also mark `foo` as This. Perhaps we should only when they are methods though, however that cannot be deduced from AST.

src/rust/ide/lib/span-tree/src/node/kind.rs
```diff
@@ -0,0 +1,303 @@
...
+#[derive(Clone,Debug,Default,Eq,PartialEq)]
+#[allow(missing_docs)]
+pub struct This {
+    pub removable : bool,
+    pub tp        : Option<String>,
```
farmaazon, line 152: Can we name it `typename`? `Tp` looks kinda obscure, especially when there are no docs for this field.

src/rust/ide/lib/span-tree/src/node/kind.rs
```diff
@@ -0,0 +1,303 @@
...
+
+
+// === Smart Setters ===
+
+#[allow(missing_docs)]
+impl This {
+    pub fn removable      (mut self)                    -> Self { self.removable = true     ; self }
+    pub fn typed          (mut self, tp:String)         -> Self { self.tp        = Some(tp) ; self }
+    pub fn with_removable (mut self, removable:bool)    -> Self { self.removable = removable; self }
+    pub fn with_tp        (mut self, tp:Option<String>) -> Self { self.tp        = tp       ; self }
```
farmaazon, line 163: here we can name function `with_type`.

src/rust/ide/lib/span-tree/src/node/kind.rs
```diff
@@ -0,0 +1,303 @@
...
+/// arguments.
+#[derive(Clone,Debug,Default,Eq,PartialEq)]
+#[allow(missing_docs)]
+pub struct InsertionPoint {
+    pub kind : InsertionPointType,
+    pub name : Option<String>,
+    pub tp   : Option<String>,
+}
+
+// === Smart Constructors ===
```
farmaazon, line 225: I see everything here is smart :D And again it conflicts with some other programming idioms: https://wiki.haskell.org/Smart_constructors

src/rust/ide/lib/span-tree/src/node/kind.rs
```diff
@@ -0,0 +1,303 @@
...
+// ==========================
+// === InsertionPointType ===
+// ==========================
+
+/// A helpful information about how the new AST should be inserted during Set action. See `action`
+/// module.
+#[allow(missing_docs)]
+#[derive(Copy,Clone,Debug,Eq,PartialEq)]
+pub enum InsertionPointType {
+    // FIXME: Why we need both before and after?
```
farmaazon, line 266: This distinction is purely for operators: in expression a + b, you can put before a (BeforeTarget), after a (AfterTarget) or after b (Append) - all those InsertionPoints are on the same level, but their _set_ action should be handled differently.

src/rust/ide/src/controller/graph.rs
```diff
@@ -652,7 +652,7 @@ impl Handle {
 
         let updated_expression = if connection.destination.var_crumbs.is_empty() {
             let port = info.port()?;
-            let only_empty_ports_after = info.chained_ports_after().all(|p| p.node.is_empty());
+            let only_empty_ports_after = info.chained_ports_after().all(|p| p.node.is_insertion_point());
```
farmaazon, line 655:
```suggestion
            let only_insertion_pointss_after = info.chained_ports_after().all(|p| p.node.is_insertion_point());
```

And fix all usages.

src/rust/ide/src/tests.rs
```diff
@@ -109,10 +110,10 @@ fn span_tree_args() {
         // The tree here should have two nodes under root - one with given Ast and second for
         // an additional prefix application argument.
         [_,second] => {
-            let Node{children,kind,parameter_info,..} = &second.node;
+            let Node{children,kind,..} = &second.node;
             assert!(children.is_empty());
-            assert_eq!(kind,&Kind::Empty(InsertType::ExpectedArgument(0)));
-            assert_eq!(parameter_info,&Some(expected_arg1_param.clone()));
+            assert_eq!(kind,&Kind::from(Kind::insertion_point().with_kind(InsertionPointType::ExpectedArgument(0))));
```
farmaazon, line 115: Long line

farmaazon (approved):

src/rust/ide/lib/span-tree/src/action.rs
```diff
@@ -170,16 +171,14 @@ impl<'a> Implementation for node::Ref<'a> {
 
 
     fn erase_impl(&self) -> Option<EraseOperation> {
-
-        match self.node.kind {
-            node::Kind::Argument{is_removable:true} |
-            node::Kind::Target  {is_removable:true} => Some(Box::new(move |root| {
+        if (self.node.kind.is_argument() || self.node.kind.is_this()) && self.node.kind.removable() {
+            Some(Box::new(move |root| {
```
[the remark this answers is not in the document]
wdanilo: As agred via phone, lets leave it this way and in the future, in case of more complex cases there, we will use matches

src/rust/ide/lib/span-tree/src/generate.rs
```diff
@@ -148,8 +150,10 @@ impl<'a> ApplicationBase<'a> {
 
 // === AST ===
 
-impl SpanTreeGenerator for Ast {
-    fn generate_node(&self, kind:node::Kind, context:&impl Context) -> FallibleResult<Node> {
+impl<T:Payload> SpanTreeGenerator<T> for Ast {
+    fn generate_node
+    (&self, kind:impl Into<node::Kind>, context:&impl Context) -> FallibleResult<Node<T>> {
```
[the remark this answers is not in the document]
wdanilo: I did what we were tlaking about - I left only small part of the function as a generic one and I refactored the body to a non-generic function

src/rust/ide/lib/span-tree/src/generate.rs
```diff
@@ -172,21 +176,24 @@ impl SpanTreeGenerator for Ast {
...
                         let node = Node {
                             size,
                             children,
-                            expression_id,
-                            kind           : node::Kind::Operation,
-                            parameter_info : None,
+                            ast_id,
+                            payload,
+                            kind : node::Kind::Operation,
                         };
```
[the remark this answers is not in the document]
wdanilo: It's harder than it looks like. I did some magic but still id love to chat aobut it more

src/rust/ide/lib/span-tree/src/node/kind.rs
wdanilo: As agreed via phone, yes it is in certain situations :)

src/rust/ide/lib/span-tree/src/node/kind.rs
wdanilo: No, because this is not typename, this is the whole type, can be complex

src/rust/ide/lib/span-tree/src/node/kind.rs
wdanilo: If we leave `tp`, this should be named `with_tp`. Leaving for now.

src/rust/ide/lib/span-tree/src/node/kind.rs
wdanilo: As agreed via phone, please add extensive documentation about these cases with nice ascii art tellign about it, as I would never understand what it is about without you drawing it to me today!

PR #877, opened 2020-10-24
Spantree refactoring - moving kind to payload

wdanilo:
DO NOT MERGE - TO BE DISCUSSED


wdanilo:
As discussed, this PR is not going to land. The idea was to move Kind to the payload, however, this has 2 major drawbacks:
1. SpanTree traversal uses Kind to determine which children to traverse, and thus, Kind should be considered part of the SpanTree structure.
2. We want payload to be mutable, while the structure of the spantree to be immutable. Kind describes the purpose of a particular node, and thus it should be immutable. This shows, that logically, it does not belong to the payload.

PR #878, opened 2020-10-25
Add `filter` and `filter_map` nodes to the FRP library

s9ferech:
### Pull Request Description
Adds a `filter` node and a `filter_map` node to the FRP library as requested in issue #813.

### Checklist

### DIFF
--- a/src/rust/lib/frp/src/lib.rs
+++ b/src/rust/lib/frp/src/lib.rs
@@ -313,8 +313,8 @@ mod dynamic_mode_tests {
     fn test_filter() {
         let passed_events = Rc::new(Cell::new(0));
         frp::new_network! { network
-            def source     = source::<bool>();
-            def filtered   = source.filter(|&value| value == true);
+            def source   = source::<bool>();
+            def filtered = source.filter(|&value| value == true);
             eval filtered ([passed_events](&value) {
                 assert_eq!(value,true);
                 passed_events.set(passed_events.get() + 1);
@@ -324,6 +324,8 @@ mod dynamic_mode_tests {
         source.emit(false);
         source.emit(true);
         source.emit(true);
+        source.emit(false);
+        source.emit(false);
         assert_eq!(passed_events.get(),2);
     }
 }
--- a/src/rust/lib/frp/src/lib.rs
+++ b/src/rust/lib/frp/src/lib.rs
@@ -308,4 +308,22 @@ mod dynamic_mode_tests {
         some_event.emit(());
         assert_eq!(passed_events.get(), 2);
     }
+
+    #[test]
+    fn test_filter() {
+        let passed_events = Rc::new(Cell::new(0));
+        frp::new_network! { network
+            def source     = source::<bool>();
+            def filtered   = source.filter(|&value| value == true);
+            eval filtered ([passed_events](&value) {
+                assert_eq!(value,true);
+                passed_events.set(passed_events.get() + 1);
+            });
+        };
+
+        source.emit(false);
+        source.emit(true);
+        source.emit(true);
+        assert_eq!(passed_events.get(),2);
+    }
 }
--- a/src/rust/lib/frp/src/lib.rs
+++ b/src/rust/lib/frp/src/lib.rs
@@ -328,4 +328,27 @@ mod dynamic_mode_tests {
         source.emit(false);
         assert_eq!(passed_events.get(),2);
     }
+
+    #[test]
+    fn test_filter_map() {
+        let passed_events = Rc::new(Cell::new(0));
+        frp::new_network! { network
+            def source        = source::<bool>();
+            def filter_mapped = source.filter_map(|&value|
+                match value {
+                    true  => Some(()),
+                    false => None,
+                });
+            eval filter_mapped ([passed_events](_) {
+                passed_events.set(passed_events.get() + 1);
+            });
+        };
+
+        source.emit(false);
+        source.emit(true);
+        source.emit(true);
+        source.emit(false);
+        source.emit(false);
+        assert_eq!(passed_events.get(),2);
+    }
 }
--- a/src/rust/lib/frp/src/lib.rs
+++ b/src/rust/lib/frp/src/lib.rs
@@ -291,64 +291,57 @@ mod dynamic_mode_tests {
     fn test_gate() {
         let passed_events = Rc::new(Cell::new(0));
         frp::new_network! { network
-            def behavior   = source::<bool>();
-            def some_event = source::<()>();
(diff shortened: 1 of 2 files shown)


src/rust/lib/frp/src/lib.rs
```diff
@@ -308,4 +308,22 @@ mod dynamic_mode_tests {
         some_event.emit(());
         assert_eq!(passed_events.get(), 2);
     }
+
+    #[test]
+    fn test_filter() {
+        let passed_events = Rc::new(Cell::new(0));
+        frp::new_network! { network
+            def source     = source::<bool>();
+            def filtered   = source.filter(|&value| value == true);
```
wdanilo, line 317: alignment you use 3 spaces before `=` instead of 1

src/rust/lib/frp/src/lib.rs
```diff
@@ -308,4 +308,22 @@ mod dynamic_mode_tests {
...
+            eval filtered ([passed_events](&value) {
+                assert_eq!(value,true);
+                passed_events.set(passed_events.get() + 1);
+            });
+        };
+
+        source.emit(false);
+        source.emit(true);
+        source.emit(true);
```
wdanilo, line 326: add one more false emit after treu amits - just as a potential corner case

src/rust/lib/frp/src/nodes.rs
```diff
@@ -303,6 +303,16 @@ impl Network {
     }
 
 
+    // === Filter ===
+
+    /// Passes exactly those incoming events that satisfy the predicate `p`.
+    pub fn filter<T:EventOutput>
+    (&self, label:Label, src:&T, p:impl'static+Fn(&Output<T>)->bool)
```
wdanilo, line 310: would `p:'static + impl Fn(&Output<T>)->bool` work here as well? If so, please use it as it's more clear what it does. In fact I haven't ever seen `impl 'static` usage anywhere else, but I'm glad it works!

src/rust/lib/frp/src/nodes.rs
```diff
@@ -1653,6 +1672,46 @@ where T1:EventOutput, T2:EventOutput, T3:EventOutput, T4:EventOutput {
 
 
 
+// ==============
+// === Filter ===
+// ==============
+
+pub struct FilterData  <T,P> { _src:T, predicate:P }
```
wdanilo, line 1679: the `_src` field is not needed. Please remove it here and in `Map` and check if it works (it should work). This was an obsolete design and I'm wondering how on earth it's still there! We are sometimes using something like `src2:watch::Ref<T2>` to indicate on which inputs values should be cached, but here you dont need it. For example, if you have `All4`, and when you have incoming event on 1st input, you want to sample all other inputs, thus you need to cache them. Here you'll have a simple streeam, so you dont need to remmebr any info about the input.

src/rust/lib/frp/src/nodes.rs
```diff
@@ -1653,6 +1672,46 @@ where T1:EventOutput, T2:EventOutput, T3:EventOutput, T4:EventOutput {
...
+pub type   OwnedFilter <T,P> = stream::Node     <FilterData<T,P>>;
+pub type   Filter      <T,P> = stream::WeakNode <FilterData<T,P>>;
+
+impl<T,P> HasOutput for FilterData<T,P>
+    where T:EventOutput, P:'static+Fn(&Output<T>)->bool {
```
wdanilo, line 1684: we align `where` to the same level as `impl`. Please fix it here and below.

wdanilo:
Amazing work @s9ferech! ❤️ Thanks for taking care about code quality and clean implementation! I've left a few minor comments and I'm waiting also for the version with `filter_map` included!

src/rust/lib/frp/src/nodes.rs
s9ferech: This would not work. But if you find that clearer then I can use an explicit type parameter as in
```
pub fn filter<T:EventOutput, P:'static+Fn(&Output<T>)->bool>
(&self, label:Label, src:&T, p:P)
```

src/rust/lib/frp/src/nodes.rs
s9ferech: I think we need the type parameter `T` in the struct to implement `HasOutput`, but but there might be another way. The `_src` field was probably included because the compiler would otherwise complain that `T` is unused. It wouldn't be able to infer lifetime constraints for `T`. I will change the definition to
```pub struct FilterData  <T,P> { phantom:PhantomData<T>, predicate:P }```
This is equivalent to the original definition without wasting memory on `_src`.

Comment (s9ferech) [reply src/rust/lib/frp/src/nodes.rs:1679]: Maybe someone should check if the lifetime constraints that we get here really make sense.

File: src/rust/lib/frp/src/lib.rs [hunk 6]
```diff
@@ -308,4 +308,47 @@ mod dynamic_mode_tests {
...
+            eval filter_mapped ([passed_events](_) {
+                passed_events.set(passed_events.get() + 1);
+            });
+        };
+
+        source.emit(false);
+        source.emit(true);
+        source.emit(true);
+        source.emit(false);
+        source.emit(false);
```
wdanilo, line 351: simplify (here and above):
```rust
let input = &[false,true,true,false,false];
for val in input {
    source.emit(*val);
}
```

src/rust/lib/frp/src/lib.rs
```diff
@@ -308,4 +308,47 @@ mod dynamic_mode_tests {
...
+        assert_eq!(passed_events.get(),2);
```
wdanilo, line 352: Instead of hardcoding 2, compute it from the `input` variable proposed above. This will be less magical. Apply to the other test as well pls.

src/rust/lib/frp/src/nodes.rs
wdanilo: I'm ok with the current style, however, the most clear to me (personal preference) in case of such a complex type, would be this style, as it provides a very short and clean look at the strucutre of the signature before getting into details about the constraints (which we use across this lib):

```rust
pub fn filter<T,F>(&self, label:Label, src:&T, f:F) -> Stream<Output<T>>
where T:EventOutput, F:'static+Fn(&Output<T>)->bool {
    ...
}
```

Btw, you are using this style below in filter_map - so lets use it here as well :)

wdanilo (commented):
Amazing job with it, thank you! Let's fix the small issues and lets merge it! :) 

Btw, @farmaazon do you know why CI is not running here?

wdanilo (approved):
❤️

PR #880, opened 2020-10-26
PR: Fix node action bar hover area beeing too large.

MichaelMauderer:
### Pull Request Description
Make the hover area around node actions tighter to the actual buttons.

### Important Notes
Re-enables debug shape visualisation shortcuts

### Checklist

### DIFF
--- a/src/rust/ide/view/graph-editor/src/component/node/action_bar.rs
+++ b/src/rust/ide/view/graph-editor/src/component/node/action_bar.rs
@@ -16,6 +16,15 @@ use ensogl_theme as theme;
 
 
 
+// ==================
+// === Constants  ===
+// ==================
+
+const BUTTON_PADDING : f32 = 0.5;
+const BUTTON_OFFSET  : f32 = 0.5;
+
+
+
 // ===============
 // === Shapes  ===
 // ===============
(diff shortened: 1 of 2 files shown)


src/rust/ide/view/graph-editor/src/component/node/action_bar.rs
```diff
@@ -112,26 +121,44 @@ impl Model {
...
 
+    fn set_size(&self, size:Vector2) {
+        self.size.set(size);
         self.icons.set_position_x(-size.x/2.0);
 
         self.place_button_in_slot(&self.icon_visibility, 0);
         self.place_button_in_slot(&self.icon_skip, 1);
         self.place_button_in_slot(&self.icon_freeze, 2);
 
+        self.layout_hover_area_to_cover_buttons(3);
```
wdanilo, line 160: lets do it in a loop, and infer the `3` number automaticly from it. You know, like `let buttons = &[&self.icon_visibility,...]` and then loop trouhh it and count it.

wdanilo (approved):

src/rust/ide/view/graph-editor/src/component/node/action_bar.rs
MichaelMauderer: Unfortunately, that doesn't work easily without some explicit dyn typing because the shapes all have different types. We'd have to create an `let buttons : [&dyn ObjectOps + Shape] = ...` (or something like that, this is just of the top of my head). If you feel strongly about this, I can try that, though.

wdanilo: Good point. Lets leave it as is!

PR #891, opened 2020-10-29
PR: Correclty update visualisation after internal change.

MichaelMauderer:
### Pull Request Description
At the moment visualizations don't get a data update after we used the cycling shortcut or the visualization chooser until they go through a show/hide cycle. This PR fixes that by adding enable/disable events whenever a visualization on a node changes. 

### Important Notes
Based on #786, might also solve #865, #864

Refactoring of `visualisation::container::FRP` to use the new macro generated FRP structs.

### Checklist


MichaelMauderer:
@farmaazon @wdanilo  Could you check whether this change makes sense from the controller side? Is it sane to emit a pair of `visualization_disabled` / `visualization_enabled` events to trigger the controller updating the data for a visualization on a node? Or should we design a better API that informs the controller about graph-editor internal changes to a node's visualization.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -976,11 +976,19 @@ impl GraphEditorModelWithNetwork {
...
+
             vis_enabled        <- vis_visible.gate(&vis_visible);
             vis_disabled       <- vis_visible.gate_not(&vis_visible);
 
+            // Ensure the graph editor knows about internal changes to the visualisation. If the
+            // visualisation changes that should indicate that the old one has been disabled and a
+            // new one has been enabled.
+            output.visualization_disabled          <+ vis_changed.constant(node_id);
+            output.visualization_enabled           <+ vis_changed.constant(node_id);
+
```
farmaazon, line 991: I'd rather opt for making "visualization_changed" a part of the GraphEditor API, and do tricks on the integration level. But this can take a bit of time, so for now please make an issue for refactoring this, and put here TODO with issue number.

farmaazon (changes requested):

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -596,28 +578,27 @@ impl Container {
...
             eval fullscreen_position.value ((p) model.fullscreen_view.set_position(*p));
-            eval model.frp.preprocessor    ((code) inputs.preprocessor_select.emit(code));
         }
 
 
         // ===  Visualisation chooser frp bindings ===
         frp::extend! { network
-            eval action_bar.visualisation_selection([model,registry,scene,action_bar](visualization_path) {
+            eval action_bar.visualisation_selection(
+            [model,registry,action_bar,frp](visualization_path) {
```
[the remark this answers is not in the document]
MichaelMauderer: I've refactored this to be a bit nicer and void the memory leak. Unfortunately, filter doesn't work here, as the `definition` needs to be wrapped in an `Option` because it does not itself implement `Default.`

src/rust/ide/view/graph-editor/src/lib.rs
MichaelMauderer: Added the task #896  and the todo in the doc.

wdanilo (approved):

MichaelMauderer:
While not blocked, this is waiting for confirmation form @farmaazon before being merged.

farmaazon (approved):

PR #894, opened 2020-10-29
Suggestion: Panic on loops in the functional reactive data flow

s9ferech:
### Pull Request Description

In the current implementation, loops in the data flow of a functional reactive network are detected and the data flow is quietly interrupted at that point to prevent infinite looping. Such an interruption would likely lead to behaviour that is hard to predict. Panicking instead could make it much easier to identify underlying bugs.

### Checklist


wdanilo:
Thank you for the suggestion! There is a problem with it though, namely, the design which breaks loops was done on purpose, is very popular in FRP implementations. However, it is true that we have never used it for a real world use-case. Moreover, I always prefer warnings than crashing the gui, so I'd suggest to print warning about it to console instead.

s9ferech:
Thanks for the answer. Out of curiosity, do you have a resource explaining the motivations behind this kind of design? Or could you give me a brief overview?

wdanilo:
Oh, the reason is pretty simple here - basically, you prefer your software to recover on production rather than crash. As FRP systems are used in many "reactive behaviours", like GUI, or device signal processing, you prefer your application to still work on client-side rather than break at some point. So collecting logs and fixing that always looks better than software that crashesh from both user nad business perspective. This is exactly the same reason why JS is used more and more to code GUI for games on PC and consoles - it is VERY forgiving. So there is higher probability that user will see "undefined" somewhere in the game rather than game crashing, which makes the final product just more stable. Does it make sense?

wdanilo:
Btw, could you use our `logger` functionality to report warnings here, please?

s9ferech:
In principle, it makes sense. I just don't have enough experience and knowledge of the Enso IDE yet to really understand this kind of trade-off.

s9ferech:
> Btw, could you use our `logger` functionality to report warnings here, please?

Do you mean the GitHub issue tracker?

s9ferech:
> Btw, could you use our `logger` functionality to report warnings here, please?

Which verbosity level would be appropriate?

wdanilo:
Definitely warning!

wdanilo:
Hey @s9ferech, are you still working on this issue? :)

s9ferech:
I am planning to get to it. But maybe someone else should just do it since I don't understand yet how you handle logging.

src/rust/lib/frp/src/stream.rs
```diff
@@ -175,19 +175,26 @@ impl<Out:Data> HasOutput for NodeData<Out> {
...
-            }
-            self.targets.borrow_mut().retain(|target| target.data.on_event_if_exists(value));
-            let mut new_targets = self.new_targets.borrow_mut();
-            if !new_targets.is_empty() {
-                let new_targets_ref : &mut Vec<EventInput<Out>> = &mut new_targets;
-                self.targets.borrow_mut().extend(mem::take(new_targets_ref));
-            }
-            self.during_call.set(false);
+        if self.during_call.get() {
+            let logger = Logger::new("stream::NodeData");
```
wdanilo, line 179: The logger name is used to navigate the user logically to the place where the error occured. I'd name the logger just `frp` in this case, as using internal structure names would not tell the user a lot.

src/rust/lib/frp/src/stream.rs
```diff
@@ -175,19 +175,26 @@ impl<Out:Data> HasOutput for NodeData<Out> {
...
+            let log_message = format!(
+                "Encountered a loop in the reactive dataflow at '{}'",
+                self.label);
+            logger.warning(&*log_message);
```
wdanilo, line 183: You could also use our macro: `warning!(&logger, "Encountered a loop in the reactive dataflow at '{self.label}'"` :)

src/rust/lib/frp/src/stream.rs
```diff
@@ -175,19 +175,26 @@ impl<Out:Data> HasOutput for NodeData<Out> {
...
+            return;
+        }
```
wdanilo, line 185: We have a rule, that we try not to use early function return if we are not forced to. It just increases the brain power to understand all exit points of a function. Thus, I'd suggest to use simple `else` statement here instead.

wdanilo (commented):
Amazing to see it implemented, thank you ❤️ I added a few minor comments - either apply them or reply if you dont agree with any of them. And then lets merge it! :)

wdanilo (approved):

wdanilo:
I'm happy to see this implemented. Small, but important improvement, good job! ❤️

PR #900, opened 2020-11-02
Node bump

wdanilo:


farmaazon (approved):
May add a comment about automatic LTS setting in workflow and link to the issue with action.

PR #906, opened 2020-11-03
PR: Edge multi connect on cmd / control press

MichaelMauderer:
### Checklist

### DIFF
--- a/src/rust/ide/view/graph-editor/src/lib.rs
+++ b/src/rust/ide/view/graph-editor/src/lib.rs
@@ -383,16 +383,43 @@ ensogl::define_endpoints! {
         debug_set_test_visualization_data_for_selected_node(),
 
 
-        set_detached_edge_targets    (EdgeTarget),
-        set_detached_edge_sources    (EdgeTarget),
+        // === Edges ===
+
+        /// Set the source of the given edge. This might connect the edge if it was detached
+        /// before.
         set_edge_source              ((EdgeId,EdgeTarget)),
+        /// Set the edge of the given edge. This might connect the edge if it was detached
+        /// before.
         set_edge_target              ((EdgeId,EdgeTarget)),
+        /// Detach the source of the given edge and create a detached edge.
         unset_edge_source            (EdgeId),
+        /// Detach the target of the given edge and create a detached edge.
         unset_edge_target            (EdgeId),
+        /// Create a edge between the two edge targets.
         connect_nodes                ((EdgeTarget,EdgeTarget)),
+
+        // FIXME[WD]: These FRP endpoints look like output endpoints, but are used in some strange
+        //            cyclic way. To be refactored / removed.
+        some_edge_targets_detached  (),
+        some_edge_sources_detached  (),
+        all_edge_targets_attached   (),
+        all_edge_sources_attached   (),
+        all_edges_attached          (), // FIXME: wrong name! Its all sources and targets of a single edge!
+
+        edge_multi_connect_on       (),
+        edge_multi_connect_off      (),
+        // Mouse down during multi-connect mode.
+        edge_multi_connect_down     (),
+        edge_multi_connect_up       (),
+
+
+        // === Nodes ===
+
         deselect_all_nodes           (),
         press_node_input             (EdgeTarget),
         press_node_output            (EdgeTarget),
+        release_node_input           (EdgeTarget),
+        release_node_output          (EdgeTarget),
         remove_all_node_edges        (NodeId),
         remove_all_node_input_edges  (NodeId),
         remove_all_node_output_edges (NodeId),
@@ -412,14 +439,6 @@ ensogl::define_endpoints! {
 
         hover_node_input            (Option<EdgeTarget>),
         hover_node_output           (Option<EdgeTarget>),
-
-        // FIXME[WD]: These FRP endpoints look like output endpoints, but are used in some strange
-        //            cyclic way. To be refactored / removed.
-        some_edge_targets_detached  (),
-        some_edge_sources_detached  (),
-        all_edge_targets_attached   (),
-        all_edge_sources_attached   (),
-        all_edges_attached          (), // FIXME: wrong name! Its all sources and targets of a single edge!
 
     }
     Output {
@@ -441,17 +460,23 @@ ensogl::define_endpoints! {
 
         edge_added         (EdgeId),
         edge_removed       (EdgeId),
+        /// The source of this edge was updated to the given `EdgeTarget`.
         edge_source_set    ((EdgeId,EdgeTarget)),
+        /// The target of this edge was updated to the given `EdgeTarget`.
         edge_target_set    ((EdgeId,EdgeTarget)),
+        /// The source of this edge was removed.
         edge_source_unset  (EdgeId),
+        /// The target of this edge was removed.
         edge_target_unset  (EdgeId),
 
-        some_edge_targets_detached (),
-        some_edge_sources_detached (),
-        all_edge_targets_attached  (),
-        all_edge_sources_attached  (),
-        all_edges_attached         (),
+        some_edge_targets_detached (bool),
+        some_edge_sources_detached (bool),
+        all_edges_attached         (bool),
+        /// Indicates whether we are in edge creation mode. This is the case if there is a detached
+        /// edge that is dragged in the graph editor.
+        edge_creation_mode         (bool),
 
+        // FIXME: This needs a doc explaining thew difference to `edge_added`/`edge_removed`.
         connection_added   (EdgeId),
         connection_removed (EdgeId),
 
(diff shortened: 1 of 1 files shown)


src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1999,7 +2062,60 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+
+    trace port_input_mouse_up_multi;
+    trace port_output_mouse_up_multi;
+
+    new_edge_source_multi <= port_output_mouse_up_multi.gate(&out.edge_creation_mode).map(f!((target) {
+        model.set_detached_edge_source(target)
+    }));
+    new_edge_target_multi <= port_input_mouse_up_multi.gate(&out.edge_creation_mode).map(f!((target) {
+        model.set_detached_edge_target(target)
+    }));
```
wdanilo, line 2079: its ver y complex. Split it to multiple vars - gate and map logic with properly named vars

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1594,12 +1661,20 @@ impl application::View for GraphEditor {
           , (Press       , "" , "alt enter"         , "exit_node")
 
           // === Node Editing ===
-          , (Press   , "" , "cmd"                   , "edit_mode_on")
-          , (Release , "" , "cmd"                   , "edit_mode_off")
-          , (Press   , "" , "cmd enter"             , "edit_selected_node")
-          , (Press   , "" , "cmd left-mouse-button" , "edit_mode_on")
-          , (Release , "" , "cmd left-mouse-button" , "edit_mode_off")
-          , (Release , "" , "enter"                 , "stop_editing")
+          , (Press   , "!edge_creation_mode" , "cmd"                   , "edit_mode_on")
```
wdanilo, line 1664: I dont think this is good usage of kb masks. Instead, please handle it inside of the FRP. We will be removign kb masks in the future - it was bad idea to introduce them in the first place :(

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1999,7 +2062,60 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+    trace port_output_mouse_up_multi;
+
+    new_edge_source_multi <= port_output_mouse_up_multi.gate(&out.edge_creation_mode).map(f!((target) {
+        model.set_detached_edge_source(target)
+    }));
+    new_edge_target_multi <= port_input_mouse_up_multi.gate(&out.edge_creation_mode).map(f!((target) {
+        model.set_detached_edge_target(target)
+    }));
+
+    re_create_edge_from_source <- new_edge_source_multi.gate(&out.edge_multi_connect)._0();
```
wdanilo, line 2081: its hard to understand all these `._0()` - create more variables with good names to make it easier to follow.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1999,7 +2062,60 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
     out.source.edge_target_set <+ new_edge_target;
 
 
+    // == Multi Connect
+    trace out.edge_creation_mode;
+
+    port_input_mouse_up_multi  <- inputs.hover_node_input.sample(&inputs.edge_multi_connect_up).unwrap();
+    port_output_mouse_up_multi <- inputs.hover_node_output.sample(&inputs.edge_multi_connect_up).unwrap();
```
wdanilo, line 2069: I don't believei ts correct. `inputs.hover_node_input` is for external control of the graph - like, backend can tell - hover this input. But this FRP endpoint should not be set by hovering the input with mouse. We should have a rule in FRP network that we never, ever emit events to `inputs` from inside of the FRP network and I think (hope) we do not do it here as well. The right solution here would be to use "outputs" instead of inputs, if our FRP network would set `outputs.hover_node_input` to something.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1999,7 +2062,60 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
     out.source.edge_target_set <+ new_edge_target;
 
 
+    // == Multi Connect
+    trace out.edge_creation_mode;
+
+    port_input_mouse_up_multi  <- inputs.hover_node_input.sample(&inputs.edge_multi_connect_up).unwrap();
```
wdanilo, line 2068: change the naming a little bit. First part - what is it, second, when it happens, like `port_input_on_up_multi`. I dropped `mouse` and its not guaranteed that it will be mouse event connected here (maybe joystick in VR?)

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1999,7 +2062,60 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+
+    new_edge_source_multi <= port_output_mouse_up_multi.gate(&out.edge_creation_mode).map(f!((target) {
+        model.set_detached_edge_source(target)
+    }));
+    new_edge_target_multi <= port_input_mouse_up_multi.gate(&out.edge_creation_mode).map(f!((target) {
+        model.set_detached_edge_target(target)
+    }));
+
+    re_create_edge_from_source <- new_edge_source_multi.gate(&out.edge_multi_connect)._0();
+    re_create_edge_from_target <- new_edge_target_multi.gate(&out.edge_multi_connect)._0();
```
wdanilo, line 2082: prefix "re_" doesnt tell much. If you want short names, you can abbreviate "create" with "mk", which is very common (thanks unixes!). The same applies to `tgt` and `src`. We are alwyas suepr ok with such helper variable names (until they become public api).

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1999,7 +2062,60 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+        let target = model.edge_target(edge).unwrap();
+        let edge   = model.new_edge_from_input(&edge_mouse_down,&edge_over,&edge_out);
+        model.set_edge_target(edge,target);
+    });
+
+    // == Connect dragged edge and (maybe) finish edge creation mode.
+    new_edge_source <= port_output_mouse_up.gate(&out.edge_creation_mode).gate_not(&out.edge_multi_connect).map(f!((target) {
+        model.set_detached_edge_source(target)
+    }));
+    new_edge_target <= port_input_mouse_up.gate(&out.edge_creation_mode).gate_not(&out.edge_multi_connect).map(f!((target) {
```
wdanilo, line 2099: these nested gates should create new, well named frp var

wdanilo (commented):
I dont like one more thing in thsi code - it doesnt seem to be connected with normal edge creation. I mean - normal edge creation just stops when you release mous button. However, in this "multi-mode" it should not. It sees here like some logic is duplicated (edge creation, etc) between single and multi mode. If I'm right, this sohuld be combined into a single, small FRP network, rather than duplicated.

wdanilo (commented):
To explain previous comment better, this code:
```rust
    new_output_edge <- create_edge_from_output.map(f_!([model,edge_mouse_down,edge_over,edge_out] {
        Some(model.new_edge_from_output(&edge_mouse_down,&edge_over,&edge_out))
    })).unwrap();
```
and this one:
```rust
    eval re_create_edge_from_target ([model,edge_mouse_down,edge_over,edge_out](edge) {
        let source = model.edge_source(edge).unwrap();
        let edge   = model.new_edge_from_output(&edge_mouse_down,&edge_over,&edge_out);
        model.set_edge_source(edge,source);
    });
```

Are just edge creation duplicated logic, and one of these codes should be removed, and edge creation should be done in one place only.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1895,11 +1970,10 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
     on_connect_follow_mode <- any(on_output_connect_follow_mode,on_input_connect_follow_mode);
     connect_drag_mode      <- any(on_connect_drag_mode,on_connect_follow_mode);
 
-    on_detached_edge    <- any(&inputs.some_edge_targets_detached,&inputs.some_edge_sources_detached);
-    has_detached_edge   <- bool(&inputs.all_edges_attached,&on_detached_edge);
-
     eval node_input_touch.down ((target)   model.frp.press_node_input.emit(target));
     eval node_output_touch.down ((target)  model.frp.press_node_output.emit(target));
+    eval node_input_touch.up ((target)   model.frp.release_node_input.emit(target));
+    eval node_output_touch.up ((target)  model.frp.release_node_output.emit(target));
```
wdanilo, line 1976: you should never ever emit INPUTS to FRP network. You are allowed to emit outputs only from inside of the network,. Inputs are alwyas considered public API and they should never be emitted not by external thing. Otherwise we will have logical cycles everywhere and this will become spaghetti

wdanilo:
As this introduces a lot of edge refactoring, but not extensive, overall refactoring, I'd close it for now. This FIX should be part of a proper, big, edge code refactoring in graph editor. To be reopned when needed. It doesnt seem to be merged anytime soon now.

PR #907, opened 2020-11-04
Correctly propagate project name updates from `Breadcrumbs` to `ProjectName`.

MichaelMauderer:
### Pull Request Description
The FRP from the `Breadcrumbs` was incorrectly connected to the FRP of the `ProjectName` and updates were not propagated correctly. This PR fixes that. It also ensures an uninitialised project name (which should never happen) is more visible by changing the default text to something more obvious. 

### Checklist

### DIFF
--- a/src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
+++ b/src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
@@ -399,7 +399,7 @@ impl Breadcrumbs {
 
             // === Project Name ===
 
-            eval frp.input.project_name((name) model.project_name.name(name));
+            eval frp.input.project_name((name) model.project_name.set_name.emit(name));
             frp.source.project_name <+ model.project_name.output.name;
 
 
--- a/src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
+++ b/src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
@@ -399,8 +399,8 @@ impl Breadcrumbs {
 
             // === Project Name ===
 
-            eval frp.project_name((name) model.project_name.frp.name(name));
-            frp.source.project_name <+ model.project_name.frp.output.name;
+            eval frp.input.project_name((name) model.project_name.name(name));
+            frp.source.project_name <+ model.project_name.output.name;
 
 
             // === GUI Update ===
--- a/src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
+++ b/src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
@@ -32,7 +32,7 @@ use logger::enabled::Logger;
 
 // Project name used as a placeholder in `ProjectName` view when it's initialized.
 // This should never be visible, but if it is we want to easily trace it back here.
-const UNINITIALISED_PROJECT_NAME: &str = "Project Name Uninitialised";
+const UNINITIALIZED_PROJECT_NAME: &str = "Project Name Uninitialized";
 /// Default line height for project names.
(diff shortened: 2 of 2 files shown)


src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
```diff
@@ -30,8 +30,9 @@ use logger::enabled::Logger;
 // === Constants ===
 // =================
 
-/// Project name used as a placeholder in `ProjectName` view when it's initialized.
-pub const UNKNOWN_PROJECT_NAME : &str = "Unknown";
+// Project name used as a placeholder in `ProjectName` view when it's initialized.
+// This should never be visible, but if it is we want to easily trace it back here.
+const UNINITIALISED_PROJECT_NAME: &str = "Project Name Uninitialised";
```
farmaazon, line 35:
```suggestion
const UNINITIALIZED_PROJECT_NAME: &str = "Project Name Uninitialized";
```

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -399,8 +399,8 @@ impl Breadcrumbs {
 
             // === Project Name ===
 
-            eval frp.project_name((name) model.project_name.frp.name(name));
-            frp.source.project_name <+ model.project_name.frp.output.name;
+            eval frp.input.project_name((name) model.project_name.name(name));
```
farmaazon, line 402: this call `model.project_name.name(name)` looks strange, and the clash of input and output node is also not desired...

Can we rename input to `set_name` @wdanilo ?

farmaazon (approved):
Tested and works.

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
MichaelMauderer: > Can we rename input to `set_name`
I would be for that. Especially since we have an in and output here.

wdanilo: lets do it

src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
```diff
@@ -30,8 +30,9 @@ use logger::enabled::Logger;
 // === Constants ===
 // =================
 
-/// Project name used as a placeholder in `ProjectName` view when it's initialized.
-pub const UNKNOWN_PROJECT_NAME : &str = "Unknown";
+// Project name used as a placeholder in `ProjectName` view when it's initialized.
+// This should never be visible, but if it is we want to easily trace it back here.
```
wdanilo, line 34: "This should never be visible, but if it is we want to easily trace it back here" - I completely do not understand this sentence. Moreover, I do not understand the reason for this constant, but maybe it would be obvious after the sentence would be fixed :)

MichaelMauderer: This is a default value for the project name. But the project name should always be initialized externally for the current project. If this value is visible, it was not set to the correct project name due to some bug. If we leave this value, e.g., as empty string, we will have a harder time tracking the behaviour "no project name visible" in case something goes wrong. This is what happened here, we saw project name "Unknown", which was the previous default value, but this was not a clear indicator of the actual bug: faulty initialization.

wdanilo: Ok, makes sense. Could you expand the comment there slightly to explain it better, a you just did to me here? <3

wdanilo (approved):

MichaelMauderer:
@wdanilo Updated the doc.

Issue #761, opened 2020-08-25
Glitches while the IDE is loading


mwu-tow:
### General Summary
While IDE is loading, several visual glitches are visible.

### Steps to Reproduce
Open the project.

### Expected Result
First loading screen appears. Then it disappears and the desired project view appears.

### Actual Result
First loading screen appears. It flushes really quickly. 
Then for some time (but longer than the loading screen) the screen is entirely blank.
Then this appears:
![vlcsnap-2020-08-26-00h12m35s446](https://user-images.githubusercontent.com/1548407/91233138-e08ced00-e730-11ea-8f2f-2b9e710fd5cf.png)

It stays this way even longer (about 3 seconds), then finally the desired view appears.

sylwiabr:
is this still valid @mwu-tow

PR #909, opened 2020-11-04
Avoid the visualisation container background showing up when loading the scene or collapsing some nodes.

MichaelMauderer:
### Pull Request Description
Avoids the background of the visualization container showing up when a new node is created on startup and when collapsing nodes.

### Important Notes
This is not a great solution. It fixes the symptom but it does probably not address the root cause of this issue. I'm not sure what exactly the root cause is, but I suspect some parenting issue. Unfortunately, I have not yet been able to pinpoint a specific fix or reproducible bug regarding the order of parenting of all the shapes involved.  

Maybe this might even be involved to #761?  (I don't see the behaviour there, but I see the empty container showing up instead)

### Checklist

### DIFF
--- a/src/rust/ide/view/graph-editor/src/component/visualization/container.rs
+++ b/src/rust/ide/view/graph-editor/src/component/visualization/container.rs
@@ -607,9 +607,10 @@ impl Container {
 
         // FIXME[mm]: If we set the size right here, we will see spurious shapes in some
         // computation heavy circumstances (e.g., collapsing nodes #805, or creating an new project
-        // #761).
-        // If we leave use the frp api and don't abort the animation, the size will stay at (0,0)
-        // until the animation has run its course and the shape stays invisible during loads.
+        // #761). This should not happen anyway, but the following is a hotfix to hide the visible
+        // behaviour. If we leave use the frp api and don't abort the animation, the size will stay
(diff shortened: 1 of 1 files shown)


src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -602,8 +602,12 @@ impl Container {
             eval_ on_selected ( action_bar.hide_icons.emit(()) );
         }
 
+        // FIXME[mm]: If we set the size right here, we will see spurious shapes in some
+        // computation heavy circumstances (e.g., collapsing nodes, or creating an new project).
+        // If we let the animation do the size change, the size will only be set after the delays.
+        // This is not optimal the optimal solution to this problem, as it also means that we have
+        // an animation on an invisible component running.
```
wdanilo, line 609: add issue link to describe it here better. Moreover, I do not fully understand the reasons for it. Could you write this in more strucured way - like in points 1. ..., 2 ... , 3... - what happens after each other and why?

MichaelMauderer: I've expanded the description.

wdanilo (approved):

PR #912, opened 2020-11-04
Fixes visualizations when using Enso's builtin JSON parser

BinarySoftware:
## Pull Request Description

As @sylwiabr found out today, table view didn't work on Tram's playground, and that was because visualization was trying to parse an already parsed object with Enso's builtin parser. Fixes the fact and also fixes it on other two visualizations.

### Checklist

### DIFF
--- a/src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/mapView.js
+++ b/src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/mapView.js
@@ -62,7 +62,10 @@ class MapViewVisualization extends Visualization {
         mapElem.setAttributeNS(null,"style","width:" + width + "px;height: " + height + "px;");
         this.dom.appendChild(mapElem);
 
-        const parsedData = eval('('+data+')' ) || JSON.parse(data);
+        let parsedData = data;
+        if (typeof data === "string") {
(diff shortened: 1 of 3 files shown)


src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/mapView.js
```diff
@@ -62,12 +62,18 @@ class MapViewVisualization extends Visualization {
         mapElem.setAttributeNS(null,"style","width:" + width + "px;height: " + height + "px;");
         this.dom.appendChild(mapElem);
 
-        const parsedData = eval('('+data+')' );
+        const parsedData = eval('('+data+')' ) || JSON.parse(data);
```
wdanilo, line 65: ok, this is scarry. What guarantees you have that data would not contain some kind of magical string which would not cause harm here? Imagine something like that - In enso, we query HTTP endpoint, and we get string back, from third party server. Then this string is sent to GUI (lets assume that by accident or somehow it is not converted to JSON) earlier and it is evaluated here. This is a BIG security concern. We should never ever use eval unless we have to. Why just using JSON parse would not be enough here?

wdanilo (approved):

PR #915, opened 2020-11-05
Put here keyword instead of module name

farmaazon:
### Pull Request Description
Searcher by default put module name as this argument when picked module method. However the name sometimes clashes with existing identifier. This is especially problematic in Main module with function main defined.

To avoid most common problem, the searcher will put "here" keyword instead of module name if the function chosen is a current module's method.

### Checklist

### DIFF
--- a/build/run.js
+++ b/build/run.js
@@ -135,7 +135,7 @@ commands.build.rust = async function(argv) {
 
         console.log('Checking the resulting WASM size.')
         let stats = fss.statSync(paths.dist.wasm.mainOptGz)
-        let limit = 4.12
+        let limit = 4.13
         let size = Math.round(100 * stats.size / 1024 / 1024) / 100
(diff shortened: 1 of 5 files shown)


wdanilo (approved):

kustosz:
> To avoid most common problem, the searcher will put "here" keyword instead of module name if the function chosen is a current module's method.

This sounds wrong, it should also check if the module name is the current module name. Otherwise `here` is equally wrong.

farmaazon:
> > To avoid most common problem, the searcher will put "here" keyword instead of module name if the function chosen is a current module's method.
> 
> This sounds wrong, it should also check if the module name is the current module name. Otherwise `here` is equally wrong.

I don't understand. If chosen function is a current module's method, then it's module name is equal to the current module name, isn't it?

PR #919, opened 2020-11-09
Show loading bar for the whole loading duration.

MichaelMauderer:
### Pull Request Description
Avoids early removal of the loading screen before the scene is ready.

### Important notes.
Also includes the implementation of a `FrameDelay` future. 

***Known issue***: The loading screen colour is fixed and not theme based. 

### Checklist

### DIFF
--- a/src/js/lib/content/src/index.html
+++ b/src/js/lib/content/src/index.html
@@ -72,6 +72,11 @@
                 border-radius: 14px;
                 box-shadow: 0 0 10px hsla(0, 0%, 0%, 0.1);
             }
+
+            #loader {
+                /* FIXME[mm]: This should be based on the theme colors. See #914.  */
+                background-color: rgb(246, 243, 241);
+            }
         </style>
         <script type="module" src="/assets/index.js" defer></script>
     </head>
--- a/src/rust/ide/src/executor/delay.rs
+++ b/src/rust/ide/src/executor/delay.rs
@@ -1,0 +1,84 @@
+//! Module defining `FrameDelay` - a `Future` that wraps a closure that should be called at a later
+//! time.
+
+pub use enso_prelude::*;
+
+use enso_protocol::prelude::Future;
+use ensogl_system_web::prelude::Closure;
+use ensogl_system_web::request_animation_frame;
+use futures::task::Context;
+use futures::task::Poll;
+use futures::task::Waker;
+use std::pin::Pin;
+
+
+// ========================
+// === Type Definitions ===
+// ========================
+
+type OnFrameClosure = Closure<dyn FnMut(f64)>;
+
+
+
+// ===================
+// === Frame Delay ===
+// ===================
+
+
+/// Delayed calls to a closure. The time delays is specified in number of render frames.
+#[derive(Debug)]
+pub struct FrameDelay<T>{
+    /// Counter that keeps track of the passage of frames. Needs to be shared with the JS runtime.
+    frames_left : Rc<Cell<u64>>,
+    /// Internal closure that is shared with the JS runtime and updates the `frames_left`.
+    on_frame    : Option<OnFrameClosure>,
+    /// Target callback that is called after the delay has passed.
+    target      : T,
+}
+
+impl<T> Unpin for FrameDelay<T>{}
+
+impl<T> FrameDelay<T> {
+    /// Constructor. Takes the number of render frames and the closure that should be called after
+    /// the render frames have been called.
+    pub fn new(frames:u64, target:T) -> Self {
+        let frames_left = Rc::new(Cell::new(frames));
+        let on_frame    = None;
+        FrameDelay{frames_left,target,on_frame}
+    }
+
+    /// Return a reference to the `on_frame` Closure. Creates it if it has not been created.
+    fn on_frame_action_init_if_uninit(&mut self, waker:&Waker) -> &OnFrameClosure {
+        if self.on_frame.is_none() {
+            let waker       = waker.clone();
+            let frames_left = Rc::clone(&self.frames_left);
+            let closure     = Closure::new(move |_:f64| {
+                let decreased = frames_left.get().saturating_sub(1);
+                frames_left.set(decreased);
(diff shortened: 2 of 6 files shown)


src/js/lib/content/src/index.html
```diff
@@ -72,6 +72,11 @@
                 border-radius: 14px;
                 box-shadow: 0 0 10px hsla(0, 0%, 0%, 0.1);
             }
+
+            #loader {
+                /* FIXME[mm]: This should be based on the theme colors. */
```
farmaazon, line 77: I saw the issue for using styles in css, can you put its id here?

src/rust/ide/src/executor/delay.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Internal closure that is shared with the JS runtime and updates the `frames_left`.
+    on_frame    : RefCell<Option<OnFrameClosure>>,
+    /// Target callback that is called after the delay has passed.
+    target      : T,
+}
+
+impl<T> FrameDelay<T> {
+    /// Constructor. Takes the number of render frames and the closure that should be called after
+    /// the render frames have been called.
+    pub fn new(n:u64, target:T) -> Self {
```
farmaazon, line 42: as this is public function, I would rename `n` to something like `frames`.

src/rust/ide/src/executor/delay.rs
```diff
@@ -0,0 +1,82 @@
...
+// ===================
+// === Frame Delay ===
+// ===================
+
+
+/// Delayed calls to a closure. The time delays is specified in number of render frames.
+#[derive(Debug)]
+pub struct FrameDelay<T>{
+    /// Counter that keeps track of the passage of frames. Needs to be shared with the JS runtime.
+    frames_left : Rc<RefCell<u64>>,
```
farmaazon, line 32: It could be `Cell`, not `RefCell`.

src/rust/ide/src/executor/delay.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Internal closure that is shared with the JS runtime and updates the `frames_left`.
+    on_frame    : RefCell<Option<OnFrameClosure>>,
```
farmaazon, line 34: Why we need to have RefCell here?

I think this structure can happily implement `Unpin` and thus `on_frame_action_init_if_uninit` be implemented on `&mut self`.

src/rust/ide/src/executor/delay.rs
```diff
@@ -0,0 +1,82 @@
...
+impl<T> FrameDelay<T> {
+    /// Constructor. Takes the number of render frames and the closure that should be called after
+    /// the render frames have been called.
+    pub fn new(n:u64, target:T) -> Self {
+        let frames_left = Rc::new(RefCell::new(n));
+        let on_frame    = default();
+        FrameDelay{frames_left,target,on_frame}
+    }
+
+    fn on_frame_action_init_if_uninit(&self, waker: &Waker) {
```
farmaazon, line 48:
```suggestion
    fn on_frame_action_init_if_uninit(&self, waker:&Waker) {
```

src/rust/ide/src/executor/delay.rs
```diff
@@ -0,0 +1,82 @@
...
+                let decreased = frames_left.borrow().saturating_sub(1);
+                frames_left.replace(decreased);
+                waker.clone().wake();
+            });
+            self.on_frame.set(closure);
+        }
+    }
+
+    /// Set the Js callback to be called on the next frame. Needs to be called on every frame.
+    fn set_on_frame_action(&self, waker: &Waker) {
```
farmaazon, line 62:
```suggestion
    fn set_on_frame_action(&self, waker:&Waker) {
```

src/rust/ide/src/executor/delay.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    fn on_frame_action_init_if_uninit(&self, waker: &Waker) {
+        if self.on_frame.borrow().is_none() {
+            let waker       = waker.clone();
+            let frames_left = Rc::clone(&self.frames_left);
+            let closure     = Closure::new(move |_:f64| {
+                let decreased = frames_left.borrow().saturating_sub(1);
+                frames_left.replace(decreased);
+                waker.clone().wake();
```
farmaazon, line 55: I'd like to have waker actually called when frame_left is equal to 0, otherwise this closure could reregister itself by request_animation_frame. Then we relieve executor from rescheduling this future on each frame.

farmaazon (changes requested):

src/rust/ide/src/executor/delay.rs
farmaazon: After discussion on PM we decided to leave it as it is - the complexity of my proposal is not adequate to the profit.

src/rust/ide/src/executor/delay.rs
```diff
@@ -0,0 +1,84 @@
...
+            self.on_frame = Some(closure);
+        }
+    }
+
+    /// Set the Js callback to be called on the next frame. Needs to be called on every frame.
+    fn set_on_frame_action(&mut self, waker:&Waker) {
+        self.on_frame_action_init_if_uninit(waker);
+        let maybe_closure = &self.on_frame;
+        // Panic Note: initialisation cannot fail, so this will never panic.
+        let closure       = maybe_closure.as_ref().expect("`on_frame` was not initialised.");
```
farmaazon, line 68: A small note: we could make on_frame_action_init_if_uninit returning the reference to closure - newly created or old one. Then the constraint will be forced by the code, not only our agreement that "initialisation cannot fail".

farmaazon (approved):

src/rust/ide/src/executor/delay.rs
MichaelMauderer: You are right that is the nicer design. I've changed it.

wdanilo:
This will not be merged anytime soon. Closing to clean the list of PRs.

PR #922, opened 2020-11-09
Clear detached edges when entering/exiting nodes.

MichaelMauderer:
### Pull Request Description
Detached edges are not cleaned up when entering/exiting nodes. This PR fixes that. 

### Checklist

### DIFF
--- a/src/rust/ide/view/graph-editor/src/lib.rs
+++ b/src/rust/ide/view/graph-editor/src/lib.rs
@@ -1342,12 +1342,13 @@ impl GraphEditorModel {
         edges
     }
 
-    pub fn clear_all_detached_edges(&self) {
-        let edges = self.edges.detached_source.mem_take();
-        edges.iter().for_each(|edge| {self.edges.all.remove(edge);});
-        let edges = self.edges.detached_target.mem_take();
-        edges.iter().for_each(|edge| {self.edges.all.remove(edge);});
+    pub fn clear_all_detached_edges(&self) -> Vec<EdgeId>{
+        let source_edges = self.edges.detached_source.mem_take();
+        source_edges.iter().for_each(|edge| {self.edges.all.remove(edge);});
+        let target_edges = self.edges.detached_target.mem_take();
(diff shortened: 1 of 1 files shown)


farmaazon (commented):
We call `clear_all_detached_edges` on model, but I wonder if the proper `edge_removed` events will be emitted? I think they should. Now apparently nobody listen to them, but one day someone could - and will be surprised that there are some cases where created edge disappears without `edge_removed` signal.

MichaelMauderer:
@farmaazon Good point! I've added the events to indicate the removed edge(s).

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2396,7 +2397,11 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
 
     node_to_enter           <= inputs.enter_selected_node.map(f_!(model.last_selected_node()));
     out.source.node_entered <+ node_to_enter;
+    removed_edge            <= out.node_entered.map(f_!(model.model.clear_all_detached_edges()));
+    out.source.edge_removed <+ removed_edge;
     out.source.node_exited  <+ inputs.exit_node;
+    removed_edge            <= out.node_exited.map(f_!(model.model.clear_all_detached_edges()));
+    out.source.edge_removed <+ removed_edge;
```
wdanilo, line 2404: In FRP, always try to merge the flows as early as possible. thus instead of attaching two separate streams to `out...` its always better to design the streams as:
```rust
removed_edge_on_enter <- ...
removed_edge_on_exit   <- ...
removed_edge <- any(&removed_edge_on_enter,&removed_edge_on_exit);
out.source.edge_removed <+ removed_edge;
```

We are not always doing it in graph editor, as the whole FRP definition is huge and is meant ot be split to several functions, but if you are working locally - like only on edge connection, deteach etc, you should design it always to mergethe streams as early as possible, please :)

wdanilo (approved):

farmaazon (approved):

PR #943, opened 2020-11-10
Quicktake: Change visualizations theme while switching IDE's theme.

BinarySoftware:
### Pull Request Description
As i thought, the easiest and best way to do this is by simply adding `dark` class to the `root` object when switching theme. All visualizations on main currently support this method, hence I implemented adding dark to root and removing in when switching back to the light side.

### Checklist


src/rust/ide/view/src/project.rs
```diff
@@ -57,8 +58,14 @@ impl Model {
...
-        if is_light { self.app.themes.set_enabled(&["dark"])  }
-        else        { self.app.themes.set_enabled(&["light"]) }
+        let root_elem = web::get_element_by_id("root").unwrap_or_else(|_| panic!("Failed to find root!"));
+        if is_light {
+            self.app.themes.set_enabled(&["dark"]);
+            root_elem.set_class_name("dark");
+        } else {
+            self.app.themes.set_enabled(&["light"]);
+            root_elem.set_class_name("");
+        }
```
wdanilo, line 68: 1. I don't like setting `.dark` ondark theme and nothing on light theme. Lets make it uniform. Lets set `.light` on light theme as well.

2. I'm not sure I like `.dark` and `.like` names. Maybe letsm ake them a little bit more explicit - what about `.dark-theme` and `.light-theme`? 

3. This does not work in `debug/interface` - everything should. Please check it always there (!)

4. Copy-pasting this code to `debug/interface` is error prone, thus create functions for it: `self.app.themes.set_dark_enabled();`etc

BinarySoftware: 4 is not necessary and it works both in IDE and in interface.

src/rust/ide/view/src/project.rs
```diff
@@ -56,9 +57,16 @@ impl Model {
         Self{app,logger,display_object,graph_editor,searcher,code_editor}
     }
 
-    fn set_style(&self, is_light:bool) {
-        if is_light { self.app.themes.set_enabled(&["dark"])  }
-        else        { self.app.themes.set_enabled(&["light"]) }
+    /// Sets style of IDE.
+    pub fn set_style(&self, is_light:bool) {
+        let root_elem = web::get_element_by_id("root").unwrap_or_else(|_| panic!("Failed to find root!"));
```
wdanilo, line 62: dont panic if you dont have to. GUI should be forgiving. Log instead.

src/rust/ide/view/src/project.rs
```diff
@@ -167,6 +175,7 @@ impl View {
         let graph                      = &model.graph_editor.frp;
         let network                    = &frp.network;
         let searcher_left_top_position = Animation::<Vector2<f32>>::new(network);
+        model.set_style(false);
```
wdanilo, line 178: the api is cryptic. I know it was before, but lets make it better. Lets do `model.set_style(theme::builtin::Light)` instead. The `theme::builtin` module will be available after merging my PR today night.

src/rust/ide/view/src/debug_scenes/interface.rs
```diff
@@ -83,7 +83,7 @@ impl DummyTypeGenerator {
 fn init(app:&Application) {
 
     ensogl_theme::dark::setup(&app);
-    // ensogl_theme::light::setup(&app);
+    ensogl_theme::light::setup(&app);
```
wdanilo, line 86: I understand now, that after commenting the light line, the `interface.rs` does not work, because dark classes in CSS would not register? If Im right, please add here the correct API usage for switching themes.

BinarySoftware: Light one is the default theme and as far as i remember only default theme generates "vars" aside from theme data. Thus it wouldnt work properly, it needs to be the first one, and then the dark one should follow it.

wdanilo:
@BinarySoftware My PR is merged, you can now fix `model.set_style(false);`.

src/rust/ensogl/lib/theme/src/lib.rs
```diff
@@ -94,6 +94,12 @@ macro_rules! define_theme {
                 app.themes.register(stringify!($name),$name);
                 app.themes.set_enabled(&[stringify!($name)]);
             }
+
+            /// Enables current theme.
+            pub fn enable(app:&Application) {
+                let mut $name = theme::Theme::new();
+                app.themes.set_enabled(&[stringify!($name)]);
+            }
```
wdanilo, line 102: This code is strange. First, it creates a mutable variable, then it drops it and runs a function with a string name. What is happening here?

src/rust/ide/view/src/project.rs
```diff
@@ -56,9 +57,30 @@ impl Model {
...
+                Err(_) => self.logger.warning("Failed to find root!"),
+            }
+        } else {
+            ensogl_theme::builtin::light::enable(&self.app);
+            match root_elem {
+                Ok(v)  => v.set_class_name("light-theme"),
+                Err(_) => self.logger.warning("Failed to find root!"),
+            }
+        }
+    }
```
wdanilo, line 76: 1. This is a public function with a suffix "internal".
2. This function takes bool, but neither its name nor the docs suggest what it is about.
3. A lot of repetitive code there, can be refactored.
4. Errors should not use exclamation marks.

src/rust/ide/view/src/project.rs
```diff
@@ -56,9 +57,30 @@ impl Model {
...
+
+    /// Sets style of IDE by theme name.
+    pub fn set_style(&self, theme:&'static str) {
+        match theme {
+            stringify!(ensogl_theme::builtin::dark) => self.set_style_internal(true),
+            &_                                      => self.set_style_internal(false),
+        }
```
wdanilo, line 83: This function is error prone and confusing. Why when calling `set_style("foo_bar_baz")` the theme will be switched to light one ? Please, use the solution I provided in my previous review instead.

src/rust/ide/view/src/project.rs
```diff
@@ -57,8 +58,34 @@ impl Model {
     }
 
     fn set_style(&self, is_light:bool) {
-        if is_light { self.app.themes.set_enabled(&["dark"])  }
-        else        { self.app.themes.set_enabled(&["light"]) }
+        if is_light { self.set_dark_style()  }
+        else        { self.set_light_style() }
+    }
```
wdanilo, line 63: What do oyu need this function for if you have now the enum `theme::Theme`? Just merge this function with `set_style_by_theme` which has a much better API than by providing a "bool"

src/rust/ide/view/src/project.rs
```diff
@@ -57,8 +58,34 @@ impl Model {
...
+    fn set_dark_style(&self) {
+        ensogl_theme::builtin::dark::enable(&self.app);
+        self.set_html_style("dark-theme");
+    }
+
+    fn set_html_style(&self, style:&'static str) {
+        let root_elem = web::get_element_by_id("root");
+        match root_elem {
+            Ok(v)  => v.set_class_name(style),
+            Err(_) => self.logger.warning("Failed to find root."),
```
wdanilo, line 87: Write it this way please:
```rust
web::get_element_by_id(&self.logger,"root",|root| root.set_class_name(style));
```

The function probably doesnt exist in `web` but it will be crazy helpful in many places, so add it there pls.

src/rust/ensogl/lib/theme/src/lib.rs
```diff
@@ -94,10 +94,25 @@ macro_rules! define_theme {
                 app.themes.register(stringify!($name),$name);
                 app.themes.set_enabled(&[stringify!($name)]);
             }
+
+            /// Enables current theme.
+            pub fn enable(app:&Application) {
+                let $name = theme::Theme::new();
+                app.themes.set_enabled(&[stringify!($name)]);
+            }
```
wdanilo, line 102: My previous comment is not fully applied here. But let the linter do the job instead of me here - please check why linter is not checking this code at all. It should gave error on unused `mut` earlier, but did not (CI was happy). It sohuld give warning now too.

wdanilo (changes requested):

src/rust/ide/view/src/project.rs
```diff
@@ -56,9 +57,26 @@ impl Model {
...
+        self.set_html_style("light-theme");
+    }
+
+    fn set_dark_style(&self) {
+        ensogl_theme::builtin::dark::enable(&self.app);
+        self.set_html_style("dark-theme");
+    }
+
+    fn set_html_style(&self, style:&'static str) {
+        web::run_on_element_by_id(&self.logger,"root",|root| root.set_class_name(style));
```
wdanilo, line 79: change the name to `with_element_by_id_or_warn` so it will match all names we use in our libs

src/rust/ide/view/src/project.rs
```diff
@@ -252,7 +271,10 @@ impl View {
             style_press            <- style_toggle_ev.gate_not(&style_was_pressed);
             style_press_on_off     <- style_press.map2(&frp.style_light, |_,is_light| !is_light);
             frp.source.style_light <+ style_press_on_off;
-            eval frp.style_light ((is_light) model.set_style(*is_light));
+            eval frp.style_light ((is_light) model.set_style({
+            if *is_light { ensogl_theme::Dark }
+            else { ensogl_theme::Light }
+            }));
```
wdanilo, line 277: indentation

src/rust/ide/view/src/project.rs
```diff
@@ -252,7 +271,10 @@ impl View {
             style_press            <- style_toggle_ev.gate_not(&style_was_pressed);
             style_press_on_off     <- style_press.map2(&frp.style_light, |_,is_light| !is_light);
             frp.source.style_light <+ style_press_on_off;
-            eval frp.style_light ((is_light) model.set_style(*is_light));
+            eval frp.style_light ((is_light) model.set_style({
```
wdanilo, line 274: unnecesary `{` and `}`

src/rust/ide/view/src/project.rs
```diff
@@ -252,7 +271,10 @@ impl View {
...
+            if *is_light { ensogl_theme::Dark }
```
wdanilo, line 275: this line doesnt make any sense.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -245,6 +245,14 @@ pub fn get_element_by_id(id:&str) -> Result<Element> {
         Error(format!("Element with id '{}' not found.",id)))
 }
 
+pub fn run_on_element_by_id<F>(logger:&Logger, id:&str, f:F) where F : FnOnce(Element){
+    let root_elem = get_element_by_id(id);
+    match root_elem {
+        Ok(v)  => f(v),
+        Err(_) => logger.warning("Failed to find root."),
```
wdanilo, line 252: the warning is incorrect.

src/rust/lib/system/web/src/lib.rs
```diff
@@ -245,6 +245,14 @@ pub fn get_element_by_id(id:&str) -> Result<Element> {
         Error(format!("Element with id '{}' not found.",id)))
 }
 
+pub fn run_on_element_by_id<F>(logger:&Logger, id:&str, f:F) where F : FnOnce(Element){
```
wdanilo, line 248: missing docs

wdanilo (changes requested):

src/rust/ide/view/src/project.rs
```diff
@@ -246,13 +266,16 @@ impl View {
...
+            let style_toggle_ev   = frp.toggle_style.clone_ref();
+            style_pressed        <- style_toggle_ev.toggle() ;
+            style_was_pressed    <- style_pressed.previous();
+            style_press          <- style_toggle_ev.gate_not(&style_was_pressed);
+            style_press_on_off   <- style_press.map2(&frp.style, |_,s| match s {
+                Theme::Light => Theme::Dark ,
+                _            => Theme::Light,
+            });
+            frp.source.style     <+ style_press_on_off;
+            eval frp.style ((style) model.set_style(style.clone()));
```
wdanilo, line 278: You dont need clone here. Pass as a reference, will be faster.

wdanilo (approved):
Good job!

wdanilo (commented):
I accepted it 4 days ago, why it's not merged? Anyway, as we entered a stage when we need to be 100% sure no regressions are made on master, @BinarySoftware please also test this thing using our scenarios and ask @farmaazon to confirm its working correctly before merging.

farmaazon:
Well I can confirm that there aren't any regressions. However I haven't found the switching themes working on linux. But it had not worked anyway, so this PR can be merged.

PR #949, opened 2020-11-13
Bump parser - Copy button & fix indentation in Code Block

BinarySoftware:
### Checklist


sylwiabr:
@BinarySoftware Mark the checked parts please in the PR description

wdanilo (approved):

src/js/lib/content/src/index.html
```diff
@@ -7,7 +7,7 @@
         <!-- FIXME Security Vulnerabilities: https://github.com/luna/ide/issues/226 -->
         <meta http-equiv="Content-Security-Policy" content="
             default-src 'self';
-            script-src  'self' 'unsafe-eval' data: https://*;
+            script-src  'self' 'unsafe-eval' 'unsafe-inline' data: https://*;
```
wdanilo, line 10: This is bad. Can we do it another way around?

wdanilo (changes requested):

src/js/lib/content/src/index.html
BinarySoftware: Not so much, this inline expression is constructed in Engine because it needs to know ID of the code block to copy :/ hence it is a inline expression.

wdanilo: We need to do it safe

BinarySoftware: It'll be done safely :)

src/rust/ide/lib/parser/build.rs
```diff
@@ -25,7 +25,9 @@ use std::path::PathBuf;
 const PARSER_PATH: &str = "./pkg/scala-parser.js";
 
 /// Commit from `enso` repository that will be used to obtain parser from.
-const PARSER_COMMIT: &str = "890b800513a5b35376fc4f52beadb0dd869d3e70";
+/// Important note: Using commit not from Engine's master but from a backported branch
+///                 as the AST in this project hasn't been updated to conform to latest scala parser.
+const PARSER_COMMIT: &str = "6e0777535847adfe8d47a6f5196e7c0cdd32083c";
```
wdanilo, line 30: What does it mean ? Can we / should we leave it this way @farmaazon ?

wdanilo (changes requested):

wdanilo:
Maciek, this is a blocker for the merge, you need to test it:
image

src/rust/ide/lib/parser/build.rs
farmaazon: To use newest parser from engine's main we should implement https://github.com/enso-org/ide/issues/732 which can be quite time-consuming.

I would only add here a issue number + `Engine's main` instead of `Engine's master`.

wdanilo:
@BinarySoftware Is this PR working? If not, it should be a draft PR. What's the plan for this PR?

src/js/lib/content/src/index.html
BinarySoftware: Done, even better - moved the js-part to the IDE so that it is loaded only once, and doesnt need to be sent with the HTML file, so the doc html itself gets much much shorter!

src/js/lib/content/src/index.js
```diff
@@ -121,6 +121,22 @@ function show_debug_screen(wasm,msg) {
...
+// =======================================
+// === Copy Documentation Example Code ===
+// =======================================
+
+function copyCode(codeBlockElem) {
+    let range = document.createRange()
+    range.selectNode(codeBlockElem)
+    window.getSelection().removeAllRanges()
+    window.getSelection().addRange(range)
+    document.execCommand('copy')
```
wdanilo, line 134: This has a few problems:
1. It is synchronious, so when used for the first time, it can display a message that a script wants to access the clipboard and it will freeze the whole GUI until someone clicks "allow" or "deny".
2. We have developed good copy handling in Rust already, which uses asynchronious callbacks and fall backs to `execCommand` in case these are not supported. Please use these instead.

src/rust/ide/view/src/documentation.rs
```diff
@@ -148,7 +160,37 @@ impl ViewModel {
...
+                        let callback = closure.as_ref().unchecked_ref();
+                        match cpy_btn.unwrap().add_event_listener_with_callback("click", callback){
+                            Ok(_)    => (),
+                            Err(err) => error!(&self.logger, "Unable to add event listener: {err:?}")
+                        }
+                        closure.forget();
+                    },
+                    None => info!(&self.logger, "No code block."),
+                },
+                None => info!(&self.logger, "No copy button."),
```
wdanilo, line 191: This info is not descriptive at all. If I see a message in a console "No code block" I would not understand it at all. Moreover, should it be "info"? Is it something you want the user to be informed about? Isnt this a debug information or a warning?

src/rust/ide/view/src/documentation.rs
```diff
@@ -148,7 +160,37 @@ impl ViewModel {
...
+        for i in 0..copy_buttons.length() {
+            let copy_button = copy_buttons.get_with_index(i);
+            let code_block  = code_blocks.get_with_index(i);
+            match copy_button {
+                Some(btn) => match code_block {
+                    Some(code) => {
+                        let cpy_btn  = btn.dyn_into::<HtmlElement>();
+                        let code_blk = code.dyn_into::<HtmlElement>();
+                        let closure  = move |_event: MouseEvent| {
+                            copyCode(&code_blk.unwrap());
```
wdanilo, line 179: unsafe unwrap, it can crash the app

src/rust/ide/view/src/documentation.rs
```diff
@@ -148,7 +160,37 @@ impl ViewModel {
...
+                        };
+                        let closure  = Closure::wrap(Box::new(closure).into_fn_mut());
+                        let callback = closure.as_ref().unchecked_ref();
+                        match cpy_btn.unwrap().add_event_listener_with_callback("click", callback){
+                            Ok(_)    => (),
+                            Err(err) => error!(&self.logger, "Unable to add event listener: {err:?}")
+                        }
+                        closure.forget();
```
wdanilo, line 187: This is leaking memory. If you use this multiple times, you will leak a lot of memory here. I was asking during the call to be super caucius regarding memory leaks here.

wdanilo (changes requested):
My biggest concerns are memory leaks here.

src/rust/ide/view/src/documentation.rs
```diff
@@ -148,7 +155,32 @@ impl ViewModel {
...
+                    let closure  = Closure::wrap(Box::new(closure).into_fn_mut());
+                    let callback = closure.as_ref().unchecked_ref();
+                    match cpy_btn.add_event_listener_with_callback("click", callback){
+                        Ok(_)    => (),
+                        Err(err) => error!(&self.logger, "Unable to add event listener: {err:?}")
+                    }
+                    closure.forget();
+                }) });
+            }) });
+        }
```
farmaazon, line 183: 1. The ViewModel structure should have a field keeping a vector of Closures. Then in this function we should gather all the closures into this instead of forgetting them, and replace old vector in structure with a new one. Therefore I would also rename this function to `set_listeners_to_copy_buttons`.
2. The loop could be much simplified and improved:
  * Instead of plain loop, lets map each element to option of closure `(0..copy_buttons.length()).iter().map(|i| ...`
  * Then, instead of nested option::map calls you could make use of `?`: 
    ```rust
    let copy_button = copy_buttons.get_with_index(i)?.dyn_into::<HtmlElement>()?;
    let code_block  = code_blocks.get_with_index(i)?.dyn_into::<HtmlElement>()?;
    ...
    ```
* After all we could call filter_map and log all Nones - as having None is not an usual situation, and I want to have an 
  error log for that.

farmaazon (changes requested):

src/rust/ide/view/src/documentation.rs
```diff
@@ -162,25 +159,33 @@ impl ViewModel {
...
-                        clipboard::write_text(inner_code);
-                    };
-                    let closure  = Closure::wrap(Box::new(closure).into_fn_mut());
-                    let callback = closure.as_ref().unchecked_ref();
-                    match cpy_btn.add_event_listener_with_callback("click", callback){
-                        Ok(_)    => (),
-                        Err(err) => error!(&self.logger, "Unable to add event listener: {err:?}")
+        (0..copy_buttons.length()).map(|i| -> Option<Closure<dyn std::ops::FnMut(web_sys::MouseEvent)>>{
+            let copy_button = copy_buttons.get_with_index(i)?;//.dyn_into::<HtmlElement>()?;
+            let code_block  = code_blocks.get_with_index(i)?;//.dyn_into::<HtmlElement>()?;
```
farmaazon:
```suggestion
            let copy_button = copy_buttons.get_with_index(i)?.dyn_into::<HtmlElement>().ok()?;
            let code_block  = code_blocks.get_with_index(i)?.dyn_into::<HtmlElement>().ok()?;
```

BinarySoftware: I've already made it, you must've reviewed some old commit haha

src/rust/ide/view/src/documentation.rs
```diff
@@ -56,6 +63,9 @@ enum InputFormat {
 
 pub use visualization::container::overlay;
 
+#[allow(unused_qualifications)]
```
wdanilo, line 66: why?

src/rust/ide/view/src/documentation.rs
```diff
@@ -68,6 +78,7 @@ pub struct ViewModel {
     /// to EnsoGL shapes, and pass them to the DOM instead.
     overlay        : component::ShapeView<overlay::Shape>,
     display_object : display::object::Instance,
+    closures       : Rc<CloneCell<Vec<OptClosure>>>
```
wdanilo, line 81: 1. "closures" is like naming a variable "var"
2. `CloneCell<Vec<...>>` has a very high performance cost. It is clonning the whole vector on each access. I don't believe you want this.

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,42 @@ impl ViewModel {
...
+                clipboard::write_text(inner_code);
+            };
+            let closure  = Closure::wrap(Box::new(closure).into_fn_mut());
+            let callback = closure.as_ref().unchecked_ref();
+            match copy_button.add_event_listener_with_callback("click", callback){
+                Ok(_)    => (),
+                Err(err) => error!(&self.logger, "Unable to add event listener: {err:?}")
+            }
+            Some(closure)
+        }).collect::<Vec<OptClosure>>();
```
wdanilo, line 185: Why you are collecting opt closures instead of collecting only closures and skipping all Nones? I don't think that collecting Nones is good idea here.

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,42 @@ impl ViewModel {
...
+            println!("{:?}",&code_block.inner_text());
+            let closure  = move |_event: MouseEvent| {
+                let inner_code = code_block.inner_text();
+                clipboard::write_text(inner_code);
+            };
+            let closure  = Closure::wrap(Box::new(closure).into_fn_mut());
+            let callback = closure.as_ref().unchecked_ref();
+            match copy_button.add_event_listener_with_callback("click", callback){
+                Ok(_)    => (),
+                Err(err) => error!(&self.logger, "Unable to add event listener: {err:?}")
```
wdanilo, line 182: this is not a sentence. Moreover, the error is not very self-explanatory, it should describe more what operation it was trying to do and why.

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,42 @@ impl ViewModel {
...
+    }
+
+    /// Append listeners to copy buttons in doc to enable copying examples.
+    fn set_listeners_to_copy_buttons(&self) {
+        let code_blocks  = self.dom.dom().get_elements_by_class_name("CodeBlock");
+        let copy_buttons = self.dom.dom().get_elements_by_class_name("copyCodeBtn");
+        let closures     = (0..copy_buttons.length()).map(|i| -> OptClosure {
+            let copy_button = copy_buttons.get_with_index(i)?.dyn_into::<HtmlElement>().ok()?;
+            let code_block  = code_blocks.get_with_index(i)?.dyn_into::<HtmlElement>().ok()?;
+            println!("{:?}",&code_block.inner_text());
```
wdanilo, line 173: This code is not cleaned

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,42 @@ impl ViewModel {
...
+        self.dom.dom().set_inner_html(&data_str);
+    }
+
+    /// Append listeners to copy buttons in doc to enable copying examples.
+    fn set_listeners_to_copy_buttons(&self) {
+        let code_blocks  = self.dom.dom().get_elements_by_class_name("CodeBlock");
+        let copy_buttons = self.dom.dom().get_elements_by_class_name("copyCodeBtn");
+        let closures     = (0..copy_buttons.length()).map(|i| -> OptClosure {
+            let copy_button = copy_buttons.get_with_index(i)?.dyn_into::<HtmlElement>().ok()?;
+            let code_block  = code_blocks.get_with_index(i)?.dyn_into::<HtmlElement>().ok()?;
```
wdanilo, line 172: there is no error handling here. Do we consider a situation when we have a code block but no copy button ok? I believe that this should be reported as an error, otherwise, we will have some day documentation not working without any errors nor warnings and we will debug it for a long tie before we will find these magical question marks here.

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,42 @@ impl ViewModel {
...
     fn push_to_dom(&self, content:String) {
         let data_str = format!(r#"<div class="docVis">{}{}</div>"#,documentation_style(),content);
-        self.dom.dom().set_inner_html(&data_str)
+        self.dom.dom().set_inner_html(&data_str);
+    }
+
+    /// Append listeners to copy buttons in doc to enable copying examples.
+    fn set_listeners_to_copy_buttons(&self) {
+        let code_blocks  = self.dom.dom().get_elements_by_class_name("CodeBlock");
+        let copy_buttons = self.dom.dom().get_elements_by_class_name("copyCodeBtn");
```
wdanilo, line 169: are these classes used anywhere else? If so, we have string literal duplciation

wdanilo, line 169: Are you sure this iterates both things in order? Please paste some docs in the docs here that prove that it happens in all browsers like that. We need to be sure about it.

src/rust/ide/view/src/documentation.rs
wdanilo, line 81: Why the `ViewModel` needs to be `CloneRef`? I did not have problems with that earlier, but now, if we clone ref it, we can easily leak accidenatally the closures, so I'm wondering why it needs to be clonable at all.

src/rust/ide/view/src/documentation.rs
wdanilo, line 182: instead of matching, there are functions to handle err outputs in the stdlib. Use them, code will be nicer. Basically, a match that does nothing (like your Ok match) is a bad code design

src/rust/ide/view/src/documentation.rs
wdanilo: Could you also explain e the live-time of the ViewModel? When it is created, who owns it and its copies, and when it will be deleted, so I could better understand how long the closures will live here? Also, add this info to the docs.

wdanilo (changes requested):

src/rust/ide/view/src/documentation.rs
farmaazon: > 2. `CloneCell<Vec<...>>` has a very high performance cost. It is clonning the whole vector on each access. I don't believe you want this.

Yes, we want. Because we never want to read this vector, only to set.

farmaazon: > Why the `ViewModel` needs to be `CloneRef`? I did not have problems with that earlier, but now, if we clone ref it, we can easily leak accidenatally the closures, so I'm wondering why it needs to be clonable at all.

The Clone-refable model used in FRP was one of the most common patterns in GUI. And yes, it caused a lot of problems, (for example not-disappearing nodes on the graph scene). But we need clone-refable structures in current FRP implementation

farmaazon: > Could you also explain e the live-time of the ViewModel? When it is created, who owns it and its copies, and when it will be deleted, so I could better understand how long the closures will live here? Also, add this info to the docs.

TBH I'm surprised with this requirement. NONE of our views' `Model`s have such detailed description and they may leak as well. Are we going to extend docs in all of the models?

src/rust/ide/view/src/documentation.rs
farmaazon: As I wrote in my previous review, after mapping to the Option<Closure> we should do the filter_map with logging error if element is None.

src/rust/ide/view/src/documentation.rs
wdanilo: > The Clone-refable model used in FRP was one of the most common patterns in GUI. And yes, it caused a lot of problems, (for example not-disappearing nodes on the graph scene)

I know you made a big mental shortcut here (based on our phone call just a minute ago), so to explain to readers, clone-refable models never caused any problems in GUI. What caused the problem was clone-refable FRP networks, which we are slowly improving. Clone-refable models are rather safe, but they can be made even safer if we would allow FRP networks to manage their lifetimes in the future.

> But we need clone-refable structures in current FRP implementation
In other places we are using just the name `Model`, not `ViewModel`, this made confusion in my head what kind of model is this. So to keep it consistent, @BinarySoftware rename it to `Model`, please

> TBH I'm surprised with this requirement. 
100% agree. If these are just normal FRP models, I'm ok with it. The only part that brought confusion was the `Viewodel` name here.

src/rust/ide/view/src/documentation.rs
BinarySoftware: https://stackoverflow.com/questions/35525811/order-of-elements-in-document-getelementsbyclassname-array
Yes it does. Top -> down.

src/rust/ide/view/src/documentation.rs
BinarySoftware: Already explained by Adam.

src/rust/ide/view/src/documentation.rs
BinarySoftware: No they're not, only in Engine

src/rust/ide/view/src/documentation.rs
BinarySoftware: tbh this is only to surpass linter inadeequate warning as the type signatuure there is really necessary ...

wdanilo: Almost impossible. Please investigate that before merge and write to me what you found out with details. This should not be used. If this is really needed, create bug report on Rust linter repo and link it here in a FIXME comment

src/rust/ide/view/src/documentation.rs
```diff
@@ -56,21 +63,25 @@ enum InputFormat {
...
-pub struct ViewModel {
+pub struct Model {
     logger : Logger,
     dom    : DomSymbol,
     size   : Rc<Cell<Vector2>>,
     /// The purpose of this overlay is stop propagating mouse events under the documentation panel
     /// to EnsoGL shapes, and pass them to the DOM instead.
     overlay        : component::ShapeView<overlay::Shape>,
     display_object : display::object::Instance,
+    closures       : Rc<CloneCell<Vec<ClickClosure>>>
```
wdanilo, line 81: the name "closures" is like naming a variable "variable". Names should semantically tell what fields are for.

src/rust/ide/view/src/documentation.rs
```diff
@@ -108,7 +119,11 @@ impl ViewModel {
         display_object.add_child(&dom);
         display_object.add_child(&overlay);
         scene.dom.layers.front.manage(&dom);
-        ViewModel {logger,dom,size,overlay,display_object}.init()
+
+        let closures: Vec<Closure<_>> = Vec::new();
+        let closures = CloneCell::new(closures);
+        let closures = Rc::new(closures);
```
wdanilo, line 125: not aligned `=` . Change the first line to `let closures = Vec::<Closure<_>>::new();`.

src/rust/ide/view/src/documentation.rs
wdanilo: So they are, or they are not? You have written that they are not and in the same sentence that they are. Moreover, why engine needs to know about classes in HTML docs?

wdanilo: As I've written in my comment, add docs to the code that prove this invariant.

src/rust/ide/view/src/documentation.rs
wdanilo: This is still not done.

wdanilo (changes requested):

src/rust/ide/view/src/documentation.rs
BinarySoftware: No they are not duplicated anywhere in this repository, by saying only in engine i mean that doc parser which is part of engine generates HTML with those classes. So the answer is No, they dont need to be refactored.

src/rust/ide/view/src/documentation.rs
```diff
@@ -108,7 +118,11 @@ impl ViewModel {
         display_object.add_child(&dom);
         display_object.add_child(&overlay);
         scene.dom.layers.front.manage(&dom);
-        ViewModel {logger,dom,size,overlay,display_object}.init()
+
+        let code_copy_closures = Vec::<Closure<_>>::new();
+        let code_copy_closures = CloneCell::new(code_copy_closures);
+        let code_copy_closures = Rc::new(code_copy_closures);
+        Model {logger,dom,size,overlay,display_object, code_copy_closures }.init()
```
wdanilo, line 125: spacing

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +147,49 @@ impl ViewModel {
...
                 InputFormat::Docstring => parser.generate_html_doc_pure(processed),
             };
             let output = output?;
-            if output.is_empty() {
-                Ok(PLACEHOLDER_STR.into())
-            } else {
-                Ok(output)
-            }
+            if output.is_empty() { Ok(PLACEHOLDER_STR.into()) }
+            else                 { Ok(output)                 }
```
wdanilo, line 155: why not `Ok(if output.is_empty() { PLACEHOLDER_STR.into() } else { output} )`? :)

src/rust/ide/view/src/documentation.rs
wdanilo: Please move these names to constants section on top of this file then.

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +147,49 @@ impl ViewModel {
...
         let data_str = format!(r#"<div class="docVis">{}{}</div>"#,documentation_style(),content);
-        self.dom.dom().set_inner_html(&data_str)
+        self.dom.dom().set_inner_html(&data_str);
+    }
+
+    /// Append listeners to copy buttons in doc to enable copying examples.
+    /// It is possible to do it with implemented method, because get_elements_by_class_name
+    /// returns top-to-bottom sorted list of elements, as found in:
+    /// https://stackoverflow.com/questions/35525811/order-of-elements-in-document-getelementsbyclassname-array
+    #[allow(unused_must_use)]
```
wdanilo, line 169: what this is needed for? Should not be needed

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +147,49 @@ impl ViewModel {
...
+            copy_button.add_event_listener_with_callback("click", callback).map_err(|err| {
+                error!(&self.logger, "Unable to add event listener to copy button: {err:?}")
+            });
+            Some(closure)
+        }).filter_map(|x| {
+            if x.is_none() {
+                error!(&self.logger, "Tried to add event listener to a non-existent copy button.")
+            }
+            x
+        }).collect::<Vec<CodeCopyClosure>>();
```
wdanilo, line 191: I don't like this solution a little bit because in case of many buttons it will just throw a bunch of the same errors. Instead, just print a single error telling that some buttons were not attached properly. You can also mention rthe indexes which failed. To do it just return `Result<usize,CodeCopyClosure>` from the lambda instead of `Option<CodeCopyClosure>`. Then filter it by all errors, collect them to vector and in case the vec is non-empty - print the error and mention all indexes.

Moreover, we try not doing such big centipede functions - just split them into more variables instead :D

src/rust/ide/view/src/documentation.rs
```diff
@@ -169,7 +210,7 @@ impl ViewModel {
     }
 
     fn display_doc(&self, content:&str, content_type: InputFormat) {
-        let html = match ViewModel::gen_html_from(content,content_type) {
+        let html = match Model::gen_html_from(content, content_type) {
```
wdanilo, line 213: rm spacing

src/rust/ide/view/src/documentation.rs
```diff
@@ -178,6 +219,7 @@ impl ViewModel {
         };
 
         self.push_to_dom(html);
+        self.set_listeners_to_copy_buttons();
```
wdanilo, line 222: `set_listeners_to_copy_buttons` -> `attach_listeners_to_copy_buttons`

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +149,54 @@ impl ViewModel {
...
+                        error!(&self.logger,"Unable to add event listener to copy button: {e:?}");
+                        None
+                    },
+                }
+            };
+            f().ok_or(i)
+        });
+        let (closures,errors) : (Vec<_>,Vec<_>) = closures.partition(Result::is_ok);
+        let closures = closures.into_iter().map(Result::unwrap).collect::<Vec<CodeCopyClosure>>();
+        let errors   = errors.into_iter().map(Result::unwrap_err).collect::<Vec<u32>>();
```
wdanilo, line 195: using `unwrap` and `unwrap_err` is not securte, as these ay panic. Of course they will not here, because the previous lines guarantees they are OK, but this is error prone - someone will refactor the previous line accidentaly and these will break. Do not use functions that may panic, do not use `unwrap*` until you really have to

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +149,54 @@ impl ViewModel {
...
+        if !errors.is_empty() {
+            error!(&self.logger, "Tried to add event listener to a non-existent copy button with indices: {errors:?}")
```
wdanilo, line 197: "copy button with indices" - can a single copy button have many indices? I nean, this sohuld be a bout copy buttonS right? Moreover, why indices vector is naed `errors` ? Also, this is not a sentence, it has no dot at the end.

src/rust/ide/view/src/documentation.rs
```diff
@@ -108,7 +119,11 @@ impl ViewModel {
         display_object.add_child(&dom);
         display_object.add_child(&overlay);
         scene.dom.layers.front.manage(&dom);
-        ViewModel {logger,dom,size,overlay,display_object}.init()
+
+        let code_copy_closures = Vec::<Closure<_>>::new();
+        let code_copy_closures = CloneCell::new(code_copy_closures);
+        let code_copy_closures = Rc::new(code_copy_closures);
```
wdanilo, line 125: Why not just `let code_copy_closures = default()` ?

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,54 @@ impl ViewModel {
...
+
+    /// Append listeners to copy buttons in doc to enable copying examples.
+    /// It is possible to do it with implemented method, because get_elements_by_class_name
+    /// returns top-to-bottom sorted list of elements, as found in:
+    /// https://stackoverflow.com/questions/35525811/order-of-elements-in-document-getelementsbyclassname-array
+    fn attach_listeners_to_copy_buttons(&self) {
+        let code_blocks  = self.dom.dom().get_elements_by_class_name(CODE_BLOCK_CLASS);
+        let copy_buttons = self.dom.dom().get_elements_by_class_name(COPY_BUTTON_CLASS);
+        let closures     = (0..copy_buttons.length()).map(|i| -> Result<CodeCopyClosure,u32> {
+            let f = || -> Option<CodeCopyClosure> {
```
wdanilo, line 173: `f` is the worst name ever.

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,54 @@ impl ViewModel {
...
+                        error!(&self.logger,"Unable to add event listener to copy button: {e:?}");
+                        None
+                    },
+                }
+            };
+            f().ok_or(i)
+        });
+        let (closures,errors) : (Vec<_>,Vec<_>) = closures.partition(Result::is_ok);
+        let ok_closures = closures.into_iter().filter_map(|t| t.ok()).collect::<Vec<CodeCopyClosure>>();
+        let err_indices = errors.into_iter().filter_map(|t| t.err()).collect::<Vec<u32>>();
```
wdanilo, line 194: dont provide so many types. Use `collect_vec` instead without explicit types here.

src/rust/ide/view/src/documentation.rs
```diff
@@ -133,22 +148,54 @@ impl ViewModel {
...
+                    Ok(_)  => Some(closure),
+                    Err(e) => {
+                        error!(&self.logger,"Unable to add event listener to copy button: {e:?}");
+                        None
+                    },
+                }
+            };
+            f().ok_or(i)
+        });
+        let (closures,errors) : (Vec<_>,Vec<_>) = closures.partition(Result::is_ok);
```
wdanilo, line 192: do you really need to provide types here?

src/rust/ide/view/src/documentation.rs
```diff
@@ -169,7 +216,7 @@ impl ViewModel {
     }
 
     fn display_doc(&self, content:&str, content_type: InputFormat) {
-        let html = match ViewModel::gen_html_from(content,content_type) {
+        let html = match Model::gen_html_from(content,content_type) {
             Ok(html)               => html,
             Err(err)               => {
```
wdanilo, line 221: indentation

wdanilo (approved):

wdanilo (approved):

src/rust/ide/view/src/documentation.rs
BinarySoftware: Yes it is necessary unfortunately

PR #973, opened 2020-11-19
Improving port highlight, fixing tab in searcher

wdanilo:
### Pull Request Description
- Implements [Connected port Highlight](https://github.com/enso-org/ide/issues/524).
- Partially implements (not for output ports yet) [The ports & connections are not colored according to type information](https://github.com/enso-org/ide/issues/765).
- Implements [After pressing <tab> to choose hint in searcher cursor disappear](https://github.com/enso-org/ide/issues/965)

### Important Notes
Moreover, this PR refactors a lot of FRP networks around edge connections and thus **it should be extensively tested** as regression possibility is high.

### Checklist


src/rust/ide/view/graph-editor/src/component/edge.rs
```diff
@@ -1207,19 +1205,22 @@ impl Edge {
...
             color_update <- all(input.set_color,input.set_dimmed);
 
-            eval color_update ([edge_color,style]((color,should_dim)) {
-                let target_color = if *should_dim {
-                   style.get_color_dim(*color)
-                } else {
-                  color.into()
-                };
-                edge_color.set_target(target_color);
+            // TODO: dimming commented for now, as it was using bad colors. Should be mixed with background instead.
```
farmaazon, line 1208: But that makes one node's button without any visual effect, doesn't it? @sylwiabr is that ok from UX perspective?

farmaazon (commented):
There is one question regarding UX

src/rust/ide/view/graph-editor/src/component/edge.rs
sylwiabr: So I will not see if the node is active or am I pressing visualization button? If so then no, it is not ok.

src/rust/ensogl/lib/core/src/application/frp.rs
```diff
@@ -256,22 +256,14 @@ macro_rules! define_endpoints {
             /// Create Frp endpoints within and the associated network.
             pub fn new() -> Self {
                 let network = $crate::frp::Network::new();
-                let input   = FrpInputs::new(&network);
-                let output  = FrpEndpoints::new(&network,input);
+                let output  = Self::extend(&network);
```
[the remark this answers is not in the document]
wdanilo: Yep, we do it everywhere like this. The outputs in FRP derefer to inputs and this is a strange design that I want to improve in the future. The current FRP struct generation has some bad names inside, but the problem is not with this particular code though.

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -665,6 +665,40 @@ impl<'a,T:Payload> RefMut<'a,T> {
     pub fn dfs_(self, mut on_node:impl FnMut(&mut Self)) {
         self.partial_dfs((),|t,_|(true,on_node(t)))
     }
+
+    /// Just like `partial_dfs` but traversing two `SpanTree`s at the same time.
```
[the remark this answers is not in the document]
wdanilo: added docs

src/rust/ide/lib/span-tree/src/node.rs
```diff
@@ -665,6 +665,40 @@ impl<'a,T:Payload> RefMut<'a,T> {
...
+    pub fn partial_double_dfs<D>
```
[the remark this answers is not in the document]
wdanilo: I've renamed it to `partial_zipped_dfs`, as it is using zipping under the hood and it feels like a better name to me. But Id be happy to discuss it more if you like!

src/rust/ide/view/graph-editor/src/component/edge.rs
wdanilo: fixed

src/rust/ensogl/lib/core/src/data/mix.rs
```diff
@@ -0,0 +1,102 @@
...
+
+// =============
+// === Space ===
+// =============
+
+/// Strongly typed value representation in the mix space.
+#[allow(missing_docs)]
+pub struct Space<T:Mixable> {
+    pub value : Repr<T>
+}
```
farmaazon, line 18: I this very case I would prefer `MixSpace` or `MixingSpace` - I'm not sure why, but - especially in this module - the `Space` alone does not look well. Perhaps the `Space` alone sounds to me like "some set of values" instead of single value in a concrete space.

src/rust/ide/view/src/debug_scenes/interface.rs
```diff
@@ -83,6 +83,7 @@ impl DummyTypeGenerator {
 
 fn init(app:&Application) {
 
+
```
farmaazon, line 86: Why we put here this line?
```suggestion
```

src/rust/lib/frp/src/nodes.rs
```diff
@@ -180,6 +180,15 @@ impl Network {
...
 
+    /// If the first input is false, the second input will be redirected to the output. Otherwise,
+    /// the third input will be redirected to the output. The redirection is persistent. The first
+    /// input doesn't have to fire to propagate the events fromm second and third input streams.
+    pub fn switch<T1,T2,T3,T>
+    (&self, label:Label, check:&T1, t2:&T2, t3:&T3) -> Stream<T>
+    where T1:EventOutput<Output=bool>, T2:EventOutput<Output=T>, T3:EventOutput<Output=T>, T:Data {
+        self.all_with3(label,check,t2,t3,|check,t1,t2| if *check {t2.clone()} else {t1.clone()})
+    }
+
```
farmaazon, line 191: The docs gave me a little suggestion, that the `check` change won't emit any signal from `switch`, because _redirection_ should not add any signal to signals received from input.

But the implementation states quite contrary: change on `check` will emit (again) signal from t1 or t2.

farmaazon (approved):

src/rust/ensogl/lib/core/src/data/mix.rs
wdanilo: But we are using this thing outside of the module as `mix::Space`, I don't want to expose it as `mix::MixSpace`. The name of the module matters.

src/rust/ide/view/src/debug_scenes/interface.rs
wdanilo: Accidental change

src/rust/lib/frp/src/nodes.rs
wdanilo: Good catch. Updated docs.

PR #981, opened 2020-11-23
Disable scene navigaton when a visualisation is active

MichaelMauderer:
### Pull Request Description
Disables the scene navigation when a visualization is active. This allows us to navigate the visualization without also navigating the scene simultaneously. 

Peek 2020-11-23 16-50

### Important Notes
Moves the `Navigator` into the `GraphEditor`.
The `Navigator` is enabled/disabled when the visualization is set active by clicking it. 

### Checklist


src/rust/ensogl/lib/core/src/display/camera/camera2d.rs
```diff
@@ -165,6 +165,10 @@ struct Camera2dData {
     dirty                  : Dirty,
     zoom_update_registry   : callback::Registry1<f32>,
     screen_update_registry : callback::Registry1<Vector2<f32>>,
+    /// Indicates whether this camera should stay in place.
+    /// Note that this is not enforced by the `Camera` itself, but should be respected by whoever
+    /// modifies the camera.
+    locked                 : bool,
```
wdanilo, line 171: I don't like this solution. Keeping a flag on camera which is not handled at all by the camera but needs to be handled by something outside, sounds like a bad responsibility design. The camera now keeps `locked` but it is not using this field for anything. This should definitely not be part of a camera implementation.

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -549,18 +550,24 @@ impl Container {
...
                         activate.emit(());
+                        return Some(true);
                     }
                 } else if let Some(deactivate) = deactivate() {
                     deactivate.emit(());
+                    return Some(false);
                 }
-            });
+                None
+            })).unwrap();
```
wdanilo, line 567: use `<=` instead fo `unwrap` here

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -549,18 +550,24 @@ impl Container {
...
+
+            is_selected_changed    <- selected.map2(&frp.source.is_selected, |&new,&old| if new != old {Some(new)} else {None}).unwrap();
```
wdanilo, line 569: 1. you should use `frp.source` only for setting values, never for reading them back
2. Line too long
3. Use `<=` instead of unwrap
4. Use `(new != old).as_some(new)`

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -380,6 +380,10 @@ ensogl::define_endpoints! {
         /// Switches the selected visualisation to/from fullscreen mode.
         toggle_fullscreen_for_selected_visualization(),
 
+        // === Scene Navigation ===
```
wdanilo, line 383: spacing before section

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -380,6 +380,10 @@ ensogl::define_endpoints! {
...
+
+        enable_navigation(),
+        disable_navigation(),
```
wdanilo, line 386: What this is about? Seems very strange. If this is a hack, please describe in comment why it is needed, what happens when its not present, and link to an issue about it.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -472,8 +476,15 @@ ensogl::define_endpoints! {
         visualization_enable_fullscreen (NodeId),
         visualization_set_preprocessor  ((NodeId,data::EnsoCode)),
 
+        visualization_selected          (NodeId),
+        visualization_deselected        (NodeId),
```
wdanilo, line 480: We have now better way of expressin it. Use this instead please:
```rust
on_visualization_select (Switch<NodeId>)
```
This will cover both `visualization_selected` and `visualization_deselected` and will simplify FRP network definition

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -472,8 +476,15 @@ ensogl::define_endpoints! {
...
+
+        some_visualisation_selected     (bool),
+
         node_being_edited (Option<NodeId>),
-        node_editing (bool)
+        node_editing (bool),
+
+        navigation_active (bool),
```
wdanilo, line 487: I'd name it `navigator_active` instead, as the namme `navigator` is used pretty often and for me, it points better at the subject of this action. Ignore if you feel otherwise.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -851,6 +862,10 @@ impl Edges {
 
 
 
+#[derive(Debug,Clone,CloneRef,Default)]
+struct Visualisations {
+    selected : SharedHashSet<NodeId>,
+}
```
wdanilo, line 868: This suggests that more than one visualization can be selected, right?

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1014,6 +1029,15 @@ impl GraphEditorModelWithNetwork {
             vis_enabled  <- vis_visible.gate(&vis_visible);
             vis_disabled <- vis_visible.gate_not(&vis_visible);
 
+            let vis_is_selected = node.model.visualization.frp.is_selected.clone_ref();
+
+            selected    <- vis_is_selected.gate(&vis_is_selected).constant(());
+            deselected  <- vis_is_selected.gate_not(&vis_is_selected).constant(());
+
+            output.source.visualization_selected   <+ selected.map(move |_| node_id);
+            output.source.visualization_deselected <+ deselected.map(move |_| node_id);
```
wdanilo, line 1038: replace with:
```rust
output.source.visualization_selected   <+ vis_is_selected.on_true().constant(node_id);
output.source.visualization_deselected <+ vis_is_selected.on_false().constant(node_id);
```

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1138,13 +1163,15 @@ impl GraphEditorModel {
         let display_object     = display::object::Instance::new(&logger);
         let nodes              = Nodes::new(&logger);
         let edges              = default();
-        let visualizations     = visualization::Registry::with_default_visualizations();
+        let vis_registry     = visualization::Registry::with_default_visualizations();
```
wdanilo, line 1166: `=` alignment

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1778,7 +1805,7 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
     let inputs         = &model.frp;
     let mouse          = &scene.mouse.frp;
     let touch          = &model.touch_state;
-    let visualizations = &model.visualizations;
+    let visualizations = &model.vis_registry;
```
wdanilo, line 1808: the var name is not updated

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1788,6 +1815,26 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
 
+    // ========================
+    // === Scene Navigation ===
+    // ========================
+
+    frp::extend! { network
+
+        no_vis_selected <- out.some_visualisation_selected.gate_not(&out.some_visualisation_selected);
+        some_vis_selected <- out.some_visualisation_selected.gate(&out.some_visualisation_selected);
+
```
wdanilo, line 1826: use `on_true` and `on_false` (available after mmerging my PR: https://github.com/enso-org/ide/pull/973)

wdanilo (changes requested):
Everything seems nice, but the hack with the camera lock needs to be changed :(

src/rust/ensogl/lib/core/src/display/camera/camera2d.rs
MichaelMauderer: We could use it inside the camera to stop setters from changing the camera when locked, but that seemed overkill. I'm not sure whether there is a better way than just having the status flag.

src/rust/ide/view/graph-editor/src/lib.rs
MichaelMauderer: I've added docs. Essentially, this is about locking the navigation of the scene, disabling panning/zooming etc.

src/rust/ide/view/graph-editor/src/lib.rs
MichaelMauderer: The problem leading to this is, that the visualizations inform us about their selection status, but we cannot know the order of FRP events we get. So, we might get, for example, "vis A select" "vis B" select" "vis A deselect" order of events, and we might want to keep track of this intermediary state. We could just have a single Option here, but this might be more useful.

src/rust/ensogl/lib/core/src/display/camera/camera2d.rs
farmaazon: And I wonder if we could apply a workaround here for now and create task for doing it properly?

@wdanilo or have you some another quick and better fix in mind?

farmaazon: Discussed during standup: We'll try approach with refactoring Navigator to be a part of Graph Editor view.

src/rust/ensogl/lib/core/src/display/navigation/navigator/events.rs
```diff
@@ -226,7 +235,8 @@ impl NavigatorEvents {
                 } else {
                     let x         = -event.delta_x() as f32;
                     let y         =  event.delta_y() as f32;
-                    let movement  = Vector2::new(x,y);
+                    let pan_speed = data.pan_speed();
+                    let movement  = Vector2::new(x,y) * pan_speed ;
```
wdanilo, line 239: space before semicolon

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -381,6 +381,17 @@ ensogl::define_endpoints! {
...
 
+        // === Scene Navigation ===
+
+
+        /// Stop the scene camera from moving around, locking the scene in place.
+        /// Can be used, e.g., if there is a fullscreen visualisation active, or navigation should
+        ///only work for a selected visualisation.
+        disable_navigator(),
+        /// Allow the scene camera to move around for navigating the scene.
+        enable_navigator(),
```
wdanilo, line 392: Over the last months I have learned that its always better to keep that as a single endpoint (`set_navigator_active(bool)`) rather than two, unless you really want to connect it to keyboard shortcuts (which we do not want). So lets use single endpoint if possible.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1132,19 +1163,21 @@ impl GraphEditorModel {
     , cursor : cursor::Cursor
     , frp    : &Frp
     ) -> Self {
-        let network        = &frp.network;
-        let scene          = app.display.scene();
+        let network        = &frp.network;let scene          = app.display.scene();
```
wdanilo, line 1166: alignment and lines glued

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1775,27 +1808,47 @@ impl Default for SelectionMode {
...
+
+        no_vis_selected <- out.some_visualisation_selected.on_false();
+        some_vis_selected <- out.some_visualisation_selected.on_true();
+
+        disable_navigator <- any_(inputs.disable_navigator,some_vis_selected);
+        enable_navigator  <- any_(inputs.enable_navigator,no_vis_selected);
+
+        eval_ disable_navigator ( model.navigator.disable() );
+        eval_ enable_navigator  ( model.navigator.enable()  );
+
```
wdanilo, line 1846: If you'd use a single frp event instead of `inputs.disable_navigator` and `inputs.enable_navigator`, then you'd be able to write (new FRP syntax ahead!) :) 

```rust
navigator_active <- inputs.set_navigator_disabled || some_vis_selected;
```
please note that Im using `inputs.set_navigator_disabled` because FRP endpoints initilize with defaults, so this will initialize with false. If we would name it `inputs.set_navigator_active`, it would initialize with false, which would give bad value when queries when `some_vis_selected` will fire.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2459,6 +2512,22 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+            if switch.is_on() {
+                model.visualisations.selected.insert(switch.value);
+            } else {
+                model.visualisations.selected.remove(&switch.value);
+            }
+        });
+
+        selection_change <- out.on_visualization_select.constant(());
+        out.source.some_visualisation_selected <+ selection_change.map(f_!([model] !model.visualisations.selected.is_empty()));
+
```
wdanilo, line 2527: rm nl

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2459,6 +2512,22 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+    frp::extend! { network
+        eval out.on_visualization_select ([model](switch) {
+            if switch.is_on() {
+                model.visualisations.selected.insert(switch.value);
+            } else {
+                model.visualisations.selected.remove(&switch.value);
+            }
+        });
+
+        selection_change <- out.on_visualization_select.constant(());
```
wdanilo, line 2525: this .constant is not needed. You are discarding the value in next line again, so this is just unnecesary FRP node = performance waste

src/rust/ide/view/graph-editor/src/lib.rs
wdanilo: Please add EXTENSIVE docs here - descirbing the architecture, the bottom up events propagation, that this set will contain a single value normally etc. This is very not obvious nad it sohuld be documented :)

wdanilo (approved):

farmaazon (approved):

PR #989, opened 2020-11-25
FRP stack trace and refactoring.

wdanilo:
### Pull Request Description
After merging the [Suggestion: Panic on loops in the functional reactive data flow](https://github.com/enso-org/ide/pull/894) PR, our GUI started throwing warnings about loops in our FRP network. The loops should not be there and can be a big potential problem in terms of nondeterministic behavior. Moreover, they cause the network to be really hard to understand, thus, they need to be eliminated. I started digging into this issue, but it was very hard to understand where these loops come from, so I have created a few utilities that help us with that (introduced in this PR):

- FRP call stack traces with information of module and line the FRP endpoint was defined in (with a flag to prune them in compile time).
- Also, the PR adds a "scam warning" to the console (see attachment). In the future, we should hide all the logs until not enabled in runtime, so the warning should be the only thing visible by the user.
- Also, I've refactored all "luna" names in different files to be "enso" names instead.

**Please note, that the recursive FRP bugs are not fixed in this PR**

image
image

### Checklist


src/rust/ide/view/graph-editor/src/component/node/output/area.rs
```diff
@@ -633,15 +633,15 @@ impl Area {
...
+        let network           = frp::Network::new("node_outputs");
         let id_map            = default();
         let frp               = Frp::new(&network,&id_map);
         let number_of_ports   = pattern_span_tree.root_ref().leaf_iter().count();
         let data              = OutputPortsData::new(scene,number_of_ports as u32);
         let data              = Rc::new(data);
         let pattern_span_tree = Rc::new(RefCell::new(pattern_span_tree));
         let scene             = scene.clone_ref();
-        let port_network      = default();
+        let port_network      = Rc::new(RefCell::new(frp::Network::new("node_outputs")));
```
farmaazon, line 644: The name is duplicated - is that by design?

src/rust/ide/view/graph-editor/src/builtin/visualization/native/raw_text.rs
```diff
@@ -44,9 +44,9 @@ impl RawText {
 
     /// Constructor.
     pub fn new(scene:&Scene) -> Self {
-        let network = default();
-        let frp   = visualization::instance::Frp::new(&network);
-        let model = RawTextModel::new(scene);
+        let network = frp::Network::new("raw_text");
```
farmaazon, line 47: "raw_text" does not suggest that this is visualization frp. (Compare with "js_visualization_instance")

src/rust/lib/frp/Cargo.toml
```diff
@@ -22,8 +24,14 @@ unicode-segmentation = { version = "1.6.0"                               }
...
 wasm-bindgen = { version = "=0.2.58" , features = ["nightly"] }
 
+
 [dependencies.web-sys]
 version = "0.3.4"
 features = [
   'KeyboardEvent',
 ]
+
+
```
farmaazon, line 34: This is the only Cargo.toml with additional spacing. And I don't think it is needed - those files are usually short, and the sections are clearly visible.

src/rust/lib/frp/src/stream.rs
```diff
@@ -7,6 +7,75 @@ use crate::data::watch;
...
+}
+
+
+// === Disabled ===
+
+/// A call stack trace for FRP events.
+#[derive(Debug,Clone,Copy,Default)]
+pub struct DisabledCallStack {
+    stack: ()
+}
```
farmaazon, line 65: Why we keep this field? AFAIK linter should disallow unused fields.

farmaazon (approved):

src/rust/ensogl/lib/core/src/application/frp.rs
```diff
@@ -207,7 +207,7 @@ macro_rules! define_endpoints {
             NORMALIZED
             $([$($global_opts)*])?
 
-            Input  { $([$($input_opts)*])?
+            Input  { $($([$($input_opts)*])?)?
```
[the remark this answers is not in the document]
wdanilo: Because it doesnt work otherwise. Look at the definition. Definition uses double optional expansion, so the body has to use one as well. It was invalid in the past!

src/rust/ensogl/lib/core/src/application/frp.rs
```diff
@@ -192,7 +192,7 @@
 /// ```
 #[macro_export]
 macro_rules! define_endpoints {
-    ( $([$($global_opts:tt)*])?
+    (   $([$($global_opts:tt)*])?
```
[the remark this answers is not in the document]
wdanilo: Yeah, this is bad. it should be indented instead in next line

Issue #939, opened 2020-11-10
Table visualization does not scroll


farmaazon:
### General Summary
Likely because of preventing default on mouse events.

PR #1001, opened 2020-11-26
Only prevent default zooming when holding ctrl key.

MichaelMauderer:
### Pull Request Description
This changes the event handling behaviour when using the mouse wheel. Right now we prevent all default events, but this prevents also some [html components from receiving events](https://github.com/enso-org/ide/issues/939). Allowing default events while zooming with CTRL key, will lead to bad behaviour: the browser will do its website zooming and our UI will also zoom. We want to avoid the website zooming, but we do want to allow panning events to be received by scrollable html components.
To achieve this, this PR allows default events when a mouse wheel event is a pan event, but keeps it deactivated when there is a zoom event on CTRL + mouse wheel. 

### Checklist

- [x] ~All code has been manually tested in the "debug/interface" scene.~ - Can't seem to see any new issues with zooming, but can't check the table visualization for working scrolling because language server is required to see the visualisation.

### DIFF
--- a/build/run.js
+++ b/build/run.js
@@ -136,7 +136,7 @@ commands.build.rust = async function(argv) {
 
         console.log('Checking the resulting WASM size.')
         let stats = fss.statSync(paths.dist.wasm.mainOptGz)
-        let limit = 4.23
+        let limit = 4.24
         let size = Math.round(100 * stats.size / 1024 / 1024) / 100
         if (size > limit) {
             throw(`Output file size exceeds the limit (${size}MB > ${limit}MB).`)
--- a/src/rust/ensogl/lib/core/src/display/navigation/navigator.rs
+++ b/src/rust/ensogl/lib/core/src/display/navigation/navigator.rs
@@ -26,6 +26,9 @@ pub struct NavigatorModel {
     resize_callback : callback::Handle,
     zoom_speed      : SharedSwitch<f32>,
     pan_speed       : SharedSwitch<f32>,
+    /// Indicates whether events handled the navigator should be stopped from propagating further
+    /// after being handled by the Navigator.
+    disable_events  : Rc<Cell<bool>>,
 }
 
 impl NavigatorModel {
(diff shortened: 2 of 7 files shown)


farmaazon (commented):
I did some testing, and found two undesired things:
1. Scrolling visualization works properly if it is active - vis scrolls, and nothing else. However, when visualization is not active, both vis and application are scrolling
2. Documentation panel scrolls with application even when it is active (I clicked into it).

It's supposed to work properly, because the event should have prevented default always when it's propagation is stopped (that is: always, except the visualization is active). And documentation should behave as a normal visualization.

Could you investigate it a little before merging?

MichaelMauderer:
@farmaazon 
> Scrolling visualization works properly if it is active - vis scrolls, and nothing else. However, when visualization is not active, both vis and application are scrolling

Can you show a video of that? I can't seem to reproduce this on my system.

> Documentation panel scrolls with application even when it is active (I clicked into it).

I think documentation was never given an "active" state so far. So clicking there does not deactivate navigator.
 I'll have a look at correcting that.

farmaazon:
> > Scrolling visualization works properly if it is active - vis scrolls, and nothing else. However, when visualization is not active, both vis and application are scrolling
> 
> Can you show a video of that? I can't seem to reproduce this on my system.
> 

[scrolling.zip](https://github.com/enso-org/ide/files/5608516/scrolling.zip)

This happens when I scroll with mouse over the visualization (without clicking on it).

wdanilo (approved):

MichaelMauderer:
@farmaazon This now also extends the searcher documentation panel to have "selected" status, that is passed through the controller to the graph editor to enable/disable the navigator.

src/rust/ide/src/view/integration.rs
```diff
@@ -867,6 +869,13 @@ impl Model {
         Ok(())
     }
 
+    fn documentation_selected_in_ui(&self, is_selected:&bool) -> FallibleResult {
+        debug!(self.logger, "Documentation selected.");
+        let graph_frp = &self.view.graph().frp;
+        graph_frp.set_navigator_disabled(is_selected);
+        Ok(())
+    }
+
```
farmaazon, line 878: I think this is wrong place: this file is integration between controllers and view, but here we integrate two views: documentation and graph.

The proper place is src/rust/ide/view/src/project.rs: the view containing both documentation panel and graph.

farmaazon (changes requested):

farmaazon:
And I tested it: now I see table and documentation scrolling working properly, but scatter plot still "scrolls" with the scene when not active. However, as it is no regression introduced by this PR, so let's merge it and I'll create a proper bug report.

farmaazon (approved):

PR #1014, opened 2020-12-01
Avoid changing edit status of project name when selecting it

MichaelMauderer:
### Pull Request Description
Avoids setting focus for the text field of the project name when it is set "selected", that is, when it is the current active breadcrumb level. 

### Important Notes
Adds some docs to clarify what "selected" in the API means. This was misleading before. 

### Checklist

- [ ] ~All code has been manually tested in the "debug/interface" scene.~ - Cannot be tested because we don't have collapsed nodes in the debug scene. 

### DIFF
--- a/src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
+++ b/src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
@@ -70,7 +70,9 @@ ensogl::define_endpoints! {
        /// Commit current project name.
        commit             (),
        outside_press      (),
+       /// Indicates that this is the currenlty active breadcrumb.
        select             (),
+       /// Indicates that this is not the currenlty active breadcrumb.
        deselect           (),
        /// Indicates the IDE is in edit mode. This means a click on some editable text should
        /// start editing it.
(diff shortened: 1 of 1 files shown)


src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
```diff
@@ -315,20 +317,16 @@ impl ProjectName {
 
             // === Selection ===
 
-            select <- any(&frp.select,&frp.input.start_editing);
-            eval_  select([text,animations]{
-                text.set_focus(true);
-                animations.color.set_target_value(selected_color);
-            });
-            frp.output.source.selected <+ select.to_true();
+            eval_  frp.select( animations.color.set_target_value(selected_color) );
```
wdanilo, line 320: doublespace after eval_

src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
```diff
@@ -315,20 +317,16 @@ impl ProjectName {
 
             // === Selection ===
 
-            select <- any(&frp.select,&frp.input.start_editing);
```
wdanilo, line 318: Earlier the `frp.input.start_editing` input was used here. It is not used anymore. Could you explain more why it was not necessary? I'm just trying to understand this change and why actually it solves the problem - any explanation why it solves it would be great

wdanilo (approved):

src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
MichaelMauderer: The actual fix is the removal of `text.set_focus(true);` on `select`. The select should not do anything to the text field but change its colour, as its only meant to be a highlight. So, `start_editing` should also not change the `select` status.

MichaelMauderer:
@wdanilo This is till missing the manual check.

PR #1033, opened 2020-12-07
Avoid updating positions of nodes during type updates.

MichaelMauderer:
### Pull Request Description
Adds a new update notification in the controller part, that indicates that the node expressions of a graph should be updated, but not the node positions.

### Checklist

- [ ] ~All code has been manually tested in the "debug/interface" scene.~ Requires language server integration. 

### DIFF
--- a/src/rust/ide/src/controller/graph.rs
+++ b/src/rust/ide/src/controller/graph.rs
@@ -68,6 +68,8 @@ pub struct NoPatternOnNode {
 pub enum Notification {
     /// The content should be fully reloaded.
     Invalidate,
+    /// The graph expressions need updating, e.g., types, names, ports.
+    ExpressionUpdate,
 }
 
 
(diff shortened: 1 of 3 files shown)


src/rust/ide/src/controller/graph.rs
```diff
@@ -68,6 +68,8 @@ pub struct NoPatternOnNode {
 pub enum Notification {
     /// The content should be fully reloaded.
     Invalidate,
+    /// The graph expressions need updating, e.g., types, names, ports.
+    ExpressionUpdate,
```
farmaazon, line 72: I'm unsure of this name. The expression means just the same as "node label" - the code. At least in controllers. But this event is actually fired not when code changes, but type or span-tree.

Therefore I would rename it to the `PortsUpdate`

farmaazon (approved):

PR #1057, opened 2020-12-11
Avoid changing edit status of project name when selecting it

MichaelMauderer:
### Pull Request Description
Avoids setting focus for the text field of the project name when it is set "selected", that is, when it is the current active breadcrumb level. 

### Important Notes
Adds some docs to clarify what "selected" in the API means. This was misleading before. 

### Checklist

- [ ] ~All code has been manually tested in the "debug/interface" scene.~ Cannot be tested because we don't have collapsed nodes in the debug scene.

### DIFF
--- a/src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
+++ b/src/rust/ide/view/graph-editor/src/component/breadcrumbs/project_name.rs
@@ -70,7 +70,9 @@ ensogl::define_endpoints! {
        /// Commit current project name.
        commit             (),
        outside_press      (),
+       /// Indicates that this is the currenlty active breadcrumb.
        select             (),
+       /// Indicates that this is not the currenlty active breadcrumb.
        deselect           (),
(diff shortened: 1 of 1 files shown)


wdanilo (approved):
I'd like @farmaazon to check it in the example scenes. For me it is not obvious why it fixes the issue, but it doesnt really matter, as this is probably very tricky event processing part.

MichaelMauderer:
> I'd like @farmaazon to check it in the example scenes. For me it is not obvious why it fixes the issue, but it doesnt really matter, as this is probably very tricky event processing part.

The main change for the fix is when `text.set_focus(true);` is called. The initial version called it when "selecting" the project name, which was wrong since that was meant to only make it highlighted as the active breadcrumb. The version that landed and was reverted completely omitted the `text.set_focus(true);` altogether which lead to some key events being missed (but not all). This version calls `text.set_focus(true);` when editing starts, ensuring that all key events are received by the text view.

farmaazon (approved):

PR #1070, opened 2020-12-16
Bundle project manager in electron package. 

MichaelMauderer:
### Pull Request Description
Add a fixed version of the project manager to our build. 

### Important Notes
Needs to be updates from time to time to use a newer version of the project manager. Build artefacts should be tested on all platforms. 

### Checklist

- [ ] ~All code conforms to the [Rust](https://github.com/enso-org/enso/blob/main/docs/style-guide/rust.md) style guide.~ No rust code.
- [ ] ~All code has been profiled where possible.~ Only affects build system
- [ ] ~All code has been manually tested in the "debug/interface" scene.~ Only affects build with language server

- [x] Build artefact tested on Linux 
- [x] Build artefact tested on macOS
- [x] Build artefact tested on Windows

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,8 +37,17 @@ jobs:
         with:
           node-version: '14.15.0'
 
-      - name: Build
-        run: node ./run dist --skip-version-validation
+      - name: Build (Ubuntu)
+        run: node ./run dist --skip-version-validation --target linux
+        if: matrix.os == 'ubuntu-latest'
+
+      - name: Build (Windows)
+        run: node ./run dist --skip-version-validation --target win
+        if: matrix.os == 'windows-latest'
+
+      - name: Build (MacOs)
+        run: node ./run dist --skip-version-validation --target macos
+        if: matrix.os == 'macos-latest'
 
       - name: Upload Artifacts (Ubuntu, AppImage)
         uses: actions/upload-artifact@v1
--- a/build/paths.js
+++ b/build/paths.js
@@ -1,4 +1,5 @@
 const path  = require('path')
+const os = require('os')
 
 
 
@@ -36,5 +37,22 @@ paths.js.root             = path.join(paths.root,'src','js')
 
 paths.rust                = {}
 paths.rust.root           = path.join(paths.root,'src','rust')
+
+function project_manager_path(root) {
+    let base_path = path.join(root, 'enso', 'bin')
+    const target_platform = os.platform()
+    switch (target_platform) {
+        case 'linux':
+            return path.join(base_path, 'project-manager')
+        case 'darwin':
+            return path.join(base_path, 'project-manager')
+        case 'win32':
+            return path.join(base_path, 'project-manager.exe')
+        default:
+            throw 'UnsupportedPlatform: ' + target_platform
+    }
+}
+
+paths.get_project_manager_path = project_manager_path
 
 module.exports = paths
--- a/build/paths.js
+++ b/build/paths.js
@@ -20,7 +20,7 @@ paths.dist                = {}
 paths.dist.root           = path.join(paths.root,'dist')
 paths.dist.client         = path.join(paths.dist.root,'client')
 paths.dist.content        = path.join(paths.dist.root,'content')
-paths.dist.bin = path.join(paths.dist.root, 'bin')
+paths.dist.bin            = path.join(paths.dist.root, 'bin')
 paths.dist.init           = path.join(paths.dist.root,'init')
 paths.dist.buildInfo      = path.join(paths.dist.root,'build.json')
 
@@ -38,21 +38,20 @@ paths.js.root             = path.join(paths.root,'src','js')
 paths.rust                = {}
 paths.rust.root           = path.join(paths.root,'src','rust')
 
-function project_manager_path(root) {
-    let base_path = path.join(root, 'bin')
+function get_project_manager_extension() {
     const target_platform = os.platform()
     switch (target_platform) {
-        case 'linux':
-            return path.join(base_path, 'project-manager')
-        case 'darwin':
-            return path.join(base_path, 'project-manager')
         case 'win32':
-            return path.join(base_path, 'project-manager.exe')
+            return '.exe'
         default:
-            throw 'UnsupportedPlatform: ' + target_platform
+            return ''
     }
 }
 
-paths.get_project_manager_path = project_manager_path
+paths.get_project_manager_path = function (root) {
+    let base_path = path.join(root, 'bin')
+    const extension = get_project_manager_extension()
+    return path.join(base_path, 'project-manager') + extension
+}
 
 module.exports = paths
--- a/build/paths.js
+++ b/build/paths.js
@@ -49,7 +49,7 @@ function get_project_manager_extension() {
 }
 
 paths.get_project_manager_path = function (root) {
-    let base_path = path.join(root, 'bin')
+    let base_path = path.join(root, 'enso', 'bin',)
     const extension = get_project_manager_extension()
     return path.join(base_path, 'project-manager') + extension
 }
--- a/build/paths.js
+++ b/build/paths.js
@@ -19,6 +19,7 @@ paths.dist                = {}
 paths.dist.root           = path.join(paths.root,'dist')
 paths.dist.client         = path.join(paths.dist.root,'client')
 paths.dist.content        = path.join(paths.dist.root,'content')
+paths.dist.bin = path.join(paths.dist.root, 'bin')
 paths.dist.init           = path.join(paths.dist.root,'init')
 paths.dist.buildInfo      = path.join(paths.dist.root,'build.json')
 
--- a/build/paths.js
+++ b/build/paths.js
@@ -39,7 +39,7 @@ paths.rust                = {}
 paths.rust.root           = path.join(paths.root,'src','rust')
 
 function project_manager_path(root) {
-    let base_path = path.join(root, 'enso', 'bin')
+    let base_path = path.join(root, 'bin')
     const target_platform = os.platform()
     switch (target_platform) {
         case 'linux':
--- a/build/run.js
+++ b/build/run.js
@@ -3,7 +3,7 @@ const fs       = require('fs').promises
 const fss      = require('fs')
 const glob     = require('glob')
 const ncp      = require('ncp').ncp
-const os = require('os')
+const os       = require('os')
 const path     = require('path')
 const paths    = require('./paths')
 const prettier = require("prettier")
@@ -74,7 +74,7 @@ function command(docs) {
 
 function run_project_manager() {
     const bin_path = paths.get_project_manager_path(paths.dist.bin)
-    console.log('Starting language server from ' + bin_path)
+    console.log('Starting the language server from ' + bin_path)
     child_process.execFile(bin_path, [], (error, stdout, stderr) => {
         console.error(stderr)
         if (error) {
@@ -317,7 +317,9 @@ optParser.options('dev', {
 })
 
 optParser.options('target', {
-    describe: 'Set the build target. Defaults to current platform',
+    describe:
+        'Set the build target. Defaults to the current platform. ' +
+        'Valid values are: "linux" "macos" and "win"',
     type: 'string',
 })
 
@@ -393,7 +395,8 @@ async function processPackageConfigs() {
 // === Main ===
 // ============
 
-async function updateBuildVersion (target) {
+async function updateBuildVersion (argv) {
+    const target =  get_target_platform(argv)
     let config        = {}
     let configPath    = paths.dist.buildInfo
     let exists        = fss.existsSync(configPath)
@@ -479,9 +482,7 @@ function get_target_platform(argv) {
 
 async function main () {
     let argv = optParser.parse()
-
-    const target = get_target_platform(argv)
-    await updateBuildVersion(target)
+    await updateBuildVersion(argv)
     await processPackageConfigs()
     let command = argv._[0]
     if(command === 'clean') {
--- a/build/run.js
+++ b/build/run.js
@@ -3,6 +3,7 @@ const fs       = require('fs').promises
 const fss      = require('fs')
 const glob     = require('glob')
 const ncp      = require('ncp').ncp
+const os = require('os')
 const path     = require('path')
 const paths    = require('./paths')
 const prettier = require("prettier")
@@ -121,7 +122,7 @@ commands.build.rust = async function(argv) {
     let crate     = argv.crate || DEFAULT_CRATE
     let crate_sfx = crate ? ` '${crate}'` : ``
     console.log(`Building WASM target${crate_sfx}.`)
-    let args = ['build','--target','web','--no-typescript','--out-dir',paths.dist.wasm.root,'--out-name','ide',crate]
+    let args = ['build','--target','web','--out-dir',paths.dist.wasm.root,'--out-name','ide',crate]
     if (argv.dev) { args.push('--dev') }
     await run_cargo('wasm-pack',args)
     await patch_file(paths.dist.wasm.glue, js_workaround_patcher)
@@ -296,6 +297,11 @@ optParser.options('dev', {
     type     : 'bool',
 })
 
+optParser.options('target', {
+    describe: 'Set the build target. Defaults to current platform',
+    type: 'string',
+})
+
 let commandList = Object.keys(commands)
 commandList.sort()
 for (let command of commandList) {
@@ -368,7 +374,7 @@ async function processPackageConfigs() {
 // === Main ===
 // ============
 
-async function updateBuildVersion () {
+async function updateBuildVersion (target) {
     let config        = {}
     let configPath    = paths.dist.buildInfo
     let exists        = fss.existsSync(configPath)
@@ -376,13 +382,17 @@ async function updateBuildVersion () {
         let configFile = await fs.readFile(configPath)
         config         = JSON.parse(configFile)
     }
-    let commitHashCmd = await cmd.run_read('git',['rev-parse','--short','HEAD'])
-    let commitHash    = commitHashCmd.trim()
-    if (config.buildVersion != commitHash) {
-        config.buildVersion = commitHash
-        await fs.mkdir(paths.dist.root,{recursive:true})
-        await fs.writeFile(configPath,JSON.stringify(config,undefined,2))
-    }
+
+    config.target = target
+
+    let commitHashCmd = await cmd.run_read('git', [
+        'rev-parse',
+        '--short',
+        'HEAD'
+    ])
+    config.buildVersion = commitHashCmd.trim()
+    await fs.mkdir(paths.dist.root, { recursive: true })
+    await fs.writeFile(configPath, JSON.stringify(config, undefined, 2))
 }
 
 async function installJsDeps() {
@@ -432,16 +442,35 @@ async function runCommand(command,argv) {
     runner()
 }
 
+function get_target_platform(argv) {
+    let target = argv.target
+    if (target === undefined) {
+        const local_platform = os.platform()
+        switch (local_platform) {
+            case 'darwin':
+                return 'macos'
+            case 'win32':
+                return 'win'
+            default:
+                return local_platform
+        }
+    }
+    return target
+}
+
 async function main () {
+    let argv = optParser.parse()
+
+    const target = get_target_platform(argv)
+    await updateBuildVersion(target)
     await processPackageConfigs()
-    updateBuildVersion()
-    let argv    = optParser.parse()
     let command = argv._[0]
-    if(command == 'clean') {
+    if(command === 'clean') {
         try { await fs.unlink(paths.dist.init) } catch {}
     } else {
         await installJsDeps()
     }
+
     await runCommand(command,argv)
 }
 
--- a/build/run.js
+++ b/build/run.js
@@ -10,6 +10,7 @@ const prettier = require("prettier")
 const stream   = require('stream')
 const yargs    = require('yargs')
 const zlib     = require('zlib')
+const child_process = require('child_process')
 
 process.on('unhandledRejection', error => { throw(error) })
 process.chdir(paths.root)
@@ -71,7 +72,17 @@ function command(docs) {
     return {docs}
 }
 
-
+function run_project_manager() {
+    const bin_path = paths.get_project_manager_path(paths.dist.bin)
+    console.log('Starting language server from ' + bin_path)
+    child_process.execFile(bin_path, [], (error, stdout, stderr) => {
+        console.error(stderr)
+        if (error) {
+            throw error
+        }
+        console.log(stdout)
+    })
+}
 
 // ================
 // === Commands ===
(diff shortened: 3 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -1,8 +1,6 @@
 name: Build (MacOS, Linux, Windows)
 
-on:
-  push:
-    branches: main
+on: [push]
```
MichaelMauderer, line 3: Note: this will be reverted once testing has finished.

MichaelMauderer:
@wdanilo I need to do more testing (see checkboxes), but I would like to get some feedback. Let me know about any JS weirdness you see. I'm not super up to date on best practices there.

.github/workflows/build.yml
```diff
@@ -37,8 +35,17 @@ jobs:
...
+        run: node ./run dist --skip-version-validation --target linux
+        if: matrix.os == 'ubuntu-latest'
+
+      - name: Build (Windows)
+        run: node ./run dist --skip-version-validation --target win
+        if: matrix.os == 'windows-latest'
+
+      - name: Build (MacOs)
+        run: node ./run dist --skip-version-validation --target macos
+        if: matrix.os == 'macos-latest'
```
wdanilo, line 48: out of curiosity, why we want to run this job 3 times instead of just 1, on MacOS with cross compilation? IT takes much more CI resources this way.

build/paths.js
```diff
@@ -19,6 +20,7 @@ paths.dist                = {}
 paths.dist.root           = path.join(paths.root,'dist')
 paths.dist.client         = path.join(paths.dist.root,'client')
 paths.dist.content        = path.join(paths.dist.root,'content')
+paths.dist.bin = path.join(paths.dist.root, 'bin')
```
wdanilo, line 23: alignment

build/paths.js
```diff
@@ -36,4 +38,21 @@ paths.js.root             = path.join(paths.root,'src','js')
...
+    let base_path = path.join(root, 'bin')
+    const target_platform = os.platform()
+    switch (target_platform) {
+        case 'linux':
+            return path.join(base_path, 'project-manager')
+        case 'darwin':
+            return path.join(base_path, 'project-manager')
+        case 'win32':
+            return path.join(base_path, 'project-manager.exe')
+        default:
```
wdanilo, line 51: 1. Try not using `returns` - this code is harder to read
2. Refactor common parts out (you are repeating the project anager name 3 times now, it would be mmuch better to compute the extension instead).

build/paths.js
```diff
@@ -36,4 +38,21 @@ paths.js.root             = path.join(paths.root,'src','js')
 paths.rust                = {}
 paths.rust.root           = path.join(paths.root,'src','rust')
 
+function project_manager_path(root) {
```
wdanilo, line 41: why this is a function instead of just var?

build/paths.js
```diff
@@ -36,4 +38,21 @@ paths.js.root             = path.join(paths.root,'src','js')
...
+        case 'darwin':
+            return path.join(base_path, 'project-manager')
+        case 'win32':
+            return path.join(base_path, 'project-manager.exe')
+        default:
+            throw 'UnsupportedPlatform: ' + target_platform
+    }
+}
+
+paths.get_project_manager_path = project_manager_path
```
wdanilo, line 56: if you want to do something liekthat in JS you can always write `paths.foo = function() { ... }` not to repeat the name several times.

build/run.js
```diff
@@ -70,7 +72,17 @@ function command(docs) {
     return {docs}
 }
 
-
+function run_project_manager() {
+    const bin_path = paths.get_project_manager_path(paths.dist.bin)
+    console.log('Starting language server from ' + bin_path)
```
wdanilo, line 77: this is not a sentence

build/run.js
```diff
@@ -70,7 +72,17 @@ function command(docs) {
...
+    child_process.execFile(bin_path, [], (error, stdout, stderr) => {
+        console.error(stderr)
+        if (error) {
+            throw error
+        }
+        console.log(stdout)
+    })
```
wdanilo, line 84: is this reporting errors / logs interactively, while running the process?

build/run.js
```diff
@@ -70,7 +72,17 @@ function command(docs) {
...
 }
 
-
+function run_project_manager() {
+    const bin_path = paths.get_project_manager_path(paths.dist.bin)
+    console.log('Starting language server from ' + bin_path)
+    child_process.execFile(bin_path, [], (error, stdout, stderr) => {
+        console.error(stderr)
+        if (error) {
+            throw error
```
wdanilo, line 81: are we sure we want to break this thing on each error ? What if during development we will send bad message from rust to the project anager, we will see it in console, then we change rust to fix it - we dont want then the `watch` script to break, right?

build/run.js
```diff
@@ -121,7 +133,7 @@ commands.build.rust = async function(argv) {
     let crate     = argv.crate || DEFAULT_CRATE
     let crate_sfx = crate ? ` '${crate}'` : ``
     console.log(`Building WASM target${crate_sfx}.`)
-    let args = ['build','--target','web','--no-typescript','--out-dir',paths.dist.wasm.root,'--out-name','ide',crate]
```
wdanilo, line 124: What was this? :O That is so strange such option was there

build/run.js
```diff
@@ -225,10 +237,18 @@ commands.watch.options  = Object.assign({},commands.build.options)
...
+        build_args.push(`--crate=${argv.crate}`)
+    }
+
+    run_project_manager()
+    build_args = build_args.join(' ')
+    let target =
+        '"' +
+        `node ${paths.script.main} build --skip-version-validation --no-js --dev ${build_args} -- ` +
+        cargoArgs.join(' ') +
+        '"'
```
wdanilo, line 250: I already hate prettier. LEts check if we can configure it soehow :(

build/run.js
```diff
@@ -296,6 +316,11 @@ optParser.options('dev', {
     type     : 'bool',
 })
 
+optParser.options('target', {
+    describe: 'Set the build target. Defaults to current platform',
```
wdanilo, line 320: not a sentence.

wdanilo: Plus, imo its worth to enuerate what platforms are supported, because platform names are not obvious

build/run.js
```diff
@@ -368,21 +393,25 @@ async function processPackageConfigs() {
...
     let config        = {}
     let configPath    = paths.dist.buildInfo
     let exists        = fss.existsSync(configPath)
     if(exists) {
         let configFile = await fs.readFile(configPath)
         config         = JSON.parse(configFile)
     }
-    let commitHashCmd = await cmd.run_read('git',['rev-parse','--short','HEAD'])
-    let commitHash    = commitHashCmd.trim()
-    if (config.buildVersion != commitHash) {
```
wdanilo, line 381: why this was removed? We should not write config file when not needed, as it may cause unneeded recompilation of JS

build/run.js
```diff
@@ -432,16 +461,35 @@ async function runCommand(command,argv) {
...
 
+function get_target_platform(argv) {
+    let target = argv.target
+    if (target === undefined) {
+        const local_platform = os.platform()
+        switch (local_platform) {
+            case 'darwin':
+                return 'macos'
+            case 'win32':
+                return 'win'
```
wdanilo, line 472: lets use full "windows". We are not making abbreviations for macos not linux

build/run.js
```diff
@@ -432,16 +461,35 @@ async function runCommand(command,argv) {
...
+                return local_platform
+        }
+    }
+    return target
+}
+
 async function main () {
+    let argv = optParser.parse()
+
+    const target = get_target_platform(argv)
```
wdanilo, line 483: I dont think thsi is the responsibility of the main function to get the target. This should be built into the `updateBuildVersion` function instead.

build/run.js
```diff
@@ -432,16 +461,35 @@ async function runCommand(command,argv) {
...
+    await updateBuildVersion(target)
     await processPackageConfigs()
-    updateBuildVersion()
-    let argv    = optParser.parse()
     let command = argv._[0]
-    if(command == 'clean') {
+    if(command === 'clean') {
```
wdanilo, line 487: yummy. I wish javascript had `====` operator :(

src/js/lib/client/package.js
```diff
@@ -21,11 +34,10 @@ let config = {
...
 
     scripts: {
-        "start": `electron ${paths.dist.content} -- `,
-        "build": "webpack ",
-        "dist": "electron-builder --publish never",
-        "dist:crossplatform": "electron-builder --mac --win --linux --publish never"
-    }
+        start: `electron ${paths.dist.content} -- `,
+        build: 'webpack ',
+        dist: 'electron-builder --publish never' + ' --' + build.target,
```
wdanilo, line 39: if your `get_build_config` fails, which it can, cause youre using if inside, then this could concatenate the string with "null" string, which is not what you want

src/js/lib/client/package.js
```diff
@@ -47,6 +59,8 @@ config.build = {
     },
     files: [
         { from: paths.dist.content, to: "." }
+    ,
+        { from: paths.dist.bin, to: '.' },
```
wdanilo, line 63: is this prettier formatting? If so, we have rather serious problem here

wdanilo, line 63: What is inside of `paths.dist.bin`? Only project manager now?

src/js/lib/client/src/index.js
```diff
@@ -1,14 +1,15 @@
 'use strict'
 
-import * as buildCfg  from '../../../../../dist/build.json'
 import * as Electron  from 'electron'
+import * as Server  from 'enso-studio-common/src/server'
+import * as assert from 'assert'
+import * as buildCfg from '../../../../../dist/build.json'
```
wdanilo, line 6: sort

src/js/lib/client/src/index.js
```diff
@@ -73,6 +74,11 @@ optParser.options('background-throttling', {
     describe : 'Throttle animations when run in background [false]',
 })
 
+optParser.options('local', {
```
wdanilo, line 77: I dont think the name `local` is self-explanatory. I would never ever think that `./enso --local ...` means "Set the path of a local project manager to use for running projects"`

src/js/lib/client/src/index.js
```diff
@@ -267,6 +273,30 @@ Electron.app.on('web-contents-created', (event,contents) => {
 
 
 
+// =======================
+// === Project Manager ===
+// =======================
+
+const child_process = require('child_process')
+const fss = require('fs')
```
wdanilo, line 281: move them to top of the file, or refactor this to separate file. Otherwise, its too easy in JS to not notice that something was already imported in this file.

src/js/lib/client/src/index.js
```diff
@@ -267,6 +273,30 @@ Electron.app.on('web-contents-created', (event,contents) => {
...
+
+async function run_project_manager() {
+    let bin_path = args.local
+    if (bin_path === undefined) {
```
wdanilo, line 285: I'd use here `if (bin_path)` which would handle `undefined`, `null`, and empty string, which is what we want I think.

src/js/lib/client/src/index.js
```diff
@@ -267,6 +273,30 @@ Electron.app.on('web-contents-created', (event,contents) => {
...
+        bin_path = paths.get_project_manager_path(root)
+    }
+    let bin_exists = fss.existsSync(bin_path)
+    assert(bin_exists, 'Could not find file: ' + bin_path)
```
wdanilo, line 289: not a sentence. Please, makle all errors and logs always sentences

src/js/lib/client/src/index.js
```diff
@@ -267,6 +273,30 @@ Electron.app.on('web-contents-created', (event,contents) => {
...
+
+    child_process.execFile(bin_path, [], (error, stdout, stderr) => {
+        console.error(stderr)
+        if (error) {
+            throw error
+        }
+        console.log(stdout)
+    })
+}
+
```
wdanilo, line 299: spacing

src/js/lib/client/src/index.js
```diff
@@ -278,6 +308,7 @@ let server     = null
 let mainWindow = null
 
 async function main() {
+    await run_project_manager()
```
wdanilo, line 311: this should hve some logging - that we are stargin project nager or something.

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
+const fss = require('fs')
```
wdanilo, line 1: no sections nor documentqtion in this file. Lets use the same docs rules as in Rust here (module docs + pub functions docs)

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+        return JSON.parse(configFile)
+    }
+}
+
+async function get_target_url(): Promise<string> {
+    const config = await get_build_config()
+    const target_platform = config.target
+    console.log('webpack target ' + target_platform)
+    const version = '0.1.2-rc.18'
+    const base_url = `https://github.com/enso-org/enso-staging/releases/download/enso-${version}/enso-project-manager-${version}`
```
wdanilo, line 30: too longl ine. Make sure every line is <100 chars. Just split it into few vars

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
+const fss = require('fs')
+const fs = fss.promises
+
+const tar = require('tar')
+const os = require('os')
+const path = require('path')
+const unzipper = require('unzipper')
+const url = require('url')
+const { http } = require('follow-redirects')
```
wdanilo, line 9: dont use require - its very old syntax. Use the import syntasx instead.

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const thisPath = path.resolve(__dirname)
+const root = path.resolve(thisPath, '..', '..', '..', '..', '..')
+const distPath = path.resolve(root, 'dist', 'bin')
+const buildInfoPath = path.resolve(root, 'dist', 'build.json')
```
wdanilo, line 14: is this not part of `paths.js`? The same applies to the above path.

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+
+async function get_build_config() {
+    let exists = fss.existsSync(buildInfoPath)
+    if (exists) {
+        let configFile = await fs.readFile(buildInfoPath)
+        return JSON.parse(configFile)
+    }
+}
+
+async function get_target_url(): Promise<string> {
```
wdanilo, line 25: this function name is not self-explanatory

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+    const base_url = `https://github.com/enso-org/enso-staging/releases/download/enso-${version}/enso-project-manager-${version}`
+    switch (target_platform) {
+        case 'linux':
+            return `${base_url}-linux-amd64.tar.gz`
+        case 'macos':
+            return `${base_url}-macos-amd64.tar.gz`
+        case 'win':
+            return `${base_url}-windows-amd64.zip`
+        default:
+            throw 'UnsupportedPlatform: ' + target_platform
```
wdanilo, line 39: 1. dont use return
2. refactor common part out

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function decompress_project_manager(source_file_path, target_folder) {
+    let decompressor
+    if (source_file_path.toString().endsWith('.zip')) {
+        decompressor = unzipper.Extract({ path: target_folder })
+    } else {
+        decompressor = tar.x({
+            strip: 1,
+            C: target_folder,
```
wdanilo, line 54: these needs docs

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+        })
+    }
+    fss.createReadStream(source_file_path)
+        .pipe(decompressor)
+        .on('finish', () => {
+            const bin_path = project_manager_path()
+            if (os.platform() != 'win32') {
+                fss.chmodSync(bin_path, '744')
```
wdanilo, line 62: this should be extracted to well naed var

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+        decompressor = tar.x({
+            strip: 1,
+            C: target_folder,
+        })
+    }
+    fss.createReadStream(source_file_path)
+        .pipe(decompressor)
+        .on('finish', () => {
+            const bin_path = project_manager_path()
+            if (os.platform() != 'win32') {
```
wdanilo, line 61: we are using `win32` name here, and in other places we were renaming osses to our names. Maybe lets create a single place with OS handling and ourrn naes, preferably encoded as enums (enums in JS are just objects with fields being strings, like `os = {windows:'windows', macos:'macos', ...}`

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+async function download_file_http(
+    file_url: string,
+    overwrite: boolean
+): Promise<void> {
+    const file_name = url.parse(file_url).pathname.split('/').pop()
+    const file_path = path.resolve(distPath, file_name)
+
+    if (fss.existsSync(file_path) && !overwrite) {
+        console.log(
+            `The ${file_path} file exists. Project manager executable will not be regenerated.`
```
wdanilo, line 76: this mentions "project manager" but the function name tells about any file

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+        .pipe(decompressor)
+        .on('finish', () => {
+            const bin_path = project_manager_path()
+            if (os.platform() != 'win32') {
+                fss.chmodSync(bin_path, '744')
+            }
+        })
+}
+
+async function download_file_http(
```
wdanilo, line 67: This sohuld be extracted to our common library in the future (after it moves to TS). Write todo about it pls

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,107 @@
...
+    const options = {
+        host: parsed.host,
+        port: 80,
+        path: parsed.pathname,
+    }
+
+    const target_file = fss.createWriteStream(file_path)
+    http.get(options, (res) => {
+        res.on('data', (data) => {
+            target_file.write(data)
```
wdanilo, line 93: can we have progress info here? Even isn the simplest form - downloading 200 or 300 b just takes time

.github/workflows/build.yml
MichaelMauderer: I'm not sure, the CI job was set up this way and I followed the existing setup.

It might better isolate the jobs in case weird things happen due to caching?

build/run.js
MichaelMauderer: I don't think it is. From the docs it sounds like this is only called on the end of the process.

build/run.js
MichaelMauderer: Depends. If the language server crashes, should we stop the watch script or try to restart the language server?

I think it might be better to also stop `watch` to have the log output from the crashing language server readily available. Otherwise, it will be gone by the time we realize what is going on from new status updates.

build/run.js
MichaelMauderer: Now it's gone, and we are better off.

build/run.js
MichaelMauderer: We need to update it always because the target OS can change now. This is no  longer only dependent on the commit has but multiple parameters. Should we actually verify that we only rewrite the file if one of them has changed?

src/js/lib/client/package.js
MichaelMauderer: I've added an exception to `get_build_config` in case it can't load the build file.

src/js/lib/client/package.js
MichaelMauderer: No, this is an artefact from removing the initial prettier formatting commit.

MichaelMauderer: Yes.

src/js/lib/client/src/index.js
MichaelMauderer: According to my sort plugin, they are sorted. What would be the preferred sorting?

src/js/lib/client/src/index.js
MichaelMauderer: That was the issue spec. 
What about `--local-project-manager` or `--project-manager`

src/js/lib/project-manager/src/build.ts
MichaelMauderer: This would be touching many places of our build system. I think we should do an extra refactor step and think about what we want to have as our command line input and then normalise our internal handling based on that.

build/run.js
wdanilo: Call note: ide/build/cmd.js

build/run.js
wdanilo: If it executes when LS terinates, its ok. Please be sure it works this way after you make interactive terminal output here.

build/run.js
wdanilo: Yep, lets compare it

build/run.js
wdanilo: Let's leave it and open maintenance issue about it

src/js/lib/client/src/index.js
wdanilo: later one!

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,160 @@
...
+// === Utils ===
+// =============
+
+async function get_build_config() {
+    let exists = fss.existsSync(buildInfoPath)
+    if (exists) {
+        let configFile = await fs.readFile(buildInfoPath)
+        return JSON.parse(configFile.toString())
+    } else {
+        throw 'Could not find build file at ' + buildInfoPath
```
wdanilo, line 27: not a sentence!

wdanilo: ```throw `Could not find build file at '${buildInfoPath}'.` ```

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,160 @@
...
+            postfix = `linux-amd64.tar.gz`
+            break
+        case 'macos':
+            postfix = `macos-amd64.tar.gz`
+            break
+        case 'win':
+            postfix = `windows-amd64.zip`
+            break
+        default:
+            throw 'UnsupportedPlatform: ' + target_platform
```
wdanilo, line 55: if then else, I beg you <3

src/js/lib/project-manager/src/build.ts
```diff
@@ -0,0 +1,160 @@
...
+            target_file.end()
+            console.log(file_url + ' downloaded to ' + file_path)
+            decompress_project_manager(file_path, distPath)
+        })
+    })
+}
+
+// ================
+// === Main ===
+// ================
```
wdanilo, line 153: alignment

wdanilo (approved):

PR #1077, opened 2020-12-20
Fixing "Improper behavior while trying to reconnect node with arguments"

wdanilo:
### Pull Request Description
- Implements fixes for [Improper behavior while trying to reconnect node with arguments](https://github.com/enso-org/ide/issues/953). The fix bases on the new invariant, that creating new symbol instances without deleting any, will always provide growing instance IDs, and thus, new symbols will always appear over old ones.
- Should be reviewed together with https://github.com/enso-org/rust-lib/pull/27 .
- Also, this PR adds a lot of documentation regarding different sprites sorting mechanisms that could be implemented, including their comparison and possible performance and memory consumption consequences.

### Checklist


src/rust/Cargo.toml
```diff
@@ -54,13 +54,13 @@ debug = true
...
+[patch.crates-io]
+enso-automata = { path = '../../../rust-lib/src/automata' }
+enso-data = { path = '../../../rust-lib/src/data' }
+enso-generics = { path = '../../../rust-lib/src/generics' }
+enso-logger = { path = '../../../rust-lib/src/logger' }
+enso-macro-utils = { path = '../../../rust-lib/src/macro-utils' }
+enso-optics = { path = '../../../rust-lib/src/optics' }
+enso-prelude = { path = '../../../rust-lib/src/prelude' }
+enso-shapely = { path = '../../../rust-lib/src/shapely/impl' }
+enso-shapely-macros = { path = '../../../rust-lib/src/shapely/macros' }
```
wdanilo, line 66: This will be removed before merge ofc

src/rust/ensogl/lib/core/src/system/gpu/data/attribute.rs
```diff
@@ -129,7 +142,7 @@ impl {
             for buffer in &self.buffers {
                 buffer.set_to_default(id.into())
             }
-            self.free_ids.push(id);
+            self.free_ids.insert(*id);
```
farmaazon, line 145: We replace popping and pushing with asymptotic complexity O(1) with removing/inserting having complexity O(n). Have you checked if it alter our IDE performance? I believe that it may hurt us when we create really many nodes, then remove the first one and add again.

And I believe that we want to guarantee the order of "consecutively added sprites" to fix the issue in topic - so this guarantee should be expressed in docs, and/or in comment here.

farmaazon (approved):
I didn't reviewed carefully the docs you wrote for `Symbol`, as it requires much more OpenGL background I believe than I have.

PR #1079, opened 2020-12-21
Errors on nodes

MichaelMauderer:
### Pull Request Description
Adds a graph editor API to indicate that nodes have an error.

Screenshot from 2020-12-22 16-03-09
Screenshot from 2020-12-22 16-08-47

Screenshot from 2020-12-22 16-02-54

### Important Notes
Provisional API that might need updating based on the language server specification of errors.

Currently, can only be seen in the debug scene as the language server API does not exist yet. 

The error is shown above the node to not overlap the visualizations.

### Checklist


src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -137,6 +150,33 @@ pub mod drag_area {
     }
 }
 
+pub mod error_shape {
```
wdanilo, line 153: this should not be in the "shape" section, as this is not node shape anymore

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -137,6 +150,33 @@ pub mod drag_area {
...
+            let width  = Var::<Pixels>::from("input_size.x");
+            let height = Var::<Pixels>::from("input_size.y");
+            let width  = width  - PADDING.px() * 2.0;
+            let height = height - PADDING.px() * 2.0;
+            let radius = RADIUS.px();
+
+            let repeat_x      = ERROR_PATTERN_REPEAT_X.px();
+            let repeat_y      = ERROR_PATTERN_REPEAT_Y.px();
+            let repeat        =  Var::<Vector2<Pixels>>::from((repeat_x,repeat_y));
+            let error_width   =  ERROR_PATTERN_STRIPE_WIDTH.px();
```
wdanilo, line 167: alignment

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -137,6 +150,33 @@ pub mod drag_area {
...
+            let stripe_red    = Rect((error_width, INFINITE.px()));
+            let pattern       = stripe_red.repeat(repeat).rotate(45.0.radians());
```
wdanilo, line 169: make it constant pls

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -137,6 +150,33 @@ pub mod drag_area {
     }
 }
 
+pub mod error_shape {
+    use super::*;
+
+    ensogl::define_shape_system! {
```
wdanilo, line 156: does it implement wider stripes on zoom out?

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -273,15 +314,17 @@ impl Deref for Node {
...
-    pub input          : input::Area,
-    pub output         : output::Area,
-    pub visualization  : visualization::Container,
-    pub action_bar     : action_bar::ActionBar,
+    pub app             : Application,
+    pub display_object  : display::object::Instance,
+    pub logger          : Logger,
+    pub main_area       : component::ShapeView<shape::Shape>,
+    pub drag_area       : component::ShapeView<drag_area::Shape>,
+    pub error_indicator : component::ShapeView<error_shape::Shape>,
```
wdanilo, line 322: why not `error_area`? I believe it is an area, looking at the screenshots, and then it would have consistent name with other areas.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -292,17 +335,28 @@ impl NodeModel {
         edge::depth_sort_hack_1(scene);
 
         output::area::depth_sort_hack(&scene);
-        let main_logger = Logger::sub(&logger,"main_area");
-        let drag_logger = Logger::sub(&logger,"drag_area");
-        let main_area   = component::ShapeView::<shape::Shape>::new(&main_logger,scene);
-        let drag_area   = component::ShapeView::<drag_area::Shape>::new(&drag_logger,scene);
+        let main_logger             = Logger::sub(&logger,"main_area");
+        let drag_logger             = Logger::sub(&logger,"drag_area");
+        let error_indicator_logger  = Logger::sub(&logger,"drag_area");
```
wdanilo, line 340: these namesare inconsistent

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -292,17 +335,28 @@ impl NodeModel {
...
+
+        let error_indicator = component::ShapeView::<error_shape::Shape>::new(&error_indicator_logger,scene);
```
wdanilo, line 342: long line. Please mmake sure that all new lines < 100 chars

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -292,17 +335,28 @@ impl NodeModel {
...
         input::area::depth_sort_hack(scene); // FIXME hack for sorting
 
         let display_object  = display::object::Instance::new(&logger);
         display_object.add_child(&drag_area);
         display_object.add_child(&main_area);
+        display_object.add_child(&error_indicator);
+        display_object.add_child(&error_text);
+
+        error_text.set_default_color.emit(color::Rgba::red());
+        let text_size:TextSize    = TEXT_SIZE.into();
```
wdanilo, line 357: alignment + you can erge this line with the next one

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -325,7 +379,8 @@ impl NodeModel {
         display_object.add_child(&output);
 
         let app = app.clone_ref();
-        Self {app,display_object,logger,main_area,drag_area,output,input,visualization,action_bar}
+        Self {app,display_object,logger,main_area,drag_area,output,input,visualization,action_bar,
+              error_indicator,error_text}
             .init()
```
wdanilo, line 384: move this init to prev line, pls

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -366,9 +421,14 @@ impl NodeModel {
         let padded_size = size + Vector2(PADDING,PADDING) * 2.0;
         self.main_area.shape.sprite.size.set(padded_size);
         self.drag_area.shape.sprite.size.set(padded_size);
+        self.error_indicator.shape.sprite.size.set(padded_size);
         self.main_area.mod_position(|t| t.x = width/2.0);
         self.drag_area.mod_position(|t| t.x = width/2.0);
 
+        self.error_indicator.set_position_x(width/2.0);
+        self.error_text.set_position_y(1.0 * height + TEXT_SIZE);
```
wdanilo, line 429: what is this `1.0 *`?

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -380,6 +440,18 @@ impl NodeModel {
     pub fn visualization(&self) -> &visualization::Container {
         &self.visualization
     }
+
+    fn set_error(&self, error:&Option<Error>) {
```
wdanilo, line 444: don't use `&Option<Error>`, use `Option<&Error>` instead. This is a much better design, which allows for a much better usage. Imagine that in one point in your program you have `&Error`. Then using the second API you can bpass it here without cloning. The same applies to situation where you have `&Option<Error>` - you can pass it to second api without perf overhead, while the API you proposed is mmuch more limited.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -380,6 +440,18 @@ impl NodeModel {
...
+        // TODO: This extra text field should not be required after #1026 has been finished.
+        //Instead we should get the error content as normal node output that is visible in the
+        //visualisation.
```
wdanilo, line 447: I think this comment shouldgo to the text field definition place. Moreover, I'm not sure this is true, as we want to display errors even with visualisations closed.

wdanilo, line 447: alignment

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -391,6 +463,7 @@ impl Node {
         let selection = Animation::<f32>::new(network);
 
         let bg_color_anim = color::Animation::new(network);
+        let error_anim    = color::Animation::new(network);
```
wdanilo, line 466: these names are inconsistent now

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -180,6 +220,7 @@ ensogl::define_endpoints! {
         set_disabled        (bool),
         set_input_connected (span_tree::Crumbs,Option<Type>,bool),
         set_expression      (Expression),
+        set_error_status    (Option<Error>),
```
wdanilo, line 223: why not just `set_error`, like `set_expression` above, and the `set_error` function in model?

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -380,6 +440,18 @@ impl NodeModel {
...
     }
+
+    fn set_error(&self, error:&Option<Error>) {
+        // TODO: This extra text field should not be required after #1026 has been finished.
+        //Instead we should get the error content as normal node output that is visible in the
+        //visualisation.
+        if let Some(error) = error.as_ref() {
+            self.error_text.set_content.emit(error.message.clone());
+        } else {
+            self.error_text.set_content.emit("".to_string());
```
wdanilo, line 451: refactor commmon part out, pls.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -450,6 +523,28 @@ impl Node {
...
+                model.set_error(error);
+                match error.as_ref() {
+                    Some(error) => {
+                        let error_data:visualization::Data = error.clone().into();
+                        model.visualization.frp.set_data.emit(error_data);
+
+                        let red:color::Lcha = color::Rgba::red().into();
+                        error_anim.target.emit(red);
+                    },
+                    None    => error_anim.target.emit(color::Lcha::transparent()),
```
wdanilo, line 538: alignment

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -450,6 +523,28 @@ impl Node {
...
+
+            eval frp.set_error_status([model,error_anim](error) {
+                model.set_error(error);
+                match error.as_ref() {
+                    Some(error) => {
+                        let error_data:visualization::Data = error.clone().into();
+                        model.visualization.frp.set_data.emit(error_data);
+
+                        let red:color::Lcha = color::Rgba::red().into();
+                        error_anim.target.emit(red);
```
wdanilo, line 536: you are doing 2 frp emmits from these block, which means that this could be split into more fine-grained FRP endpoints. Please do it. Emitting more than one thing is almost always a bad design. Moreover, then you can not emit things at all, and use our new syntax here, like `error_anim.target <+ something`.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -450,6 +523,28 @@ impl Node {
...
+                    },
+                    None    => error_anim.target.emit(color::Lcha::transparent()),
+                }
+            });
+
+            eval error_anim.value([model](value) {
+                let rgba = color::Rgba::from(value);
+                model.error_indicator.shape.color_rgba.set(rgba.into());
```
wdanilo, line 544: you can make it 1 line, then also remove curly braces, and also `[model]` part.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -464,6 +559,7 @@ impl Node {
             );
         }
 
+        frp.set_error_status.emit(None);
```
wdanilo, line 562: why is it needed?

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -362,6 +362,8 @@ ensogl::define_endpoints! {
         stop_editing(),
         /// Remove all nodes from the graph.
         collapse_selected_nodes(),
+        /// Indicate whether this node had an error or not.
+        set_node_error_status((NodeId,Option<node::error::Error>)),
```
wdanilo, line 366: don't use double parenshere, our macros will handle that for you

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2424,11 +2426,25 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+
+
+    // === Set Node Error ===
+    frp::extend! { network
+
+    eval inputs.set_node_error_status([model]((node_id, error)) {
+        if let Some(node) = model.nodes.get_cloned_ref(node_id) {
+            node.set_error_status.emit(error)
+        }
+    });
```
wdanilo, line 2439: Try not using `emit` fro inside `eval`. Currently, our FRP primitives are so rich, that it is not needed almost ever. Just use `<= ... . map(... model.nodes.get_cloned_ref(node_id))` and then zip it with error, and then `node.set_error_status <+ ...`

wdanilo (changes requested):
Good to see it implemented! Few comments:

1. I'm surprised by the number of places where the code was not cleaned stylistically – remember to do self-review, please. It was not happening with previous PRs.

2. The resulting image looks bad, the lines are very pixelated, this needs to be fixed, however, I do not know what it is caused by (the equations look good to me). Definitely, something is broken here. Such an effect normally means that we are "shortening" the distance between pixels. Are you sure that without repeat, the single stripe would be of the same width than with the repeat? Also, is single stripe rendered correctly? Please show me that in DM on Discord and lets chat.
image

3. No dark-mode screenshots! Please make some.

4. Also, please make bigger screenshots with the whole GUI and few nodes visible, so we can see how it looks like with normal zoom and with other nodes. Also, please test uch wider stripes below nodes.

src/rust/ide/view/graph-editor/src/component/node.rs
MichaelMauderer: I find `indicator` more apt. It has no area properties, instead, it something that can visually indicate that there is an error. It's only states are on/off (with animations) and we don't care about area-based interactions with it.

src/rust/ide/view/graph-editor/src/component/node.rs
MichaelMauderer: To indicate that this can also clear the error status by setting `None`. Just `set_error` does not imply that.

src/rust/ide/view/graph-editor/src/component/node.rs
MichaelMauderer: If we don't have it we get a black error indicator from the colour animation.

src/rust/ide/view/graph-editor/src/lib.rs
MichaelMauderer: I'm not sure how you mean. Nodes can't be passed through the FRP network as they don't have a default value, so we need to unwrap in normal rust code.

MichaelMauderer:
>     2. The resulting image looks bad, the lines are very pixelated, this needs to be fixed, however, I do not know what it is caused by (the equations look good to me). Definitely, something is broken here. Such an effect normally means that we are "shortening" the distance between pixels. Are you sure that without repeat, the single stripe would be of the same width than with the repeat? Also, is single stripe rendered correctly? Please show me that in DM on Discord and lets chat.
>        image

This is most likely just the screen grab being overblown in the view here.
> 
>     3. No dark-mode screenshots! Please make some.

Added!

>     4. Also, please make bigger screenshots with the whole GUI and few nodes visible, so we can see how it looks like with normal zoom and with other nodes.
Added!

src/rust/ensogl/lib/core/src/display/shape/primitive/glsl/shape.glsl
```diff
@@ -290,6 +296,14 @@ Shape set_color(Shape shape, Srgba t) {
     return shape;
 }
 
+Shape withInfiniteBounds (Shape s) {
+    Id       id    = s.id;
+    Color    color = s.color;
+    BoundSdf sdf   = s.sdf;
+    sdf.bounds = infinite();
```
wdanilo, line 303: alignment

src/rust/ensogl/lib/core/src/display/shape/primitive/glsl/shape.glsl
```diff
@@ -322,3 +336,7 @@ vec2 scale (vec2 position, float value) {
 vec2 cartesian2polar (vec2 position) {
   return vec2(length(position), atan(position.y, position.x));
 }
+
+vec2 repeat (vec2 position, vec2 t) {
+    return mod(position, t);
+}
```
wdanilo, line 342: "t" is not thebest name either ;)

src/rust/ensogl/lib/core/src/display/shape/primitive/shader/canvas.rs
```diff
@@ -298,6 +298,20 @@ impl Canvas {
         let value = value.into();
         self.grow(num, s, -value)
     }
+
+    /// Repeat the shape with the given tile size.
+    pub fn repeat<T:Into<Var<Vector2<Pixels>>>>
+    (&mut self,  num:usize, s:Shape, tile_size:T) -> Shape {
```
wdanilo, line 304: double space

src/rust/ide/view/graph-editor/src/component/node.rs
wdanilo: call note: `set_error`

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -380,6 +450,16 @@ impl NodeModel {
     pub fn visualization(&self) -> &visualization::Container {
         &self.visualization
     }
+
+    fn set_error(&self, error:Option<&Error>) {
+        let message = if let Some(error) = error {
+            error.message.clone()
+        } else {
+            "".to_string()
+        };
```
wdanilo, line 459:
```
error.map(|t| t.message.clone()).unwrap_or_default()
```

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -450,6 +531,25 @@ impl Node {
...
+                        model.visualization.frp.set_data.emit(error_data);
+                        color::Lcha::from(color::Rgba::red())
+                    },
+                    None => color::Lcha::transparent(),
+                }
+            }));
+
+            eval error_color_anim.value((value) {
+                model.error_indicator.shape.color_rgba.set(color::Rgba::from(value).into());
+            });
```
wdanilo, line 550: rm curly braces

src/rust/ide/view/graph-editor/src/component/node/error.rs
```diff
@@ -0,0 +1,34 @@
...
+
+// =============
+// === Error ===
+// =============
+
+/// Error information to be displayed on a node.
+///
+/// Note:
+/// Tis is a dummy implementation that can and should be extended once we have the proper
+/// error information from the language server. See #1026 for more information.
```
wdanilo, line 16: lets use full links

wdanilo (approved):

PR #1108, opened 2021-01-12
Fix output edge coloring

farmaazon:
### Pull Request Description
Update the output root type id if the whole node's expression type has changed.

### Checklist

- [ ] ~~All code has been manually tested in the "debug/interface" scene.~~

### DIFF
--- a/src/rust/ide/view/graph-editor/src/lib.rs
+++ b/src/rust/ide/view/graph-editor/src/lib.rs
@@ -1529,8 +1529,11 @@ impl GraphEditorModel {
     fn set_node_expression_usage_type(&self, node_id:impl Into<NodeId>, ast_id:ast::Id, maybe_type:Option<Type>) {
         let node_id  = node_id.into();
         if let Some(node) = self.nodes.get_cloned_ref(&node_id) {
+            // TODO[ao]: we must update root output port according to the whole expression type
+            //     due to a bug in engine https://github.com/enso-org/enso/issues/1038.
             if node.view.model.output.whole_expr_id().contains(&ast_id) {
-                node.view.model.output.set_expression_usage_type(span_tree::Crumbs::default(), maybe_type.clone());
+                let crumbs = span_tree::Crumbs::default();
+                node.view.model.output.set_expression_usage_type(crumbs,maybe_type.clone());
             }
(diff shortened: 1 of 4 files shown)


src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1528,6 +1529,9 @@ impl GraphEditorModel {
     fn set_node_expression_usage_type(&self, node_id:impl Into<NodeId>, ast_id:ast::Id, maybe_type:Option<Type>) {
         let node_id  = node_id.into();
         if let Some(node) = self.nodes.get_cloned_ref(&node_id) {
+            if node.view.model.output.whole_expr_id().contains(&ast_id) {
+                node.view.model.output.set_expression_usage_type(span_tree::Crumbs::default(), maybe_type.clone());
+            }
```
wdanilo, line 1534: The lines below are a little bit duplicating this work. `node.view.frp.set_expression_usage_type` is implemented as:

```rust
    fn set_expression_usage_type(&self, crumbs:&Crumbs, tp:&Option<Type>) {
        match crumbs.endpoint {
            Endpoint::Input  => self.input.set_expression_usage_type(&crumbs.crumbs,tp),
            Endpoint::Output => self.output.set_expression_usage_type(&crumbs.crumbs,tp),
        }
    }
```

So I'd prefer this to work this way, that instead of sending event to some nested structure like you do, you should make special case if `node.view.model.output.whole_expr_id().contains(&ast_id)` and modify the `crumbs` variable, so the line `node.view.frp.set_expression_usage_type.emit((crumbs,maybe_type));` will be redirected correctly. Also, next to this special case write comment that this is a hack and link to the issue (similar to the comment in the output ports implementation).

farmaazon: But in this special case we'll need to call `set_expression_usage_type` twice anyway - because we need to update both input and output port's type. Of course, I can create crumbs for input and output separately and call `set_expression_usage_type` but I think it is strange to create Crumbs structure only to be decomposed right away.

wdanilo: Ok, so just add comment there about that this is a fix for a problem

wdanilo (approved):

PR #1113, opened 2021-01-15
Use bordered windows for the electron app on non-MacOS platforms.

MichaelMauderer:
### Pull Request Description
By default, add frames to windows on non-MacOS platforms. Can  still be explicitly overridden as desired. 

### Checklist

- [x] All code has been manually tested in the IDE. (With Win/KDE)


src/js/lib/client/src/index.js
```diff
@@ -383,6 +382,15 @@ function urlParamsFromObject(obj) {
 function createWindow() {
     let webPreferences     = secureWebPreferences()
     webPreferences.preload = path.join(root,'preload.js')
+
+    // Note: this is a conditional default. We want borders on Win/Linux and borderless on Mac.
```
wdanilo, line 386: Describe why! We want borderless everywhere, but there are bugs that prevent us from doing it. This should be documented. Moreover, it should be reported in electron repo and linked here.

src/js/lib/client/src/index.js
```diff
@@ -383,6 +382,15 @@ function urlParamsFromObject(obj) {
...
+    if (args.frame === undefined && process.platform === 'darwin') {
+        args.frame = false
+    }
+    if (args.frame === undefined && process.platform !== 'darwin') {
+        args.frame = true
+    }
```
wdanilo, line 392: this could be refactored to something simpler:
```js
if (args.frame === undefined) {
    args.frame = proces.platform !== 'darwin'
}

```

wdanilo: Btw, the placeent of this code is really bad. Why function `createWindow` is responsible for mutating command line arguments and adding some defaults there? Function `createWindow` should just create a window.

wdanilo (changes requested):

src/js/lib/client/src/index.js
```diff
@@ -177,6 +176,15 @@ function parseCmdArgs() {
 
 let args = parseCmdArgs()
 
+// Note: this is a conditional default to avoid issues with some window managers affecting
+// interactions at the top of a borderless window. Thus, we want borders on Win/Linux and
+// borderless on Mac. See https://github.com/enso-org/ide/issues/1101 for details.
```
wdanilo, line 181: Could you also add issue on Electron bug tracker for this? Currently, the `-webkit-app-region  : drag;` prevents events to be send to JS on Windows and Linux, but works on MacOS. This is used to define custom dragregion, as described here https://www.electronjs.org/docs/api/frameless-window#draggable-region

src/js/lib/client/src/index.js
```diff
@@ -177,6 +176,15 @@ function parseCmdArgs() {
...
+if (args.frame === undefined) {
+    const show_border = process.platform !== 'darwin'
+    args.frame = show_border
+}
```
wdanilo, line 185: Why do we use so many names for the same thing? If we are using name `frame`, lets stick to it, instead of naming it "border" as well.

wdanilo (changes requested):

wdanilo (approved):

PR #1123, opened 2021-01-18
Css comment with issue

lolczak:
### Pull Request Description
Changelog:
- Css comment with the issue
- Loader adjustments to the Cloud UI

### Checklist


wdanilo:
Who has marked it?
image

wdanilo (approved):

lolczak:
> Who has checked it?
> image

Sorry, checked too many checkboxes.

wdanilo:
@lolczak Are you going to finish the checks in this PR? 
image

lolczak:
> @lolczak Are you going to finish the checks of this PR? We are unsure if we should wait for it or not.
> image

Done.

src/js/lib/common/src/loader.js
```diff
@@ -17,8 +17,10 @@ export class ProgressIndicator {
     constructor(cfg) {
         this.dom                = html_utils.new_top_level_div()
         this.dom.id             = 'loader'
-        this.dom.style.position = 'absolute'
-        this.dom.style.zIndex   = 1
+        this.dom.style.position = 'fixed'
+        this.dom.style.top      = 0
+        this.dom.style.left     = 0
+        this.dom.style.zIndex   = 100
```
wdanilo, line 23: I'm ok with all the changes, but this is bugging me a little bit. Why are we setting the `zIndex` to 100? It is not obvious why it is needed and either needs a good docs with explanation here, or it should be removed. I'd rather remove it and change the `zIndex` where it is really needed - in the cloud version, because this setting is cloud-only, so keeping it in GUI doesnt make too much sense as it introduces confusion why it is needed at all. What do you think about it? Beside that, I'm ok with all other changes here.

lolczak: I added a meaningful const name.

lolczak: We don't have a cloud version of the IDE and we cannot set it in the cloud. Moreover, we don't want to have a tight coupling between the cloud UI and the progress bar that is an internal state of the IDE. It also breaks the encapsulation of the IDE. In order to do it as you want I would have to violate too many golden rules of software engineering.

wdanilo (approved):

src/js/lib/common/src/loader.js
wdanilo: Ok, lets do it this way. If you could add there a comment telling that this is needed for the cloud it would be cool, otherwise someone will chagne it without knowing the consequences.

lolczak: done

wdanilo (approved):

PR #1156, opened 2021-01-27
docs: fix backslash working as escape char

akavel:
A backslash in markdown works as escaping char, so to emit a real backslash (in ctrl-\ keyboard shortcut), a double-backslash needs to be used.


wdanilo (approved):

wdanilo:
Thank you!

PR #1158, opened 2021-01-27
Making CI config more generic

wdanilo:
### Pull Request Description
This PR implements new CI configuration:
- It is generalized, all sub-steps are written in a reusable way (in JavaScript, the GitHub YAML is generated).
- The CI configuration is now much more optimized and uses job dependencies (less resource hungry, faster to build).
- The CI supports now automatic releases and tags creation based on commits to branches `unstable` and `stable` and the entries in `CHANGELOG.md`. See docs (or below) for more information.
- The CDN integration is done.
- There is a new option in CI. If you want to trigger the building of artifacts on your local branch, just add `[ci build]` to your commit message!

### Checklist


build/run.js
```diff
@@ -1,30 +1,29 @@
...
+const path          = require('path')
+const paths         = require('./paths')
+const prettier      = require("prettier")
+const stream        = require('stream')
+const yargs         = require('yargs')
+const zlib          = require('zlib')
 const child_process = require('child_process')
+const release       = require('./release')
+const workflow      = require('./workflow')
+const { promisify } = require('util')
```
farmaazon, line 16: Can we sort those "imports" alphabetically?

docs/CONTRIBUTING.md
```diff
@@ -227,3 +227,35 @@ have prepared several scripts which maximally automate the process:
...
+  `[major].[minor].[patch]-[sfx]`, where `[sfx]` is one of `alpha.[n]`, `beta.[n]`, or `rc.[n]`, 
+  where `[n]` is an unstable build number.
+
+- **stable**
+  Contains only those commits which can be considered stable product releases. Each commit to this
+  branch will result in a stable release of the product and will be published on GitHub as a 
+  release. The build version and build description will be automatically fetched from the newest 
+  `CHANGELOG.md` entry and will fail if the version will not be of the form 
+  `[major].[minor].[patch]`.
+
```
farmaazon, line 261: This does not tell the full story how the CHANGELOG file should look like: That each version is a single-`#` section with the necessary "Enso " prefix and the version as described.

Do I understand correctly, that once we do merge to stable branch we modify changelog to merge all alphas, betas and rcs into a single changelog for a new stable version?

farmaazon (changes requested):
* I like this JS code, it is very clear.
* Currently we don't make the release build on the feature branches, and thus we don't check the wasm binary size limit before merging to master (blocker)
* I think we also should check on feature branches if CHANGELOG file have a proper format.
* I would also add a new check to the PR template to make sure the CHANGELOG was updated if applicable.

wdanilo:
> * I like this JS code, it is very clear.
Thanks!

> * Currently we don't make the release build on the feature branches, and thus we don't check the wasm binary size limit before merging to master. We should do it on at least one platform (blocker).
Good point, done.

> * I think we also should check on feature branches if CHANGELOG file have a proper format.
We do it on every execution of the `./run` script to update our JS configs, so it's already done.

> * I would also add a new check to the PR template to make sure the CHANGELOG was updated if applicable.
Done!

build/release.js
```diff
@@ -86,7 +86,7 @@ function changelogEntries() {
     let entries  = []
     for (let section of sections) {
         if (!section.startsWith(prefix)) {
-            throw `Improper changelog entry header: ${section}`
+            throw `Improper changelog entry header: ${section}. It should start with`
```
farmaazon, line 89: Looks like something is missing.

docs/README_TEMPLATE.md
```diff
@@ -0,0 +1,163 @@
...
+[79272]: http://github.com/ticket
+[79273]: http://github.com/ticket
+[79274]: http://github.com/ticket
+[79275]: http://github.com/ticket
+
+
+
+<br/>
+# Enso 47.0.0-alpha.8 (2049-01-24)
+...
```
farmaazon, line 163: This file is hilarious :D however I wonder if we need so long a template?

farmaazon (approved):
Only minor comment - to consider, no need to be applied

PR #1166, opened 2021-01-29
Integrate setPreprocessor function with Engine

farmaazon:
### Pull Request Description
In this PR a new method of "ExecututionContext" model is created allowing visualization modification, and the "on_preprocessor_changed" endpoint of visualization is exposed in GraphEditor and integrated with the new controller methods.

### Important Notes
A new debug shortcut has been added in a histogram visualization which sets a predefined preprocessor discarding the value and returning simple list, according to the Acceptance Criteria of the issue.

### Checklist


src/rust/ide/src/controller/graph/executed.rs
```diff
@@ -116,28 +116,35 @@ impl Handle {
...
-    /// See `expression_info_registry` in `ExecutionContext`.
+    /// See [`model::ExecutionContext::expression_info_registry`].
     pub fn computed_value_info_registry(&self) -> &ComputedValueInfoRegistry {
         self.execution_ctx.computed_value_info_registry()
     }
 
+    /// Modify preprocessor code in visualization. See also
+    /// [`model::ExecutionContext::modify_visualization`].
+    pub async fn set_visualization_preprocessor
+    (&self, id:VisualizationId, code:String) -> FallibleResult{
```
wdanilo: add space before `{`

src/rust/ide/src/ide/integration.rs
```diff
@@ -1133,6 +1147,24 @@ impl Model {
             }
         });
     }
+
+    fn visualization_preprocessor_changed(&self, node_id:graph_editor::NodeId, code:&graph_editor::data::EnsoCode)
+    -> FallibleResult {
+        if let Some(visualization) = self.visualizations.get_copied(&node_id) {
+            let logger      = self.logger.clone_ref();
+            let controller  = self.graph.clone_ref();
+            let code_string = AsRef::<String>::as_ref(code).clone();
```
wdanilo: what does this `clone` do? Doesn't it just copy a reference?

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -394,13 +394,16 @@ impl ContainerModel {
...
     }
 
-    fn set_visualization(&self, visualization:Option<visualization::Instance>) {
-        if let Some(visualization) = visualization {
-            let size = self.size.get();
-            visualization.set_size.emit(size);
-            self.view.add_child(&visualization);
-            self.visualization.replace(Some(visualization));
+    fn set_visualization
+    (&self, visualization:visualization::Instance, preprocessor:&frp::Any<EnsoCode>) {
```
wdanilo: couldyou change the types to be `language::enso::Code` (of course `language` part can be skipped in imports), as we've been talking on DM, please?

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -394,13 +394,16 @@ impl ContainerModel {
...
+        let size = self.size.get();
+        visualization.set_size.emit(size);
+        let _vis_network = visualization.network();
+        frp::extend! { _vis_network
+            preprocessor <+ visualization.on_preprocessor_change;
```
wdanilo: I think here is a logical bug. If I' not wrong and if I understand this code correctly, when switching a visualization, this code attaches `on_preprocessor_change` output from the new visualization to some outer `preprocessor` FRP input. I see two things here:
1. We do not use in any other file such FRP code that is a function that accepts ref to an `frp` input and attaches things to it. Such logic should rather be expressed in FRP itself because of 2 reasons:
    - It is really hard to track it down.
    - It is easy to make errors this way (see point 2).
2. What happens if we switch visualizations several times? I understand that this will still be bound to old visualizations as well, right? If so, old visualizations can run in parallel and can send `set_preprocessor` events that can then set the preprocessor of other visualization. Even if this is not true, because in our current case old visualizations are destroyed, then I don't like the implicit assumptions of this function - it assumes that the old visualization will be removed before this function is evaluated (if not, then old visualization before being destroyed can still emit `set_preprocessor` event). (Also, we will want to be able to display more than one visualization attached to the same node - however, this may be done as a special case of specially displayed nodes on the stage, so it may not be important here. Still, worth noting.)

wdanilo (changes requested):
I added few notes to the code, some important. Also, there are few things that need to be fixed / I'd love to chat about:
1. The docs about how to define JS visualizations are not updated. They should be updated in detail, as we will use them as user-facing docs. They should include information about how to set preprocessor function, how it is being evaluated by the engine (in what context/mode), how to change preprocessor function during JS runtime. etc.
2. The docs about the preprocessor function should also appear somewhere in Rust-based visualizations to explain the Rust API of setting the preprocessor. They should contain similar information to docs from point 1.
3. I think that this issue should be part of this PR: https://github.com/enso-org/ide/issues/1167 . Right now, we cannot specify the context that the preprocessor is run, and thus, the preprocessor is unusable for anything more serious than toy examples. Without the ability to specify the right context, this just doesn't work correctly (it is not a lack of a feature, it is a broken behavior). Even worse, if we implement it in a separate task, it will introduce a breaking change and all visualizations created between these tasks may need an update (we know how hard it is after qualified names support). It does not seem like a big task at all, as this is just information passing in the same way as we set preprocessor, so it should easily be added to the code / this PR. Also, docs should be very clear about the default mode. What is the default evaluation mode right now?

farmaazon:
> 3. I think that this issue should be part of this PR: #1167 . Right now, we cannot specify the context that the preprocessor is run, and thus, the preprocessor is unusable for anything more serious than toy examples. Without the ability to specify the right context, this just doesn't work correctly (it is not a lack of a feature, it is a broken behavior). Even worse, if we implement it in a separate task, it will introduce a breaking change and all visualizations created between these tasks may need an update (we know how hard it is after qualified names support). It does not seem like a big task at all, as this is just information passing in the same way as we set preprocessor, so it should easily be added to the code / this PR. Also, docs should be very clear about the default mode. What is the default evaluation mode right now?

Currently we run the visualizations in the context of `Main` module of currently opened project. And currently we support only visualization in currently opened project, so I think we set the proper default according to the @kustosz 's description. The only compatibility broken will be the default Module of builtin visualization, but that will be not visible to user.

Thus the #1167 won't break the old visualization, and thence I would insist on doing it at it was planned.

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
farmaazon: > We do not use in any other file such FRP code that is a function that accepts ref to an frp input and attaches things to it. 

We do. Look at `new_node` function of `GraphEditorModelWithNetwork`. And I don't know how I could express modifying FRP in FRP. Because that's it: we receive a new visualization instance with any arbitrary FRP network, and we want to connect it's frp with our frp.

> What happens if we switch visualizations several times? I understand that this will still be bound to old visualizations as well, right? If so, old visualizations can run in parallel and can send set_preprocessor events that can then set the preprocessor of other visualization. Even if this is not true, because in our current case old visualizations are destroyed, then I don't like the implicit assumptions of this function - it assumes that the old visualization will be removed before this function is evaluated (if not, then old visualization before being destroyed can still emit set_preprocessor event)

It is not an assumption of this function. The old visualization is destroyed in the same place, a line 406. So it will be destroyed before any other code will be run (remember, we are in a single-threaded environment). So the function itself does not assume anything.

And TBH I don't know how to do it better: to have an endpoint in container which emits me the newest preprocessor regardless what instance is currently attached.

src/rust/ide/src/ide/integration.rs
farmaazon: No. Calling clone on a reference will clone the content, otherwise such a code wouldn't work as expected
```rust
fn foo(x:&SomeBigStruct) -> SomeBigStruct {
    let c = x.clone();
   ...
}
```
And I checked it: if I remove clone, I get an error saying "expected String, got reference".

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
farmaazon: After all, we decided to make this "connection" as a separate network, managed solely by Visualization Container, so any connections should not leak after instance change.

src/rust/ide/view/graph-editor/src/component/visualization/instance.rs
```diff
@@ -32,42 +34,46 @@ pub struct Frp {
...
+    /// This event should be emitted when the received data are incorrect, or cause an internal
+    /// error.
     pub data_receive_error    : frp::Source<Option<DataError>>,
-    pub preprocessor_change: frp::Source<EnsoCode>,
-    pub activate              : frp::Source,
-    pub deactivate            : frp::Source,
+    /// This event should be emitted  should be emitted to set a new code of the preprocessor. The
+    /// preprocessor is a function called on the Engine side before sending data to IDE, allowing us
+    /// do some compression or filtering for the best performance. See also _Lazy Visualization_
+    /// section in `docs/product/visualizations.md`
```
MichaelMauderer: Is there special syntax to create a doc link here?

src/rust/ide/view/graph-editor/src/component/visualization/instance.rs
```diff
@@ -21,8 +21,10 @@ use ensogl::display::Scene;
 #[derive(Clone,CloneRef,Debug)]
 #[allow(missing_docs)]
 pub struct FrpInputs {
-    pub set_size  : frp::Source<Vector2>,
-    pub send_data : frp::Source<Data>,
+    pub set_size   : frp::Source<Vector2>,
+    pub send_data  : frp::Source<Data>,
+    pub activate   : frp::Source,
+    pub deactivate : frp::Source,
 }
```
MichaelMauderer, line 32: Should this be refactored to use the FRP macro?

farmaazon: I'm not sure. The macro makes the `source` field crate-local. But we want to allow anyone to extend the instance network as they please, when implementing "rust" visualizations.

docs/product/visualizations.md
```diff
@@ -251,19 +252,27 @@ In particular:
...
+
+For example, imagine you want to display a heatmap of 10 million points on 
+a map, and these points change rapidly. Sending such amount of information via
+WebSocket could be too much, and you (as the visualization author) might decide 
+that the visualization image should be generated on the server, and your 
+visualization is meant only to display the resulting image. In such a scenario, 
+you can define in your visualization an Enso function which will compute the 
+image on the server!
+
+
```
wdanilo, line 275: I love the example in the docs!

This has a few grammar mistakes (like the rest of the docs here) - would you be so nice and log-in to `https://app.grammarly.com/` and copy-paste this doc there? It will show all errors with auto-fix buttons. In fact, we would use Grammarly for all comments. Its so bad its not integrated with intellij

src/rust/ide/src/controller/visualization.rs
```diff
@@ -137,7 +137,7 @@ impl Handle {
                 let js_code    = self.language_server_rpc.read_file(&path).await?.contents;
                 let identifier = visualization.clone();
                 let error      = |_| VisualizationError::InstantiationError {identifier}.into();
-                let module     = data::builtin_library(); // FIXME: provide real library name.
+                let module     = enso::builtin_library(); // FIXME: provide real library name.
```
wdanilo: is this fixme still valid?

src/rust/ide/src/ide/integration.rs
wdanilo: Oh, got it. Can we then simplify the code and add a method `to_string` to `enso::Code` please?

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -394,13 +399,20 @@ impl ContainerModel {
...
-            self.view.add_child(&visualization);
-            self.visualization.replace(Some(visualization));
+    fn set_visualization
+    (&self, visualization:visualization::Instance, preprocessor:&frp::Any<enso::Code>) {
+        let size = self.size.get();
+        visualization.set_size.emit(size);
+        frp::new_network! { vis_frp_connection
+            // We need an additional "copy" node here. Doing simple
+            // `preprocessor <+ visualization.on_preprocessor_change` will not create any node in
+            // this network, so the connection will leak after setting new instance.
```
wdanilo: IMO this comment is not enough. If I did not know what is happening here, I'd not understand what "leak" would mean. This comment should be clear that we are creating a new network here that will be dropped when switching visualizations, so this connection should be dropped then as well.

src/rust/ide/view/graph-editor/src/component/visualization/container.rs
```diff
@@ -520,13 +532,14 @@ impl Container {
         frp::extend! { network
             eval  frp.set_visibility    ((v) model.set_visibility(*v));
             eval_ frp.toggle_visibility (model.toggle_visibility());
+            let   preprocessor = &frp.source.preprocessor;
```
wdanilo: multiplespaces after `let`

src/rust/ide/view/graph-editor/src/component/visualization/instance.rs
wdanilo: Michael is right, this should use FRP macro. Adam is right as well :D The macro should allow thesethings to be public (I wondering if it does not already). Adam, could you add fixme here, please? :)

wdanilo (approved):

src/rust/ide/view/graph-editor/src/component/visualization/instance.rs
farmaazon: I don't know, but I put here a simple link to dev.enso.org

src/rust/ide/view/graph-editor/src/component/visualization/instance.rs
farmaazon: THere is another problem: the Frp structures created by macro owns the Network. But here we create "Visualization API" in existing network (the specific visualization creates and owns its Network).

src/rust/ide/src/controller/visualization.rs
farmaazon: Yes, it is. Actually the code is funny: we set here a wrong name, but then in integration.rs we don't care about that field, but just set the Project.Main. I'll extend the FIXME and add the issue number.

src/rust/ide/src/ide/integration.rs
farmaazon: Actually, the to_string is already available, as we deref to &str :)

PR #1175, opened 2021-02-04
CI Improvements

wdanilo:
### Pull Request Description
- Adding checks if `CHANGELOG.md` was modified. See docs for more info and options.
- Adding `sha256` checksum generation.
- Removing `.zip` outputs from electron-builder. Zip archives were of almost the same size (saving of <3Mb) than the whole package and their generation was long. Other types, like `.tar.gz` do not save much and their generation is even longer (which probably is also broken in electron-builder, as the generation of `.tar.gz` can take literally hours!).
- Bringing back the docs generation CI (fixes regression of previous PR).


farmaazon (approved):

build/workflow.js
```diff
@@ -13,9 +13,11 @@ const yaml  = require('js-yaml')
 // === Constants ===
 // =================
 
-const NODE_VERSION      = '14.15.0'
-const RUST_VERSION      = 'nightly-2019-11-04'
-const WASM_PACK_VERSION = '0.9.1'
+const NODE_VERSION             = '14.15.0'
```
MichaelMauderer, line 16: Are we not getting this version programmatically somewhere?

MichaelMauderer (approved):

build/workflow.js
wdanilo: Good question and good observation. Yes, we repeat things in CI, we always did. I'll add a task to fix it, but it will take some good refactoring one day.

PR #1209, opened 2021-02-17
Don't insert imports when method calls are created through searcher

mwu-tow:
### Pull Request Description
When method call is inserted as a suggestion from searcher, the import of the relevant module won't be added anymore.
It is not needed and may cause trouble with name collisions, like #1145.
Issue #1178

### Checklist


farmaazon (approved):

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ read the notes of the `Enso 2.0.0-alpha.1` release.
 - [Assigning intermediate expressions to values][1067]. Nodes added with searcher will have their 
   values automatically assigned to a newly generated variables. This allows Enso Engine to cache
   intermediate values, improving visualization performance.
+- [Not adding spurious imports][1209]. Fixed case where IDE would generate spurious import 
+  declarations when inserting method calls.
```
wdanilo: Could you describe the issue a little bit more using simpler words and maybe bigger description? The changelog is for users snd they should understand what it is about. Even "inserting method calls" is not really understandable fully, as in the GUI its not obvious to every user what that's mean.

I would go this way (written fast, contains errors, just a scratch description):
```
[Not adding unnecessary library imports][1209]. Fixed situation when IDE was adding unnecessary libraries imports when selecting hints from node searcher. This makes the generated textual code easier to read.
```

wdanilo (approved):

farmaazon (changes requested):
Unfortunately, the reality is more complex than I expected. Actually there is scenario which works wrong with the current solution: press tab without any node selected, pick Table.new and press enter.

the new is a method, but it is a method of not imported module "Table". So actually we should skip importing only when we pick a method for this argument.

kustosz:
@farmaazon @mwu-tow it's even more complex! Even if you have the `this`, some methods are extension methods. For an example, `Json.to_table` is a method in defined `Table`, even though `Json` comes from `Base`. So the rule for "import is not necessary" is:
1. The method being called comes from the same module as `this`. In this case, never import.
2. The method being called comes from a different module than `this`. Then check if the method's module is already imported and don't import if it aready is.

farmaazon:
As for point 2: in scope of this task is only checking if we literally have the same import (as we do with other suggestions).

kustosz:
@farmaazon sure, point 1 is the most important one, because it could lead you towards overoptimization – i.e. not importing extension methods, which would break programs (and not just make them ugly by too many imports)

kustosz:
I have not read this code btw. it's very possible you've already taken everything into account :P Just saw this discussion and randomly commented

farmaazon:
No, no, the point 1 in your comment is important (cc @mwu), and no, we don't handle it in our code.

src/rust/ide/src/model/suggestion_database/entry.rs
```diff
@@ -103,25 +114,45 @@ impl Entry {
...
+        let code = match this_expr {
+            Some(this_expr) => format!("{}{}{}",this_expr,ast::opr::predefined::ACCESS,self.name),
+            None            => self.name.clone(),
+        };
+
+        CodeToInsert {code,imports}
+    }
+
+    /// Check if this is a non-extension module method.
+    // TODO [mwu] Currently we have no means to recognize module extension methods.
```
farmaazon: Could you add an issue for TODOs here? And add a note, that module extension methods will be treated as non-module methods.

src/rust/ide/src/model/suggestion_database/entry.rs
```diff
@@ -103,25 +114,45 @@ impl Entry {
...
-            self.name.clone()
+    /// Describes code insertion to be done when Searcher input suggestion is picked.
+    pub fn code_to_insert
+    (&self, current_module:Option<&module::QualifiedName>, generate_this:bool) -> CodeToInsert {
+        let mut imports = BTreeSet::new();
+
+        // Regular methods are found through this-lookup. Other entities need to be visible.
+        // Import if not already defined in this module.
+        if !self.is_regular_method() && !current_module.contains(&&self.module) {
+            imports.insert(self.module.clone());
```
farmaazon: I don't understand the comment.

src/rust/ide/src/model/suggestion_database/entry.rs
```diff
@@ -164,6 +195,20 @@ impl Entry {
     pub fn invocation_info(&self) -> span_tree::generate::context::CalledMethodInfo {
         self.into()
     }
+
+    /// Check if this is a regular method (i.e. non-extension method).
+    pub fn is_regular_method(&self) -> bool {
+        println!("{:?}",self);
```
farmaazon: println left.

farmaazon (commented):
Please, update the branch so I could easily test it with all new changes.

farmaazon (approved):

PR #1277, opened 2021-03-01
Update mixpanel integration to fix events beeing sent in invalid way.

MichaelMauderer:
### Pull Request Description
* Refactor mix panel integration to avoid double encoding the event names. 
* Add information about build (git revision, git status, debug build info) 
* 
### Checklist

- [ ] ~The `CHANGELOG.md` was updated with the changes introduced in this PR.~
- [ ] ~The documentation has been updated if necessary.~


build/run.js
```diff
@@ -482,6 +482,7 @@ async function updateBuildVersion (argv) {
     if (config.buildVersion !== commitHash || config.target !== target){
         config.target = target
         config.buildVersion = commitHash
+        config.args = argv
```
farmaazon: Why we need this? I don't see any usage in this PR.

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,84 @@ pub use wasm_bindgen::prelude::*;
...
+            window.enso.remoteLog(msg,data)
+        } catch (error) {
+            console.error(\"Error while logging message. \" + error );
+        }
+
     } else {
         console.warn(\"Failed to send log message.\")
     }
 }
 ")]
```
farmaazon: 1. This became a bit of js. I'd prefer to put it in a js file, not inline snippet
2. I don't know js very well, but is there _any_ chance to reduce all repeated ifs an tries here? For example, to use a helper wrapping some lambda in those if and try?

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,84 @@ pub use wasm_bindgen::prelude::*;
...
+/// Send the provided public event with data to our logging service.
+pub fn remote_log
+<T:Loggable>(message:&str, data:AnonymousData<T>) {
+    _remote_log(JsValue::from(message.to_string()), data.0.get());
+}
+
+/// Send the provided public event to our logging service.
+pub fn register(message:&str) {
+    _register(JsValue::from(message.to_string()));
 }
```
farmaazon: The docs of `register` and `remote_log_event` are identical. So I don't know the difference between those functions. Same for `remote_log` and `remote_log_data_field`.

src/rust/ide/src/ide.rs
```diff
@@ -74,6 +74,12 @@ pub struct Ide {
 impl Ide {
     /// Constructor.
     pub async fn new(application:Application, view:ide_view::project::View, project:model::Project) -> FallibleResult<Self> {
+
+        #[cfg(debug_assertions)]
+        analytics::remote_log_data_field("debug_mode", "debug_mode_is_active", analytics::AnonymousData("true"));
+        #[cfg(not(debug_assertions))]
+        analytics::remote_log_data_field("debug_mode", "debug_mode_is_active", analytics::AnonymousData("false"));
+
```
farmaazon: It this should be the first log, I probably put it in initialization function. Please note, that the view is already created and could already make some logs (however it shouldn't because there should be not any user interaction yet - still I would keep things consistent).

farmaazon (approved):

src/js/lib/content/src/index.js
```diff
@@ -179,20 +179,37 @@ const MAX_MESSAGE_LENGTH = 500;
 
 function remoteLog(event,data) {
     if (mixpanel) {
-        event = JSON.stringify(event).substr(0,MAX_MESSAGE_LENGTH)
-        if (data !== undefined) {
+        event = event.substr(0,MAX_MESSAGE_LENGTH)
+        if (data !== undefined && data !== null) {
```
wdanilo, line 191: function to check whether soething is undefined or null should be extracted and used everywhere. In the code below we chach `(data !== undefined)` - we should also check if it was null. Basically, checking for both is a good practice always.

src/js/lib/content/src/index.js
```diff
@@ -179,20 +179,37 @@ const MAX_MESSAGE_LENGTH = 500;
 
 function remoteLog(event,data) {
```
wdanilo, line 188: I also have question to the line above (doesn't need to be fixed in this PR, but Id love to know the answer). Is Mixpanel synchronous initialization the only option? (I hope not!). If the internet connection is slow, the whole GUI freezes on Mixpanel initialization :( We should instead buffer logs and send them after the initialization is done. Please either fix it or create issue for it.

src/js/lib/content/src/index.js
```diff
@@ -179,20 +179,37 @@ const MAX_MESSAGE_LENGTH = 500;
...
     } else {
         console.warn(`Failed to log the event '${event}'.`)
     }
 }
 
-window.enso.remote_log = remoteLog
+function register(data) {
+    if (mixpanel) {
+        if (data !== undefined) {
+            data = JSON.stringify(data).substr(0,MAX_MESSAGE_LENGTH)
```
wdanilo: I would add some indicator that the message was trimmed, like `+ ' ...'`.

src/js/lib/content/webpack.config.js
```diff
@@ -1,12 +1,20 @@
...
 const root     = path.resolve(thisPath,'..','..','..','..')
 const distPath = path.resolve(root,'dist')
 const wasmPath = path.resolve(distPath,'wasm')
 
+const child_process = require('child_process');
+function git(command) {
+    return child_process.execSync(`git ${command}`, { encoding: 'utf8' }).trim();
+}
+
+const BUILD_INFO = JSON.parse(require('fs').readFileSync('../../../../dist/build.json', 'utf8'));
```
wdanilo: I don't like doing it this way because of few reasons:
1. This should be either in a separate section, or should have lovercase name, like other file path constants above.
2. Instead of using such string, it is always recommended to use `path.resolve` instead. 
3. Why do we use so many `../` in this string, if the `root` path is already resolved above?

src/js/lib/content/webpack.config.js
```diff
@@ -28,6 +36,11 @@ module.exports = {
             path.resolve(thisPath,'src','style.css'),
             path.resolve(wasmPath,'ide.wasm'),
         ]),
+        new webpack.DefinePlugin({
+            GIT_HASH: JSON.stringify(git('rev-parse HEAD')),
+            GIT_STATUS: JSON.stringify(git('status --short --porcelain')),
+            BUILD_INFO: JSON.stringify(BUILD_INFO),
```
wdanilo, line 43: nice one

src/js/lib/content/src/index.js
```diff
@@ -179,20 +179,37 @@ const MAX_MESSAGE_LENGTH = 500;
...
+        console.warn(`Failed to register data '${data}'.`)
+    }
+}
+
+window.enso.remoteLog = remoteLog
+window.enso.register = register
 
 window.setInterval(() =>{remoteLog("alive");}, 1000 * 60)
 
+window.enso.remoteLog("git_hash", {hash: GIT_HASH})
```
wdanilo, line 207: Please add comment here where these magic variables come from (webpack plugin)

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,84 @@ pub use wasm_bindgen::prelude::*;
...
 
 /// Send the provided public event with data to our logging service.
-pub fn remote_log_data
-<T1:Loggable,T2:Loggable>(message: AnonymousData<T1>, data: AnonymousData<T2>) {
-    _remote_log(message.0.get(), data.0.get());
+pub fn remote_log_data_field
+<T:Loggable>(message:&str, field_name:&str, data:AnonymousData<T>) {
+    let msg = JsValue::from(message.to_string());
+    let field_name = JsValue::from(field_name.to_string());
+    _remote_log_data_field(msg, field_name, data.0.get());
```
wdanilo: alignment (space after `,`)

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,84 @@ pub use wasm_bindgen::prelude::*;
...
+}
+
+/// Send the provided public event with data to our logging service.
+pub fn remote_log
+<T:Loggable>(message:&str, data:AnonymousData<T>) {
+    _remote_log(JsValue::from(message.to_string()), data.0.get());
```
wdanilo: alignment

src/rust/ide/src/ide/integration.rs
```diff
@@ -324,19 +324,19 @@ impl Integration {
...
-            eval_ editor_outs.node_exited([]{analytics::remote_log(analytics::AnonymousData("graph_editor::node_exit_request"))});
-            eval_ editor_outs.on_edge_endpoints_set([]{analytics::remote_log(analytics::AnonymousData("graph_editor::edge_endpoints_set"))});
-            eval_ editor_outs.visualization_enabled([]{analytics::remote_log(analytics::AnonymousData("graph_editor::visualization_enabled"))});
-            eval_ editor_outs.visualization_disabled([]{analytics::remote_log(analytics::AnonymousData("graph_editor::visualization_disabled"))});
-            eval_ on_connection_removed([]{analytics::remote_log(analytics::AnonymousData("graph_editor::connection_removed"))});
-            eval_ searcher_frp.used_as_suggestion([]{analytics::remote_log(analytics::AnonymousData("searcher::used_as_suggestion"))});
-            eval_ project_frp.editing_committed([]{analytics::remote_log(analytics::AnonymousData("project::editing_committed"))});
+            eval_ editor_outs.node_editing_started([]{analytics::remote_log_event("graph_editor::node_editing_started")});
```
wdanilo, line 327: What hapenned with all the `AnonymousData` things here? We were designing them for a purpose. Why they are deleted? Also, why the name change from `remote_log` to `remote_log_event`?

wdanilo (approved):
I'm OK with merging it earlier, but I'd love to get answers / fixes to my comments (can be in separate PR)

build/run.js
MichaelMauderer: Removed.

src/rust/ide/lib/analytics/src/remote_log.rs
MichaelMauderer: This has been greatly reduced. Especially since one of the methods was unused. 
So, I'd opt for keeping it as a snippet in its new form.

src/rust/ide/lib/analytics/src/remote_log.rs
MichaelMauderer: Updated, with better naming now.

src/rust/ide/src/ide.rs
MichaelMauderer: Moved.

src/js/lib/content/src/index.js
MichaelMauderer: This is now only in one method, which is re-used to do all the logging.

src/js/lib/content/src/index.js
MichaelMauderer: Let's create an issue. I have not researched this deeply, but what we do here is the way described in the docs. At first, glance there seems to be no async alternative, so we'd have to wrap this ourselves.

src/js/lib/content/webpack.config.js
MichaelMauderer: Simplified to make use of the `root`.

src/rust/ide/src/ide/integration.rs
MichaelMauderer: The message title is not meant to be `data` but a simple static event name in the source code. So, it cannot contain sensitive data and does not need to be wrapped.

`remote_log_event` is used to indicate that only the presence of an event is logged, while data needs to be logged with `remote_log_value`

src/js/lib/content/src/index.js
MichaelMauderer: Added.

src/js/lib/content/src/index.js
wdanilo: I was browsing Mixpanel docs and it seems that their API should be async (with callbacks). So Im wondering why it happens. Anyway, it needs to be investigated, that's right. A hint - in chrome dev tools you can limit the internet speed per tab to see how it works with a low speed! Wpuld you be so nice and create issue for it, please?

src/rust/ide/src/ide/integration.rs
wdanilo: I don't "feel" it tbh. I mean, I understand the purpose, but in such a scenario I think we can remove `AnonymousData` altogether. IMO either we keep it secure by writing in docs it should be secure (like in case of `remote_log_event`), or we use some newtype machinery in code to be sure everybody remembers about it. I can think of theoretical event names that violate the anonymous data collection rules. I know this example is not perfect, but `"node_created_by_a_woman"` would violate it. And of course, we would catch it during code review, but we would probably catch it as well during code review in case of `remote_log_value`. I think that we should either use `AnonymousData` everywhere in the API or nowhere, but I'm of course very OK with chatting ore about it! CC @farmaazon , I'd love to hear your thoughts about it too.

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,44 @@ pub use wasm_bindgen::prelude::*;
...
+                window.enso.remoteLog(msg)
+            } else {
+                window.enso.remoteLog(msg,value)
+            }
+        } catch (error) {
+            console.error(\"Error while logging message. \" + error );
+        }
     } else {
         console.warn(\"Failed to send log message.\")
     }
```
wdanilo: I don't like one thing in this wrapper a lot. Imagine that we will refactor the code and rename the function `window.enso.remoteLog` to `window.luna.remoteLog` (or just anything else). This wrapper will silently fail and discovery during development why logs are not working anymore could take days. IMO, this should just be:

```js
try {
    window.enso.remoteLog(msg)
} catch (error) {
    console.error(\"Error while logging message. \" + error );
}
```

If `window.enso.remoteLog` is not present, it means that something is deeply broken and we should know about it.

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,44 @@ pub use wasm_bindgen::prelude::*;
...
+    _remote_log(msg,data)
+}
 ")]
 extern "C" {
-     #[allow(unsafe_code)]
-     fn _remote_log(msg:JsValue,data:JsValue);
+    #[allow(unsafe_code)]
+    fn _remote_log_value(msg:JsValue,field_name:JsValue,value:JsValue);
+    #[allow(unsafe_code)]
+    fn _remote_log(msg:JsValue,value:JsValue);
```
wdanilo: btw, why do we keep names with `_` on the beginning? We could just keep them in a local `js` module instead.

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,44 @@ pub use wasm_bindgen::prelude::*;
...
-    if (window !== undefined && window.enso !== undefined && window.enso.remote_log !== undefined) {
-        window.enso.remote_log(msg,{data:data})
+export function _remote_log(msg, value) {
+    if (window !== undefined && window.enso !== undefined && window.enso.remoteLog !== undefined) {
+        try {
+            if (value === undefined && value !== null) {
+                window.enso.remoteLog(msg)
+            } else {
+                window.enso.remoteLog(msg,value)
+            }
```
wdanilo: Wait, this can be simplified a lot - just use `window.enso.remoteLog(msg,value)` - `value` can be undefined, you don't need to check it!

src/rust/ide/lib/analytics/src/remote_log.rs
```diff
@@ -9,27 +9,44 @@ pub use wasm_bindgen::prelude::*;
...
+    fn _remote_log_value(msg:JsValue,field_name:JsValue,value:JsValue);
+    #[allow(unsafe_code)]
+    fn _remote_log(msg:JsValue,value:JsValue);
 }
 
 /// Send the provided public event to our logging service.
-pub fn remote_log<T:Loggable>(message: AnonymousData<T>) {
-    _remote_log(message.0.get(), JsValue::NULL);
+pub fn remote_log_event(message:&str) {
+    _remote_log(JsValue::from(message.to_string()),JsValue::UNDEFINED);
```
wdanilo: Why do you pass `JsValue::UNDEFINED` here? Just don't pass it. In JS it will automatically be undefined there!

src/rust/ide/src/lib.rs
```diff
@@ -92,6 +92,13 @@ pub fn entry_point_ide() {
 
     // FIXME: This code is temporary. It's used to remove the loader UI.
     ensogl_text_msdf_sys::run_once_initialized(|| {
+
+        // Logging of build information.
+        #[cfg(debug_assertions)]
+            analytics::remote_log_value("debug_mode", "debug_mode_is_active", analytics::AnonymousData("true"));
+        #[cfg(not(debug_assertions))]
+            analytics::remote_log_value("debug_mode", "debug_mode_is_active", analytics::AnonymousData("false"));
```
wdanilo: I'd change it to `analytics::remote_log_value("debug_mode","debug_mode_is_active",true);` - this should be simple to be allowed and we will log more and ore bool events everywhere.

src/rust/ide/view/graph-editor/src/component/visualization/container/visualization_chooser.rs
```diff
@@ -143,9 +143,11 @@ impl VisualizationChooser {
 
             eval frp.source.chosen_entry([](entry){
                 if let Some(entry) = entry{
-                    let event = analytics::AnonymousData("graph_editor::visualization_chooser::vis_selected");
-                    let data  = analytics::AnonymousData(|| format!("{:?}", entry.name));
-                    analytics::remote_log_data(event,data);
+                    let event     = "graph_editor::visualization_chooser::vis_selected";
+                    let name:&str =  entry.name.as_ref();
```
wdanilo: spacing

wdanilo (approved):

src/js/lib/content/src/index.js
MichaelMauderer: https://github.com/enso-org/ide/issues/1280

src/rust/ide/src/ide/integration.rs
MichaelMauderer: I was considering to require a `&'static str` which almost enforces the use of a string literal here. How about we create an issue for the discussion about proper refactoring and handling of this?

farmaazon (approved):

src/rust/ide/lib/analytics/src/remote_log.rs
MichaelMauderer: Because the Rust wrapper requires it. We could probably also add a new wrapper method that does not need it, but since none of that is exposed outside this file it doesn't really matter, I think. This means we have the more flexible method and avoid having another JS wrapper.

PR #1328, opened 2021-03-15
Add --no-data-gathering as a startup option

s9ferech:
### Pull Request Description

This solves issue #1303. The option will disable sending any data to Mixpanel.

### Important Notes

There will still be a request to Mixpanel during initialisation (src/js/lib/content/index.js, line 175). No user specific data is transmitted during this initialisation, but we could circumvent it as well, if it it a problem.

### Checklist

- [ ] ~The documentation has been updated if necessary.~
- [ ] ~All code conforms to the [Rust](https://github.com/enso-org/enso/blob/main/docs/style-guide/rust.md) style guide.~
- [ ] ~All code has been manually tested by the PR owner against our [test scenarios](https://docs.google.com/spreadsheets/d/1RatJDM_f9_3bvYhl3Bpq2d8SyKgtVdrV1RkGxPU17c8/edit?ts=5faa7049#gid=0).~


src/js/config.js
```diff
@@ -1,2 +1,6 @@
+/**
+ * Configuration options that are only used by the JavaScript part of the project
```
wdanilo: Docs need to be sentences - missing dot.

src/js/lib/client/src/index.js
```diff
@@ -172,7 +172,12 @@ optParser.options('crash-report-host', {
     describe    : 'The address of the server that will receive crash reports. ' +
                   'Consists of a hostname, optionally followed by a ":" and a port number',
     requiresArg : true,
-    default: cfg.defaultLogServerHost
+    default     : cfg.defaultLogServerHost
+})
+
+optParser.options('no-data-gathering', {
+    describe    : 'Disable the sharing of any usage data with the Enso developers',
```
wdanilo: I would name it "Disable anonymous data collection" - it's shorter, a more popular name for that, and also doesn't mention "Enso" (we have a rule that we do not use product name too often if that's possible in order to keep the messages shorter + code easier to refactor).

src/js/lib/content/src/index.js
```diff
@@ -450,18 +429,28 @@ function ok(value) {
...
+        noDataGathering: false,
     }
     let urlParams = new URLSearchParams(window.location.search);
     let urlConfig = Object.fromEntries(urlParams.entries())
     let config    = Object.assign(defaultConfig,inputConfig,urlConfig)
     API[globalConfig.windowAppScopeConfigName] = config
 
-    //initCrashHandling()
+    API.remoteLog = config.noDataGathering
+                    ? function () {}
```
wdanilo: create var `let remoteLogDisabled = () => {}` instead the line above. We are trying to minimize the number of multi-line expressions in the codebase

src/js/lib/content/src/index.js
```diff
@@ -450,18 +429,28 @@ function ok(value) {
...
+                    : remoteLog
+    window.setInterval(() =>{API.remoteLog("alive");}, 1000 * 60)
```
wdanilo: please expose the `60 seconds` as a constant variable on top of the file (in constants section)

src/js/lib/log-server/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "scripts": {
+      "start": "node server.js",
+      "test": "mocha"
+    },
+    "devDependencies": {
+      "axios": "^0.21.0",
+      "mocha": "^8.2.1",
+      "mock-fs": "^4.13.0"
+    }
+  }
```
wdanilo: no nl

src/js/lib/log-server/package.json
```diff
@@ -0,0 +1,31 @@
+{
```
wdanilo, line 1: I don't understand this change. Why this file is added now / why it was missing before? This PR doesn't seem to have any connection to this file. (This question is the only blocker from merging it to me)

wdanilo (changes requested):
I have one blocker question, beside that, it looks fine. I'd feel better if also @MichaelMauderer would review the logic, as he was implementing it, and maybe there are some things that I overlooked.

src/js/lib/client/src/index.js
MichaelMauderer: "Disable anonymous data collection" could be misleading, though. It could imply that non-anonymous data gathering is happening.

MichaelMauderer (commented):
> There will still be a request to Mixpanel during initialisation (src/js/lib/content/index.js, line 175). No user specific data is transmitted during this initialisation, but we could circumvent it as well, if it it a problem.

I think it would be great if this could also disable the initial connection to mixpanel. That would ensure that not outgoing connection is required when using this cmd option. Doing an initial connection can delay the IDE startup in no- or low-connectivity environments.

src/js/lib/client/src/index.js
s9ferech: I changed it to "Disable anonymous data collection". I assume it will be clear enough.

src/js/lib/log-server/package.json
s9ferech: Indeed, the file is not directly related to the purpose of this PR. I added it, so I can test if the crash reporting still works.

The file existed before but was deleted by accident, as far as I can tell. Or it was deleted with a good reason that I currently don't understand. I am not very familiar with Lerna. Is it correct that this package should have its own `package.json`? Or should that configuration go somewhere else?

src/js/lib/client/src/index.js
wdanilo: @MichaelMauderer oh good point. @s9ferech please confirm it with @iamrecursion - she always knows how to write sentences that are not ambiguous (maybe this one is correct, just confirm it with Ara please)

src/js/lib/log-server/package.json
wdanilo: no nl

src/js/lib/log-server/package.json
wdanilo: Im surprised the file was not there tbh. Could you please confirm it works before merging this PR? I mean - could you confirm this server compiles and runs based on this file config?

wdanilo (approved):

src/js/lib/log-server/package.json
s9ferech: I was also surprised. Maybe it got lost because we have `package.json` in our `.gitignore`.

The server does run with that config.

src/js/lib/client/src/index.js
iamrecursion: No. That one is definitely ambiguous. I would go with the one that is here originally.

PR #1335, opened 2021-03-16
Database Visualizations

radeusgd:
### Pull Request Description

Implements visualizations for the Database library:
- the Table visualization is enhanced to be able to automatically materialize database tables (so now it supports both dataframes and databases).
  - It also limits the output to at most 1000 elements and displays the count of hidden elements below the table.
- the SQL visualization that pretty prints the query with interpolated values.

### Important Notes

- This PR should be merged after the backing changes, that is PR https://github.com/enso-org/enso/pull/1582 (and https://github.com/enso-org/enso/pull/1595) is released so that the IDE can see these changes within the engine that it is using.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,10 @@
   that the connection was lost and IDE must be restarted. In future IDE will try
   to automatically reconnect.
 - [Database Visualizations][1335]. Visualizations for the Database library have
-  been added: the Table visualization now automatically executes the underlying
-  query to display its results in a table and a SQL Query visualization allows
-  to see the query that is going to be run in the database.
+  been added. The Table visualization now automatically executes the underlying
+  query to display its results in a table. In addition, the SQL Query
+  visualization allows the user to see the query that is going to be run against
+  the database.
 
 <br/>![Bug Fixes](/docs/assets/tags/bug_fixes.svg)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   will be lost. In this build we added notification in statusbar to signalize
   that the connection was lost and IDE must be restarted. In future IDE will try
   to automatically reconnect.
+- [Database Visualizations][1335]. Visualizations for the Database library have
+  been added: the Table visualization now automatically executes the underlying
+  query to display its results in a table and a SQL Query visualization allows
+  to see the query that is going to be run in the database.
 
 <br/>![Bug Fixes](/docs/assets/tags/bug_fixes.svg)
 
--- a/src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
+++ b/src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
@@ -1,0 +1,97 @@
+loadScript('https://cdnjs.cloudflare.com/ajax/libs/sql-formatter/4.0.2/sql-formatter.min.js')
+
+class SqlVisualization extends Visualization {
+    static inputType = "Any"
+    static label = "SQL"
+
+    constructor(api) {
+        super(api)
+        this.setPreprocessorModule('Standard.Base')
+        this.setPreprocessorCode(`
+           x ->
+               stmt = x.to_sql.prepare
+               dialect = x.connection.dialect.name
+               Json.from_pairs [['sql', stmt.first], ['interpolations', stmt.second], ['dialect', dialect]] . to_text
+        `)
+    }
+
+    onDataReceived(data) {
+        while (this.dom.firstChild) {
+            this.dom.removeChild(this.dom.lastChild);
+        }
+        let parsedData = data
+        if (typeof data === 'string') {
+            parsedData = JSON.parse(data)
+        }
+
+        const style = `
+        <style>
+        span {
+            padding-left: 3px;
+            padding-right: 3px;
+            padding-top: 1px;
+            padding-bottom: 1px;
+            border-radius: 5px;
+        }
+        pre {
+            margin-left: 12px;
+        }
+        .string {
+            background-color: rgba(255, 100, 200, 0.8);
+        }
+        .number {
+            background-color: rgba(120, 255, 120, 0.8);
+        }
+        .boolean {
+            background-color: rgba(110, 110, 240, 0.8);
+        }
+        .other {
+            background-color: rgba(200, 200, 200, 0.8);
+        }
+        </style>
+        `
+
+        function renderParameter(param) {
+            const item = param[0]
+            if (typeof item === 'string') {
+                return '<span class="string">"' + item + '"</span>'
+            } else if (typeof item === 'number') {
+                return '<span class="number">' + item + '</span>'
+            } else if (typeof item === 'boolean') {
+                return '<span class="boolean">' + item + '</span>'
+            } else {
+                return '<span class="other">' + JSON.stringify(item) + '</span>'
+            }
+        }
+
+        const params = parsedData.interpolations.map(renderParameter)
+        let language = "sql"
+        if (parsedData.dialect == "postgresql") {
+            language = "postgresql"
+        }
+        const formatted = sqlFormatter.format(parsedData.sql, {params: params, language: language})
+
+        const width = this.dom.getAttributeNS(null, 'width')
+        const height = this.dom.getAttributeNS(null, 'height')
+        const elem = document.createElement('div')
+        elem.setAttributeNS(null, 'id', 'vis-sql-view')
+        elem.setAttributeNS(null, 'class', 'scrollable')
+        elem.setAttributeNS(null, 'viewBox', '0 0 ' + width + ' ' + height)
+        elem.setAttributeNS(null, 'width', '100%')
+        elem.setAttributeNS(null, 'height', '100%')
+        const viewStyle = `width: ${width - 10}px;
+             height: ${height - 10}px;
+             overflow: scroll;
+             padding:2.5px;`
+        elem.setAttributeNS(null, 'style', viewStyle)
+        elem.innerHTML = style + '<pre>' + formatted + '</pre>'
+        this.dom.appendChild(elem)
+    }
+
+    setSize(size) {
+        this.dom.setAttributeNS(null, "width", size[0])
+        this.dom.setAttributeNS(null, "height", size[1])
+    }
+}
+
+return SqlVisualization
--- a/src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
+++ b/src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
@@ -1,268 +1,88 @@
 loadScript('https://cdnjs.cloudflare.com/ajax/libs/sql-formatter/4.0.2/sql-formatter.min.js')
 
+/**
+ * A visualization that pretty-prints generated SQL code and displays type hints related to
+ * interpolated query parameters.
+ */
 class SqlVisualization extends Visualization {
-    static inputType = 'Any' // 'Standard.Database.Data.Table.Table | Standard.Database.Data.Column.Column' // TODO change this once sum types are properly supported
+    // TODO Change the type below once #837 is done:
+    // 'Standard.Database.Data.Table.Table | Standard.Database.Data.Column.Column'
+    static inputType = 'Any'
     static label = 'SQL Query'
 
     constructor(api) {
         super(api)
         this.setPreprocessorModule('Standard.Visualization.Sql.Visualization')
         this.setPreprocessorCode(`x -> here.prepare_visualization x`)
-
-        // mock theme
-        function hash(s) {
-            let sum = 0
-            for (let i = 0; i < s.length; ++i) {
-                sum = (sum + 31 * s.charCodeAt(i)) % 255
-            }
-            return sum
-        }
-        function getColor(name) {
-            return [hash('r' + name), hash('g' + name), hash('b' + name), 1]
-        }
-        this.theme = {
-            getColorForType: getColor,
-            getForegroundColorForType: function (x) {
-                return [255, 255, 255, 1]
-            },
-        }
+        this.theme = themeMock // TODO remove once themes are merged (see below)
     }
 
     onDataReceived(data) {
         while (this.dom.firstChild) {
             this.dom.removeChild(this.dom.lastChild)
         }
+
         let parsedData = data
         if (typeof data === 'string') {
             parsedData = JSON.parse(data)
         }
 
-        const style = `
-        <style>
-        .sql {
-            font-family: DejaVuSansMonoBook, sans-serif;
-            font-size: 12px;
-            margin-left: 7px;
-        }
-        .interpolation {
-            border-radius: 6px;
-            padding:1px 2px 1px 2px;
-            display: inline;
-        }
-        .mismatch-parent {
-            position: relative;
-            display: inline-flex;
-            justify-content: center;
-        }
-        .mismatch-mouse-area {
-            display: inline;
-            position: absolute;
-            width: 150%;
-            height: 150%;
-            align-self: center;
-            z-index: 0;
-        }
-        .mismatch {
-            z-index: 1;
-        }
-        .modulepath {
-            color: rgba(150, 150, 150, 0.9);
-        }
-        .tooltip {
-            font-family: DejaVuSansMonoBook, sans-serif;
-            font-size: 12px;
-            opacity: 0;
-            transition: opacity 0.2s;
-            display: inline-block;
-            white-space: nowrap;
-            background-color: rgba(249, 249, 249, 1);
-            box-shadow: 0 0 16px rgba(0, 0, 0, 0.16);
-            text-align: left;
-            border-radius: 6px;
-            padding: 5px;
-            position: absolute;
-            z-index: 99999;
-            pointer-events: none;
-        }
-        </style>
-        `
-
-        function renderColor(color) {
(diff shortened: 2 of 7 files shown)


radeusgd:
Below we can see an example with some filtering applied to the table. The Table visualization displays the materialized results:
image

Alternatively we can switch to the SQL visualization which shows the query that will be used to compute the results:
image

The interpolated values that come from Enso are highlighted and their color is based on their type (currently only basic types are supported and that is inferred from their Enso type, not from SQL - we have access to the SQL type so maybe we should switch to that - the only issue is that the representation of SQL types is based on implementation-defined numeric identifiers for each type which I didn't want to hardcode, but if we want to switch to this, I could approximate the datatype name on the backend side and send it as text to the visualization).

> _Note for clarity:_ the `jdbc:` prefix when opening a DB connection is not necessary, skipping it will also work.

wdanilo:
Looks cool. Let's have a video call with @sylwiabr to see how it works and to chat about missing things. Great job with that!

wdanilo:
We are waiting for this to be finished.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,274 @@
+loadScript('https://cdnjs.cloudflare.com/ajax/libs/sql-formatter/4.0.2/sql-formatter.min.js')
+
+class SqlVisualization extends Visualization {
+    static inputType = 'Any' // 'Standard.Database.Data.Table.Table | Standard.Database.Data.Column.Column' // TODO change this once sum types are properly supported
```
wdanilo: 1. Too long line, max 100 chars everywhere - apply to ALL files
2. In comments always add links to issues describing them.
3. Comments should be valid english sentences. missing dot

Apply to all your code.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,274 @@
...
+    static label = 'SQL Query'
+
+    constructor(api) {
+        super(api)
+        this.setPreprocessorModule('Standard.Visualization.Sql.Visualization')
+        this.setPreprocessorCode(`x -> here.prepare_visualization x`)
+
+        // mock theme
```
wdanilo: comments should be valid sentences. Also, they should be understandable by people not familiar with codebase, thisdefinitely is not. It needs to describe why it is a mock and what issue it is waiting for with a link to the issue.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,274 @@
...
+    onDataReceived(data) {
+        while (this.dom.firstChild) {
+            this.dom.removeChild(this.dom.lastChild)
+        }
+        let parsedData = data
+        if (typeof data === 'string') {
+            parsedData = JSON.parse(data)
+        }
+
+        const style = `
```
wdanilo: Do you need to recreate this string on every data received? Please refacotr all parts of the codebase that are not needed to be recomputed on new data, so this will work faster

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,274 @@
...
+        const theme = this.theme
+
+        function splitTypeName(name) {
+            var ix = name.lastIndexOf('.')
+            if (ix < 0) {
+                return ['', name]
+            }
+
+            return [name.substr(0, ix + 1), name.substr(ix + 1)]
+        }
```
wdanilo: all these functions should have comments describing their behavior. Or better naming. If I see name `splitTypeName` I do not understand what does it do - split from what, into what? Better naming is preferred. IF its not possible, musch better and detailed docs. Thesame applies to other functions, like a super-long function `rendeParameter`. Its namedoesnt tell anything about what it does, and is looks complex.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,274 @@
...
+            const actualType = param.actual_type
+            let value = param.value
+
+            if (actualType == 'Builtins.Main.Text') {
+                value = "'" + value.replaceAll("'", "''") + "'"
+            }
+
+            const actualTypeColor = theme.getColorForType(actualType)
+            const fgColor = actualTypeColor
+            let bgColor = [...fgColor.slice(0, 3), bgAlpha]
```
wdanilo: magic numbers. Do not use literals like that. Extract them to well named variables. You can use `0` as it means start, but `3` is magical

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,274 @@
...
+                const received = splitTypeName(actualType)
+                const expected = splitTypeName(expectedType)
+                let message =
+                    'Received <span class="modulepath">' +
+                    received[0] +
+                    '</span><span style="color: ' +
+                    renderColor(actualTypeColor) +
+                    '">' +
+                    received[1] +
+                    '</span><br>'
```
wdanilo: these are VERY complex expressions. All such multi-line expressions (in the whole new code) please split into well named variables - semantically named. So do `let someMeaningfulNAme = received[0]` etc.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,274 @@
...
+        function splitTypeName(name) {
+            var ix = name.lastIndexOf('.')
+            if (ix < 0) {
+                return ['', name]
+            }
+
+            return [name.substr(0, ix + 1), name.substr(ix + 1)]
+        }
+
+        function renderParameter(param) {
```
wdanilo: Functions should have 50 lines of code. Max 100. In case of bigger functions, split them down into well named sub-functions. This one is HUGE

wdanilo:
I'm not sure regarding the code quality - the functions are super long, lack of docs in many places, long expressions are not split into variables etc. I would like @farmaazon to review it from a perspective of someone that does not know what this code does and leave comments - this code should be understandable to every reader easily, without browisng connected code - the logic should be easy to follow. I feel it is not. Ada, would you mind looking at it please?

radeusgd:
> I'm not sure regarding the code quality - the functions are super long, lack of docs in many places, long expressions are not split into variables etc. I would like @farmaazon to review it from a perspective of someone that does not know what this code does and leave comments - this code should be understandable to every reader easily, without browisng connected code - the logic should be easy to follow. I feel it is not. Ada, would you mind looking at it please?

I think I could refactor it a bit to make the functions a bit smaller and add docs before a re-review?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
radeusgd: Added comments and changed names a bit.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
radeusgd: This mock theme will be removed anyway before the PR is merged, as we discussed that this PR should be merged after #1337

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
radeusgd: Ok, I changed this to `const baseChannels = 3` since that `3` represents the `RGB` channels that are kept when changing opacity; it's also now a separate function so it is easier to see what it does.

I assume that I can use `2` without treating it like a magic number when I'm dividing things by two, right? (Like I want to compute the middle of something, so I have `width / 2`)

radeusgd:
Some screenshots on what the visualizations look like:

radeusgd:
Thanks to #1337 we now have proper colors!

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
   will be lost. In this build we added notification in statusbar to signalize
   that the connection was lost and IDE must be restarted. In future IDE will try
   to automatically reconnect.
+- [Database Visualizations][1335]. Visualizations for the Database library have
+  been added: the Table visualization now automatically executes the underlying
+  query to display its results in a table and a SQL Query visualization allows
+  to see the query that is going to be run in the database.
```
wdanilo: confirm this sentecnes with Ara. I think there are errors like "a SQL" instead of "an SQL".

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
+loadScript('https://cdnjs.cloudflare.com/ajax/libs/sql-formatter/4.0.2/sql-formatter.min.js')
```
wdanilo, line 72: no file docs

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+ * interpolated query parameters.
+ */
+class SqlVisualization extends Visualization {
+    // TODO Change the type below once #837 is done:
+    // 'Standard.Database.Data.Table.Table | Standard.Database.Data.Column.Column'
+    static inputType = 'Any'
+    static label = 'SQL Query'
+
+    constructor(api) {
+        super(api)
```
wdanilo, line 89: thats right! Our API is super <3 Everyone should know that!

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+        this.setPreprocessorModule('Standard.Visualization.Sql.Visualization')
+        this.setPreprocessorCode(`x -> here.prepare_visualization x`)
+    }
+
+    onDataReceived(data) {
+        while (this.dom.firstChild) {
+            this.dom.removeChild(this.dom.lastChild)
+        }
```
wdanilo: refacto to "reoveAllChildren"

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+
+            let language = 'sql'
+            if (parsedData.dialect == 'postgresql') {
+                language = 'postgresql'
+            }
+
+            const formatted = sqlFormatter.format(parsedData.code, {
+                params: params,
+                language: language,
+            })
```
wdanilo, line 118: this line would probably fail if parsedData is for example `undefined`, `null`, or a number. We should make our visualizations as error prone as possible, so please, make more checks regarding the input data!

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+
+            const codeRepresentation = '<pre class="sql">' + formatted + '</pre>'
+            visHtml += codeRepresentation
+        }
+
+        const containers = this.createContainers()
+        const parentContainer = containers[0]
+        const scrollable = containers[1]
```
wdanilo: Why this function returns an array instead of an object? This looks very error prone to me.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+             padding:2.5px;`
+        scrollable.setAttributeNS(null, 'style', viewStyle)
+        parentContainer.appendChild(scrollable)
+        return [parentContainer, scrollable]
+    }
+
+    setSize(size) {
+        this.dom.setAttributeNS(null, 'width', size[0])
+        this.dom.setAttributeNS(null, 'height', size[1])
+    }
```
wdanilo, line 176: again, why such strangedesign with arrays instead of objects?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/**
+ * Splits a qualified type name into a module prefix and the typename itself.
+ */
+function splitQualifiedTypeName(name) {
```
wdanilo: I would name it `splitModuleNae` instead and return object containing field `moduleName` instead of "prefix". This function name sounds like it splits qualified type name, and it does not do it.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+    if (ix < 0) {
+        return {
+            prefix: '',
+            name: name,
+        }
+    }
+
+    return {
+        prefix: name.substr(0, ix + 1),
+        name: name.substr(ix + 1),
```
wdanilo: is this correct? Does one of the parts contain "." ? If so, this should not be contained in one of the parts, as this is surprising design

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+    }
+}
+
+/**
+ * Renders a 4-element array representing a color into a CSS-compatible rgba string.
+ */
+function renderColor(color) {
```
wdanilo: this is not rendering. Rendering is rasterizing or biuilding visual representation (like DO structure later in this file). this sohuld be named `convertArrayToRgba`

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+    const g = 255 * color.green
+    const b = 255 * color.blue
+    const a = color.alpha
+    return 'rgba(' + r + ',' + g + ',' + b + ',' + a + ')'
+}
+
+/** Changes the alpha component of a color (represented as a 4-element array),
+ * returning a new color.
+ */
+function changeAlpha(color, newAlpha) {
```
wdanilo: "change" sounds like it accepts lambda. This should be named "replaceAlpha".

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
...
+    message += 'Expected ' + expectedPrefix + expectedSuffix + '<br>'
+    message += 'The database may perform an auto conversion.'
+    return message
+}
+
+const textType = 'Builtins.Main.Text'
+const customSqlTypePrefix = 'Standard.Database.Data.Sql.Sql_Type.'
+
+/** Specifies opacity of interpolation background color. */
+const interpolationBacgroundOpacity = 0.3
```
wdanilo: all such things should be moved to "Constants" section on top of the file

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
```diff
@@ -0,0 +1,395 @@
+loadScript('https://cdnjs.cloudflare.com/ajax/libs/sql-formatter/4.0.2/sql-formatter.min.js')
+
+/**
```
wdanilo, line 78: Also, this file is huuuuge and does not contain sections. It should contain at least few sections.

wdanilo (approved):
This code is MUCH nicer. I was able to create a proper review of it now! Few inor things, Im accepting it. After fixing them, feel free to merge it <3

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
radeusgd: Well, `parsedData` is what is returned by the visualization preprocessor which is controlled by the visualization itself, so I think we have control over what `parsedData` should look like.

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
radeusgd: This is a function that every `Visualization` must override provided from the Rust API, so it is not my design :) It is the same as in most other visualizations. Interestingly though, now I have noticed that the documentation indeed states that this is an object with `width` and `height` but I have checked the Rust code and it seems to pass arrays.

I guess it should be filed as an issue?

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/sql.js
radeusgd: I called it based on what it takes instead of what it returns.

I was thinking about `moduleName` result, but please note that the `prefix` also includes the trailing dot (so `prefix = "Foo.Bar."` and for `moduleName` that dot would not really be part of a module name). This design is simpler because I need to split just to be able to apply different styles to these componenents and so I want to have this dot as part of the prefix. If I didn't return the prefix with the dot, I would need to separately handle the case of an unqualified name which has no prefix (to avoid adding an unnecessary dot there), so I think it would slightly complicate the design.

PR #1348, opened 2021-03-18
Fix pasting multiline code in single-lined text area

s9ferech:
### Pull Request Description

When pasting multiline code in a single-lined text area, only the first line will be inserted.

This fixes issue #1336.

### Important Notes

This seems to duplicate some logic from `key_to_string` on text areas, which also handles newlines specially in single-line situations. Maybe they should be combined in some way?

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,9 +43,9 @@
 - [Handle syntax errors in custom-defined visualizations][#1341]. The IDE is now
   able to run properly, even if some of the custom-defined visualisations inside
   a project contain syntax errors.
-- [Fix pasting multiline code in nodes][1348]. Since nodes contain only a
-  single line, the first copied line will be inserted and all additional lines
-  will be ignored.
+- [Fix pasting multiline code in nodes][1348]. Since nodes contain only a single
+  line, the first copied line will be inserted and all additional lines will be
+  ignored.
 
 #### EnsoGL (rendering engine)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,8 +43,8 @@
 - [Handle syntax errors in custom-defined visualizations][#1341]. The IDE is now
   able to run properly, even if some of the custom-defined visualisations inside
   a project contain syntax errors.
-- [Fix pasting multiline code in nodes][1348]. Since nodes contain only a single
-  line, the first copied line will be inserted and all additional lines will be
+- [Fix issues with pasting multi-line text into single-line text fields][1348].
+  The first copied line will be inserted and all additional lines will be
   ignored.
 
 #### EnsoGL (rendering engine)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@
 - [Disable area selection][1318]. The area selection was visible despite being
   non-functional. To avoid confusion, area selection has been disabled until it
   is [correctly implemented][479].
+- [Fix pasting multiline code in nodes][1348]. Since nodes contain only a
+  single line, the first copied line will be inserted and all additional lines
+  will be ignored.
 
 #### EnsoGL (rendering engine)
 
--- a/src/rust/ensogl/lib/text/src/component/area.rs
+++ b/src/rust/ensogl/lib/text/src/component/area.rs
@@ -569,7 +569,7 @@ impl Area {
             eval_ sels_cut (m.buffer.frp.delete_left());
 
             eval_ input.paste (m.paste());
-            eval input.paste_string([m](s) m.paste_string(s));
+            eval input.paste_string((s) m.paste_string(s));
 
 
             eval_ m.buffer.frp.text_change (m.redraw(true));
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -43,6 +43,9 @@
 - [Handle syntax errors in custom-defined visualizations][#1341]. The IDE is now
   able to run properly, even if some of the custom-defined visualisations inside
   a project contain syntax errors.
+- [Fix pasting multiline code in nodes][1348]. Since nodes contain only a single
+  line, the first copied line will be inserted and all additional lines will be
+  ignored.
```
wdanilo: The changelog description should be more user-facing. This is a technical description about what happens under the hood. Instead, I'd go in the direction of "[Fixed issues with pasting multi-line text into single-line text fields.][...]" But also, please confirm with any native speaker (Joe, Ara), that this sentence sounds good (there is also a small grammar problem in yours if I'm correct - the "only" should be put at the end of the first part of the sentence). It's important that our public communication is 100% correct English.

src/rust/ensogl/lib/text/src/component/area.rs
```diff
@@ -569,7 +569,16 @@ impl Area {
...
             eval_ input.paste (m.paste());
-            eval input.paste_string ((s) m.buffer.frp.paste(m.decode_paste(s)));
+            eval input.paste_string([m](s) {
+                let mut fragments = m.decode_paste(s);
+                if m.single_line.get() {
+                    fragments = fragments
+                        .iter()
+                        .map(|fragment| fragment.lines().next().unwrap_or("").to_string())
+                        .collect();
+                }
```
wdanilo: Few things:

1. I'm not sure the logic is correct. What are "Fragments"? You are iterating over fragments and for each fragment you are taking the first line. So there may still be multiple lines there from different fragments?
2. This code is really hard to read. Of course its easy when you know what it is doing, but the code should be structured in such a way, that when reading the logic, you should look at a code fragment and after 1-2 seconds you should know what it is doing. Here, you should refactro this logic to a function that is called `dropAllButFirstLine` and use it instead.
3. According to style guide this chained multiline expression should be divided into multiple, well named variables that describe the steps used there. But also, just FYI, you don't have to use the full name `fragment` in variable name, you can use single letter there, as `fragments.iter(|f| ...)` is very easy to read and very understandable.

CHANGELOG.md
s9ferech: I would suggest that a native speaker goes over all the texts once before the release. Does that make sense?

src/rust/ensogl/lib/text/src/component/area.rs
s9ferech: 1. I renamed `fragments` to `chunks`, which seems to be consistent with other parts of the code. I assume you get multiple chunks when you copy text from a table and Enso would only use all of them if you have multiple cursors. Is that right?
2. & 3. I refactored it.

CHANGELOG.md
wdanilo: Not really, the README is a publicly available document on our GitHub. People will be reading that before release.

src/rust/ensogl/lib/text/src/component/area.rs
wdanilo: 1. So how does it work when pasting fro table with multiline texts?

CHANGELOG.md
s9ferech: Okay. I updated the text anyway, so the "only" is not there anymore.

src/rust/ensogl/lib/text/src/component/area.rs
s9ferech: Before this PR, it pasted one full table row per cursor, including all cells of that row and all possible new-line characters inside those cells.

Now it still pastes one row per cursor, but truncates each row at its first line-break character, which can appear in any cell or not at all. It will still paste multiple cells per row though.

wdanilo: Oh, this is per-cursor! That is what I was asking about. Wuld you be so nice and add there docs abut that, so no one will forget how ti works in the future, please?

wdanilo (approved):

src/rust/ensogl/lib/text/src/component/area.rs
```diff
@@ -569,7 +569,7 @@ impl Area {
             eval_ sels_cut (m.buffer.frp.delete_left());
 
             eval_ input.paste (m.paste());
-            eval input.paste_string ((s) m.buffer.frp.paste(m.decode_paste(s)));
+            eval input.paste_string([m](s) m.paste_string(s));
```
wdanilo: you don't need `[m]`. Our FRP rules look if your expression starts with `foo.` if it starts, then they use `[foo]` implicitly if not provided explicitly.

src/rust/ensogl/lib/text/src/component/area.rs
```diff
@@ -954,10 +954,30 @@ impl AreaModel {
...
+    /// line.
+    fn paste_string(&self, s: &str) {
+        let mut chunks = self.decode_paste(s);
+        if self.single_line.get() {
+            for f in &mut chunks {
+                Self::drop_all_but_first_line(f);
+            }
+        }
+        self.buffer.frp.paste(chunks);
+    }
```
wdanilo, line 971: Awesome <3

wdanilo (approved):

PR #1355, opened 2021-03-22
Fullscreen Visualization

farmaazon:
### Pull Request Description
This PR enables again the shortcut for FS visualization. It also contains a lot of fixes in visualizations

### Important Notes
* Scatterplot and histogram are redrawn after each resize.
* Most visualizations don't keep zoom and pan after going fullscreen. Fixing that is planned for https://github.com/enso-org/ide/issues/1354

### Checklist


src/js/lib/content/src/style.css
```diff
@@ -20,6 +20,9 @@
 .visualization {
     z-index: 2;
     border-radius: 14px;
+    /* visualizations are put into the fullscreen-vis layer which has pointer-events set to none, so we need to
+       override it */
+    pointer-events: auto;
 }
```
wdanilo: is the `visualization` class attached to fullscreen visualizations only? If so, it has a bad name.

src/rust/ensogl/lib/core/src/display/object/class.rs
```diff
@@ -342,22 +342,26 @@ impl<Host> Model<Host> {
...
 
-    fn take_removed_children_and_hide_orphans(&self, host:&Host) {
+    fn take_removed_children_and_update_their_visibility(&self, host:&Host) {
         if self.dirty.removed_children.check_all() {
             debug!(self.logger, "Updating removed children.", || {
                 for child in self.dirty.removed_children.take().into_iter() {
                     if let Some(child) = child.upgrade() {
-                        if child.is_orphan() {
+                        if !child.has_visible_parent() {
                             child.set_vis_false(host);
```
wdanilo, line 360: Wouldnt this change call "set_vis_false" on a child that was already hidden sometimes? For example, if a hidden child was moved to a visible parent, which just got hidden the next frame? If this could happen, this is blocker from merging. I'd love to chat why it never happens (if it never happens). We should add tests about it.

src/rust/ensogl/lib/core/src/display/scene.rs
```diff
@@ -458,9 +458,12 @@ impl Dom {
 #[derive(Clone,CloneRef,Debug)]
 pub struct DomLayers {
     /// Back DOM scene layer.
-    pub back: DomScene,
+    pub back : DomScene,
+    /// Back DOM with fullscreen visualization. Kept separately from `back`, because use different
+    /// Camera
```
wdanilo: this has several problems:
1. It's not gramatically correct ('use')
2. does not have dot at the end.
3. States that this is a "back DOM", which is not specific enough, as this is a back DOM **scene**.
4. Does not explain what a "different camera" is and why do we use a different camera here. It should be explained here soduring future refactoring it would be clear what it is about.

src/rust/ensogl/lib/core/src/display/scene.rs
```diff
@@ -469,25 +472,30 @@ pub struct DomLayers {
...
-        canvas.set_style_or_warn("z-index"       , "1"       , &logger);
+        canvas.set_style_or_warn("z-index"       , "2"       , &logger);
         canvas.set_style_or_warn("pointer-events", "none"    , &logger);
         front.dom.set_class_name("front");
         front.dom.set_style_or_warn("z-index", "1", &logger);
         back.dom.set_class_name("back");
         back.dom.set_style_or_warn("pointer-events", "auto", &logger);
         back.dom.set_style_or_warn("z-index"       , "0"   , &logger);
+        fullscreen_vis.dom.set_class_name("fullscreen_vis");
+        // fullscreen_vis.dom.set_style_or_warn("pointer-events", "auto", &logger);
```
wdanilo: commented code

src/rust/ensogl/lib/core/src/display/scene.rs
```diff
@@ -614,7 +622,7 @@ pub struct HardcodedLayers {
 
     pub viz_fullscreen : Layer,
     pub breadcrumbs    : Layer,
-    layers             : Layers,
+    layers             : Layers,
```
wdanilo: You left spaces on the end of the line, which is strange, as IDEs should trim it on save

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -689,8 +689,16 @@ class Histogram extends Visualization {
      * Sets size of the main parent DOM object.
      */
     setSize(size) {
+        console.warn('setting size', size)
```
wdanilo: Such things should not be left in a code

src/rust/ide/view/graph-editor/src/component/visualization/container/fullscreen.rs
```diff
@@ -0,0 +1,99 @@
...
+// === Shapes ===
+// ==============
+
+/// Container background shape definition.
+///
+/// Provides a backdrop and outline for visualisations. Can indicate the selection status of the
+/// container.
+/// TODO : We do not use backgrounds because otherwise they would overlap JS
+///        visualizations. Instead we added a HTML background to the `View`.
+///        This should be further investigated while fixing rust visualization displaying. (#526)
```
wdanilo, line 27: Does it mean that this code in the `background` module is not used anywhere now?

src/rust/ide/view/graph-editor/src/component/visualization/container/fullscreen.rs
```diff
@@ -0,0 +1,99 @@
...
+    // background     : background::View,
+    pub background_dom : DomSymbol
+}
+
+impl Panel {
+    /// Constructor.
+    pub fn new(logger:&Logger, scene:&Scene) -> Self {
+        let logger         = Logger::sub(logger,"fullscreen_view");
+        let display_object = display::object::Instance::new(&logger);
+        // scene.layers.viz_fullscreen.add_exclusive(&display_object);
```
wdanilo: commented code

src/rust/ide/view/graph-editor/src/component/visualization/container/fullscreen.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+
+
+// ======================
+// === FullscreenView ===
+// ======================
+
+/// View of the visualization container meant to be used in fullscreen mode. Its components are
+/// rendered on top-level layers of the stage.
```
wdanilo: I dont think this is strictly true. Components of full screen visualization panel, like HTML visualizations will not be rendered on top-level layers of the stage, will they be?

src/rust/ide/view/graph-editor/src/component/visualization/container/fullscreen.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn new(logger:&Logger, scene:&Scene) -> Self {
+        let logger         = Logger::sub(logger,"fullscreen_view");
+        let display_object = display::object::Instance::new(&logger);
+        // scene.layers.viz_fullscreen.add_exclusive(&display_object);
+
+        // FIXME : StyleWatch is unsuitable here, as it was designed as an internal tool for shape system (#795)
+        let styles   = StyleWatch::new(&scene.style_sheet);
+        let bg_color = styles.get_color(ensogl_theme::graph_editor::visualization::background);
+        let bg_color = color::Rgba::from(bg_color);
+        let bg_hex   = format!("rgba({},{},{},{})",bg_color.red*255.0,bg_color.green*255.0,bg_color.blue*255.0,bg_color.alpha);
```
wdanilo: too long line

src/rust/ide/view/graph-editor/src/component/visualization/container/fullscreen.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        Self {logger,display_object,background_dom}
+    }
+}
+
+impl display::Object for Panel {
+    fn display_object(&self) -> &display::object::Instance {
+        &self.display_object
+    }
+}
```
wdanilo: no nl

src/rust/ide/view/src/project.rs
```diff
@@ -129,6 +132,23 @@ impl Model {
...
+            *self.fullscreen_vis.borrow_mut() = Some(visualization);
+        }
+    }
+
+    fn hide_fullscreen_visualization(&self) {
+        if let Some(visualization) = std::mem::take(&mut *self.fullscreen_vis.borrow_mut()) {
+            self.display_object.remove_child(&visualization);
+            self.display_object.add_child(&self.graph_editor);
+        }
+    }
```
wdanilo, line 151: these should have docs of what hacks they are doing and why do we do the hacks with a link to issue which will provide correct solution (the scene cutting one)

wdanilo: Also, does this work in debug-scene? It is implemented in such a module that I'm not sure it works there.

wdanilo (changes requested):

src/js/lib/content/src/style.css
farmaazon: no, but is attached to the div element we move to the fullscreen-vis. I will change the comment to "may be put"

src/rust/ensogl/lib/core/src/display/object/class.rs
farmaazon: Yes, it may happen. But if the visible parent got hidden, then it will call set_vis_false, and thus the moved child will be also hidden.

src/rust/ide/view/src/project.rs
farmaazon: It works in interface.

src/rust/ide/view/graph-editor/src/component/visualization/container/fullscreen.rs
farmaazon: Yes. Similarly, the background module in `controller.rs`. This is code just moved from there.

src/rust/ensogl/lib/core/src/display/object/class.rs
wdanilo: So this is bad. `set_vis_false` sohuld be guaranteed to be called only once and never be called on a thing already hidden. Otherwise it can break some rendering assumptions. The same with showing things.

wdanilo (approved):

src/rust/ensogl/lib/core/src/display/object/class.rs
farmaazon: but set_vis_false checks if the instance is visible. I will add a test to make sure that on_hide callback is not called redundantly.

PR #1358, opened 2021-03-22
Add a theming API for visualizations

s9ferech:
### Pull Request Description

This solves issue #1337.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@
   a project contain syntax errors.
 - [Users can opt out of anonymous data gathering.][1328] This can be done with
   the `--no-data-gathering` command-line flag during the startup of the IDE.
+- [Provide a theming API for JavaScript visualizations][1358]. This API can be
+  used to obtain the colors associated with a type or a class GUI elements.
 
 #### EnsoGL (rendering engine)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,9 +46,9 @@
 - [Users can opt out of anonymous data gathering.][1328] This can be done with
   the `--no-data-gathering` command-line flag during the startup of the IDE.
 - [Provide a theming API for JavaScript visualizations][1358]. It is now
-  possible to use the Enso theming engine while developing custom
-  visualizations in JavaScript. You can query it for all IDE colors, including
-  the colors used to represent types.
+  possible to use the Enso theming engine while developing custom visualizations
+  in JavaScript. You can query it for all IDE colors, including the colors used
+  to represent types.
 - [You can now start the IDE service without window again.][1353] The command
   line arguyment `--no-window` now starts all the required backend services
   again, and prints the port on the command line, allowing you to open the IDE
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,8 +45,10 @@
   a project contain syntax errors.
 - [Users can opt out of anonymous data gathering.][1328] This can be done with
   the `--no-data-gathering` command-line flag during the startup of the IDE.
-- [Provide a theming API for JavaScript visualizations][1358]. This API can be
-  used to obtain the colors associated with a type or a class GUI elements.
+- [Provide a theming API for JavaScript visualizations][1358]. It is now
+  possible to use the Enso theming engine while developing custom
+  visualizations in JavaScript. You can query it for all IDE colors, including
+  the colors used to represent types.
 
 #### EnsoGL (rendering engine)
 
--- a/src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/binding.rs
+++ b/src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/binding.rs
@@ -4,8 +4,13 @@ use crate::prelude::*;
 
 use crate::component::visualization::foreign::java_script::PreprocessorCallback;
 use crate::component::visualization::instance::PreprocessorConfiguration;
+use crate::component::type_coloring;
+use crate::Type;
+use crate::display::style;
 
 use ensogl::display::DomSymbol;
+use ensogl::display::shape::StyleWatch;
+use ensogl::data::color;
 use fmt::Formatter;
 use wasm_bindgen::prelude::*;
 use web_sys::HtmlDivElement;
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,8 @@
   a project contain syntax errors.
 - [Users can opt out of anonymous data gathering.][1328] This can be done with
   the `--no-data-gathering` command-line flag during the startup of the IDE.
+- [Provide a theming API for JavaScript visualizations][1358]. This API can be
+  used to obtain the colors associated with a type or a class GUI elements.
```
wdanilo: This should be more user-facing. Like "It is now possible to use Enso theming engine while developing custom visualizations in JavaScript. You can query it for all IDE colors, including colors for a particular type". (please also check it with Ara or Joe (native speakers) to see if its 100% dramatically correct, as CHANGELOG is a public document).

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/binding.rs
```diff
@@ -86,10 +94,57 @@ impl JsConsArgs {
...
     }
 }
+
+/// The theming API that we expose to JS visualizations
+#[wasm_bindgen]
+#[derive(Clone, Debug)]
+pub struct JsTheme {
+    styles: StyleWatch
+}
+
```
wdanilo: The whole JsTheme thing should be in a separate section. We also try to introduce first things that are used later, s this should go to earlier lines in the file.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/binding.rs
```diff
@@ -86,10 +94,57 @@ impl JsConsArgs {
...
+fn format_lcha(lcha: color::Lcha) -> String {
+    let rgba = color::Rgba::from(lcha);
+    format!("rgba({:.0}, {:.0}, {:.0}, {})",
+        (rgba.red * 255.0).round(),
+        (rgba.green * 255.0).round(),
+        (rgba.blue * 255.0).round(),
+        rgba.alpha)
+}
```
wdanilo: could you add method `to_javascript_string` to both `Rgb` and `Rgba` instead, please?

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/binding.rs
```diff
@@ -86,10 +94,57 @@ impl JsConsArgs {
...
+    pub fn get(&self, path: &str) -> Option<String> {
+        if let style::Data::Color(lcha) = self.styles.get(path)? {
+            Some(format_lcha(lcha))
+        } else {
+            None
+        }
+    }
+}
+
+fn format_lcha(lcha: color::Lcha) -> String {
```
wdanilo: the name `format_lcha` is not semantic - it doesn't tell what it does. if you want to refactor it to a separate function it should be named `lcha_to_javascript_string`. However, after some changes proposed in this PR this would not be needed anymore, as you would be able to do `color.to_rgba().to_javascript_string()` instead.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/instance.rs
```diff
@@ -159,7 +159,9 @@ impl InstanceModel {
         let logger                        = Logger::new("Instance");
         let root_node                     = Self::create_root(scene,&logger)?;
         let (preprocessor_change,closure) = Self::preprocessor_change_callback();
-        let init_data                     = JsConsArgs::new(root_node.clone_ref(), closure);
+        let init_data                     = JsConsArgs::new(root_node.clone_ref(),
+                                                            StyleWatch::new(&scene.style_sheet),
+                                                            closure);
```
wdanilo: distribute that to variables please. Also, in case it would be impossible somewhere else to distribute things to variables, we use 4 spaces indentation, not aligned with rest of args.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/binding.rs
```diff
@@ -86,10 +94,57 @@ impl JsConsArgs {
...
+        let tp = Type::from(tp_name.to_string());
+        let lcha = type_coloring::compute(&tp,&self.styles);
+        format_lcha(lcha)
+    }
+
+    /// Takes a qualified type name and returns the color that should be used for foreground
+    /// (e.g. text) that is shown on top of the background color returned by getColorForType.
+    pub fn getForegroundColorForType(&self, _tp_name: &str) -> String {
+        "white".to_string()
+    }
```
wdanilo: this should query our theme for colr. We have in theme somewhere the foreground color. Its not always white.

wdanilo (approved):

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/binding.rs
radeusgd: I would actually suggest that we get these values as an array of 4 numbers (RGBA) instead of a string.

For example, in the DB visualization we need to change the alpha to use for the background. If it is a string, it would get complicated and require parsing, while an array can be easily modified and then rendered into a desired format.

s9ferech: @radeusgd and I decided to pass the raw numbers to the visualisation, so this should be obsolete.

s9ferech:
I still have two questions:
1. Should I update the documentation as part of this PR?
2. Do we need to reload the visualisations on style changes?

radeusgd:
> I still have two questions:
> 
> 1. Should I update the documentation as part of this PR?
> 2. Do we need to reload the visualisations on style changes?

I'm not sure about (1) (I guess @wdanilo or @farmaazon will have a say here).

As for (2), I think it would make sense - if the colors for types can change with the theme, the visualization also needs to reflect this change to be consistent. Maybe we could allow some kind of callback telling the visualization to refresh its color scheme, but I think that would make it more complex for visualization creators, so IMO refreshing is the simplest solution here. What do you think?

wdanilo:
1. Of course. Docs are always part of the PR.
2. Not now. We do not have proper theme reloading now, this will be a separate issue.

PR #1363, opened 2021-03-23
Show visualization on output port hover.

MichaelMauderer:
### Pull Request Description
Introduces a quick preview for visualizations and errors. Hovering a node output will first show a tooltip with the type information and then after some time, will show the visualization of the node. The preview visualization will be located above other nodes, whereas the normal view, will be shown below nodes.  
Some notes: errors will show the preview visualization immediately. Nodes without type information will show the visualization immediately. If you press and hold `ctrl` tooltips are skipped and the preview visualization shows up immediately. 

https://user-images.githubusercontent.com/1428930/112172088-b206c000-8bf4-11eb-8c98-6e78aa58c8f2.mp4

### Checklist


CHANGELOG.md
```diff
@@ -11,6 +11,15 @@
...
   that the connection was lost and IDE must be restarted. In future IDE will try
   to automatically reconnect.
+- [Visualization preview on output port hover.][1363] There is now a quick
+  preview for visualizations and error descriptions. Hovering a node output will
+  first show a tooltip with the type information and then after some time, will
+  show the visualization of the node. The preview visualization will be located
+  above other nodes, whereas the normal view, will be shown below nodes. Errors
+  will show the preview visualization immediately. Nodes without type
+  information will also show the visualization immediately. You can enter a
+  quick preview mode by pressing ctrl, which will show the preview visualization
```
wdanilo: this should be command on mac

CHANGELOG.md
```diff
@@ -11,6 +11,15 @@
...
+  immediately for all nodes.
```
wdanilo: I dont think thisis true - it will show the visualization immediately for the hovered node, not for all nodes, right?

src/rust/ensogl/lib/core/src/animation/frp/animation/delayed.rs
```diff
@@ -0,0 +1,82 @@
...
+            transition.target <+ delay_end.constant(1.0);
+
+            // Reset the animation and delay.
+            delay.stop_and_rewind      <+ frp.reset.constant(0.0);
+            transition.stop_and_rewind <+ frp.reset.constant(0.0);
+
+            // Output bindings.
+            frp.source.value    <+ transition.value;
+            frp.source.on_reset <+ transition.value.map(|t| (t - 0.0).abs() < std::f32::EPSILON).on_true();
+            frp.source.on_end   <+ transition.value.map(|t| (t - 1.0).abs() < std::f32::EPSILON).on_true();
```
wdanilo: too long - please check all new lines in the codebase and check if they cross the 100chars boundary

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -254,6 +258,8 @@ ensogl::define_endpoints! {
         set_expression_usage_type        (Crumbs,Option<Type>),
         set_output_expression_visibility (bool),
         set_vcs_status                   (Option<vcs::Status>),
+        /// Indicate whether preview visualisations should be delayed or immediate.
+        quick_preview_vis                    (bool),
```
wdanilo: spacing of `(bool)`

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -405,7 +411,7 @@ impl NodeModel {
         display_object.add_child(&visualization);
         display_object.add_child(&input);
 
-        let error_visualization = builtin_visualization::Error::new(&scene);
+        let error_visualization = builtin_visualization::Error::new(scene.clone_ref());
```
wdanilo: why do we need to provide scene clone to `Error`? Also, I think our APIs should always allow passing references to clone-refable things and just do clonesinside, not to force users of the API to manually create clone refs on every call side (it is just API simpler to use).

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -570,34 +576,78 @@ impl Node {
...
             eval visualization_enabled_frp ((enabled) model.action_bar.set_action_visibility_state(enabled));
-            no_error_set          <- not(&is_error_set);
+
+            // Show preview visualisation after some delay, depending on whether we show an error
+            // or are in quick preview mode. Also, omit the delay if we don't have a tooltip to show.
+            has_tooltip <- model.output.frp.tooltip.map(|tt| tt.has_content());
+            preview_show_delay <- all3(&frp.quick_preview_vis,&is_error_set,&has_tooltip).map(
+                |(quick_preview,is_error,has_tooltip)| {
+                if !has_tooltip {
+                    return 0.0;
```
wdanilo: I'd be thankful if we do not use in the code early-return statements if not needed necessarily. The functional, dataflow program structure is more obvious to the reader.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -570,34 +576,78 @@ impl Node {
...
+            layer_on_hover     <- preview_visible.on_false().map(|_| visualization::Layer::Default);
+            layer_on_not_hover <- preview_visible.on_true().map(|_| visualization::Layer::Front);
+            layer              <- any(layer_on_hover,layer_on_not_hover);
+            model.visualization.frp.set_layer <+ layer;
+            eval layer ((l) model.error_visualization.frp.set_layer.emit(l));
+
+
+            update_error <- all(frp.set_error,visualization_visible);
+            eval update_error([model]((error,visible)){
+                if *visible {
```
wdanilo, line 663: use `.asSome(...)`

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1001,22 +1010,24 @@ impl Deref for GraphEditorModelWithNetwork {
...
 
+/// Context data required to create a new node.
+#[derive(Debug)]
+struct NodeContext<'a> {
+    pointer_style  : &'a frp::Source<cursor::Style>,
+    tooltip_update : &'a frp::Source<tooltip::Style>,
+    output_press   : &'a frp::Source<EdgeEndpoint>,
+    input_press    : &'a frp::Source<EdgeEndpoint>,
+    output         : &'a FrpEndpoints,
+}
```
wdanilo: 1. Very cool.
2. I'd name it `NewNodeData` or something likethat to indicate its purpose

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1876,6 +1887,9 @@ impl application::View for GraphEditor {
           // , (DoublePress , "!node_editing" , "space" , "double_press_visualization_visibility")
           , (Release     , "!node_editing" , "space" , "release_visualization_visibility")
           , (Press       , ""              , "cmd i" , "reload_visualization_registry")
+          , (Press       , ""              , "ctrl" , "enable_quick_visualization_preview")
+          , (Release     , ""              , "ctrl" , "disable_quick_visualization_preview")
```
wdanilo: this does not work good on mac! On mac, `ctrl` is used only in special situations,l like killing a process. Use `cd` instead (which maps to ctrl on linu and windows).

wdanilo (approved):
Michael, it looks amazing. I would definitely change two behaviors:
1. `Nodes without type information will show the visualization immediately` - this should not be here. The thing is that we do not show vis preview immediately in order not to have blinking in GUI, so no type information should not affect our visuals.
2. The error should be shown not only on output port hover, but just on this node hover.

wdanilo:
Michael, could you please also move the popup below the mouse, like it is on main now? I was playing with main today and it looks so beautiful and consistent with the popup below the mouse, not on the right? If you could just make it configurable (I think it is), it would be cool, so we will play with both versions

MichaelMauderer:
> Michael, could you please also move the popup below the mouse, like it is on main now? I was playing with main today and it looks so beautiful and consistent with the popup below the mouse, not on the right? If you could just make it configurable (I think it is), it would be cool, so we will play with both versions

Done. This can be configured on a per-tooltip basis. (See last commit on this branch)

mwu-tow (changes requested):
I've run the test scenarios. Here are the issues I've written down:

# BASIC SCENARIO

1. When I press tab and start typing a new node (with no prior selection), the tooltip appear, despite not having content.
Tooltip should not appear while we are creating node.
Also, tooltip should never appear empty. It should either not appear in such cases or include some diagnostics, as to why it is empty.

2. Histogram visualization has broken panning. When I move mouse with MMB pressed, it quickly goes wild.

3. "Fit all" button in histogram stops working after too much panning is done.

4. Likewise, visualization zooming is broken when dragging with RMB pressed.

5. Engine reports execution failures: `Execution failed in context f8ff239e-dc10-4521-b96c-ec507a128412. Error: Methods cannot be overloaded, but you have tried to overload Histogram.enso_project.`.

6. It is impossible to enter the node with `func1` after adding a new defaulted parameter. (test scenario 1): error in console `Error while performing UI action on controllers: The suggestion with id 1567 has not been found in the database.` 

7. It is impossible to connect additional input to `func1` as there is no placeholder for it on node.

8. Due to above issues some testings steps have been omitted. Once issues are addressed, the full test scenario should be once more checked.



# TRAMS SCENARIO
1. The table visualization does not appear on `fields.get "vehicles"`. Can be triggerred though..

2. There is no such method as `Geo.points`. 

3. The geo map visualization not only does not open bubt it is impossible to be selected (missing from visualziations list).

4. Further steps of the scenario are too sketchy for me to follow. Will need to chat with @sylwiabr  later.



Some well known issues are omitted for clarity,

farmaazon:
> I've run the test scenarios. Here are the issues I've written down:
> 
> # BASIC SCENARIO
> 2. Histogram visualization has broken panning. When I move mouse with MMB pressed, it quickly goes wild.
> 3. "Fit all" button in histogram stops working after too much panning is done.

I spotted those two also on my branch, so those are likely regressions on develop.

> 6. It is impossible to enter the node with `func1` after adding a new defaulted parameter. (test scenario 1): error in console `Error while performing UI action on controllers: The suggestion with id 1567 has not been found in the database.`
> 7. It is impossible to connect additional input to `func1` as there is no placeholder for it on node.

Those are engine issues and were reported some time ago.

> # TRAMS SCENARIO
> 3. The geo map visualization not only does not open bubt it is impossible to be selected (missing from visualziations list).

This also was reported as a develop regression

MichaelMauderer:
>     1. When I press tab and start typing a new node (with no prior selection), the tooltip appear, despite not having content.
>        Tooltip should not appear while we are creating node.
>        Also, tooltip should never appear empty. It should either not appear in such cases or include some diagnostics, as to why it is empty.
This is somewhat fixed. The really proper solution would be to check whether the node has an output value from the engine, but that information is not currently available to the node. So now this only skips nodes that do not yet have content to avoid the pop-up on empty nodes.

>     2. Histogram visualization has broken panning. When I move mouse with MMB pressed, it quickly goes wild.

Regression unrelated to this PR.

>     3. "Fit all" button in histogram stops working after too much panning is done.

Regression unrelated to this PR.

>     4. Likewise, visualization zooming is broken when dragging with RMB pressed.

Regression unrelated to this PR.

>     5. Engine reports execution failures: `Execution failed in context f8ff239e-dc10-4521-b96c-ec507a128412. Error: Methods cannot be overloaded, but you have tried to overload Histogram.enso_project.`.

As Adam mentioned on stand-up: unrelated to this PR and under investigation.

MichaelMauderer:
> # TRAMS SCENARIO
> 
>     1. The table visualization does not appear on `fields.get "vehicles"`. Can be triggerred though..

Regression on develop. When opening the table vis on this node it opens up as empty for me.

>     2. There is no such method as `Geo.points`.

I believe there is only a `Geo.point` method. But I don't see either used in the Scenario. Can you point out where this happens?

farmaazon:
@MichaelMauderer 

> I believe there is only a `Geo.point` method. But I don't see either used in the Scenario. Can you point out where this happens?

Yes, it was fixed when me and Michał went through the scenario together recently.

mwu-tow (changes requested):
@MichaelMauderer 
I did the following:
* start a new unnamed project
* move nodes
* save
* reopen IDE
* hover nodes

The first nodes has tooltip with type. The second node (sort) has no tooltip. Though it likely is caused by a factor external to this PR, as node output port is also missing colors. Still might be worth investigating, as this PR makes the issue more visible.

***

Observed this when I was about to perform the copying step of the scenario.

obraz

***

Visualization automatically appears when I try to add a node after File.read.


https://user-images.githubusercontent.com/1548407/112964968-1a9bf280-9149-11eb-99c9-4bc7bb7ab0dc.mp4



***

I haven't tested further, once those issues are resolved (or confirmed as known issues unrelated to the PR) I will retest.
Please also merge develop, it should bring at least some fixes.

MichaelMauderer:
@mwu-tow 

> @MichaelMauderer
> I did the following:
> 
>     * start a new unnamed project
> 
>     * move nodes
> 
>     * save
> 
>     * reopen IDE
> 
>     * hover nodes
> 
> 
> The first nodes has tooltip with type. The second node (sort) has no tooltip. Though it likely is caused by a factor external to this PR, as node output port is also missing colors. Still might be worth investigating, as this PR makes the issue more visible.

This is expected behaviour, if the node has no type information (which would explain why it has no colour either). If there is no type information to display, no tooltip is shown. But otherwise this is a bug that should be reported as "Node that should have a type does not have a type".


> Observed this when I was about to perform the copying step of the scenario.
> 
> obraz

I believe this has been addressed on Discord?

> Visualization automatically appears when I try to add a node after File.read.
> Enso.2021-03-30.11-06-55.mp4

It looks like you are still hovering another node, so this is expected behaviour unless we want to prevent hovering while editing a node. 

> I haven't tested further, once those issues are resolved (or confirmed as known issues unrelated to the PR) I will retest.
Please also merge develop, it should bring at least some fixes.

Branch is updated to newest develop.

mwu-tow:
@MichaelMauderer 
> It looks like you are still hovering another node, so this is expected behaviour unless we want to prevent hovering while editing a node.

To me it looks like bug, even if it was designed this way. User must hover node to click on it to select. Then they open searcher to create a new node. And after a moment a strange popup appears, obscuring everything. They should not be required to move mouse away from node before opening searcher. This looks to me like a very common scenario.

Also, observe behavior on the video closely. At the moment when visualization appears, the output port *is not hovered*. It was hovered for a brief moment before, however mouse was moved and the type tooltip was hidden.

MichaelMauderer:
> > It looks like you are still hovering another node, so this is expected behaviour unless we want to prevent hovering while editing a node.
> 
> To me it looks like bug, even if it was designed this way. User must hover node to click on it to select. Then they open searcher to create a new node. And after a moment a strange popup appears, obscuring everything. They should not be required to move mouse away from node before opening searcher. This looks to me like a very common scenario.
> 
> Also, observe behavior on the video closely. At the moment when visualization appears, the output port _is not hovered_. It was hovered for a brief moment before, however mouse was moved and the type tooltip was hidden.

@mwu-tow You are right! It seems the FRP endpoint I hooked into was not only triggered when hovering the output port but some additional area. This is fixed now.

mwu-tow:
The issue seems resolved to me.

mwu-tow (approved):

PR #1383, opened 2021-03-25
`onHide` callback for Visualizations

s9ferech:
### Pull Request Description

This addresses issue #1356. JavaScript visualizations can define a `onHide` method that will be called whenever the visualization is hidden.

### Checklist


s9ferech:
@wdanilo @farmaazon, I am not very happy with this PR since I need to trigger the `on_hide` function explicitly from three different places in the visualization container. They cover the cases where the visualization is just hidden in the GUI, where the visualization is replaced by another one and where the corresponding node is deleted.

I would rather register an `on_hide` callback on the visualization itself as a `display::object::Instance`. But those callbacks do not get called when the `Instance` is dropped without being set to invisible first. And that how we remove visualizations when they are replaced or when their node is deleted.

The clean solution, in my view, would be to make sure that we always call the `on_hide` callback of `display::object::Instance` when the `Instance` gets dropped or that we do not drop such `Instance`s before they have been removed from the scene by hand. But both approaches seem like they require larger changes.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/instance.rs
```diff
@@ -182,6 +192,13 @@ impl InstanceModel {
         self.root_node.set_size(size);
     }
 
+    fn on_hide(&self) {
+        if let Some(f) = &*self.on_hide {
+            let context = &JsValue::NULL;
+            let _ = f.call0(context);
```
farmaazon: Why we skip the result? If this is Result, I think we should log a warning if it is an Error.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/instance.rs
```diff
@@ -165,8 +165,18 @@ impl InstanceModel {
...
-        Ok(InstanceModel{object,on_data_received,set_size,root_node,logger,preprocessor_change})
+        Ok(InstanceModel {
+            object,
+            on_data_received,
+            set_size,
+            on_hide,
+            root_node,
+            logger,
+            preprocessor_change
+        })
```
farmaazon: According to our guidelines, you should rather wrap line, instead putting all fields in a separate line. Look for SceneData `new` constructor and see an example there.

farmaazon (changes requested):
As I think about it, I'd prefer to set on_hide callback of visualization instance's display_object. That's better, because in the current solution the instance owner has responsibility of defining when visualization is hidden, which looks like a bad design - even now, not each visualization is inside the container.

AFAIK the display object should call on_hide callback also when destroyed - am I right @wdanilo ?

s9ferech:
> As I think about it, I'd prefer to set on_hide callback of visualization instance's display_object. That's better, because in the current solution the instance owner has responsibility of defining when visualization is hidden, which looks like a bad design - even now, not each visualization is inside the container.

I absolutely  agree. This was the first thing that I tried.

> AFAIK the display object should call on_hide callback also when destroyed - am I right @wdanilo ?

To me it looks like this cannot currently work because the `on_hide` callback of display objects gets the host as argument and we do not necessarily have access to the host during destruction.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/instance.rs
s9ferech: I just did this to be consistent with the other JS calls in the module

wdanilo:
> @wdanilo @farmaazon, I am not very happy with this PR since I need to trigger the `on_hide` function explicitly from three different places in the visualization container. They cover the cases where the visualization is just hidden in the GUI, where the visualization is replaced by another one and where the corresponding node is deleted.

This is very bad, this should definitely not b called manually and needs to be designed properly. Work with Adam on the solution he mentioned. In case it would not work, ping me on Discord, and Ill help you implement it.



> As I think about it, I'd prefer to set on_hide callback of visualization instance's display_object. That's better, because in the current solution the instance owner has responsibility of defining when visualization is hidden, which looks like a bad design - even now, not each visualization is inside the container.
> 
> AFAIK the display object should call on_hide callback also when destroyed - am I right @wdanilo ?

You are 100% right. When you are destroying our shape its width and height are set to 0 and it is still kept there by the rendering engine for fast re-use and the "on_hide" callback is called then.

s9ferech:
I pushed another version. @farmaazon, is this roughly the solution that you had in mind?

src/rust/ensogl/lib/core/src/display/object/class.rs
```diff
@@ -573,6 +573,16 @@ pub struct Instance<Host=Scene> {
...
+impl<Host> Drop for Instance<Host> {
+    fn drop(&mut self) {
+        // If this is the last reference, remove it from its parent. If there is a parent, then this
+        // will create a new reference which will be dropped on the parent's next update.
+        if Rc::strong_count(&self.rc) == 1 {
+            self.unset_parent();
+        }
+    }
+}
+
```
farmaazon, line 586: Is that needed?
When we destroy Instance, we also destroy the parent_biding, which should remove the node from the parent upon dropping.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/definition.rs
```diff
@@ -41,6 +41,7 @@ pub mod field {
 pub mod method {
     pub const ON_DATA_RECEIVED : &str = "onDataReceived";
     pub const SET_SIZE         : &str = "setSize";
+    pub const ON_HIDE             : &str = "onHide";
```
farmaazon: Alignment.

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/instance.rs
```diff
@@ -169,8 +169,17 @@ impl InstanceModel {
...
+        let on_hide                       = get_method(&object.as_ref(), method::ON_HIDE).ok();
+        if let Some(f) = on_hide {
+            display_object.set_on_hide(move |_| {
+                let context = &JsValue::NULL;
+                let _ = f.call0(context);
+            });
+        }
         Ok(InstanceModel{object,on_data_received,set_size,root_node,logger,preprocessor_change,
-                         scene})
+                         display_object, scene})
```
farmaazon:
```suggestion
                         display_object,scene})
```

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/instance.rs
```diff
@@ -169,8 +169,17 @@ impl InstanceModel {
         let set_size                      = Rc::new(set_size);
         let object                        = Rc::new(object);
         let scene                         = scene.clone_ref();
+        let display_object                = display::object::Instance::new(Logger::new(""));
+        display_object.add_child(root_node.display_object());
+        let on_hide                       = get_method(&object.as_ref(), method::ON_HIDE).ok();
+        if let Some(f) = on_hide {
+            display_object.set_on_hide(move |_| {
+                let context = &JsValue::NULL;
+                let _ = f.call0(context);
```
farmaazon:
```suggestion
                let context = &JsValue::NULL;
                let _       = f.call0(context);
```

src/rust/ensogl/lib/core/src/display/object/class.rs
s9ferech: It is needed, so we do not drop the object immediately and can call `on_hide` when the parent gets updated. Should I clarify this in the comment?

src/rust/ide/view/graph-editor/src/component/visualization/foreign/java_script/definition.rs
s9ferech:
```suggestion
    pub const SET_SIZE         : &str = "setSize";
    pub const ON_HIDE          : &str = "onHide";
```

src/rust/ensogl/lib/core/src/display/object/class.rs
farmaazon: Ach, now I see. ParentBind will not keep the instance, and the parent won't be able to upgrade Weak.

s9ferech: Right

farmaazon (approved):
The last thing: add an info in docs of method setting on_hide callback, that the callback may be called after dropping all display object instances.

PR #1394, opened 2021-03-26
Histogram & Scatterplot : Fix panning with normal mouse.

BinarySoftware:
### Pull Request Description
Adam with Michał today found yet another bug with those visualizations - if panning with middle mouse button it would behave unexpectedly and wildly - jump a thousand values from place. now that's fixed and works as gracefully as panning with touchpad.

### Checklist


src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
```diff
@@ -316,6 +317,18 @@ class Histogram extends Visualization {
...
+                d3.event.sourceEvent.buttons === midButtonClicked
+            ) {
+                const distanceScale = d3.zoomIdentity.translate(
+                    d3.event.sourceEvent.movementX / 2,
+                    d3.event.sourceEvent.movementY / 2
+                )
+                transformedScale.x = distanceScale.rescaleX(transformedScale.x)
+                if (distanceScale.rescaleY(transformedScale.y).domain()[0] >= 0) {
+                    transformedScale.y = distanceScale.rescaleY(transformedScale.y)
+                }
```
farmaazon: This function became very repetitive. And I don't understand all lines, for example: why we need to divide by two? was that just picked for the best UX, or is there other reason? And why we have `if` for x rescale, but for y not?

farmaazon (approved):

src/rust/ide/view/graph-editor/src/builtin/visualization/java_script/histogram.js
BinarySoftware: We have if for y rescale because we dont want the user to get lower than 0 on the histogram - it cant show values below 0, so that's why. 2 is because this way the mouse stays exactly above the panned element, and the velocity is correct.

wdanilo (approved):

farmaazon (changes requested):
I tested it and there are still some issues:
1. in scatterplot I still get the error `Uncaught TypeError: Cannot read property 'offsetX' of null` when clicking "fit-to-all". This can be made as a separate issue if it's hard.
2. zoom does not work as in IDE. When I zoom in by right click and move mouse back to starting position, in IDE I go back to the initial zoom. in visualizations i need to move _over_ the starting position to get the view zoomed out back.

PR #1395, opened 2021-03-28
Fix build failing when executing `run watch` on a clean build.

MichaelMauderer:
### Pull Request Description
Fix a concurrency issue between our different build tasks when running the `watch` command.

For the `watch` command to work, we need to set up the language server, run it, execute the watch process of the JS build (which includes doing the initial build) and executing the watch process for our rust build (which includes doing the initial rust build). The old setup did `((Build Language Server . Run Language Server. Watch Rust) | (Initialize JS Dependencies . Watch JS))` ([CSS notation](https://en.wikipedia.org/wiki/Calculus_of_communicating_systems)) but since setting up the language server requires the JS dependencies this meant interference of the build process. This was not a process in other commands as they never allowed running of these steps in parallel. Also, the timing between starting the watch process for JS and Rust is important because if they start in the wrong order the process ends up hanging and requires restarting.

The new sequence is now `(Initialize JS Dependencies . Build Language Server . Run Language Server .  Build Rust . Build JS . (Watch Rust | Watch JS))` , which adds more build time but starts up correctly.

### Important Notes
We should consider moving to a task-based build system (e.g., cargo make) that has better support for specifying task dependencies while avoiding the pitfalls of having to code the build in JS and makes setup more re-usable. 

The correct way to do this efficiently would be to separate out the language server (and other build steps) as a pre-step to starting the watch processes. We could then independently do `(Build JS | Build Rust | Set Up Language Server) . Start Language Server . Start Backend . Start IDE`

### Checklist

- [x] All code has been manually tested ~in the IDE~.


wdanilo (approved):
I dont have any comments but I really like the solution here. Very clean approach. Nice one!

PR #1427, opened 2021-04-02
Show node output type right under the node, not in pop-up.

s9ferech:
### Pull Request Description

This solves issue #1401.

(outdated screenshot)
Screenshot 2021-04-02 091401

The type appears on hover: (outdated recording)
main

It does not appear when a visualization is visible or we have an error: (outdated recordings)
visualization
error

Types are abbreviated for a short moment before the full appears:
abbreviation

### Checklist


src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -689,21 +686,25 @@ impl Node {
...
-
+        
         model.error_visualization.set_layer(visualization::Layer::Front);
-        frp.set_error.emit(None);
+
+        // Those three lines initialize the values that are necessary for port labels to show.
+        // All three are needed in this order, for reasons that I do not understand.
+        frp.set_error(None);
+        frp.enable_visualization();
+        frp.disable_visualization();
```
s9ferech: This part seems problematic. But I don't know how we initialize networks properly.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -434,8 +456,33 @@ impl Model {
...
+            set_label_x <- all_with(&frp.size,&label.width,
+                f!([port_count,port_index](port_size,label_width) {
+                    let corner_circumference  = node::RADIUS * 2.0 * PI;
+                    let corner_segment_length = corner_circumference * 0.25;
+                    let center_segment_length = port_size.x - node::RADIUS * 2.0;
+                    let shape_border_length   = center_segment_length + corner_segment_length * 2.0;
+                    let relative_x = (port_index as f32 + 0.5) / (port_count as f32) - 0.5;
+                    let center_x = relative_x * shape_border_length;
+                    center_x - label_width/2.0
+                }));
```
s9ferech: This part duplicates some logic from the shape definitions, but it seems difficult to avoid that.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -434,8 +456,33 @@ impl Model {
...
+            label_opacity.target    <+ label_visibility.on_false().constant(0.0);
+            label_color             <- all_with(&color.value,&label_opacity.value,
+                |color,&opacity| color.opaque.with_alpha(opacity).into());
+            label.set_color_all     <+ label_color;
+            label.set_default_color <+ label_color;
+
+
             // === Size ===
 
+            frp.source.size <+ frp.set_size;
```
s9ferech, line 516: This is quite redundant. Can this be avoided?

s9ferech:
Currently, the labels clash with outgoing wires. How should that be solved?

CHANGELOG.md
```diff
@@ -46,6 +46,9 @@
   visualization has been adjusted, allowing to display the values of the
   precomputed bins, which is useful when the dataset is relatively big, and it's
   cheaper to send the precomputed bins rather than the entire dataset.
+- [Hovering over the output port of a node shows the node's output type][1427].
+  This allows discovering the result type of a node which can help with
+  debugging and development.
```
farmaazon: I suppose the description should be altered, as we already released the pop-ups.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -649,6 +645,7 @@ impl Node {
             layer_on_not_hover <- preview_visible.on_true().map(|_| visualization::Layer::Front);
             layer              <- any(layer_on_hover,layer_on_not_hover);
             model.visualization.frp.set_layer <+ layer;
+            eval layer ((l) model.error_visualization.frp.set_layer.emit(l));
```
farmaazon: I think I included such line in one of my commits.

src/rust/ide/view/graph-editor/src/component/node.rs
farmaazon: Well, the root cause is that each frp node has output being an output value until some input events arrives.

Knowing that, let's look at the line: 
```
            model.output.set_port_label_visibility
                <+ visualization_visible.not().and(&no_error_set);
```
the visualization_visible is properly false, however no_error_set could not - until someone emits signal about new error. I'm not sure why disabling/enabling visualizations is necessary - @wdanilo do you have any thoughts?

If we want to have port types displayed by default, we should directly fire model.output.set_port_label_visibility with true during initialization

CHANGELOG.md
wdanilo: This information should be a nice sounding sentence to the user. It shoudl inform him about a new feature in the product. Please work with Ara on formulating it in this spirit.

src/rust/ide/view/graph-editor/src/component/node.rs
wdanilo: What this line has to do with "showing node output type under the node" (which is the subject of this PR)?

src/rust/ide/view/graph-editor/src/component/node.rs
wdanilo: Few things here:

1. Do not leave such comments in the code, please. This is a PR, not a Draft PR. It means that you propose these changes. So you propose that we add to code a comment "All three are needed in this order, for reasons that I do not understand.". Until this issue is resolved, this should be marked as a Draft PR, not a real PR.
2. `frp.enable_visualization();frp.disable_visualization();` looks bad. Take a look at Adam's explanation above and design the FRP netwrok initialization based on what Adam has described.

src/rust/ide/view/graph-editor/src/component/node/output/area.rs
```diff
@@ -122,6 +121,7 @@ ensogl::define_endpoints! {
         set_hover                 (bool),
         set_expression            (node::Expression),
         set_expression_visibility (bool),
+        set_port_label_visibility (bool),
```
wdanilo: What port labels have to do with types under the node? This change seems completely unrelated to the subject of this PR.

src/rust/ide/view/graph-editor/src/component/node/output/area.rs
```diff
@@ -130,12 +130,12 @@ ensogl::define_endpoints! {
...
-        on_port_type_change  (Crumbs,Option<Type>),
-        port_size_multiplier (f32),
-        body_hover           (bool),
-        tooltip              (tooltip::Style),
+        on_port_press         (Crumbs),
+        on_port_hover         (Switch<Crumbs>),
+        on_port_type_change   (Crumbs,Option<Type>),
+        port_size_multiplier  (f32),
+        body_hover            (bool),
+        port_label_visibility (bool),
```
wdanilo: What port labels have to do with types under the node? This change seems completely unrelated to the subject of this PR.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -34,8 +33,7 @@ const PORT_OPACITY_NOT_HOVERED : f32 = 0.25;
 const SEGMENT_GAP_WIDTH        : f32 = 2.0;
 const HOVER_AREA_PADDING       : f32 = 20.0;
 const INFINITE                 : f32 = 99999.0;
-
-const TOOLTIP_LOCATION : Placement = Placement::Bottom;
+const LABEL_OFFSET             : f32 = -30.0;
```
wdanilo: What port labels have to do with types under the node? This change seems completely unrelated to the subject of this PR.

**EDIT:** Ah, these are not port labels. They were meant to be output type labels! The naming have to be changed eveywhere to reflect that.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -365,35 +365,42 @@ impl display::Object for PortShapeView {
...
 
 #[derive(Clone,Debug,Default)]
 pub struct Model {
-    pub frp    : Option<Frp>,
-    pub shape  : Option<PortShapeView>,
-    pub index  : usize,
-    pub length : usize,
+    pub frp            : Option<Frp>,
+    pub shape          : Option<PortShapeView>,
+    pub label          : Option<text::Area>,
```
wdanilo: This is port model. So in port model we have a field label. Is this label of a port?

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -365,35 +365,42 @@ impl display::Object for PortShapeView {
...
+    pub display_object : Option<display::object::Instance>,
+    pub index          : usize,
+    pub length         : usize,
+    port_count         : usize,
+    port_index         : usize,
```
wdanilo, line 443: Do you really need to store this information?

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
wdanilo: If its redundant, why not just delete it?

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
wdanilo: Please talk with Michael, who was implementing the existing equations if reusing them could be possible. Maybe we should not use `Var<f32>` in the equations but just `f32`? I mean - this might be possible, depending on usage. Please confirm this with Michael - for me, it seems like refactoring this logic should be doable, as we are really computing the same things there.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -445,16 +492,17 @@ impl Model {
                 |usage_tp,def_tp| usage_tp.clone().or_else(|| def_tp.clone())
             );
 
+            label.set_content <+ frp.tp.map(f!([](tp)
+                if let Some(tp) = tp {
+                    tp.to_string()
+                } else {
+                    "".to_string()
+                }
+            ));
```
wdanilo: I think you're looking for this
`label.set_content <+ frp.tp.map(|tp| tp.map_or_else(||default(),|s| s.to_string()))`

wdanilo: Btw, please add a function `map_or_default` to `Option` to our Prelude, and then use it like 
`label.set_content <+ frp.tp.map(|tp| tp.map_or_defaukt(|s| s.to_string()))`

This will be insanely useful function.

wdanilo (changes requested):
Also, when creating such visual PRs, please show the result in action - create a screen recording includign different corner casees - moving your maouse with different zoom levels, connecting the nodes, dragging the connection over these texts, etc. We need to see all possible corner cases to understand if this behaves as it should!

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
s9ferech: I think I only tried to be consistent with other code and I find it slightly weird to use something called `set_size` in places where I only read the size.

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: Firing `model.output.set_port_label_visibility` might not change anything since it would get overwritten by the network connection that you mentioned. I'll have to find the roots from where the incorrect values propagate.

Generally, I find it a bit problematic that I have to care about the consistent initialization of networks. The connections that I create already describe how it should be initialized. Is there a better way to do this?

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
s9ferech: We need it to compute the position of the label whenever the width of the node changes. I think wee need to store it.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
s9ferech: I have created a [PR for the prelude](https://github.com/enso-org/rust-lib/pull/35). Who should review that?

s9ferech:
There is still some code left in `node` and `graph-editor` that can be used to display tooltips. Will we keep that or should it be deleted? @farmaazon @wdanilo

wdanilo:
Ok, so here's the visual review:

1. Something is wrong with spacing. On some screenshots, the text is very close to nodes:
    image
    while on some it is far away:
    image
2. Let's move the type information closer to the node, so when two nodes are created by the searcher, the type will appear between them. Please make the offset configurable in the theme editor.
3. Please make also the offset at which the node searcher makes new nodes configurable in the theme editor (if its not done yet).
4. Let's simplify the view - after hovering over the output port, lets display only the last part of the type, like `Vector`. Let's display the full type after keeping the mouse cursor there for a moment (approx 2-3 seconds - configurable in theme editor).

After having that done, I'll try to adjust the themes values and then I'd love everyone interested in this topic to play with it and have a conversation about this solution vs the popups one (surprisingly, this is the most controversial / hard design part so far).

src/rust/ide/view/graph-editor/src/component/node.rs
farmaazon: Well, as I know, the FRP node output value is not overwritten be setting up network connection _per se_, but only when the event go through it. So we need to intialize the output value to have proper value before any input connection will be fired.

s9ferech:
> 4. Let's simplify the view - after hovering over the output port, lets display only the last part of the type, like `Vector`. Let's display the full type after keeping the mouse cursor there for a moment (approx 2-3 seconds - configurable in theme editor).

Currently, this seems to be easy since we seem to only display single identifiers there. But will it stay like that? I assume that we will display full types, composed from multiple identifiers and operators, in the future.

wdanilo:
What is the status of this task ?

s9ferech:
> What is the status of this task ?

I made all the changes that you requested. It is ready for review. There is one new recording that shows the abbreviated types and the new label offset in action.

src/rust/ide/view/src/project.rs
```diff
@@ -236,6 +231,12 @@ impl View {
         //   See: https://github.com/enso-org/ide/issues/795
         app.themes.update();
 
+        let style_sheet                    = &model.app.display.scene().style_sheet;
+        let styles                         = StyleWatch::new(style_sheet);
+        let default_gap_between_nodes_path = ensogl_theme::project::default_gap_between_nodes;
+        let default_gap_between_nodes      = styles.get_number(default_gap_between_nodes_path);
+        frp.source.default_gap_between_nodes.emit(default_gap_between_nodes);
```
wdanilo: This is not using the StyleWatchFRP, so it would not update when the theme changes. Is this correct or am I missing something here? If its correct, using the frp version would be pretty simple, as we are close to frp network here.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -109,6 +107,24 @@ impl AllPortsShape {
...
+}
+
+fn corner_segment_length<T>(corner_radius: T) -> T
+    where T: Mul<f32,Output=T> {
+    let corner_circumference  = corner_radius * 2.0 * PI;
+    corner_circumference * 0.25
+}
+
+fn center_segment_length<T>(corner_radius: T, width: T) -> T
+    where T: Mul<f32,Output=T> + Sub<Output=T> {
```
wdanilo: where should not be indented

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -109,6 +107,24 @@ impl AllPortsShape {
...
 }
 
+fn shape_border_length<T>(corner_radius: T, width: T) -> T
+    where T: Clone + Mul<f32,Output=T> + Sub<Output=T> + Add<Output=T> {
+    let corner_segment_length = corner_segment_length(corner_radius.clone());
+    let center_segment_length = center_segment_length(corner_radius, width);
+    center_segment_length + corner_segment_length * 2.0
+}
+
+fn corner_segment_length<T>(corner_radius: T) -> T
```
wdanilo: spacing - space after `:`

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -109,6 +107,24 @@ impl AllPortsShape {
     }
 }
 
+fn shape_border_length<T>(corner_radius: T, width: T) -> T
```
wdanilo: please add docs to these functions - they are not obvious. For example, what is "border" here? And what is border length? In the next function, what is "segment" - these should be described in docs.

wdanilo (commented):
This looks super ocol. The implementation looks good as well to me! Just a couple of comments. Also, could you not remove the popup and make it configurable - either using popup or tdisplayin it below the node? This is tricky as hell to get it right here, and Id prefer to have both options for now. Regarding the popup - please keep the abbreviated thing there as well, as the abbreviation is amazing!

src/rust/ide/view/src/project.rs
s9ferech: I fixed it.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -109,6 +113,42 @@ impl AllPortsShape {
     }
 }
 
+/// The length of the port shape's inner border (the part that touches the node)
```
wdanilo: This is not a sentence, missing dot. The same applies to a lit of other lines in other docs. Please fix this in all new docs.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
wdanilo: Felix, please do not mark such conversations as "resolved", as it was not resolved. You've replied, marked it as "resolved" and it was hidden. I'm reopenning it.

I still don't understand it. What do you mean by "This is quite redundant" then? I do not understand your last comment - could you elaborate, please? Let's not leave in code anything that is redundant. This needs to be fixed before merge.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -436,6 +504,17 @@ impl Model {
...
             // === Size ===
 
+            frp.source.size <+ frp.set_size;
+            eval frp.size ((&s)
+                shape.set_size(s + Vector2(HOVER_AREA_PADDING,HOVER_AREA_PADDING) * 2.0));
+            set_type_label_x <- all_with(&frp.size,&type_label.width,
+                f!([port_count,port_index](port_size,type_label_width) {
+                    let shape_border_length = shape_border_length(node::RADIUS, port_size.x);
+                    let relative_x          = (port_index as f32 + 0.5) / (port_count as f32) - 0.5;
+                    let center_x            = relative_x * shape_border_length;
```
wdanilo: What are these variables used for? The names `relative_x` and `center_x` do not explain what they are used for. What is `relative_x` relative to? Could you please improve the names so their meaning would be obvious?

Also, the logic is strange IMO. Why are you adding and removing `0.5`? For example, what if `port_index` is `1` and `port_count` is `2`? Then your equation is `(1 + 0.5) / 2 - 0.5 = 1.5 / 2.0 - 0.5 = 0.25` - TBH I don't understand reasoning behind it. This should be explained much better.

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -406,18 +455,37 @@ impl Model {
...
+        let network            = &frp.network;
+        let events             = shape.events();
+        let opacity            = Animation::<f32>::new(network);
+        let color              = color::Animation::new(network);
+        let type_label_opacity = Animation::<f32>::new(network);
+        let port_count         = self.port_count;
+        let port_index         = self.port_index;
+        let full_type_timer    = DelayedAnimation::new(network);
+        full_type_timer.set_delay(FULL_TYPE_ONSET_DELAY_MS);
+        full_type_timer.set_duration(0.0);
```
wdanilo, line 497: What are you creating animation for if you are setting its duration to 0 ?

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
```diff
@@ -34,8 +35,11 @@ const PORT_OPACITY_NOT_HOVERED : f32 = 0.25;
 const SEGMENT_GAP_WIDTH        : f32 = 2.0;
 const HOVER_AREA_PADDING       : f32 = 20.0;
 const INFINITE                 : f32 = 99999.0;
+const FULL_TYPE_ONSET_DELAY_MS : f32 = 2000.0;
 
 const TOOLTIP_LOCATION : Placement = Placement::Bottom;
+const SHOW_TOOLTIP     : bool      = false;
+const SHOW_TYPE_LABEL  : bool      = true;
```
wdanilo: Please add comments here describing what these constants are used for (explaining that we have left 2 implementations for now).

wdanilo (approved):

src/rust/ide/view/graph-editor/src/component/node/output/port.rs
s9ferech: I use it just for the delay. When the delay is over, I want to switch the value instantly. I copied this from how the way how visualizations are shown on hover. Is there a better way to do this?

farmaazon (approved):

PR #1447, opened 2021-04-08
Add background to breadcrumbs and status bar

s9ferech:
### Pull Request Description

This adds a background to the breadcrumbs and statusbar, solving issue #1398. Background color and corner radius for each can be set through the theme.

<kbd></kbd>

### Important Notes

The implementation is strongly coupled with the internals of breadcrumbs and the status bar. This is necessary to compute the width and height of those components and to track their changes. We could improve this in the future through introduction of a generic layout mechanism.

### Checklist

### DIFF
--- a/src/rust/ensogl/lib/components/src/shadow.rs
+++ b/src/rust/ensogl/lib/components/src/shadow.rs
@@ -2,17 +2,43 @@
 use crate::prelude::*;
 
 use ensogl_core::data::color;
-use ensogl_core::display::DomSymbol;
+use ensogl_core::display::{DomSymbol, style};
 use ensogl_core::display::shape::*;
 use ensogl_core::display::shape::AnyShape;
 use ensogl_core::system::web::StyleSetter;
 use ensogl_theme as theme;
 
 
+/// Defines the apearance of a shadow
+#[derive(Debug,Copy,Clone)]
+pub struct Parameters {
+    base_color : color::Rgba,
+    fading     : color::Rgba,
+    size       : f32,
+    spread     : f32,
+    exponent   : f32,
+    offset_x   : f32,
+    offset_y   : f32
+}
+
+/// Loads shadow parameters from the given style, at the given path. The structure of the style
+/// definition should be analogous to that at `ensogl_theme::shadow`.
+pub fn parameters_from_style_path(style:&StyleWatch,path:impl Into<style::Path>) -> Parameters {
+    let path: style::Path = path.into();
+    Parameters {
+        base_color: style.get_color(&path),
+        fading: style.get_color(&path.sub("fading")),
+        size: style.get_number(&path.sub("size")),
+        spread: style.get_number(&path.sub("spread")),
+        exponent: style.get_number(&path.sub("exponent")),
+        offset_x: style.get_number(&path.sub("offset_x")),
+        offset_y: style.get_number(&path.sub("offset_y"))
+    }
+}
 
 /// Return a shadow for the given shape. Exact appearance will depends on the theme parameters.
 pub fn from_shape(base_shape:AnyShape, style:&StyleWatch) -> AnyShape {
-    let alpha         = Var::<f32>::from(1.0);
+    let alpha = Var::<f32>::from(1.0);
     from_shape_with_alpha(base_shape,&alpha,style)
 }
 
@@ -20,29 +46,35 @@ pub fn from_shape(base_shape:AnyShape, style:&StyleWatch) -> AnyShape {
 /// The color will be multiplied with the given alpha value, which is useful for fade-in/out
 /// animations.
 pub fn from_shape_with_alpha(base_shape:AnyShape,alpha:&Var<f32>,style:&StyleWatch) -> AnyShape {
-    let shadow_size   = style.get_number(theme::shadow::size);
-    let shadow_spread = style.get_number(theme::shadow::spread);
-    let shadow_off_x  = style.get_number(theme::shadow::offset_x).px();
-    let shadow_off_y  = style.get_number(theme::shadow::offset_y).px();
+    let parameters = parameters_from_style_path(style, theme::shadow);
+    from_shape_with_parameters_and_alpha(base_shape,parameters,alpha)
+}
 
-    let shadow_grow   = Var::<f32>::from(shadow_size);
-    let shadow        = base_shape.grow(shadow_grow);
-    let shadow        = shadow.translate((shadow_off_x,shadow_off_y));
+/// Return a shadow for the given shape and shadow parameters.
+pub fn from_shape_with_parameters(base_shape:AnyShape,parameters:Parameters) -> AnyShape {
+    let alpha = Var::<f32>::from(1.0);
+    from_shape_with_parameters_and_alpha(base_shape,parameters,&alpha)
+}
 
-    let base_color    = style.get_color(theme::shadow);
-    let base_color    = Var::<color::Rgba>::from(base_color);
+/// Return a shadow for the given shape, shadow parameters and alpha. As in `from_shape_with_alpha`,
+/// the shadow colors will be multiplied with the given alpha value.
+pub fn from_shape_with_parameters_and_alpha
+(base_shape:AnyShape,parameters:Parameters,alpha:&Var<f32>)
+-> AnyShape {
+    let grow          = Var::<f32>::from(parameters.size);
+    let shadow        = base_shape.grow(grow);
+    let shadow        = shadow.translate((parameters.offset_x.px(),parameters.offset_y.px()));
+
+    let base_color    = Var::<color::Rgba>::from(parameters.base_color);
     let base_color    = base_color.multiply_alpha(&alpha);
 
-    let fading_color  = style.get_color(theme::shadow::fading);
-    let fading_color  = Var::<color::Rgba>::from(fading_color);
+    let fading_color  = Var::<color::Rgba>::from(parameters.fading);
     let fading_color  = fading_color.multiply_alpha(&alpha);
-
-    let exp           = style.get_number(theme::shadow::exponent);
 
     let shadow_color  = color::gradient::Linear::<Var<color::LinearRgba>>
     ::new(fading_color.into_linear(),base_color.into_linear());
-    let shadow_color  = shadow_color.sdf_sampler().size(shadow_size)
-        .spread(shadow_spread).exponent(exp);
+    let shadow_color  = shadow_color.sdf_sampler().size(parameters.size).spread(parameters.spread)
+        .exponent(parameters.exponent);
     let shadow        = shadow.fill(shadow_color);
     shadow.into()
 }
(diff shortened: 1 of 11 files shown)


src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -10,30 +10,47 @@ pub use breadcrumb::Breadcrumb;
...
+                                             + breadcrumb::PADDING
+                                             + breadcrumb::VERTICAL_MARGIN
+                                             + VERTICAL_MARGIN;
+
+const MACOS_TRAFFIC_LIGHTS_CONTENT     : f32 = 52.0;
+const MACOS_TRAFFIC_LIGHTS_SIDE_OFFSET : f32 = 13.0;
+const MACOS_TRAFFIC_LIGHTS_WIDTH       : f32 =
+    MACOS_TRAFFIC_LIGHTS_CONTENT + 2.0 * MACOS_TRAFFIC_LIGHTS_SIDE_OFFSET;
+// This should be as large as the shadow around the background.
+const MAGIC_SHADOW_MARGIN              : f32 = 40.0;
```
s9ferech: I use this constant to estimate how much space the drop shadow will need. The implementations of list view, nodes and tooltip seem to do something similar. It would be good if we could get this from the definition of the shadow or if we didn't need to mention this value at all. @farmaazon do you have a suggestion what is the best way to solve this?

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -10,30 +10,47 @@ pub use breadcrumb::Breadcrumb;
...
+const VERTICAL_MARGIN                  : f32 = GLYPH_WIDTH;
+const HORIZONTAL_MARGIN                : f32 = GLYPH_WIDTH;
+const TEXT_SIZE                        : f32 = 12.0;
+const HEIGHT                           : f32 = VERTICAL_MARGIN
+                                             + breadcrumb::VERTICAL_MARGIN
+                                             + breadcrumb::PADDING
+                                             + LINE_HEIGHT
+                                             + breadcrumb::PADDING
+                                             + breadcrumb::VERTICAL_MARGIN
+                                             + VERTICAL_MARGIN;
```
wdanilo: we are aligning lines only if there are glued together. In such a scenario it sees very strange, so I'd skip aligning this top block to bottom one

wdanilo: Btw, I love how you have written it - keeping the visual hierarchy of what goes after what. Very clear.

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -181,15 +223,33 @@ impl BreadcrumbsModel {
...
+    fn breadcrumbs_container_size(&self) -> f32 {
         self.breadcrumbs.borrow().iter().map(|breadcrumb| breadcrumb.width()).sum()
     }
 
-    fn relayout_for_project_name_width(&self, width:f32) {
-        self.breadcrumbs_container.set_position_x(width.round());
+    fn update_layout(&self) {
+        let is_macos     = ARGS.platform.map(|p|p.is_macos()) == Some(true);
+        let is_frameless = ARGS.frame == Some(false);
+        let x_offset     = if is_macos && is_frameless { MACOS_TRAFFIC_LIGHTS_WIDTH }
```
wdanilo: Why did you move this logic into breadcrumbs? I don't think this component is responsible for making the spacing on the left side. We will use breadcrumbs in the future in multiple panels, not only on top of the window. So to correctly position breadcrumbs is the responsibility of its parent, not this component itself

wdanilo (commented):
Just minor code comments for now. I'd love to see it again after the MacOS traffic lights problem gets resolved that I was writing about in the comment. In case you feel that the proposed design is bad, please reach out to me and lets talk about it! :) 

Also, from visual perspective:
1. The backgrounds were meant to not go outside of the window, they should be contained inside (the shadows can go outside, but the bqckgrounds should be fully visible). There should be a padding from left side of window border and bottom side of window border to the background of the status bar. The same applies to breadcrumbs. This way:
  image
2. Make the shadow of these panels controllable from the theme editor as well. We might want to adjust them slightly. Make in themes `status_bar.background.shadow` and all subsequent parameters (size, spread, fading, etc) and set them just to the generic ones, like:
```
    background {
        shadow = shadow, shadow;
        shadow {
            size = shadow::size, shadow::size;
            ...
        }
    }
...
```

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
s9ferech: Because in my implementation, the breadcrumb object stretched to contain the traffic lights. Also, I found the positioning a bit weird as it was before. The breadcrumbs seemed to position themselves in the corner of the screen, but we added this offset from the outside. It just isn't what I would expect.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1273,9 +1273,8 @@ impl GraphEditorModel {
         let is_macos     = ARGS.platform.map(|p|p.is_macos()) == Some(true);
         let is_frameless = ARGS.frame == Some(false);
         let x_offset     = if is_macos && is_frameless { MACOS_TRAFFIC_LIGHTS_WIDTH }
-                           else                        { MACOS_TRAFFIC_LIGHTS_SIDE_OFFSET };
+                           else                        { 0.0 };
```
s9ferech: I changed this because I don't see why we should have the offset if the traffic lights are not their. Please correct me if there is a reason for the offset. Also, someone with Mac OS should check if the new design still works there.

src/rust/ensogl/lib/components/src/shadow.rs
```diff
@@ -2,47 +2,79 @@
 use crate::prelude::*;
 
 use ensogl_core::data::color;
-use ensogl_core::display::DomSymbol;
+use ensogl_core::display::{DomSymbol, style};
```
wdanilo: we're not using such imports (Style guide)

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -57,10 +69,27 @@ mod background {
...
-            let gray     = 34.0/255.0;
-            let bg_color = color::Rgba::new(gray,gray,gray,1.0);
-            Plane().fill(bg_color).into()
+        (style:Style) {
+            let width             = Var::<Pixels>::from("input_size.x");
+            let height            = Var::<Pixels>::from("input_size.y");
+
+            let corner_radius     = style.get_number
+                                    (ensogl_theme::graph_editor::breadcrumbs::background
+                                    ::corner_radius);
```
wdanilo: extract `ensogl_theme::graph_editor::breadcrumbs` as `breadcrumbsTheme` and make these nulti-line expressions single line

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -57,10 +69,27 @@ mod background {
...
+            let shape_width       = width - MAGIC_SHADOW_MARGIN.px() * 2.0;
+            let shape_height      = height - MAGIC_SHADOW_MARGIN.px() * 2.0;
+            let shape             = Rect((&shape_width,&shape_height))
+                                    .corners_radius(corner_radius.px());
```
wdanilo: extract variables if possible to have one line expressions here and in other places. Its easier to read them if the components are well named variables

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -138,7 +167,7 @@ pub struct BreadcrumbsModel {
     display_object        : display::object::Instance,
     background            : background::View,
     project_name          : ProjectName,
-    root                   : display::object::Instance,
+    root: display::object::Instance,
```
wdanilo: why this change?

src/rust/ide/view/graph-editor/src/lib.rs
wdanilo: This was added here because we want to have the same padding of breadcrumbs fron the left side of the window as the padding between left side of the window and traffic lights (in fact we want to use the same padding for top padding too). Please adjust it so these paddings are the same.

src/rust/ide/view/src/status_bar.rs
```diff
@@ -75,6 +83,40 @@ pub mod process {
...
+
+            let corner_radius     = style.get_number
+                                    (ensogl_theme::application::status_bar::background::corner_radius);
+            let shape_width       = width  - MAGIC_SHADOW_MARGIN.px() * 2.0;
+            let shape_height      = height - MAGIC_SHADOW_MARGIN.px() * 2.0;
+            let shape             = Rect((&shape_width,&shape_height))
+                                    .corners_radius(corner_radius.px());
+
+            let bg_color          = style.get_color
+                                    (ensogl_theme::application::status_bar::background);
```
wdanilo: make all these utliline expressions single lines please

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
s9ferech: This was just a refactoring accident.

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -181,15 +214,27 @@ impl BreadcrumbsModel {
...
+        self.breadcrumbs_container.set_position_x(
+            HORIZONTAL_PADDING + self.project_name.width.value().round());
+
+        let background_width  = HORIZONTAL_PADDING
+                              + self.project_name.width.value().round()
+                              + self.breadcrumbs_container_size()
+                              + HORIZONTAL_PADDING;
+        let background_height = HEIGHT;
+        self.background.size.set(Vector2(background_width+MAGIC_SHADOW_MARGIN*2.0,
+            background_height+MAGIC_SHADOW_MARGIN*2.0));
```
wdanilo: pls refactor it to var and make this one-liners

src/rust/ide/view/graph-editor/src/component/breadcrumbs.rs
```diff
@@ -181,15 +214,27 @@ impl BreadcrumbsModel {
...
+    fn breadcrumbs_container_size(&self) -> f32 {
         self.breadcrumbs.borrow().iter().map(|breadcrumb| breadcrumb.width()).sum()
     }
 
-    fn relayout_for_project_name_width(&self, width:f32) {
-        self.breadcrumbs_container.set_position_x(width.round());
+    fn update_layout(&self) {
+        self.project_name.set_position_x(HORIZONTAL_PADDING);
+        self.breadcrumbs_container.set_position_x(
+            HORIZONTAL_PADDING + self.project_name.width.value().round());
```
wdanilo: pls refactor it to var and make this one-liners

wdanilo (approved):

PR #1465, opened 2021-04-12
Improve rendering of zoomed HTML elements.

s9ferech:
### Pull Request Description

Before, we used the `perspective` CSS atribute and modified the Z coordinate of JS visualizations to zoom in and out. This had two drawbacks:
1. It caused a bug with mouse interaction on visualizations, that I can't fully explain. (#1429).
2. It lead to a bad visual quality. (#1432)

With this PR, the `perspective` attribute does not get set. Instead, we handle zooming explicitly through the `scale` transform. This solves both issues.

### Important Notes

There are two drawbacks:
1. Performance might suffer, since visualizations get rendered at high quality on every move.
2. Visualizations wobble slightly during movement.

The right solution for those problems might be to temporarily set the `will-change: transform` CSS attribute while the user moves the camera.

### Checklist


s9ferech:
It would be good if someone could confirm that I am using the camera transformation correctly.

s9ferech:
I created a [minimal example](https://jsfiddle.net/s9ferech/k0jo4m97/19/) to demonstrate how the visual quality suffers when using 3D transformations in CSS. I ran it on windows in Chrome version 89.0.4389.114 and it looked like this:

Capture

The odd lines are sharp, the even ones blurry. The example is supposed to demonstrate how this depends on the content's original width and its distance to the virtual camera. I it looks differently in other browsers or on displays with other pixel densities, it should likely suffice to adapt the `width` or `translateZ` values to achieve a similar result as me.

When I use the `scale` transform to scale objects to the same size, instead of `translateZ`, then everything looks perfectly sharp.

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,18 +191,21 @@ impl DomScene {
...
 
         match camera.projection() {
             Projection::Perspective{..} => {
-                js::setup_perspective(&self.data.dom,&near.into());
-                setup_camera_perspective(&self.data.view_projection_dom,near,&trans_cam);
+                // We do not use the CSS `perspective` property. As a consequence, the Z translation
+                // of the view matrix will have no visible impact and we have to handle zoom
+                // explicityly. We do this for two reasons:
+                // 1. We avoid a bug with D3 visualizations.
+                //    (https://github.com/enso-org/ide/issues/1429)
```
wdanilo: This is not explained at all. What does it mean that we avoid a bug in D3? What bbug? Where is it described? Why perspective 3D is causing it? This needs to be described in detail and properly documented so when reading the code we understand what is causing such behavior.

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,18 +191,21 @@ impl DomScene {
...
+                // 2. In Chrome and other browsers, we get a much better visual quality.
+                //    (https://github.com/enso-org/ide/issues/821)
```
wdanilo: This is not specific - what other browsers? And on what systems? On MacOS its working fine. If we dont specify such details, someone will test it in the future, for example on MacOS and will think that this issue was fixed in the browsers, while it was not. We should wirte everything we know in comments.

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,18 +191,21 @@ impl DomScene {
...
-        let half_dim   = camera.screen().height / 2.0;
-        let fovy_slope = camera.half_fovy_slope();
-        let near       = half_dim / fovy_slope;
+        let zoom = camera.zoom();
 
         match camera.projection() {
             Projection::Perspective{..} => {
-                js::setup_perspective(&self.data.dom,&near.into());
-                setup_camera_perspective(&self.data.view_projection_dom,near,&trans_cam);
+                // We do not use the CSS `perspective` property. As a consequence, the Z translation
```
wdanilo: if we do not use perspective property, why in the CSS there is 3D matrix left?

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -29,46 +29,44 @@ mod js {
...
+            // align objects to pixel boundaries. The drawbacks are that it has an impact on
+            // performance and that things wobble slightly when the camera is manipulated. To avoid
+            // this, one could temporarily set the CSS attribute `will-change: transform` during
+            // camera movements,
+            let scale  = `scale(${zoom})`;
             let matrix3d    = arr_to_css_matrix3d(matrix_array);
-            let transform   = translateZ + matrix3d + 'translate(50%,50%)';
+            let transform   = scale + matrix3d + 'translate(50%,50%)';
             dom.style.transform = transform;
         }
```
wdanilo: these changes does not make sense. This function is called "setup_camera_perspective". After these changes it is not setuping a perspective projection anymore.

EDIT: the name would make still sense if there was additional translate component. When I was writing this comment I thought it was skipped on purpose

CHANGELOG.md
```diff
@@ -123,6 +123,8 @@
   time.][1428]
 - [Improved performance of the graph editor, particularly when opening a project
   for a first time.][1445]
+- [Visualizations will be sharp and handle mouse input correctly]
+  [https://github.com/enso-org/ide/pull/1465]
```
wdanilo: This is not a proper description in CHANGELOG. I mean, when they "will be" sharp? This sentence only tells that they will be sharp some day, but does not specify when. It also does not tell what "correctly means". Also, it doesn't tell anything about what is happening now to visualizations and why they are not sharp.

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -29,46 +29,44 @@ mod js {
...
+            // Since the transform here only consists of scaling and translation, with no real 3D
+            // transformations, Chrome will render the transformed object at best quality, and will
+            // align objects to pixel boundaries. The drawbacks are that it has an impact on
+            // performance and that things wobble slightly when the camera is manipulated. To avoid
+            // this, one could temporarily set the CSS attribute `will-change: transform` during
+            // camera movements,
+            let scale  = `scale(${zoom})`;
             let matrix3d    = arr_to_css_matrix3d(matrix_array);
-            let transform   = translateZ + matrix3d + 'translate(50%,50%)';
+            let transform   = scale + matrix3d + 'translate(50%,50%)';
```
wdanilo: This change breaks how camera is working - it does not work in 3D, when camera is rotated. The scale would need another trnaslate component here. But lets first check the things I mentioned at the root comment, and then lets see how to fix it

wdanilo (changes requested):
Felix, I have few issues with this PR:

- It breaks the rendering consistency. After rotating the camera, the rendering gives wrong results:
   - In Perspective projection you can replace the matrix + z-transform by matrix + scale + XY-transform (you are missing the last component).
   - The proposed solution works only with Perspective and Orthographic projections, but would not work with non-affine projections, where standard methods would work. This is less important than the previous point, but we should be aware of it. 
   - It changes how functions work while leaving their old names. For example, the function `setup_camera_perspective` was setuping a real perspective in the past, while now it does not.
   - It leaves comments which do not explain what is really happening, like "this is to avoid a bug with D3 visualizations", but without any further explanation of what this bug is, what it was caused by, why this change fixes it, etc. Such an explanation should be provided. In case you are not sure why this fixes the bug, this should be written there and another task to research should be created and linked there.

Let's do the following steps:
1. Let's compare the quality of rendering of the following scenarios (with text) (let's do it in both Chrome and Firefox) - please create JS fiddle which can be controlled with keyboard (to switch the modes):
    a. z-rotate -> scale
    b. scale -> z-rotate
    c. z-rotate -> z-translate
    e. z-translate -> z-rotate
    f. y-rotate -> scale
    g. scale -> y-rotate
    h. y-rotate -> z-translate
    i. z-translate -> y-rotate
    j. identity-matrix -> scale 
    k. scale -> identity-matrix
    l. identity-matrix -> z-translate
    m. z-translate -> identity matrix

2. In case we will find a way to actually keep a good quality by such simulations, we can re-write the perspective renderer to use these simulations there. In case we will not get good quality, lets do point 3.
3. Lets do a check - if the camera is not rotated, we will simulate z-transform with scale. If the camera is rotated, we will use normal CSS3D transformations (as we use now).

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,18 +191,21 @@ impl DomScene {
...
 
-        let trans_cam  = camera.transform_matrix().try_inverse();
-        let trans_cam  = trans_cam.expect("Camera's matrix is not invertible.");
+        let trans_cam  = camera.view_matrix();
         let trans_cam  = trans_cam.map(eps);
         let trans_cam  = inverse_y_translation(trans_cam);
-        let half_dim   = camera.screen().height / 2.0;
-        let fovy_slope = camera.half_fovy_slope();
-        let near       = half_dim / fovy_slope;
+        let zoom = camera.zoom();
```
wdanilo, line 197: I'm not sure this works as expected now. The lines you have removed were re-computing the zoom after for example scene dimension change. So, when you are resizing the window, you want objects to be exactly the same size they were earlier, so camera needs to recompute its zoom. This needs to be pixel perfect, without even 0.5 px transformation in any direction, because it will cause blurred renders in EnsoGL vector renderer. These equations were designed to recompute zoom with high precision

s9ferech:
I created a [JsFiddle](https://jsfiddle.net/s9ferech/pdyt8vgo/7/)  demonstrating different transformation methods.

s9ferech:
We have created some [benchmarks ](https://github.com/enso-org/dom-benchmarks) to compare the performance and image quality of different display methods for visualizations.

wdanilo:
Felix, as discussed on Discord, we need to make a very good description of why are we want to choose a particular solution. What we need here is a detailed guide of how we went from the tests to the conclusions, in a form of a markdown file. It should tell:
1. what benchmarks we did;
2. what observations we made;
3. what steps we've taken to get the final conclusions. 

It has to be written very, very precisely, so when new browsers will be released, we would be able to follow exactly the same steps to see if we should change what we use. 

Such important design decisions, like the way we render 50% of our IDE have to base on well-documented facts. Until we got well-documented facts with well-described reasoning - from tests to conclusions, we will not progress with it anyway.

s9ferech:
Here is a list of some potential problems if we rely too much on fixed benchmarks to make the decision:
1. We do not test all configurations, but just a few that we could think of. Especially in the scenario that browsers evolve and provide better methods, it is very likely that the necessary configuration is not included in our benchmarks.
2. The benchmarks may not be representative for the performance of realistic visualizations in the context of the IDE.
3. The benchmark results on a small number of hardware/OS/browser combinations may not be representative for all settings.
3. We do not just optimize for FPS, but for many factors. We need to judge which solution provides the best trade-off. One factor is user experience which is hard to specify and may require creative solutions. Another factor is that we should prefer to use constructs the way they were intended to be used and we should avoid unnecessary complexity. Even if we focus exclusively on performance, the problem is not simple: How is performance when we have all visualization on screen at once vs when most visualizations are off-screen? How is it for a large number of visualizations vs a small number with high internal complexity?
4. If we rely on benchmarks without understanding them then we might pick a solution that only works in special cases.

My confidence in the suggested solution comes from the fact that
1. I have a good explanation, why it is appropriate,
2. I did a lot of research and experimentation trying to find flaws or better solutions and
3. the benchmarks in our repo do not contradict the explanations and do not show any configurations that perform significantly better without unacceptable flaws.

I think you have to consider all three points if you revise the decision in the future. Number 2 is a big one if you are really ambitious to find better solutions. Number 3 contributes a rather small part. It should probably also involve that you update the benchmarks to reflect your new understanding and that you check if the results are in any way surprising.

wdanilo:
1. That's a good point, but if I understand correctly, our benchmarks test all available configurations now, don't they? I mean, sure, this needs to evolve over time, but it should be a good, extensible framework for us. If I'm not right, what possible configs are we missing (we should add them to benchmarks)?
2. Why? Could you elaborate? The visualizations will be either a single WebGL canvas or a lot of SVG elements. These are 2 possible scenarios only and we can test them in benchmarks.
3. Having that written up we will test that on macOS, Windows, and Linux - these should give us a "good enough" approximation. Sure, it can vary between hardware - this should be noted in the markdown too, that's a good point.
4. I don't understand the reasoning here. We DO optimize for FPS and quality. During movement of the camera - FPS is more important, on a still image, the quality. The questions if offscreen vis change the results is a good question - I'd propose to make additional tests for it (not necessarily in the benchmarks, maybe side-by-side the benchmarks).
4. What do you mean by "without understanding them" ? If we test possible configurations and we pick the one which performs best, we should be fine. If you'd like to really understand why we get these results, then I think reading the source code of Chrome rendering implementation is the only option, but this would be a total overkill. Did you mean something else? 

Regarding the solution you proposed (the next 1-3 points above):
1. Good, then let's write it down in a scientific way - going from problems, through facts, tests, till conclusions. 
2. The problem I have here is that you did it and this is not reproducible. Anyone from our team or community should see it and understand ALL steps and all reasoning why this solution was chosen. So we need to write it down as a series of well-documented and logically connected steps.
3. I don't really understand it. What are "unacceptable flaws", are these blurred fonts? If so, then this is bad - the benchmarks should cover these. Of course they do not need to cover integer-pixels, as jumping during movement is something we definitely o not want to have. However, blurred fonts are not "unacceptable flaw", especially when we are moving things around. This is one of such things that need to be described really, really well, before making the conclusions.

To sum this up - I understand that you did a lot of research, that's good. Now, before making a decision, you need to show that to the team. In a very, very logical order - starting with problems, through all tests, results, to the final conclusions. This HAS TO base on some **framework**. My proposition is that the framework is simple:

1. We show all possible configurations in benchmarks (macOS + Windows + Linux)
2. We choose the best ones.
3. We check which ones have flaws (like blurred fonts)
4. Based on that we propose a solution (a single one or a hybrid one - for moving and still camera). In my opinion, this will guarantee that we will choose the best performing options here.

If you have a better framework - let's discuss it, if not, let's progress with it. Anyway, let's move with this task ASAP now, as we already spend a lot of time on it and the results are not gathered anywhere now (in a logical order that anyone can follow).

**So the steps to do now are:**
1. Let's agree on the framework - based on what I've written and what you've written, please form a bullet-point-based framework of how we should progress (my proposal Is above).
B. Let's create a markdown file according to the framework - from problems, through steps, benchmarks, till final solution proposal.
C. Let's Implement the solution.

s9ferech:
> 1. That's a good point, but if I understand correctly, our benchmarks test all available configurations now, don't they? I mean, sure, this needs to evolve over time, but it should be a good, extensible framework for us. If I'm not right, what possible configs are we missing (we should add them to benchmarks)?

Some things that we are currently missing: setting `contain` or wrapping the root in a shadow DOM or the children in SVGs. But it is okay that we don't have them because the benchmark would like not measure the potential benefits anyway. There are also more exotic solutions like rendering visualizations to a canvas and including that in the WebGL scene. I don't know if there are more because it is an open world. Personally I would not even have tried `render-style: preserve-3d`, but it is good that we did.

> 2. Why? Could you elaborate? The visualizations will be either a single WebGL canvas or a lot of SVG elements. These are 2 possible scenarios only and we can test them in benchmarks.

In the IDE, we won't have as many visualizations and they will be spread out through the scene, which can both affect the performance of different approaches. On the other hand, the rest of the IDE will run in parallel in the same window and they might interact in ways that the benchmarks don't describe. Maybe event handlers on the visualizations will impact performance or maybe D3 does anything else that has an impact. Again, there are so many things that I can't list them.

> 4. I don't understand the reasoning here. We DO optimize for FPS and quality. During movement of the camera - FPS is more important, on a still image, the quality. The questions if offscreen vis change the results is a good question - I'd propose to make additional tests for it (not necessarily in the benchmarks, maybe side-by-side the benchmarks).

What I wanted to say is that a problem specification will likely be incomplete because there are so many things to consider. But even if we fix a specification then we can't express it in a single number. The choice of the right approach can be a complex judgement.

> 5. What do you mean by "without understanding them" ? If we test possible configurations and we pick the one which performs best, we should be fine. If you'd like to really understand why we get these results, then I think reading the source code of Chrome rendering implementation is the only option, but this would be a total overkill. Did you mean something else?

I mean that picking the best number from the results, even assuming that we only optimize FPS, does not suffice. Just to give some examples: The behavior of the browser might depend on the size of the screen or the number and size of visualizations. (Ideally, we will choose a solution that is robust against such changes.)

> 2. The problem I have here is that you did it and this is not reproducible. Anyone from our team or community should see it and understand ALL steps and all reasoning why this solution was chosen. So we need to write it down as a series of well-documented and logically connected steps.

Yes, this is a problem, but it cannot be changed. The only step in our documentation would be "Go on the internet and see what you find." The best evidence that I could refer to would be some unanswered questions on Stack Overflow. But those alone are hardly useful, especially not in a few years from now.

> 3. I don't really understand it. What are "unacceptable flaws", are these blurred fonts? If so, then this is bad - the benchmarks should cover these. Of course they do not need to cover integer-pixels, as jumping during movement is something we definitely o not want to have. However, blurred fonts are not "unacceptable flaw", especially when we are moving things around. This is one of such things that need to be described really, really well, before making the conclusions.

I simplified that point a bit. If we accept blurred fonts then the proposed solution is among the best in our benchmark results. If we do not accept them, then blurred fonts and stuttering in Firefox would be the two relevant flaws, if I don't miss anything right now.

> To sum this up - I understand that you did a lot of research, that's good. Now, before making a decision, you need to show that to the team. In a very, very logical order - starting with problems, through all tests, results, to the final conclusions. This HAS TO base on some **framework**. My proposition is that the framework is simple:
> 
> 1. We show all possible configurations in benchmarks (macOS + Windows + Linux)
> 2. We choose the best ones.
> 3. We check which ones have flaws (like blurred fonts)
> 4. Based on that we propose a solution (a single one or a hybrid one - for moving and still camera). In my opinion, this will guarantee that we will choose the best performing options here.

With this framework, we would have to justify how we avoid the problems that I outline above.

My proposed framework would be that we start with the configuration that we want to use, explaining why it seems reasonable and maybe discussing some other approaches that might look promising but that we did not choose. This is essentially the information that I have in my notes. We can try to find some hard evidence where you think that it is needed. The benchmarks can then be used as a concrete justification, to check if our performance is among the best.

If you look at this solution again in the future, you will either just run the benchmarks and check if it is still among the best, or you will already have an idea in mind how the solution could be improved. In that case, you should examine how it fits into our explanations, do some research on the new configuration and finally add it to the benchmarks to confirm that it is really better.

s9ferech:
> 2. We choose the best ones.

I have one more question about this, before I write it up. How do we define the best one?

wdanilo:
Lets talk tomorrow after standup. Ive got few small questions to the results - I need to confirm the thoughts and then lets implement the proposed solution! :)

s9ferech:
We will implement it as described [here](https://github.com/enso-org/dom-benchmarks), such that we switch between configurations when movements start or end.

s9ferech:
Here is a recording of the demo scene on the current commit.
demo

s9ferech:
This GIF shows how things get blurry while zooming but sharper when they stand still. Only in the end, when the zoom is really strong, the text stays blurred.
zooming

This screenshot shows one of the worst cases where text can also look blurred with lower zoom:
blurred

For comparison, this screenshot shows how it would look with perfectly sharp text:
sharp

wdanilo:
Felix, I don't understand this part:
image

You did not mention that we have still blur problems with not extreme zoom levels. Few questions:
1. What happened here?
2. Why does it happen? Is this implementation not fixing all of such cases?
3. What can we do to prevent it? It looks really bad.

wdanilo (changes requested):
Felix, I don't understand this part:
image

You did not mention that we have still blur problems with not extreme zoom levels. Few questions:
1. What happened here?
2. Why does it happen? Is this implementation not fixing all of such cases?
3. What can we do to prevent it? It looks really bad.

I've got one important request - ALL cases should be really well described and well documented. If you write "This screenshot shows one of the worst cases where text can also look blurred with lower zoom" - there is a big problem with this description. It just does not explain WHAT IS THE CASE - when it happens, why it happens, etc. Anyway, this looks really bad and we need to fix it. Maybe we have some misunderstanding here? I was asking today during the call a lot about blurring problems and you mentioned that they occur only with extreme zoom levels. If they occur with normal zoom levels, then this task should not be considered fixed IMO.

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,22 +113,104 @@ impl DomScene {
...
+        // === Reset configuration ===
+
+        self.data.view_projection_dom.set_style_or_panic("transform", "");
+        self.data.view_projection_dom.set_style_or_panic("left", "0px");
+        self.data.view_projection_dom.set_style_or_panic("top", "0px");
+        self.data.dom.set_style_or_panic("perspective", "");
+
+
+        let view_matrix = camera.view_matrix();
+        let view_matrix = view_matrix.map(eps);
```
MichaelMauderer: Is this still part of `// === Reset configuration ===`? Why does the `view_matrix` need mapping?

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,22 +113,104 @@ impl DomScene {
...
+        // In CSS, the y axis points downwards. (In EnsoGL upwards)
+        let mut trans_cam = flip_y_axis(view_matrix);
```
MichaelMauderer: Why does flipping the axis turn it from a `view_matix` into a transformed camera?

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,22 +113,104 @@ impl DomScene {
...
+                    // scene. This has to be applied on right (that is, in world space) to keep it
+                    // unaffected by the other transformations.
+                    let transform = translateZ + matrix3d.as_str() + "translate(50%,50%)";
+                    self.data.view_projection_dom.set_style_or_panic("transform", transform);
+
+                    let left = target_x * scale;
+                    let top = target_y * scale;
+                    self.data.view_projection_dom.set_style_or_panic("position", "relative");
+                    self.data.view_projection_dom.set_style_or_panic("left", format!("{}px", left));
+                    self.data.view_projection_dom.set_style_or_panic("top", format!("{}px", top));
```
MichaelMauderer: This is a lot of code in a match arm. Maybe it should be a separate method?

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,22 +113,104 @@ impl DomScene {
...
+        let mut trans_cam = flip_y_axis(view_matrix);
 
         match camera.projection() {
             Projection::Perspective{..} => {
-                js::setup_perspective(&self.data.dom,&near.into());
-                setup_camera_perspective(&self.data.view_projection_dom,near,&trans_cam);
+                // In order to achieve a better visual quality when the camera is in fron of the z=0
+                // plane, we simulate translation in the z direction by scaling the scene around
+                // the origin, as discussed at https://github.com/enso-org/ide/pull/1465.
+                // In that situation, we also simulate x and y translation through `top` and `left`.
```
MichaelMauderer: What does it mean to simulate through "top" and "left"? Is there a verb missing here?

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,22 +113,104 @@ impl DomScene {
...
+                let x_index = (0, 3);
+                let y_index = (1, 3);
+                let z_index = (2, 3);
+                let target_x = *trans_cam.index(x_index);
+                let target_y = *trans_cam.index(y_index);
+                let target_z = *trans_cam.index(z_index);
+
+                // Put the camera in front of the origin and adjust the field of view, such that one
+                // px unit at z=0 maps to one px on screen.
+                self.data.dom.set_style_or_panic("perspective", format!("{}px", camera.z_zoom_1()));
```
MichaelMauderer: I think we also have `set_style_or_warn`, which might be safer to use as it doesn't crash the IDE if something goes wrong.

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,22 +113,104 @@ impl DomScene {
...
+                // The position of the scene relative to the camera after setting "perspective".
+                let current_z = -camera.z_zoom_1();
```
MichaelMauderer: Why does it need negating?

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -193,22 +113,104 @@ impl DomScene {
...
+
+                let MAXIMUM_SCALE_FACTOR = 1000.0;
+                // Our method to simulate translation through scaling works only when `target_z` is
+                // less than 0. We cut off at some point before that and fall back to real 3D
+                // transformations. We determine that point through the maximum scale factor that
+                // that we want to allow. (See also the computation of `scale` below)
```
MichaelMauderer: Can you explain why this no longer works if `z >= 0.0`?

s9ferech:
> Felix, I don't understand this part:
> image
> 
> You did not mention that we have still blur problems with not extreme zoom levels. Few questions:
> 
> 1. What happened here?
> 2. Why does it happen? Is this implementation not fixing all of such cases?
> 3. What can we do to prevent it? It looks really bad.
> 
> I've got one important request - ALL cases should be really well described and well documented. If you write "This screenshot shows one of the worst cases where text can also look blurred with lower zoom" - there is a big problem with this description. It just does not explain WHAT IS THE CASE - when it happens, why it happens, etc. Anyway, this looks really bad and we need to fix it. Maybe we have some misunderstanding here? I was asking today during the call a lot about blurring problems and you mentioned that they occur only with extreme zoom levels. If they occur with normal zoom levels, then this task should not be considered fixed IMO.

This happened when I simply zoomed to the zoom level that you can see on the screenshot. We talked about it, but I did notice how bad it could get until I took those screenshots. I believe that it also depends on the visualization that you open and size of the browser window. What happens is just that some heuristic in the browser decides that this is an appropriate resolution to render things. We can prevent this, but we would have to use another configuration than the one that was chosen by our benchmarks. It would likely take a bit of time and some design changes to implement.

s9ferech:
@wdanilo We discussed last week on a call that we should not set the CSS `perspective` property, since it seems to be the cause for the remaining problems. But if we do not set it then we will not get correct results when the camera is tilted or `DomSymbols` reach out of the z=0 plane in any way. There are multiple methods to solve this:

1. We could detect those problematic situations and then switch `perspective` on at runtime. For the camera tilt, that is easy. For transformations of `DomSymbols`, it might be possible with reasonable effort if we make sure that we do not set `transform-style: preserve-3d` on them, and, thereby, make sure that they are just flat surfaces in our 3D scene. Then we would have to check every transformation that gets applied to them.

2. As a simpler solution, you suggested that we forbid 3d transformations on `DomSymbols` by default, but add the ability to manually switch into a mode where `perspective` is set and 3D transformations are allowed.

3. As a variation of 2, we could introduce a form of `DomSymbols` that are, by definition, 2D objects on a flat DOM surface within our scene.

3. We could just us orthographic projection for the whole scene for now. This would be the most elegant solution if it turns out that we would not have to fake 3D transformations by hand, as we do now. It would leave our current perspective mode intact and switching would be easily possible.

4. We could treat JS visualizations as "billboards" that are always directed towards the camera.

Can you please write down which way you want to do it, with enough detail, such that it can be implemented?

wdanilo:
@s9ferech Thanks for writing it down.

1. I believe this solution is pretty complex, especially regarding the discovery of symbols transformations deeply in the display object hierarchy. As we've got pretty tight time constraints, I'd not try doing it right now.

2. This was not exactly my suggestion. My suggestion was to keep the perspective and have an API to disable it. I still feel this might be the easiest solution to follow, but before we make a decision, let me cover the rest of the points and ask few questions.

3. I really like this idea. How would it work? Would you keep 2 roots of DOM scenes then - one with perspective and one without? Then would you keep all 2D symbols in one root and others in the second one?

4. Questions:
    a. Are you talking about using orthographic projection for DOM elements only?
    b. ` if it turns out that we would not have to fake 3D transformations by hand, as we do now` - what do you mean by that? I need further explanation of this sentence.

5. Creating something like `DomSprite` object is also an option, an elegant one. We definitely do not want to remove possibility of having them in 3D, so a separate `DomSprite` type will work here.

What are your time estimates for 2, 3, 4, and 5?

s9ferech:
@wdanilo

> 2. This was not exactly my suggestion. My suggestion was to keep the perspective and have an API to disable it. I still feel this might be the easiest solution to follow, but before we make a decision, let me cover the rest of the points and ask few questions.

I think think would be very easy to implement from where we are right now. But it would also require that we polish the documentation of the existing code, which might take more time. I think this would take some hours. But it would be quite dirty because you could still apply 3D transformations after unsetting `perspective`, they would just lead to results that look slightly wrong.

> 3. I really like this idea. How would it work? Would you keep 2 roots of DOM scenes then - one with perspective and one without? Then would you keep all 2D symbols in one root and others in the second one?

I guess there are different options how you could approach this. Generally, I would think of it as a DOM "canvas" placed inside our 3D scene. This canvas can either be a `display::Object`, or just positioned statically at z=0. And then you could create visualizations as DOM nodes on that canvas. But I believe it would take many days to design and implement this.

> 4. Questions:
>    a. Are you talking about using orthographic projection for DOM elements only?

I am talking about using it for everything.

>    b. ` if it turns out that we would not have to fake 3D transformations by hand, as we do now` - what do you mean by that? I need further explanation of this sentence.

In this PR, we currently fake 3D transformations in 2D to get better image quality. I expect that this will be unnecessary if we use the orthographic projection. The browsers will likely render it in good quality anyway.

About the implementation: I am not familiar with the WebGL side, but this method would likely also be easy to implement. It would require a new implementation of zooming, but we should likely reimplement that anyway (#1613). I estimate that this solution would take slightly longer than 2, with some uncertainty for the WebGL part. But I would prefer the result because it is simpler and does not need any kind of hack.

> 5. Creating something like `DomSprite` object is also an option, an elegant one. We definitely do not want to remove possibility of having them in 3D, so a separate `DomSprite` type will work here.

I think this would not take as much work as 3, but still a few days, since it needs some design work and changes how we handle many things.

wdanilo:
@s9ferech 

> I guess there are different options for how you could approach this. Generally, I would think of it as a DOM "canvas" placed inside our 3D scene. This canvas can either be a display::Object or just positioned statically at z=0. And then you could create visualizations as DOM nodes on that canvas. But I believe it would take many days to design and implement this.

I was asking about the DOM layout - about how do you imagine the implementation. Right now, we have a canvas and a div next to it. The div is the root for all DOM elements that we operate. My question was if you imagine keeping yet another div next to the one that exists and you plan to set "perspective" only on one of them.

> In this PR, we currently fake 3D transformations in 2D to get better image quality. I expect that this will be unnecessary if we use the orthographic projection. The browsers will likely render it in good quality anyway.

When you were researching how to fix the quality issue, that was not among the proposed solutions - why? Have we missed it, or something changed and we have a better picture of the whole problem space now? 

> About the implementation: I am not familiar with the WebGL side, but this method would likely also be easy to implement. It would require a new implementation of zooming, but we should likely reimplement that anyway (#1613). I estimate that this solution would take slightly longer than 2, with some uncertainty for the WebGL part. But I would prefer the result because it is simpler and does not need any kind of hack.

The WebGL thing would be also super simple. It would require adding an Orthographic projection to the camera and just output the right matrix. Everything else will just work. 

HOWEVER, I do not agree with:
1. That it would require a new implementation of zooming - why? Currently, zooming is done by moving the camera closer to / further from the scene. This should work with an orthographic camera without a change.
2. I don't agree with the conclusion that we need to reimplement zooming - this is a way too "hard" conclusion for now. I understand that this is a hypothesis, but I also think that this hypothesis is wrong and I have replied why I think so in the linked thread - I'd love to talk more about it, maybe I'm missing something.

But basically, I really, really like this solution.

> I think this would not take as much work as 3, but still, a few days, since it needs some design work and changes how we handle many things.

Got it, thanks!

-----------

Based on all of this, I believe the orthographic camera solution is the most beautiful one. I'm a little bit sad that we spent so much time on all the research if this could be the chosen solution from the beginning and it could save us a few days. On the other hand, probably the research was needed in order to discover that removing the perspective projection will fix the issue! Could you please confirm this? In your description you've written `The browsers will likely render it in good quality anyway` - this needs to be confirmed before we implement it. Also, if this would be the desired solution, I would implement that in a separate PR - keeping in this PR the "faked zoom" because it is still an absolutely fantastic work. Btw, are the docs already updated and are describing whats happening properly everywhere?

s9ferech:
> @s9ferech
> 
> > I guess there are different options for how you could approach this. Generally, I would think of it as a DOM "canvas" placed inside our 3D scene. This canvas can either be a display::Object or just positioned statically at z=0. And then you could create visualizations as DOM nodes on that canvas. But I believe it would take many days to design and implement this.
> 
> I was asking about the DOM layout - about how do you imagine the implementation. Right now, we have a canvas and a div next to it. The div is the root for all DOM elements that we operate. My question was if you imagine keeping yet another div next to the one that exists and you plan to set "perspective" only on one of them.

We would probably have one root for every "DOM canvas" within the EnsoGL scene.

> > In this PR, we currently fake 3D transformations in 2D to get better image quality. I expect that this will be unnecessary if we use the orthographic projection. The browsers will likely render it in good quality anyway.
> 
> When you were researching how to fix the quality issue, that was not among the proposed solutions - why? Have we missed it, or something changed and we have a better picture of the whole problem space now?

I just did not think of it as an option. Likely because I assumed the perspective projection was a requirement.

> HOWEVER, I do not agree with:
> 
> 1. That it would require a new implementation of zooming - why? Currently, zooming is done by moving the camera closer to / further from the scene. This should work with an orthographic camera without a change.

Moving the "camera" closer does not change the outcome of the orthographic projection.

> 2. I don't agree with the conclusion that we need to reimplement zooming - this is a way too "hard" conclusion for now. I understand that this is a hypothesis, but I also think that this hypothesis is wrong and I have replied why I think so in the linked thread - I'd love to talk more about it, maybe I'm missing something.

I have commented there.

> On the other hand, probably the research was needed in order to discover that removing the perspective projection will fix the issue! Could you please confirm this?

I am sorry to disappoint you, but the fact that perspective projection is the root cause of our problems, was the one thing that I knew long before I even started working on the issue 😅

> Btw, are the docs already updated and are describing whats happening properly everywhere?

There are some docs describing the current state of the PR. But they could be improved. Michael already gave some feedback that could be included. Should I do that?

s9ferech:
I just checked: With orthographic projection, we seem get good quality in Chrome, Firefox and Edge under windows 10, if the view matrix does not modify the w coordinate (which it likely does not do in our current implementation).

wdanilo:
> Moving the "camera" closer does not change the outcome of the orthographic projection.

Oh, you are of course right. For a moment I confused it. Ok, this is bad. In order to make the navigator work correctly with orthographic projection as smooth as its done right now it would require a lot of research (probably many days and many hours of talking). The navigator now works as a "camera on a rubber band" that has a smooth movement even if you change the movement while it is already beating animated. We want to keep it this way. The idea that Z-axis movement of the camera makes the objects smaller and bigger is very important for us. 

Also, we will be using perspective projection a lot, for example, while doing animations of entering nodes with nodes "dropping on the scene" soon (after entering a node, nodes will be dropping on the scene by changing their z position from something bigger to 0, making them bigger -> smaller during this animation), so the longer I think about it, the more I realize that this idea will cause much more trouble than good.

---------

Ok. Felix, after considering all of the solutions above I believe that:

The best solution would be here to create `DomSpriteSymbol` (similar to the current `DomSymbol`) which will always be facing the camera. This is the standard way of how you do it in OpenGL. 

In order to do it properly we would need to create the following actions:
1. Create `DomSpriteSymbol` type, similar to `DomSymbol` currently, which will just always face the camera.
2. Add a new dom node to [DomScene](https://github.com/enso-org/ide/blob/develop/src/rust/ensogl/lib/core/src/display/scene/dom.rs) that will be next to the existing dom node, but will not have perspective turned on.
3. Rename the function [manage](https://github.com/enso-org/ide/blob/develop/src/rust/ensogl/lib/core/src/display/scene/dom.rs#L179) in `DomScene` to `managePerspectiveSymbol`.
4. Create a new function `manageSpriteSymbol` in `DomScene` which will add the `DomSpriteSymbol` to a DOM layer that does not have perspective projection enabled.
5. Create a function `manage` in some trait in `DomScene` such, that when we will pass `DomSymbol` it will use `managePerspectiveSymbol` and if we use it with `DomSpriteSymbol` it will use `manageSpriteSymbol` instead. This way, the API will not change in any place and we will just have another type o DOM symbols.

And this should work. If you believe that there is something not clear / hard to implement, let's have a call about it.

---------

In the absolutely worst-case scenario, if the above would be hard to implement because of any reason (I believe it should not), let's create the API that explicitly disables perspective projection in DOM scene (of course IFF this does not make the objects "desynchronized" with WebGL objects in any way with the camera not tilted on X and Y axes).

s9ferech:
I just noticed that this implementation would make it useless to apply `will-change: transform` on the DOM scene root node, as we had planned. This is because we would have to transform all the visualizations individually. I think it would be fine, but it could be a performance problem if we have a big number of visualizations.

If we want to fix this then we have to use a solution like number 2 or 3 from above, where all visualizations would be guaranteed to lie on one plane in our 3D space.

s9ferech:
We just had a call and agreed on a variant of solution 2:

We expose a method to unset the `perspective` property on the scene's root DOM node. When it is disabled then we will provide high quality graphics. But the user of the API will have to guarantee that the camera is always directed in the z direction and all DOM objects lie in the z=0 plane.

wdanilo:
This will take time to review. Leaving it for a couple of days for now.

MichaelMauderer:
This will be revisited at a later date.

PR #1536, opened 2021-04-29
Correctly interpret string arguments as booleans in electron arguments.

MichaelMauderer:
### Pull Request Description
Previously arguments were passed to the main function as strings when using the electron app, or as booleans when using the development environment. Since `false` is a truthy value that was not further checked, this lead to wrong settings. 
This PR adds a conversion of `true` and `false` to their boolean values, and adds an assertion that ensures that arguments are actually booleans. 

### Checklist

### DIFF
--- a/src/js/lib/content/src/index.js
+++ b/src/js/lib/content/src/index.js
@@ -7,6 +7,7 @@ import * as html_utils    from 'enso-studio-common/src/html_utils'
 import * as animation     from 'enso-studio-common/src/animation'
 import * as globalConfig  from '../../../../config.yaml'
 import cfg                from '../../../config'
+import assert             from "assert";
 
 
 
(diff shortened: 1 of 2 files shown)


src/js/lib/content/src/index.js
```diff
@@ -438,6 +439,36 @@ function ok(value) {
     return value !== null && value !== undefined
 }
 
+/// Check whether the value is a string with value `"true"`/`"false"`, if so, return the
+// appropriate boolean instead. Otherwise, return the original value.
+function tryAsBoolean(value) {
```
wdanilo: I think the name of the function could be more descriptive, like `parseBooleanOrLeaveAsIs` I know its longer, but "try" in name tells me that we are returning something like Option, but here we are returning the original value. It could also be `tryAsBooleanOrLeaveAsIs`, but `parse` is better as it tells that we are working with string input.

src/js/lib/content/src/index.js
```diff
@@ -438,6 +439,36 @@ function ok(value) {
...
+    return value
+}
+
+/// Turn all values that have a boolean in string representation (`"true"`/`"false"`) into actual
+/// booleans (`true/`false``).
+function ensureBooleans(config) {
+    for (const key in config) {
+        config[key] = tryAsBoolean(config[key])
+    }
+}
```
wdanilo: I'd call it `parseAllBooleans` instead - I think this name is more descriptive.

wdanilo (approved):

mwu-tow:
I think this should be connected to https://github.com/enso-org/ide/issues/1512
Also converting other well-known types (like numbers) might have sense here.

MichaelMauderer:
> I think this should be connected to #1512
> Also converting other well-known types (like numbers) might have sense here.

I think, ideally we would convert the whole file (and all JS files bit by bit) to TypeScript and implement this properly using a `Config` class that gets instantiated and validates its fields.

MichaelMauderer:
Closed in favour of #1539

Issue #1068, opened 2020-12-15
Expose profiling information in Controllers


farmaazon:

farmaazon:
Engine will provide a similar API as for visualizations: we'll have methods like "attach profiler" "detach profiler" and "profile info" notification.

sylwiabr:
I think it is unblocked now, @farmaazon ?

Issue #1031, opened 2020-12-06
Text coloring is very slow


wdanilo:
### General Summary
Currently, the text coloring is implemented in a very naive way. In `component/area.rs` we can see:
```rust
eval input.set_color_bytes ((t) {
    m.buffer.frp.set_color_bytes.emit(*t);
    m.redraw();
});
```
This basically means that for each color set, the whole text is re-drawn (thrown away and created again). This is particularly causing a slowdown with color animations - showing a single node may cause its expression to be re-drawn even 500x times.

### Expected Result
- Text coloring should update existing glyphs colors.

### Actual Result
- All glyphs are re-created.

PR #1546, opened 2021-05-06
Add profiling mode to GUI

s9ferech:
### Pull Request Description

This adds a profiling mode with running time indicators to the graph editor, following the description of issue #1069. This PR includes just the visual part. Retrieval of the profiling information is handled in Issue #1068.

The profiling mode can be entered and exited through <kbd>ctrl</kbd>+<kbd>p</kbd>. (Holding <kbd>ctrl</kbd> causes the node labels to appear temporarily.)

(Outdated recording)
Enabling and disabling the profiling mode.

The appearance has an edge case when the node gets smaller than the running time label (Outdated recording):
Edge case on small nodes.

### Checklist


s9ferech:
I assume this should not be included in the CHANGELOG before we finish the epic, right?

docs/product/shortcuts.md
```diff
@@ -38,6 +38,7 @@ further investigation.
 | <kbd>ctrl</kbd>+<kbd>q</kbd>                                | Close the application (Linux) |
 | <kbd>alt</kbd>+<kbd>F4</kbd>                                | Close the application (MacOS, Windows, Linux) |
 | <kbd>ctrl</kbd>+<kbd>w</kbd>                                | Close the application (Windows, Linux) |
+| <kbd>ctrl</kbd>+<kbd>p</kbd>                                | Toggle profiling mode |
```
farmaazon: As it is not fully working yet, add the appropriate mark (see the preamble of this file).

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -12,12 +12,15 @@ pub mod output;
...
 pub mod vcs;
+mod profiling_indicator;
 
 pub use error::Error;
 pub use expression::Expression;
 
 use crate::prelude::*;
 
+use profiling_indicator::ProfilingIndicator;
+use crate::Mode as EditorMode;
```
farmaazon: Mixed imports.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -370,7 +396,7 @@ pub struct NodeModel {
 
 impl NodeModel {
     /// Constructor.
-    pub fn new(app:&Application, registry:visualization::Registry) -> Self {
+    pub fn new(app:&Application,registry:visualization::Registry) -> Self {
```
farmaazon:
```suggestion
    pub fn new(app:&Application, registry:visualization::Registry) -> Self {
```
we do put space in argument list (to clearly separate arguments).

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -101,6 +104,24 @@ fn traffic_lights_gap_width() -> f32 {
...
+pub enum Mode {
+    Normal,
+    Profiling
+}
+
+impl Default for Mode {
+    fn default() -> Self {
+        Mode::Normal
+    }
+}
```
farmaazon: Docs

s9ferech:
The profiling mode only grays out edges currently. Should anything else be included?

s9ferech:
The minimum and maximum running times are currently recomputed on every change, to keep things simple. If this should be too costly then I would suggest to use an existing priority queue implementation rather than the algorithm in the issue description. That should likely provide some simplicity and best performance.

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -592,11 +600,16 @@ impl Area {
...
+                    pointer_style_over  <- map4
+                        (&mouse_over,&frp.set_ports_active,&port.tp,&frp.editor_mode,
+                        move |_,(_,edge_tp),port_tp,editor_mode| {
                             let tp    = port_tp.as_ref().or_else(||edge_tp.as_ref());
                             let color = tp.map(|tp| type_coloring::compute(tp,&styles));
                             let color = color.unwrap_or(any_type_sel_color);
+                            let color = match editor_mode {
+                                EditorMode::Normal    => color,
+                                EditorMode::Profiling => any_type_sel_color
+                            };
```
farmaazon: We do the color computation every time, even if it is discarded afterwards (because we are in Profiling mode). Instead, lets compute the color only when we actually use it.

src/rust/ide/view/graph-editor/src/component/node/profiling_indicator.rs
```diff
@@ -0,0 +1,194 @@
...
+                    _                                                      => false,
+                }
+            });
+
+            color.target_alpha <+ visibility.map(|&is_visible| {
+                if is_visible {
+                    1.0
+                } else {
+                    0.0
+                }
```
farmaazon:
```suggestion
                if is_visible { 1.0 } else { 0.0 }
```

src/rust/ide/view/graph-editor/src/component/node/profiling_indicator.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        frp::extend! { network
+
+            // === Visibility ===
+
+            visibility <- all_with(&frp.set_editor_mode,&frp.set_execution_status,|mode,status| {
+                match (mode,status) {
+                    (EditorMode::Profiling,ExecutionStatus::Finished {..}) => true,
+                    _                                                      => false,
+                }
```
farmaazon: use `matches!` macro instead.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -61,6 +62,8 @@ use ensogl::prelude::*;
 use ensogl::system::web;
 use ensogl_theme as theme;
 
+use std::f32;
+
```
farmaazon: remove the blank space before

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -101,6 +104,24 @@ fn traffic_lights_gap_width() -> f32 {
 
 
 
+// ============
+// === Mode ===
+// ============
+
+#[derive(Debug,Copy,Clone,CloneRef,PartialEq)]
```
farmaazon: Let's implement Eq as well.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -937,6 +967,77 @@ impl Edges {
...
+            min_duration (f32),
+            max_duration (f32),
+        }
+    }
+
+    #[derive(Debug,Clone,CloneRef,Default)]
+    pub struct ExecutionStatuses {
+        frp       : Frp,
+        // If performance becomes an issue then we could try to store the durations in efficient
+        // priority queues instead, one to determine the minimum and one to determine the maximum.
```
farmaazon: We could also use https://docs.rs/bimap/0.6.1/bimap/btree/struct.BiBTreeMap.html (I already used bimap crate somewhere else in the project).

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -937,6 +967,77 @@ impl Edges {
...
+                NodeExecutionStatus::Finished {duration} => self.durations.insert(node,duration),
+                _ => self.durations.remove(&node),
+            };
+            self.update_min_max();
+        }
+
+        pub fn remove(&self,node:NodeId) {
+            self.durations.remove(&node);
+            self.update_min_max();
+        }
```
farmaazon: This API and logic can also be done in FRP.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1817,10 +1934,13 @@ impl GraphEditorModel {
...
-        opt_color.unwrap_or_else(||
-            color::Lcha::from(styles.get_color(theme::code::types::any::selection))
-        )
+        let opt_color     = edge_type.map(|t|type_coloring::compute(&t,&styles));
+        let neutral_color = color::Lcha::from(styles.get_color(theme::code::types::any::selection));
+        let type_color    = opt_color.unwrap_or(neutral_color);
+        match self.frp.mode.value() {
+            Mode::Normal    => type_color,
+            Mode::Profiling => neutral_color
+        }
```
farmaazon: Same as before: let's compute color only when it's needed.

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: I think I do not understand the style guide in this regard. Can you tell me how it should be done?

src/rust/ide/view/graph-editor/src/lib.rs
s9ferech: If we want to avoid additional dependencies then I would rather use `BTreeSet` that seems more convenient and more efficient at first sight.

s9ferech: Do you agree?

src/rust/ide/view/graph-editor/src/lib.rs
s9ferech: I am not sure how you would do the logic in FRP. Do you think it should be done that way?

src/rust/ide/view/graph-editor/src/component/node/profiling_indicator.rs
s9ferech: `matches!` seems to be marked as unstable. Should I still use it?

src/rust/ide/view/graph-editor/src/component/node.rs
farmaazon: first section: `use crate::prelude::*` and other preludes if any
second section: all imports from crate: `use crate::<...>`. Usually all things imported this way should be imported using "absolute path" (starting with `crate::`)
third section: imports from other crates.

In this case we have `use profiling_indicator::ProfilingIndicator;` looking like import from another crate (but if it is from the current one, it should be imported with full path).

src/rust/ide/view/graph-editor/src/lib.rs
farmaazon: It is already a dependency. And why BTreeSet is more convenient? the BiBTreeMap will replace both the "direct" map nodeId->time and the set of times, and will synchronize them. And it should not be inefficient as it uses the BTreeSet itself.

src/rust/ide/view/graph-editor/src/lib.rs
farmaazon: I would do it this way:

1. in define_endpoints macro above add inputs "set" and "remove" with proper inputs
2. in add FRP nodes mapping the "set" and "remove" inputs, so they change `durations` set accordingly and return the min/max times.
3. redirect results from nodes defined in 2 to the outputs.

Then, instead of emitting `set` you can redirect some other FRP output with `<+`

And yes, I think it could be profitable; some time ago we had the debug printer for FRP graphs, which we may restore some day. And then we will see the relationship between calling set and updaing min/max time.

src/rust/ide/view/graph-editor/src/lib.rs
s9ferech: Thinking about it again, `BTreeSet` would not work of course, because we lose the connection to the nodes. But I could not find any reasonable way to access the maximum through the interface of `BiBTreeMap`. Did I miss something?

wdanilo:
1. Regarding not including it in README - why? Every task when merged should be included in the readme if it implements something that can be described from the user perspective.
2. I'm not checking the code yet, as the visual part is not implemented according to the description. The description states that the background of nodes should get the color according to its performance, there is no description of some additional borders around the nodes. When nodes get colored background, everything else should be grayed out (everything that has colors).
3. I replied to comments on the issue description, please see them as well.
4. Is one of the requirements described by Sylwia also implemented here? (the ability to profile selected nodes). If so, how can we enable it? Is there a shortcut for it?
5. Is there an icon implementation? The task states "an icon displayed in the top-right corner of the app" - the description used the word "shortcut or icon", but that was incorrect - there should be shortcut and icon - we've been talking about it during the estimation meeting.
6. Regarding the comment:

> The minimum and maximum running times are currently recomputed on every change, to keep things simple. If this should be too costly then I would suggest using an existing priority queue implementation rather than the algorithm in the issue description. That should likely provide some simplicity and the best performance.

I do not understand it tbh. The algorithm described in the issue is not recomputing minimum and maximum running times on every change - and the algorithm was part of the requirements. If you want to change the requirements, let's talk about it before implementation, please. Also, we care a lot about performance here - we've been talking about it during planning / estimation meeting - I told then that If there is any more efficient algorithm, let's talk about it before implementing the task. Could you describe the algorithm that uses the priority queue step by step? If it has better performance, it should be implemented instead. However, again, we should have been talking about it before implementing this task, not after, as it would mean throwing away some part of the work - am I correct here?

wdanilo (changes requested):
It's not implemented accroding to the spec, see my comment above please.

src/rust/ide/view/graph-editor/src/lib.rs
farmaazon: You can use the https://docs.rs/bimap/0.6.1/bimap/btree/struct.BiBTreeMap.html#method.right_values iterator - it is in ascending order and double ended, so you can quickly get the first and last element.

s9ferech: Thanks! I did not notice that it was double ended.

s9ferech:
> 1. Regarding not including it in README - why? Every task when merged should be included in the readme if it implements something that can be described from the user perspective.

The only thing that currently happens from the user's perspective is that <kbd>ctrl</kbd>+<kbd>P</kbd>. Switches the colors off. The profiling mode can only be used when the whole epic is finished. I would suggest to update the CHANGELOG then.

> 2. I'm not checking the code yet, as the visual part is not implemented according to the description. The description states that the background of nodes should get the color according to its performance, there is no description of some additional borders around the nodes. When nodes get colored background, everything else should be grayed out (everything that has colors).

To make sure that I understand it right, by background you mean the area inside the node that is behind the text and currently white?

> 4. Is one of the requirements described by Sylwia also implemented here? (the ability to profile selected nodes). If so, how can we enable it? Is there a shortcut for it?

I created another issue for that, as you requested. It is not implemented yet.

> 5. Is there an icon implementation? The task states "an icon displayed in the top-right corner of the app" - the description used the word "shortcut or icon", but that was incorrect - there should be shortcut and icon - we've been talking about it during the estimation meeting.

It is not there yet. I have a question about the design before I implement this. The icon should probably be used as indicator to let users know when they are in profiling mode. We could try to do this through color or present it as a switch between modes. Maybe we will also add more modes later. What is your opinion on this?

> 6. Regarding the comment:
> 
> > The minimum and maximum running times are currently recomputed on every change, to keep things simple. If this should be too costly then I would suggest using an existing priority queue implementation rather than the algorithm in the issue description. That should likely provide some simplicity and the best performance.
> 
> I do not understand it tbh. The algorithm described in the issue is not recomputing minimum and maximum running times on every change - and the algorithm was part of the requirements. If you want to change the requirements, let's talk about it before implementation, please. Also, we care a lot about performance here - we've been talking about it during planning / estimation meeting - I told then that If there is any more efficient algorithm, let's talk about it before implementing the task. Could you describe the algorithm that uses the priority queue step by step? If it has better performance, it should be implemented instead. However, again, we should have been talking about it before implementing this task, not after, as it would mean throwing away some part of the work - am I correct here?

I implemented it this way, because it was almost no additional effort and makes it a bit easier to discuss the algorithm. There are slightly different ways how the priority queues could be used. What Adam suggested, would be to use a `BiBTreeMap` that maps node IDs to running times. Then `set` and `remove` will just keep this map updated and we can always get the min and max through the `right_values` iterator. I would likely make the updates faster that we need when we remove a node or reset its running time. But it would also make the other cases a little bit slower. It is probably not worth to discuss the different algorithms in detail. Let me just know which one you want and I will implement it. The one in the task description is also reasonable.

s9ferech:
I just implemented that syntax highlighting for types will be disabled in profiling mode and the performance during the transition is incredibly bad, roughly 5-10 FPS in the interface demo scene. @farmaazon @wdanilo Does one of you have an idea why that could be?

wdanilo:
> I just implemented that syntax highlighting for types will be disabled in profiling mode and the performance during the transition is incredibly bad, roughly 5-10 FPS in the interface demo scene. @farmaazon @wdanilo Does one of you have an idea why that could be?

1. Are you using an optimized build?
2. If you have such problems, please always record the profiling in chrome and attach it here, otherwise it would be just guessing.

wdanilo:
> I implemented it this way, because it was almost no additional effort and makes it a bit easier to discuss the algorithm. There are slightly different ways how the priority queues could be used. What Adam suggested, would be to use a BiBTreeMap that maps node IDs to running times. Then set and remove will just keep this map updated and we can always get the min and max through the right_values iterator. I would likely make the updates faster that we need when we remove a node or reset its running time. But it would also make the other cases a little bit slower. It is probably not worth to discuss the different algorithms in detail. Let me just know which one you want and I will implement it. The one in the task description is also reasonable.

The idea to use a priority map here is a really nice one. And you are right - it would be much faster in the general case. It will have worse performance than the described algorithm for just updates, but for deletion, etc, it will behave much, much faster. Let's go with it then, it sounds like a much better design!.

> It is not there yet. I have a question about the design before I implement this. The icon should probably be used as indicator to let users know when they are in profiling mode. We could try to do this through color or present it as a switch between modes. Maybe we will also add more modes later. What is your opinion on this?

Let's implement it as part of this task. The icon should be gray on the right top corner of the app and when clicked, it should turn red / orange and be this color while the profiling mode is on. Also, remember that we have buttons implementation in the IDE - there is one created by Michael and one created by Michal - chat with them to choose the right abstraction, please! :)

Regarding the design, I'll attach something in couple of minutes here.

wdanilo:
@s9ferech let's use this icon, but without the outer circle (feel free to adjust the width of the borders to make it look crisp in small size):

image

wdanilo:
> To make sure that I understand it right, by background you mean the area inside the node that is behind the text and currently white?

That is correct. Lets use strong colors then and lets use white text then on nodes. This will allow users to track the performance easy, even when zooming far away from the canvas :)

s9ferech:
> @s9ferech let's use this icon, but without the outer circle (feel free to adjust the width of the borders to make it look crisp in small size):
> 
> image

@wdanilo Do you have a suggestion what's the best way to produce the needle shape? Maybe we should add rounded triangles to the primitives?

s9ferech:
I checked the performance during mode switching in the optimized binary (compiled with `node run dist`). It is a bit better with optimizations, but still a big problem. The profiling info, that I recorded with Chrome, can currently be found [here](https://drive.google.com/file/d/1hSJAj1ae3ecTxFcYD-jDl5MQNTS41zeu/view?usp=sharing). @wdanilo

wdanilo:
@s9ferech 

> I checked the performance during mode switching in the optimized binary (compiled with node run dist). It is a bit better with optimizations, but still a big problem. The profiling info, that I recorded with Chrome, can currently be found here. @wdanilo

What does it mean "a big problem"? Earlier it was 5 FPS, what is it now? Please, always be very precise when describing such issues, otherwise we would spend a lot more time just to understand where we are. Also, when attaching profiling information, please attach the information which does not have removed names of functions, as this does not tell us anything (optimized build removes the names). 

Anyway, on the first sight it looks like during a frame, the colors were refreshed hundreds if not thousands of times. It looks like some FRP events are emitted way too often. This normally means, that there is a bug in the code and instead of emitting 1 event, someone is emitting events in a loop, for example, for each component multiplied by number of components. But to be sure, we would need to see a correct profiling file.

Also, I've got other suspections. Does it happen if you do not change the color of the text?


> @wdanilo Do you have a suggestion what's the best way to produce the needle shape? Maybe we should add rounded triangles to the primitives?

I believe that triangle + grow should make it. If I'm correct, the triangle SDF is implemented in such a way that the corners will get rounded when growing the triangle. Please check it out and tell me if Im correct.

-------

Also, regarding the problem with performance:
1. Please check the questions I posted above to understand where it originates from.
2. Let's make a small design change (which actually will probably improve the whole thing as well) and in case it will appear that the performance problem is connected with letters coloring, it will not cause the problem anymore. The idea is that instead of changing the background color and making the letters white, do this:
   a. Show a new background over the node (over the text of the node), so when the background will appear, the node code will NOT be visible.
   b. Show the profiling information (the time) ON that background.
   c. So in the end, when profiling mode is on, the expressions on nodes will not be visible, only colors and profiling information. 
    d. This background (and timing info) should disappear when moving mouse over a node.

s9ferech:
> Also, I've got other suspections. Does it happen if you do not change the color of the text?

It only happens when the text color is changed. I am now pretty sure that at least the major part of the problem comes from issue #1031.

s9ferech:
> What does it mean "a big problem"? Earlier it was 5 FPS, what is it now?

"Big problem" means that it needs to be fixed. In small scenes, like the interface demo, I still got 60 FPS. But in slightly larger scenes, it got noticable. If it is still relevant then I can record some profiling info, but I think that is not necessary since I have already found the source of the problem.

s9ferech:
> > @wdanilo Do you have a suggestion what's the best way to produce the needle shape? Maybe we should add rounded triangles to the primitives?
> 
> I believe that triangle + grow should make it. If I'm correct, the triangle SDF is implemented in such a way that the corners will get rounded when growing the triangle. Please check it out and tell me if Im correct.

This does not seem to work. It stays a triangle when you apply `grow`. What would your other suggestion be? @wdanilo

wdanilo:
@s9ferech interesting, it means that our triangle implementation is highly optimized (on the other hand, Im not surprised :P). It is here btw: https://github.com/enso-org/ide/blob/develop/src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs#L268 

The mathematically correct equation is much slower and can be found below. Also, we are missing some interesting shapes, like "Uneven Capsule" one. See it here: https://iquilezles.org/www/articles/distfunctions2d/distfunctions2d.htm

I think that the best way here would be to implement the "Uneven Capsule" shape and just use it - see how Triangle was implemented. In case you'll have any problems, ping me on Discord and I'll help!

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/enso-org/ide">enso-org/ide</a></strong></div><div>Enso – a visual and textual functional programming language. - enso-org/ide</div></blockquote>
<blockquote><div><strong><a href="https://iquilezles.org">Inigo Quilez</a></strong></div><div>Articles on computer graphics, math and art</div></blockquote>

s9ferech:
I have made the requested changes. Here is how it looks now (outdated):

Demo of the design.

JS visualizations get grayed out in profiling mode, but implementing this for EnsoGL visualizations would likely take much more work. I suggest to create a separate issue for that.

(Outdated recording)
A JS visualizations gets grayed out when entering profiling mode.

Output ports are still be shown in color inside profiling mode, which seems good to me, since they only appear on hover.

(Outdated screenshot)
Output port in profiling mode

s9ferech:
In the theme definition, we currently have `code::syntax` and `graph_editor::node::text`, which seem to be intended for the same purpose. Should they be merged?

wdanilo:
There is something bad hapenning with parentheses and the right side of the node (See black dots there):
image

docs/product/shortcuts.md
```diff
@@ -38,6 +38,7 @@ further investigation.
 | <kbd>ctrl</kbd>+<kbd>q</kbd>                                | Close the application (Linux) |
 | <kbd>alt</kbd>+<kbd>F4</kbd>                                | Close the application (MacOS, Windows, Linux) |
 | <kbd>ctrl</kbd>+<kbd>w</kbd>                                | Close the application (Windows, Linux) |
+| :warning: <kbd>ctrl</kbd>+<kbd>p</kbd>                                | Toggle profiling mode |
```
wdanilo: alignment

src/rust/ensogl/lib/components/src/toggle_button.rs
```diff
@@ -213,7 +213,7 @@ impl<Shape:ColorableShape+'static> ToggleButton<Shape>{
             // === State ===
 
             toggle <- any(frp.toggle,icon.mouse_down);
-            frp.source.state <+ toggle.toggle();
+            frp.source.state <+ frp.state.not().sample(&toggle);
```
wdanilo, line 216: why this change? IMO this sohuld work the same way

src/rust/ensogl/lib/core/src/data/color/space/def.rs
```diff
@@ -520,6 +520,15 @@ impl Lcha {
     pub fn to_javascript_string(&self) -> String {
         Rgba::from(self).to_javascript_string()
     }
+
+    /// Convert the color to gray by setting chroma to zero.
+    pub fn desaturate(&self) -> Lcha {
```
wdanilo: `Lcha` is `Copy` so Im wonderingwhy linting doesn't tell it should be `self` not `&self` - please check it out, maybe linter is broken here somehow?

If this is copy, you can simplify the code below with `..self`

src/rust/ensogl/lib/core/src/display/scene/dom.rs
```diff
@@ -175,6 +175,12 @@ impl DomScene {
         self.data.dom.set_style_or_warn("z-index", z.to_string(), &self.logger);
     }
 
+    /// Sets the CSS property `filter: grayscale({value})` on this element. A value of 0.0 displays
+    /// the element normally. A value of 1.0 will make the element completely gray.
+    pub fn set_grayscale(&self, value:f32) {
```
wdanilo: I'll name it `filter_grayscale` so it woukd be more connected to css naming

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
```diff
@@ -271,6 +271,41 @@ define_sdf_shapes! {
...
+
+        float max_radius = max(radius_top,radius_bottom);
+        float min_x = -max_radius;
+        float max_x = max_radius;
+        float min_y = -radius_bottom;
+        float max_y = inner_height + radius_top;
+        BoundingBox bounds = bounding_box(min_x,max_x,min_y,max_y);
+
+        return bound_sdf(dist,bounds);
+    }
```
wdanilo: a new EnsoGL shape, YAY! <3

src/rust/ensogl/lib/theme/src/lib.rs
```diff
@@ -308,6 +314,12 @@ define_themes! { [light:0, dark:1]
                 stripe_width = 10.0 , 10.0;
                 stripe_angle = 45.0 , 45.0;
             }
+            profiling {
+                lightness = code::types::lightness , code::types::lightness;
+                chroma    = code::types::chroma    , code::types::chroma;
+                min_hue   = 0.38                   , 0.38;
+                max_hue   = 0.07                   , 0.07;
```
wdanilo: this sohuld be named `min_time_hue` and `max_time_hue`

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -12,12 +12,15 @@ pub mod output;
 pub mod error;
 #[deny(missing_docs)]
 pub mod vcs;
+mod profiling_indicator;
```
wdanilo: we are not using such names - we do not have "selection_indicator", we just have "selection"

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -12,12 +12,15 @@ pub mod output;
...
 
 pub use error::Error;
 pub use expression::Expression;
 
 use crate::prelude::*;
 
+use crate::component::node::profiling_indicator::ProfilingIndicator;
```
wdanilo: Instead of using repeated names liek this we are always using qualified imports and non repeated names. Always prefer qualified imports over unqualified ones. In this case, this would be `profiling_indicator::ProfilingIndicator` which is bad. It should be `profiling::Indicator`. Another thing is that "Indicator" is not very descriptive - IMO better name should be used here. I cannot help with it, as I do not understand what the "indicator" really is.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -205,6 +208,24 @@ pub mod error_shape {
...
+pub enum ExecutionStatus {
+    Running,
+    Finished { duration: f32 }
+}
+
+impl Default for ExecutionStatus {
+    fn default() -> Self {
+        ExecutionStatus::Running
+    }
+}
```
wdanilo: no docs

wdanilo: Plus, it sohuld go to separate file `node::execution::Status`

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -395,13 +421,15 @@ impl NodeModel {
...
         let error_indicator_logger  = Logger::sub(&logger,"error_indicator");
 
-        let error_indicator = error_shape::View::new(&error_indicator_logger);
-        let backdrop        = backdrop::View::new(&main_logger);
-        let background      = background::View::new(&main_logger);
-        let drag_area       = drag_area::View::new(&drag_logger);
-        let vcs_indicator   = vcs::StatusIndicator::new(app);
-        let display_object  = display::object::Instance::new(&logger);
+        let error_indicator     = error_shape::View::new(&error_indicator_logger);
+        let profiling_indicator = ProfilingIndicator::new(app);
```
wdanilo: this "profiling_indicator" looks like a separate shape - if so, why it is done as yet another shape? Node already has a background with a color - why the background is just not changing color instead?

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -12,12 +12,15 @@ pub mod output;
...
 pub mod vcs;
+mod profiling_indicator;
 
 pub use error::Error;
 pub use expression::Expression;
 
 use crate::prelude::*;
 
+use crate::component::node::profiling_indicator::ProfilingIndicator;
+use crate::Mode as EditorMode;
```
wdanilo: I dont like the mixed repsonsibilities here. Graph editor knows about nodes, so it is a hiher-level abstraction for nodes. Nodes should not know about graph editor, but because of this import they know about it, as "EditorMode" is a graph editor concept. Instead, nodes should define their own "display mode". With the proposed design nodes know about graph editor and graph editr knows about nodes, which is a bad design decision

wdanilo: I'd be ok if the display mode would be a separate entity not connected to graph editor, like `crate::display::Mode`, but as long as we call ti `EditorMode` it is a no-go for me

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -587,7 +617,16 @@ impl Node {
...
-       }
+
+
+            // === Editor Mode ===
+
+            model.input.set_editor_mode               <+ frp.set_editor_mode;
+            model.profiling_indicator.set_editor_mode <+ frp.set_editor_mode;
+            model.vcs_indicator.set_visibility        <+ frp.set_editor_mode.map(|&mode| {
+                !matches!(mode,EditorMode::Profiling {..})
+            });
```
wdanilo: I do not understand the reasoning for this solution:
1. It uses a separate shape which is much slower than just reusing the current background.
2. It does not use animations, while the current background has animations built-in (as far as I'm aware). So this is just slower and more complex than reusing the current background

Maybe this is proper solution, but I cannot know that, as the "ProfilingIndicator" does not have any docs, so I don't understand what it is, while its name is not descriptive.

s9ferech:
> There is something bad hapenning with parentheses and the right side of the node (See black dots there):

The black dots were just an artifact from bad GIF compression. I edited my message with a better version. @wdanilo

But the black parentheses are a bit strange. They are black because they are not covered by the syntax highlighting. I believe this is an issue with the debug scene, not the node implementation, but I might look into that if you think that it is a problem.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -601,9 +640,12 @@ impl Node {
             frp.source.error <+ frp.set_error;
             is_error_set <- frp.error.map(|err| err.is_some());
             no_error_set <- not(&is_error_set);
-            error_color_anim.target <+ frp.error.map(f!([style](error)
-                Self::error_color(error,&style))
-            );
+            error_color_anim.target <+ all_with(&frp.error,&frp.set_editor_mode,f!([style](error,&mode)
+                match mode {
+                    EditorMode::Normal    => Self::error_color(error,&style),
+                    EditorMode::Profiling => Self::error_color(error,&style).desaturate(),
```
wdanilo: this code is repetitive, please refactor the common part.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -663,15 +705,55 @@ impl Node {
...
+                    match status {
+                        ExecutionStatus::Running => color::Lcha::from(bgg),
+                        ExecutionStatus::Finished {duration} => {
+                            let hue_delta         = max_hue - min_hue;
+                            let relative_duration = (duration - min) / (max - min);
+                            let hue               = min_hue + relative_duration * hue_delta;
+                            let alpha             = 1.0;
+                            color::Lcha::new(lightness,chroma,hue,alpha)
+                        }
+                    }
```
wdanilo: this should be refactored to a model utility. We should keep in FRP the minimum network definition, while all bigger computations should be refactored out.

src/rust/ide/view/graph-editor/src/component/node.rs
```diff
@@ -663,15 +705,55 @@ impl Node {
...
+                });
+
+            bg_color_anim.target <+ all_with4(&bgg,&frp.set_disabled,&frp.set_editor_mode,&profiling_color,
+                f!([model](bgg,disabled,&mode,profiling_color) {
+                    model.input.frp.set_disabled(*disabled);
```
wdanilo: emitting FRP event from FRP lambda is an anti-pattern, please do not emit events from lambdas nor model functions. This should be done using normal FRP connections outside of this lambda. Currently it has also a very strange behavior - when the "editor_mode" changes, you're re-emitting the disabled state - is it on purpose?

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -387,6 +395,18 @@ impl Area {
             // === Expression Type ===
 
             eval frp.set_expression_usage_type (((a,b)) model.set_expression_usage_type(a,b));
+
+
+            // === Editor Mode ===
```
wdanilo: spacing

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -665,29 +688,65 @@ impl Area {
...
+                            };
+                            let selected_rgba  = styles.get_color(theme::code::types::selected);
+                            let disabled_rgba  = styles.get_color(path.sub("disabled"));
+                            let expected_rgba  = styles.get_color(path.sub("expected"));
+                            let selected_color = color::Lcha::from(selected_rgba);
+                            let disabled_color = color::Lcha::from(disabled_rgba);
+                            let expected_color = color::Lcha::from(expected_rgba);
+                            if      is_selected     { selected_color }
+                            else if is_disabled     { disabled_color }
                             else if is_expected_arg { expected_color }
```
wdanilo: this is again a very complex function which makes the FRP network hard to read. I'm OK with keeping small chunks of code in FRP network definition, but all bigger functions should be refactored to the model and named correctly. Its hard to understand now what it really does. If it was refactored and named nicely, the code would e much easier to follow.

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -665,29 +688,65 @@ impl Area {
...
+                                    theme::code::syntax::profiling::HERE.path(),
+                                _ =>
+                                    theme::code::syntax::HERE.path(),
+                            };
+                            let selected_rgba  = styles.get_color(theme::code::types::selected);
+                            let disabled_rgba  = styles.get_color(path.sub("disabled"));
+                            let expected_rgba  = styles.get_color(path.sub("expected"));
+                            let selected_color = color::Lcha::from(selected_rgba);
+                            let disabled_color = color::Lcha::from(disabled_rgba);
+                            let expected_color = color::Lcha::from(expected_rgba);
```
wdanilo: Earlier, the theme editor was queried once for the colors, now it is queried every time a node is selected, disabled, edtro mode is changed, base color changed etc. This is a major performance regression. Please take extra care of performance in your code. Also, this solution does not take care of updating colors when the theme changes (after theme changes this will not get recomputed). Please use StyleWatchFRP for it here.

src/js/package-lock.json
```diff
@@ -19,18 +19,18 @@
             }
         },
         "@babel/helper-validator-identifier": {
-            "version": "7.12.11",
-            "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.12.11.tgz",
-            "integrity": "sha512-np/lG3uARFybkoHokJUmf1QfEvRVCPbmQeUQpKow5cQ3xWrV9i3rUHodKDJPQfTVX61qKi+UdYk8kik84n7XOw==",
+            "version": "7.14.0",
```
wdanilo: What hapenned in this PR that so many JS deps were changed? I don't see any relevant description of this change in the PR description.

src/rust/ensogl/lib/core/src/data/color/space/def.rs
```diff
@@ -520,6 +520,15 @@ impl Lcha {
     pub fn to_javascript_string(&self) -> String {
         Rgba::from(self).to_javascript_string()
     }
+
+    /// Convert the color to gray by setting chroma to zero.
```
wdanilo: The description is not correct. This function can convert the color to white or black as well, depending on lightness, so it is not always converting the color to gray.

src/rust/ensogl/lib/text/src/component/area.rs
```diff
@@ -103,9 +102,7 @@ pub mod selection {
             let alpha          = alpha_weight.mix(blinking_alpha,SELECTION_ALPHA);
             let shape          = Rect((1.px() * rect_width,1.px() * rect_height));
             let shape          = shape.corners_radius(SELECTION_CORNER_RADIUS.px());
-            let color          = style.get_color(theme::code::syntax::selection);
-            let color          = format!("srgba({},{},{},{})",color.red.glsl(),color.green.glsl()
-                ,color.blue.glsl(),alpha.glsl());
+            let color          = format!("srgba({}.x,{}.y,{}.z,{})",color_rgb,color_rgb,color_rgb,alpha.glsl());
```
wdanilo: This line is too long. Please fix all improper spacing and too long lines in this PR.

src/rust/ensogl/lib/text/src/component/area.rs
```diff
@@ -135,6 +132,7 @@ pub struct Selection {
     position       : DEPRECATED_Animation<Vector2>,
     width          : DEPRECATED_Animation<f32>,
     edit_mode      : Rc<Cell<bool>>,
+    set_color      : frp::Any<color::Rgb>,
```
wdanilo: We have a macro that allows us to define FRP inputs / outputs, which cares of any / stream / sampler conversions. We are using it everywhere but here. What is the reason this was not defined with the macro? IS this done on purpose? If so, please explain it in the docs.

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -20,7 +21,9 @@ use text::Text;
 use crate::Type;
 use crate::component::type_coloring;
 use crate::node::input::port;
+use crate::node::ExecutionStatus;
 use crate::node;
+use crate::Mode as EditorMode;
```
wdanilo: use qualified names rather than renaming. If such renaming is needed, its suggests me that this thing is put in badly named module.

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -665,29 +688,65 @@ impl Area {
...
-                base_color <- frp.tp.map(f!([styles](t) type_coloring::compute_for_code(t.as_ref(),&styles)));
+                base_color <- all_with3(&self.editor_mode,&self.set_execution_status,&frp.tp,
+                    f!([styles](&mode,&status,t) {
+                        match (mode,status) {
+                            (EditorMode::Normal,_) =>
+                                type_coloring::compute_for_code(t.as_ref(),&styles),
+                            (EditorMode::Profiling,ExecutionStatus::Running) =>
+                                color::Lcha::from(styles.get_color(theme::code::syntax::base)),
+                            (EditorMode::Profiling,ExecutionStatus::Finished {..}) =>
+                                color::Lcha::from(styles.get_color(theme::code::syntax::profiling::base)),
```
wdanilo: this reads colors from theme editor every time events are emitted which is slow. Always think of performance implications of such code, please. The correct solution is to use StyleWatchFRP instead and reading the themes once, not on every emitted event.

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -665,29 +688,65 @@ impl Area {
...
+                            else                    { base_color }
+                        }));
+                    editing_color <- all_with(&self.editor_mode,&self.set_execution_status,
+                        f!([styles](&mode,&status) {
+                            let path = match (mode,status) {
+                                (EditorMode::Profiling,ExecutionStatus::Finished {..}) =>
+                                    theme::code::syntax::profiling::base,
+                                _ =>
+                                    theme::code::syntax::base,
+                            };
```
wdanilo: this code can have a much nicer layout - please take care of the layout of the code, so its easy to read. For example:

```
use theme::code::syntax;
let is_profiled = mode.is_profiling() && status.is_finished();
let path        = if is_profiled { syntax::profiling::base } else { syntax::base };
```

Please check out other parts of the code in this PR and try to improve their readability as well.

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -665,29 +688,65 @@ impl Area {
...
+                            color::Lcha::from(styles.get_color(path))
```
wdanilo: this is slow (see other comments mentioning StyleWatchFRP)

src/rust/ide/view/graph-editor/src/component/node/input/area.rs
```diff
@@ -703,11 +762,22 @@ impl Area {
...
+                        (&port_tp,&frp.set_connected,&self.editor_mode,
+                        f!([styles](port_tp,(is_connected,edge_tp),editor_mode) {
+                            if *is_connected {
+                                match editor_mode {
+                                    EditorMode::Normal => {
+                                        let tp = port_tp.as_ref().or_else(||edge_tp.as_ref());
+                                        select_color(&styles,tp)
+                                    },
+                                    EditorMode::Profiling => {
+                                        styles.get_color(theme::code::types::any::selection).into()
```
wdanilo: this is slow (see other comments mentioning StyleWatchFRP)

src/rust/ide/view/graph-editor/src/component/node/profiling_indicator.rs
```diff
@@ -0,0 +1,137 @@
+use crate::prelude::*;
```
wdanilo: this whole file is missing docs. Please fix linter config for it.

src/rust/ide/view/graph-editor/src/component/node/profiling_indicator.rs
```diff
@@ -0,0 +1,137 @@
...
+
+            color.target_color <+ all_with3
+                (&frp.set_execution_status,&frp.set_min_global_duration,
+                &frp.set_max_global_duration,f!([styles](&status,&min,&max) {
+                    let theme         = ensogl_theme::graph_editor::node::profiling::HERE.path();
+                    let lightness     = styles.get_number(theme.sub("lightness"));
+                    let chroma        = styles.get_number(theme.sub("chroma"));
+                    let min_hue       = styles.get_number(theme.sub("min_hue"));
+                    let max_hue       = styles.get_number(theme.sub("max_hue"));
+                    let hue_delta     = max_hue - min_hue;
```
wdanilo: Several issues here:
1. this is slow (see other comments mentioning StyleWatchFRP)
2. By using the dynamic API (`theme.sub`) you are stepping out of typechecker here, so after refactoring, this might crash at runtime. Please do not use `theme.sub` until you really need.

src/rust/ide/view/graph-editor/src/component/node/profiling_indicator.rs
```diff
@@ -0,0 +1,137 @@
...
+                    let running_color = color::Lch::new(lightness,chroma,max_hue);
+                    match status {
+                        ExecutionStatus::Running             => running_color,
+                        ExecutionStatus::Finished {duration} => {
+                            let relative_duration = (duration - min) / (max - min);
+                            let hue               = min_hue + relative_duration * hue_delta;
+                            color::Lch::new(lightness,chroma,hue)
```
wdanilo: This code is repeated! You are computing profiling colors in two separate places which is very error prone and would e very hard to understand and change in the future. Please take care of always computing things in one place, rather than in multiple places. Right now if I change the equation here, the colors of "indicator" (whatever it is) will be different than color of the node background and debugging it will take a lot of time.

src/rust/ide/view/graph-editor/src/component/node/profiling_indicator.rs
```diff
@@ -0,0 +1,137 @@
...
+
+            eval label.width((&width) label.set_position_x(-width/2.0));
+
+
+            // === Content ===
+
+            label.set_content <+ frp.set_execution_status.map(|&status|
+                match status {
+                    ExecutionStatus::Running             => "".to_string(),
+                    ExecutionStatus::Finished {duration} => format!("{} ms", duration)
```
wdanilo: Please take care of formatting of long durations, like seconds or minutes. If that will be seconds, just format t like `4.3 s` etc.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -57,9 +61,12 @@ use ensogl::display::object::Id;
 use ensogl::display::shape::StyleWatch;
 use ensogl::display;
 use ensogl::gui::cursor;
+use ensogl_gui_components::toggle_button;
+use ensogl_gui_components::toggle_button::ToggleButton;
 use ensogl::prelude::*;
 use ensogl::system::web;
 use ensogl_theme as theme;
+use std::f32;
```
wdanilo: please use full qualified name instead of importing it - it is less confusing then when reading the code.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -101,6 +108,26 @@ fn traffic_lights_gap_width() -> f32 {
...
 
 
+// ============
+// === Mode ===
+// ============
+
+/// Represents the current global mode of the graph editor. In profiling mode, edges should not be
+/// colored and profiling information on nodes should become visible.
+#[derive(Debug,Copy,Clone,CloneRef,PartialEq,Eq)]
+pub enum Mode {
```
wdanilo: This name does not make sens in this module. It is `graph_editor::Mode` - what is a mode of graph editor ? When reading the name, it should be clear that this is about display mode.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -966,6 +1002,87 @@ impl Edges {
...
 
 
+// =========================
+// === ExecutionStatuses ===
+// =========================
+
+/// [`ExecutionsStatuses`] can be used to collect the execution statuses of all nodes in a graph. It
+/// exposes their minimum and maximum running times through its FRP endpoints. The structure needs
+/// to be updated whenever a node is added or deleted or changes its execution status.
+mod execution_statuses {
```
wdanilo: Lets not use such nested modules. If you need it, probably better idea is to extract it to a file. The only exception are shapes definitions, but in all other cases, this is much harder to find. If I see in code `use foo::bar::bax` and I don't see `bax` file in folder `bar`, its really hard to grasp whats going on. Also, this just makes the code longer in this file, which doesnt make us benefit out f it.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -966,6 +1002,87 @@ impl Edges {
...
+    impl Deref for ExecutionStatuses {
+        type Target = Frp;
+
+        fn deref(&self) -> &Self::Target {
+            &self.frp
+        }
+    }
+
+    impl ExecutionStatuses {
+        pub fn new() -> Self {
```
wdanilo: missing odcs here and in many places in this PR.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1282,6 +1413,8 @@ pub struct GraphEditorModel {
     visualisations     : Visualisations,
     frp                : FrpEndpoints,
     navigator          : Navigator,
+    execution_statuses : ExecutionStatuses,
+    profiling_button   : ToggleButton<profiling_icon::DynamicShape>,
```
wdanilo: Graph editor should not know about such details as "toggle button" and some icon - this should be refactored as a separate type `ProfilingButton` - just like `component::Breadcrumb`. It should be defined in a separate file and just imported here.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1322,6 +1457,22 @@ impl GraphEditorModel {
...
+
+        let styles = StyleWatch::new(&self.scene().style_sheet);
+        let color_scheme = toggle_button::ColorScheme {
+            non_toggled     : Some(styles.get_color(theme::graph_editor::profiling_button::non_toggled).into()),
+            toggled         : Some(styles.get_color(theme::graph_editor::profiling_button::toggled).into()),
+            hovered         : Some(styles.get_color(theme::graph_editor::profiling_button::hovered).into()),
+            toggled_hovered : Some(styles.get_color(theme::graph_editor::profiling_button::toggled_hovered).into()),
+            ..default()
+        };
+        self.profiling_button.set_color_scheme(&color_scheme);
```
wdanilo: The initialization of the profiling button should be refactored to a module containing it to prevent this file from growing and to partition resposnibilities.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1844,13 +1997,19 @@ impl GraphEditorModel {
...
         // FIXME : StyleWatch is unsuitable here, as it was designed as an internal tool for shape system (#795)
         let styles    = StyleWatch::new(&self.scene().style_sheet);
-        let edge_type = self.edge_hover_type()
-            .or_else(|| self.edge_target_type(edge_id))
-            .or_else(|| self.edge_source_type(edge_id));
-        let opt_color = edge_type.map(|t|type_coloring::compute(&t,&styles));
-        opt_color.unwrap_or_else(||
-            color::Lcha::from(styles.get_color(theme::code::types::any::selection))
-        )
+        let neutral_color = color::Lcha::from(styles.get_color(theme::code::types::any::selection));
```
wdanilo: This is slow - see other comments about StyleWatchFRP to learn more why

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -3090,6 +3267,57 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
         eval quick_visualization_preview((value) model.nodes.set_quick_preview(*value));
     }
 
+
+
+    // =============
+    // === Modes ===
```
wdanilo: Please add info what are these "modes" here. Are these graph "display modes" ? I mean, this should be understandable while browsing the codebase and graph editor "mode" is not descriptive enough IMO.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -3090,6 +3267,57 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+    // =============
+
+    let profiling_mode_transition = Animation::new(network);
+    frp::extend! { network
+        out.source.mode <+ frp.toggle_profiling_mode.map2(&frp.mode,|_,&mode| {
+            match mode {
+                Mode::Profiling => Mode::Normal,
+                _               => Mode::Profiling,
```
wdanilo: please make FRP functions short. Refactor this as method on `Mode`. Please apply that to all lambdas in FRP network definitions that you can apply to (just make them as short and clear as possible)

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -3090,6 +3267,57 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+            }
+        });
+
+        eval out.mode([model](_) {
+            for edge_id in model.edges.keys() {
+                model.refresh_edge_color(edge_id);
+            }
```
wdanilo: As previously-  refattor it to a model function, like `refresh_all_edges_colors` - the FRP network def should be as short as possible so its easy to follow the FRP definition

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -3090,6 +3267,57 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+    let profiling_mode_transition = Animation::new(network);
+    frp::extend! { network
+        out.source.mode <+ frp.toggle_profiling_mode.map2(&frp.mode,|_,&mode| {
+            match mode {
+                Mode::Profiling => Mode::Normal,
+                _               => Mode::Profiling,
+            }
+        });
+
+        eval out.mode([model](_) {
```
wdanilo: in such cases use `eval_` and skip `[model]` part (after applying the second comment)

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -3090,6 +3267,57 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+
+    // === Profiling Button ===
+
+    frp::extend! { network
+        out.source.mode <+ model.profiling_button.state.map(|&toggled| {
+            if toggled { Mode::Profiling } else { Mode::Normal }
+        });
+
+        model.profiling_button.set_state <+ out.mode.on_change().map(|&mode| {
+            matches!(mode,Mode::Profiling)
```
wdanilo: In such cases add methods like `mode.is_profiling()` - it is easier to use and very useful in many places.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -3090,6 +3267,57 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
+        });
+
+        eval scene.frp.camera_changed([model,scene](_) {
+            let screen = scene.camera().screen();
+            model.profiling_button.set_position_x(screen.width/2.0 - 16.0);
+            model.profiling_button.set_position_y(screen.height/2.0 - 16.0);
+        });
```
wdanilo: Is this how we set up position of breadcrumbs? I'm asking cause it is rather a bad idea to do. Thsi should just be attached to a camera that does not move - scene has many layers that do not move.

src/rust/ide/view/graph-editor/src/profiling_icon.rs
```diff
@@ -0,0 +1,56 @@
+use crate::prelude::*;
```
wdanilo: 1. No docs
2. No ections
3. Too long lines
4. No newline at the end of the file
5. misaligned lines

src/rust/lib/frp/src/nodes.rs
```diff
@@ -494,10 +494,19 @@ impl Network {
     /// Specialized version `all_with`.
     pub fn all_with5<T1,T2,T3,T4,T5,F,T>
     (&self, label:Label, t1:&T1, t2:&T2, t3:&T3, t4:&T4, t5:&T5, f:F) -> Stream<T>
-    where T1:EventOutput, T2:EventOutput, T3:EventOutput, T4:EventOutput, T5:EventOutput,
-          T:Data, F:'static+Fn(&Output<T1>,&Output<T2>,&Output<T3>,&Output<T4>,&Output<T5>)->T {
+        where T1:EventOutput, T2:EventOutput, T3:EventOutput, T4:EventOutput, T5:EventOutput,
+              T:Data, F:'static+Fn(&Output<T1>,&Output<T2>,&Output<T3>,&Output<T4>,&Output<T5>)->T {
```
wdanilo: earlier it was correct (here and in other places)

wdanilo (changes requested):

wdanilo:
Also, few additional things:
1. On the screenshot you have provided it doesn't seem that the background has color animation - it should have it.
2. The visualizations should not get grayscale - sorry that I didn't mention it explicitly. It's my fault as I've described that "everything" in graph editor should get grayscaled, but I forget that my sentence might be applied to visualizations as well.
3. The output ports should get grayscale as well - otherwise, you will drag edges from colored ports, and edges will be grayscale, which would look strange - please post also a screencast showing as many interactions as possible (like dragging edges in the profiling mode) - this way we will discover more what looks good and bad.
4. We are definitely over-using the space above / below nodes. After thinking for a while, I'd put the time info on the left side of the node and hide it if the node name is shown. The reason for that is that node names are not really useful in profiling mode, so this is a good place to actually place it. Also, when posting the screencast, please how this feature looks like with several nodes connected with each other n "standard layout" - when nodes are close to each other in vertical spacing (like after being created with node searcher).

src/js/package-lock.json
s9ferech: `node run watch` just updated them. Likely because I switched to a new node version.

src/rust/ensogl/lib/components/src/toggle_button.rs
s9ferech: It does not work the same, because `frp.source.state` is also set through other connections.

src/rust/ensogl/lib/core/src/data/color/space/def.rs
s9ferech: You can take `self` by value, but the `..self` does not work that easily, because `Lcha` is a nested struct type with three levels.

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: It changes the color of the existing background. The `profiling_indicator` provides a label with the running time.

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: The name `display` is already taken. Do you have another name that you would be happy with?

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: Those lines hide or show the existing VCS indicators in profiling mode. They have nothing to do with the profiling indicator. You seem to be right that there is no hide or show animation for them. Should I add one?

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: We are currently having a call about it and are thinking about the name `view::Mode`.

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: We discussed to skip this now.

iamrecursion:
A quick question. I assume that you're calculating a weighted average of the profiling updates from the engine? Otherwise it's very possible that the values could be quite noisy.

iamrecursion:
Another question: how do you determine the colours?

s9ferech:
> A quick question. I assume that you're calculating a weighted average of the profiling updates from the engine? Otherwise it's very possible that the values could be quite noisy.

I have updated the description to be a bit clearer about this. It is not part of this PR to retrieve the information from the engine. That was part of another issue. So, the question does not concern this PR.

s9ferech:
> Another question: how do you determine the colours?

From the running times that we get. The fastest node is green, the slowest is red and everything else is interpolated linearly in-between.

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: Did you have any new thoughts on this? @wdanilo

iamrecursion:
We need to think about colours due to user perception. Think about schemes such as [`parula`](https://uk.mathworks.com/help/matlab/ref/colormap.html) used in Matlab.

s9ferech:
It could make sense to use a color scheme that only highlights the slow nodes and displays the fast nodes in greyscale.

I would also suggest that we use green (or whatever the end of the scale is) only for a running time of zero. If we have two nodes, one with 10 s and one with 11 s, then it seems wrong to color the 10 s node in green. It should not be much different to the 11 s node.

src/rust/ide/view/graph-editor/src/component/node.rs
s9ferech: If we plan to only use this only for the profiling mode then it could make sense to move it to `profiling::Mode::On` and `profiling::Mode::Off`. Otherwise, I would suggest `display_mode::DisplayMode::Normal` and `display_mode::DisplayMode::Profiling`. Or just leaving it where it is and renaming `Mode` to `DisplayMode`.

wdanilo: I'd rather give us possibility to implement easily other modes without major refacotirng so I'd not go with `profiling::Mode::On`. On the other hand, the naming `display_mode::DisplayMode` is what we try not to do, as it is repeating itself and preventing nice qualified usage. I think that `crate::View::Mode::Normal` is much better (as we discussed on the call).

s9ferech:
Here are two new recordings of the profiling mode.
Profiling mode in a normal program
Profiling mode in the interface demo scene.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2972,13 +3044,19 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
-    eval out.on_edge_target_unset (((id,_)) model.refresh_edge_color(*id));
+    let neutral_color = styles_frp.get_color(theme::code::types::any::selection);
+    _eval <- all_with(&out.on_edge_source_set,&neutral_color,f!(((id,_),neutral_color)
+            model.refresh_edge_color(*id,neutral_color.into())));
+    _eval <- all_with(&out.on_edge_target_set,&neutral_color,f!(((id,_),neutral_color)
+            model.refresh_edge_color(*id,neutral_color.into())));
+    _eval <- all_with(&out.on_edge_source_unset,&neutral_color,f!(((id,_),neutral_color)
+            model.refresh_edge_color(*id,neutral_color.into())));
+    _eval <- all_with(&out.on_edge_target_unset,&neutral_color,f!(((id,_),neutral_color)
+            model.refresh_edge_color(*id,neutral_color.into())));
```
wdanilo: There is a major issue with this code, but IMO easy to fix. Think what would happen when theme changes. The `neutral_color` will emit new event and thus all of these FRP nodes will fire (not only those, in other places as well), which means, that `model.refresh_edge_color` will be fired MANY times then. This is bad. After changing theme, we should refresh colors of edges only once per edge.

wdanilo (commented):
Felix, amazing job with the code! ❤️

I have almost no comments - the only one is a major issue but super simple to be fixed. Let's talk tomorrow about it. If it will be fixed, there is an acceptance from my side! Please ask someone to test it before the merge as I'm unable currently. It is important to very carefully test all edges, connections, color changes, and animations and see if there are any regressions regarding their behavior / animations.

s9ferech:
@MichaelMauderer, would you have time to test this PR to make sure that it works in all edge cases?

wdanilo:
Felix, please ping me (and mark me as re-requested for the review) when this will be ready for final review.

src/rust/ide/view/graph-editor/src/component/node/profiling.rs
```diff
@@ -0,0 +1,251 @@
...
+// ============
+// === Frp  ===
+// ============
+
+ensogl::define_endpoints! {
+    Input {
+        set_status              (Status),
+        set_min_global_duration (f32),
+        set_max_global_duration (f32),
+        set_view_mode         (view::Mode),
```
wdanilo: spacing

src/rust/ide/view/graph-editor/src/component/node/profiling.rs
```diff
@@ -0,0 +1,251 @@
...
+/// time of any node on the stage through `set_min_global_duration` and `set_max_global_duration`.
+/// The color of the label will reflect the status and be determined by [`Status::display_color`].
+/// The necessary theme will be taken from the application's style sheet. The origin of the label,
+/// as a `display::Object` should be placed on the node's center.
+#[derive(Clone,CloneRef,Debug)]
+pub struct ProfilingLabel {
+    root    : display::object::Instance,
+    label   : text::Area,
+    frp     : Frp,
+    styles  : StyleWatchFrp,
```
wdanilo: spacing

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1172,7 +1196,23 @@ impl GraphEditorModelWithNetwork {
             output.source.visualization_enabled <+
                 vis_enabled.map2(&metadata,move |_,metadata| (node_id,metadata.clone()));
             output.source.visualization_disabled <+ vis_disabled.constant(node_id);
+
+
+            // === Mode ===
```
wdanilo: change the title to View Mode pls

src/rust/ide/view/graph-editor/src/profiling.rs
```diff
@@ -0,0 +1,89 @@
+//! [`ProfilingStatuses`] can be used to collect the profiling statuses of all nodes in a graph. It
+//! exposes their minimum and maximum running times through its FRP endpoints. The structure needs
+//! to be updated whenever a node is added or deleted or changes its profiling status.
+
+use crate::prelude::*;
```
wdanilo, line 5: Add sections to the file. Each file should contain at least one section.

src/rust/ide/view/graph-editor/src/view.rs
```diff
@@ -0,0 +1,43 @@
+//! Provides the `Mode` enum to represent the graph editors view mode.
+
+use crate::prelude::*;
```
wdanilo, line 3: Add sections to the file. Each file should contain at least one section.

wdanilo (approved):

wdanilo:
I don't see this checked:
image

Please make sure that someone tested that before the merge :)

farmaazon (changes requested):
I spotted one regression during testing: I cannot change the project name.

s9ferech:
@farmaazon 
> I spotted one regression during testing: I cannot change the project name.

I cannot reproduce this problem. Can you describe the exact steps?

farmaazon:
@s9ferech 

> @farmaazon
> 
> > I spotted one regression during testing: I cannot change the project name.
> 
> I cannot reproduce this problem. Can you describe the exact steps?

I first, did all the steps of scenario https://docs.google.com/spreadsheets/u/1/d/1RatJDM_f9_3bvYhl3Bpq2d8SyKgtVdrV1RkGxPU17c8/edit#gid=0 to the point where the project name should be changed. I was unable to edit it, neither by ctrl+click nor by double click.

<blockquote><div>Google Docs</div><div><strong><a href="https://docs.google.com/spreadsheets/u/1/d/1RatJDM_f9_3bvYhl3Bpq2d8SyKgtVdrV1RkGxPU17c8/edit?usp=embed_facebook">Scenarios</a></strong></div><div>Basic

Action to be done,Result,Release,Status,Why does not work?
DONE: 24/28 (86%)
Double click Enso executable.
,Unnamed project opens with two connected nodes `0 . up_to 100 . to_vector . map noise` and second `sort` with open scatterplot visualizations. Connections should have colors. Open vi...</div></blockquote>

s9ferech:
> I first, did all the steps of scenario https://docs.google.com/spreadsheets/u/1/d/1RatJDM_f9_3bvYhl3Bpq2d8SyKgtVdrV1RkGxPU17c8/edit#gid=0 to the point where the project name should be changed. I was unable to edit it, neither by ctrl+click nor by double click.

I am still not able to reproduce this. It might be helpful if you find a minimal sequence of actions that causes this.

farmaazon:
> > I first, did all the steps of scenario https://docs.google.com/spreadsheets/u/1/d/1RatJDM_f9_3bvYhl3Bpq2d8SyKgtVdrV1RkGxPU17c8/edit#gid=0 to the point where the project name should be changed. I was unable to edit it, neither by ctrl+click nor by double click.
> 
> I am still not able to reproduce this. It might be helpful if you find a minimal sequence of actions that causes this.

Well, actually just opening project and trying to rename it shows the problem.

If it is not reproducible on your side, then I will look into it tomorrow.

<blockquote><div>Google Docs</div><div><strong><a href="https://docs.google.com/spreadsheets/u/1/d/1RatJDM_f9_3bvYhl3Bpq2d8SyKgtVdrV1RkGxPU17c8/edit?usp=embed_facebook">Scenarios</a></strong></div><div>Basic

Action to be done,Result,Release,Status,Why does not work?
DONE: 24/28 (86%)
Double click Enso executable.
,Unnamed project opens with two connected nodes `0 . up_to 100 . to_vector . map noise` and second `sort` with open scatterplot visualizations. Connections should have colors. Open vi...</div></blockquote>

mwu-tow:
@s9ferech @farmaazon I have the same issue on develop, however only if I build with gnu-flavored rustc toolchain. Also, it seems to happen only in `dist` command.

s9ferech:
That is very interesting. For a long time we had a problem that the error stripes appeared in front of nodes on some systems, but not on others, right? Could that be related? I think the breadcrumbs contain some invisible shapes as mouse targets. Maybe the objects are in the wrong order when you run it and the mouse events do not arrive at the right one. Does that sound possible?

mwu-tow:
@s9ferech Interesting theory, might be possible. TBH recently I haven't seen the error stripes issue but if such ordering issues can appear and disappear on its own, then there might be actually something wrong in that area. Relying on some unspecified behavior or something else that makes objects in different order depending on build environment.

farmaazon (approved):
The bug will be fixed once #1620 will be closed.

PR #1556, opened 2021-05-07
Deselect visualization when its node gets removed

s9ferech:
### Pull Request Description

This fixes issue #1352. The graph editor will register visualizations as deselected when their node is removed.

### Checklist


src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1405,6 +1405,7 @@ impl GraphEditorModel {
         let node_id = node_id.into();
         self.nodes.remove(&node_id);
         self.nodes.selected.remove_item(&node_id);
+        self.frp.source.on_visualization_select.emit(Switch::Off(node_id));
```
MichaelMauderer: I'm not 100% sure abut this, but it might be better to do this before the node is actually removed. Otherwise, the `node_id` might no longer be viable.

s9ferech: I changed it, to be sure.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -1403,6 +1403,7 @@ impl GraphEditorModel {
     /// implementation.
     fn remove_node(&self, node_id:impl Into<NodeId>) {
         let node_id = node_id.into();
+        self.frp.source.on_visualization_select.emit(Switch::Off(node_id));
```
wdanilo: Emiting frp events from model functions is rather a bad pattern, as it is really hard to understand where the events come from. Instead, check where this model function is called in the FRP network and handle it directly there.

This should be handled next to the lines:

```
output.source.on_visualization_select <+ selected.constant(Switch::On(node_id));
output.source.on_visualization_select <+ deselected.constant(Switch::Off(node_id));
```

so it would be easy to track when visualization is deselectected when reading the code (all deselecting will be coded in one place then)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 #### Visual Environment
 
 - [Delete key will delete selected nodes][1538].
+- [It is possible to move around after deleting a selected visualization][1556].
```
wdanilo: what does it mean "deleting a [...] visualization"? It is possible to delete a node with an attached visualization, but it is not possible to delete a visualization per see. Moreover, this should be described much more user friendly / in more explanatory way, like "There was a bug that after a node with a selected visualization was deleted, it was impossible to pan the stage. The error is fixed now."

wdanilo (changes requested):

src/rust/ide/view/graph-editor/src/lib.rs
s9ferech: I changed this, but it cannot be handled next to the lines that you suggested, because we need only one connection per graph editor, but those lines are run for every node.

wdanilo (approved):

PR #1578, opened 2021-05-21
Fix Cursor Position in Cloud and Bug in Creating New Projects

farmaazon:
### Checklist

- [ ] ~~The `CHANGELOG.md` was updated with the changes introduced in this PR.~~

### DIFF
--- a/src/rust/ensogl/lib/core/src/display/scene.rs
+++ b/src/rust/ensogl/lib/core/src/display/scene.rs
@@ -320,7 +320,11 @@ pub struct Mouse {
 
 impl Mouse {
     pub fn new
-    (scene_frp:&Frp, variables:&UniformScope, current_js_event:&CurrentJsEvent, logger:Logger)
+    ( scene_frp        : &Frp
+    , root             : &web::dom::WithKnownShape<web::HtmlDivElement>
+    , variables        : &UniformScope
+    , current_js_event : &CurrentJsEvent
+    , logger           : Logger)
     -> Self {
         let scene_frp       = scene_frp.clone_ref();
         let target          = PointerTarget::default();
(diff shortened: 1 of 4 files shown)


src/rust/ensogl/lib/core/src/display/scene.rs
```diff
@@ -328,7 +328,7 @@ impl Mouse {
         let position        = variables.add_or_panic("mouse_position",Vector2::new(0,0));
         let hover_ids       = variables.add_or_panic("mouse_hover_ids",target.to_internal(&logger));
         let target          = Rc::new(Cell::new(target));
-        let body            = web::dom::WithKnownShape::new(&web::document().body().unwrap());
```
wdanilo, line 331: The name of the dom to embed EnsoGL is passed as an argument to scene or world constructor (cannot check now I’m on phone). This will not always be named “root” and hard coding the name I’m several places is rather a bad design. Please refactor it out so the name would not be hardcoded and will be passed from Scene instead.

src/rust/ide/src/controller/ide/desktop.rs
```diff
@@ -97,7 +97,7 @@ impl ManagingProjectAPI for Handle {
             let with_suffix           = (1..).map(|i| format!("{}_{}", UNNAMED_PROJECT_NAME, i));
             let mut candidates        = std::iter::once(without_suffix).chain(with_suffix);
             // The iterator have no end, so we can safely unwrap.
-            let name    = candidates.find(|c| names.contains(c)).unwrap();
+            let name    = candidates.find(|c| !names.contains(c)).unwrap();
```
wdanilo, line 100: This change doesn’t seem to be related to cursor, is it? If not, what it is about?

wdanilo (changes requested):

src/rust/ide/src/controller/ide/desktop.rs
farmaazon: It's related to the second part of PR titile: `Bug in Creating New Projects`

wdanilo: Yes, but the bug is not described anywhere. I'd like to understand what it was and how this change fixes it. Any comment about it would be helpful. Also, is this something that should be added to the changelog?

wdanilo (approved):

src/rust/ide/src/controller/ide/desktop.rs
farmaazon: Well, the full history: in this place we look for the first name which is not already taken by another project from list [`Unnamed`, `Unnamed_1`, `Unnamed_2`, ...]. First, I put here `skip_while` + `next`, but, after merging develop with compiler version, the linter asked me for replacing it with `find`, _but_ I missed the negation - so instead looking for first free name, the code looked for first taken name.

PR #1614, opened 2021-06-09
Scrollbar and ScrollArea

MichaelMauderer:
### Pull Request Description
Add a re-usable scrollbar component for internal use. Can be tested in the `slider` demo scene at `http://localhost:8080/?entry=slider`

[ci no changelog needed]

https://user-images.githubusercontent.com/1428930/121358730-1f440a80-c933-11eb-8f4e-4fa792a03f35.mp4

### Checklist


s9ferech:
I tried the scroll bars and there are some things that are non-standard and could be improved in my opinion:

1. When I move the cursor beyond the end of the scroll bar and return, it does not stick to the same position on the handle anymore.
scroll bar
2. The handle moves also when I move the mouse orthogonally to the bar.
orthogonal movement
3. Clicking on the bar, outside the handle, should move the handle by a bit less than its length, which will lead to scrolling a bit less than the viewport size. Alternatively, it could jump directly to the cursor position.
4. The handle should be highlighted on hover to give a hint that you can interact with it.

s9ferech:
I ran our test scenarios on this and everything works, except that I could not edit the project name. The same issue was observed on an [unrelated branch](https://github.com/enso-org/ide/pull/1546#pullrequestreview-678604540). It is likely already present on develop.

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,152 @@
...
+        let scene                = app.display.scene();
+        let mouse                = &scene.mouse.frp;
+        let track_position_lower = Animation::new(&network);
+        let track_position_upper = Animation::new(&network);
+
+        let base_frp = selector::Frp::new(model, style, network, frp.resize.clone().into(), mouse);
+
+        model.use_track_handles(false);
+        model.show_left_overflow(false);
+        model.show_right_overflow(false);let style_track_color = style.get_color(theme::component::slider::track::color);
```
wdanilo: some lines are to long - please check all lines in this PR

wdanilo: Also, here we have 2 liens glued together

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,152 @@
...
+            normalised_track_center <- normalised_track_bounds.map(|bounds| bounds.center());
+
+            // Slider Updates
+            update_slider <- all(&normalised_track_bounds,&frp.resize);
+            eval update_slider(((value,size)) model.set_background_range(*value,*size));
+
+            // Mouse IO - Clicking
+            click_delta <- base_frp.background_click.map2(&normalised_track_center,
+                f!([model](click_value,current_value) {
+                    let rotation  =  Rotation2::new(-model.rotation().z);
```
wdanilo: spacing

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,152 @@
...
+                    let rotated   = rotation * click_value;
+                    let direction = if rotated.x > *current_value { 1.0 } else { -1.0 };
+                    const SCALE: f32 = 0.05;
```
wdanilo: what is this SCALE? It looks magical and has no docs

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,152 @@
...
+
+            new_value_lower <- new_value_absolute.map(|b| b.start);
+            new_value_upper <- new_value_absolute.map(|b| b.end);
+
+            track_position_lower.target <+ new_value_lower.gate(&is_in_bounds);
+            track_position_upper.target <+ new_value_upper.gate(&is_in_bounds);
+
+            track_position_lower.skip <+ should_animate.on_false();
+            track_position_upper.skip <+ should_animate.on_false();
+
```
wdanilo: This looks like pretty complex FRP - it seems it includes dragging etc. Why such things as dragging are not completely handled by slider instead? I was thinking that this component would just wrap slider and just set a few FRP values to it, without its own complex FRP utility. Also, there are no docs describing what really these functions do, which makes it much harder to understand - it contains some dragging, some mouse IO, while all of these things should be handled by the slider component.

wdanilo:
Also, from the visual point of view, our sliders should not have background, should not have shadows and should have rounded endings:

image


Also, the points raised by Felix are good too!

wdanilo:
Also, on the Felix's gif, there is a visual artifact visible (this is a max scroll):

image

src/rust/ensogl/lib/components/src/scrollbar.rs
MichaelMauderer: The actual drag implementation and underlying logic is shared. But this is the logic how to interpret it, which is different in each component: (1) single number selectors have clicking at the target and dragging only on the background (2) range slider need to act on dragging either the track piece or the edges of the track piece (3) the scrollbar needs to have dragging of the track piece and "jumping" when clicking on some other part of the background. 
That is essentially what is encoded here. Maybe some things could be refactored and abstracted, but not too much.

wdanilo: Why cant these things be just coded in (moved to) the slider implementation so you'd be able to choose which mode you wants with FRP settings? Just like choosing colors, we could tell "I want this slider to jump to the place when clicking the background". This should be part of the sldier configuration rather than extracted here imo. Can we move it there, please?

s9ferech:
I made some heavy design changes to the scrollbars and added scroll areas to the PR because they are closely related:
scroll area in action
The recording demonstrates hovering over the bars, dragging them, clicking on them and using the mouse wheel. At the current state, content outside the scroll area is not clipped yet. That feature can be added during the next days, as soon as we have general clipping abilities.

src/rust/ensogl/example/src/slider.rs
```diff
@@ -13,7 +13,6 @@ use ensogl_text_msdf_sys::run_once_initialized;
 use ensogl_theme as theme;
 
 
-
```
farmaazon: restore this

src/rust/ensogl/lib/components/src/scroll_area.rs
```diff
@@ -0,0 +1,161 @@
+//! This module provides the [`ScrollArea`] component. It displays two scrollbars, for horizontal
+//! and vertical scrolling. Content can be added to the `content` attribute. The content size has
+//! to be set through `set_content_height` and `set_content_width`. The component is anchored at
+//! the top left corner. All scroll coordinates describe the point of the `content` object at that
+//! corner.
```
farmaazon: I'd prefer to have this description in the struct docs, and here refer to the struct. Usually when browsing documentation I rather look at the structures' descriptions, not modules'.

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,267 @@
...
+/// Amount the scrollbar moves on a single click, relative to the viewport size.
+const CLICK_JUMP_PERCENTAGE : f32 = 0.80;
+/// Width of the scrollbar in px.
+pub const WIDTH             : f32 = 11.0;
+/// The amount of padding on each side inside the scrollbar.
+const PADDING               : f32 = 2.0;
+/// The thumb will be displayed with at least this size to make it more visible and dragging easier.
+const MIN_THUMB_SIZE        : f32 = 12.0;
+/// After an animation, the thumb will be visible for this time, before it hides again.
+const HIDE_DELAY            : f32 = 1000.0;
```
farmaazon, line 39: I would put those in styles.

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,267 @@
...
+        frp::extend! { network
+
+            // Scrolling and Jumping
+
+            frp.scroll_to <+ frp.scroll_by.map2(&thumb_position.target,|delta,pos| *pos+*delta);
+
+            // We will use this to reveal the scrollbar on scrolling. It has to be defined before
+            // the following nodes that update `thumb_position.target`.
+            active <- frp.scroll_to.map2(&thumb_position.target,|&new:&f32,&old:&f32| {
+                let error_margin = 0.1;
```
farmaazon: make it a constant on the top of the file.

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,267 @@
...
+        let thumb_color       = color::Animation::new(network);
+        let activity_cool_off = DelayedAnimation::new(network);
+        activity_cool_off.frp.set_delay(HIDE_DELAY);
+        activity_cool_off.frp.set_duration(0.0);
+
+        frp::extend! { network
+            resize <- frp.set_length.map(|&length| Vector2::new(length,WIDTH));
+        }
+
+        let base_frp = selector::Frp::new(model, style, network, resize.clone(), mouse);
```
farmaazon, line 111: What is the difference between base_frp and frp?

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,267 @@
...
+            // active). This will instantly reveal the scrollbar and hide it after the delay has
+            // passed.
+            activity_cool_off.frp.reset <+ active;
+            activity_cool_off.frp.start <+ active;
+
+            recently_active <- bool(&activity_cool_off.frp.on_end,&activity_cool_off.frp.on_reset);
+
+            // The signed distance between the cursor and the edge of the scrollbar. If the cursor
+            // is further left or right than the ends of the scrollbar then we count the distance as
+            // infinite.
```
farmaazon: Please explain _why_ we do the infinite thingy.

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,267 @@
...
+                        0.0
+                    } else if active || dragging {
+                        1.0
+                    } else {
+                        if dist <= 0.0 {
+                            1.0
+                        } else {
+                            (0.7 - dist / 20.0).max(0.0).min(1.0)
+                        }
+                    }
```
farmaazon: it deserves its own function. And maybe a bit of explanation of the 0.7 - dist / 20.0 magic.

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,267 @@
...
+
+            // === Dragging ===
+
+            drag_started <- base_frp.is_dragging_track.on_change().on_true().constant(());
+            drag_offset  <- all4(&mouse_position,&inner_length,&frp.set_max,&frp.thumb_position)
+                .sample(&drag_started)
+                .map(|(mouse_px,length_px,max,thumb_pos)| {
+                    let thumb_position_px = thumb_pos / max * length_px;
+                    mouse_px.x - thumb_position_px
+                });
```
farmaazon: according to our coding guidelines, the chain should be split to many `<-` instructions.

src/rust/ensogl/lib/components/src/selector/model.rs
```diff
@@ -64,6 +64,14 @@ pub struct Model {
...
     /// a group.
     pub root               : display::object::Instance,
+
+    background_color                  : Rc<RefCell<color::Rgba>>,
+    track_color                       : Rc<RefCell<color::Rgba>>,
+    background_left_corner_roundness  : Rc<Cell<bool>>,
+    background_right_corner_roundness : Rc<Cell<bool>>,
+    padding                           : Rc<Cell<f32>>,
+
+    pub app : Application,
```
farmaazon, line 74: Usually such parameters are just the FRP outputs of components. For example: the ListView does not keep its size and scroll position in the model, but as the frp outputs anyone could read.

src/rust/ensogl/lib/components/src/selector/shape.rs
```diff
@@ -50,10 +51,10 @@ pub mod background {
     use super::*;
 
     ensogl_core::define_shape_system! {
-        (style:Style,corner_left:f32,corner_right:f32) {
+        (style:Style,corner_left:f32,corner_right:f32,color:Vector4,show_shadow:f32) {
             let background = Background::new(&corner_left,&corner_right,style);
-            let color      = style.get_color(theme::component::slider::background);
-            let shadow     = shadow::from_shape(background.shape.clone(),style);
+            let shadow     = shadow::from_shape_with_alpha(background.shape.clone(),
+                             &show_shadow,style);
```
farmaazon:
```suggestion
            let shadow     = shadow::from_shape_with_alpha(background.shape.clone(),
                &show_shadow,style);
```

farmaazon (changes requested):

src/rust/ensogl/lib/components/src/scroll_area.rs
s9ferech: I agree and will change it.

src/rust/ensogl/lib/components/src/scrollbar.rs
s9ferech: I would rather not put this in styles before we have a better implementation of reactive networks. You have seen that it is really messy to use styles currently. Would you accept that?

src/rust/ensogl/lib/components/src/scrollbar.rs
s9ferech: The way how all the slider components (including the scrollbars in this PR) are implemented, they share some common logic. And that is behind the `base_frp`. The `frp` then adds some component-specific behavior on top. Does that make it clear?

src/rust/ensogl/lib/components/src/selector/model.rs
s9ferech: So, you would change it to be more consistent with those components?

src/rust/ensogl/lib/components/src/scrollbar.rs
farmaazon: yes

src/rust/ensogl/lib/components/src/scrollbar.rs
farmaazon: Then add a TODO which refer to the GH issue of better implementation of reactive networks.

src/rust/ensogl/lib/components/src/selector/model.rs
farmaazon: Yes, if there are no specific reason for keeping them as they are. But if there is, please describe that reason in a comment.

s9ferech: It is not possible the way how things are currently structured. There is not FRP network available. I would have to refactor things. I don't think it's worth it.

src/rust/ensogl/lib/components/src/scrollbar.rs
s9ferech: Is there already such an issue? I can't find it.

farmaazon: If not, let's create it.

src/rust/ensogl/lib/components/src/scrollbar.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+
+
+// ===========================
+// === Scrollbar Component ===
+// ===========================
+
+/// Scrollbar component that can be used to implement scrollable components. See module description
+/// for details.
```
farmaazon: Again, I would put here an extended description, and refer to it from the module docs.

farmaazon (approved):

s9ferech:
Here is a recording that shows the interactions in a slightly better quality.
https://user-images.githubusercontent.com/8134934/124300195-318e1e80-db56-11eb-9fea-50f28d3790c0.mp4

<blockquote></blockquote>

wdanilo:
@s9ferech Looks absolutely stunning <3

wdanilo (approved):

PR #1619, opened 2021-06-09
Uploading files

farmaazon:
[ci no changelog needed]

### Pull Request Description

This PR covers the controller part of uploading files by drag'n'drop.

### Important Notes
To see it working, the project must use engine version 0.2.12-SNAPSHOT.2021-06-15 or greater.

This does not include the GUI part; the created node has no progress indicator yet, and its expression changes during upload are visible (including "uploading error"). However the minimal integration is there.

### Checklist

- [ ] ~~The `CHANGELOG.md` was updated with the changes introduced in this PR.~~
- [ ] ~~All code has been manually tested in the "debug/interface" scene.~~


src/rust/ide/lib/enso-protocol/src/binary/client.rs
```diff
@@ -13,11 +14,10 @@ use crate::binary::message::MessageFromServerOwned;
 use crate::binary::message::VisualisationContext;
 use crate::language_server::types::Path;
 
-use json_rpc::error::RpcError;
 use json_rpc::TransportEvent;
 use json_rpc::Transport;
 use mockall::automock;
-
+use crate::types::Sha3_224;
```
mwu-tow: misplaced import

src/rust/ide/lib/enso-protocol/src/binary/client.rs
```diff
@@ -280,6 +302,7 @@ mod tests {
         let mut mock_reply = MessageFromServer::new(FromServerPayloadOwned::Error {
             code : mock_error_code,
             message : mock_error_message.clone(),
+            data : None,
```
mwu-tow: vertical align

src/rust/ide/lib/enso-protocol/src/binary/serialization.rs
```diff
@@ -273,7 +409,8 @@ impl<'a> DeserializableUnionField<'a, OutboundMessage<'a>> for FromServerPayload
                 let payload = message.payload_as_error().unwrap();
                 Ok(FromServerPayload::Error {
                     code: payload.code(),
-                    message: payload.message().unwrap_or_default(),
+                    message: payload.message(),
+                    data: Option::<message::ErrorPayload>::deserialize(payload)?
                 })
```
mwu-tow, line 414: vertical align

src/rust/ide/lib/enso-protocol/src/binary/serialization.rs
```diff
@@ -291,9 +428,30 @@ impl<'a> DeserializableUnionField<'a, OutboundMessage<'a>> for FromServerPayload
...
+            OutboundPayload::READ_BYTES_REPLY => {
+                let payload = message.payload_as_read_bytes_reply().unwrap();
+                Ok(FromServerPayload::ReadBytesReply {
+                    checksum : message::EnsoDigest::deserialize(payload.checksum())?,
+                    bytes    : payload.bytes(),
+                })
+            }
+            OutboundPayload::CHECKSUM_BYTES_REPLY => {
+                let payload = message.payload_as_checksum_bytes_reply().unwrap();
+                let checksum = match payload.checksum() {
```
mwu-tow: align `=`

src/rust/ide/lib/enso-protocol/src/binary/serialization.rs
```diff
@@ -291,9 +428,30 @@ impl<'a> DeserializableUnionField<'a, OutboundMessage<'a>> for FromServerPayload
...
+                    Some(checksum) => message::EnsoDigest::deserialize(checksum)?,
+                    None           => default(), //TODO shall the checksum be required?
```
mwu-tow: unresolved TODO?

src/rust/ide/lib/enso-protocol/src/binary/serialization.rs
```diff
@@ -335,7 +517,8 @@ impl<'a> DeserializableUnionField<'a, OutboundMessage<'a>> for FromServerPayload
                 let payload = message.payload_as_error().unwrap();
                 Ok(FromServerPayloadOwned::Error {
                     code: payload.code(),
-                    message: payload.message().unwrap_or_default().to_string(),
+                    message: payload.message().to_string(),
+                    data: Option::<message::ErrorPayload>::deserialize(payload)?
```
mwu-tow: align

src/rust/ide/lib/enso-protocol/src/binary/serialization.rs
```diff
@@ -353,13 +536,46 @@ impl<'a> DeserializableUnionField<'a, OutboundMessage<'a>> for FromServerPayload
...
+            OutboundPayload::READ_BYTES_REPLY => {
+                let payload = message.payload_as_read_bytes_reply().unwrap();
+                Ok(FromServerPayloadOwned::ReadBytesReply {
+                    checksum : message::EnsoDigest::deserialize(payload.checksum())?,
+                    bytes    : payload.bytes().to_vec(),
+                })
+            }
+            OutboundPayload::CHECKSUM_BYTES_REPLY => {
+                let payload = message.payload_as_checksum_bytes_reply().unwrap();
+                let checksum = match payload.checksum() {
```
mwu-tow: align

src/rust/ide/lib/enso-protocol/src/language_server/response.rs
```diff
@@ -38,6 +38,13 @@ pub struct FileInfo {
     pub attributes: FileAttributes,
 }
 
+/// Response of `file_checksum` method.
+#[derive(Hash,Debug,Clone,PartialEq,Eq,Serialize,Deserialize)]
+pub struct FileChecksum {
+    #[allow(missing_docs)]
+    pub checksum: Sha3_224,
```
mwu-tow:
```suggestion
    pub checksum:Sha3_224,
```

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+
+
+
+// =================
+// === Constants ===
+// =================
+
+const DATA_DIR_NAME               : &str = "data";
+const CONTENT_ROOT_NOT_FOUND_CODE : i64  = 1001;
+const FILE_NOT_FOUND_CODE         : i64  = 1003;
```
mwu-tow: Consider placing this somewhere in enso protocol.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    /// Upload next chunk. Returns true if all data has been uploaded.
+    ///
+    /// After uploading, the checksum of the uploaded file is compared with the file content digest,
+    /// and an error is returned if they do not match.
+    pub async fn upload_chunk(&mut self) -> FallibleResult<bool> {
+        match self.file.data.next_chunk().await {
+            Ok(Some(data)) => {
+                info!(self.logger, "Received chunk of {self.file.name} of size {data.len()}");
```
mwu-tow: consider `debug!`

mwu-tow: and merging both

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+                trace!(self.logger, "Data: {data:?} uploading to {self.remote_path:?}");
+                self.bin_connection.write_file(&self.remote_path,&data).await?; // TODO replace with write bytes.
```
mwu-tow: prettify TODO — describe why and what it wats for

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+// ==================================
+
+/// The handler for nodes created by dragging and dropping files into IDE.
+///
+/// It is responsible for creating node, uploading file and updating the node's metadata.
+#[derive(Clone,CloneRef,Debug)]
+pub struct NodeFromDroppedFileHandler {
+    logger   : Logger,
+    project  : model::Project,
+    graph    : controller::Graph
```
mwu-tow:
```suggestion
    logger  : Logger,
    project : model::Project,
    graph   : controller::Graph
```

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+        Self{logger,project,graph}
+    }
+
+    /// Create a node from dropped file and start uploading file.
+    ///
+    /// The function returns once the node is created; the uploading process is scheduled in the
+    /// global executor. The node's metadata will be updated with the uploading progress and
+    /// error messages if any.
+    pub fn create_node_and_start_uploading
+    (self, file:FileToUpload<impl DataProvider + 'static>, position:Position) -> FallibleResult {
```
mwu-tow: Is it necessary to consume `self`?

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+        let node = self.graph.add_node(Self::new_node_info(&file,position))?;
+        executor::global::spawn(async move {
+            if let Err(err) = self.upload_file(node,file).await {
+                error!(self.logger, "Error while uploading file: {err}");
+                self.update_metadata(node, |md| md.error = Some(err.to_string()));
```
mwu-tow: If we do not want this to appear on undo stack, we need to opt-out by creating an aborted transaction.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+        self.update_metadata(node, |md| md.remote_name = Some(remote_name.clone()));
+        self.graph.set_expression(node,Self::uploading_node_expression(&remote_name))?;
+        let file_size       = file.size;
+        let remote_path     = self.data_path().append_im(&remote_name);
+        let bin_connection  = self.project.binary_rpc();
+        let json_connection = self.project.json_rpc();
+        let mut process     = FileUploadProcess::new
+            (&self.logger,file,bin_connection,json_connection,remote_path);
+
+        while process.bytes_uploaded < file_size {
```
mwu-tow: Should upload while there is more file in provider

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+            process.upload_chunk().await?;
+            self.update_metadata(node, |md| md.bytes_uploaded = process.bytes_uploaded);
```
mwu-tow, line 235: be undo-aware

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+        }
+        self.graph.set_expression(node,Self::uploaded_node_expression(&remote_name))?;
```
mwu-tow: This will not play nicely with undo. Please create a separate issue for this.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+        }
+        Ok(())
+    }
+
+    fn update_metadata(&self, node:ast::Id, f:impl FnOnce(&mut UploadingFile)) {
+        let update_md = Box::new(|md:&mut NodeMetadata| {
+            if let Some(uploading_md) = &mut md.uploading_file {
+                f(uploading_md)
+            } else {
+                warning!(self.logger, "Cannot update upload progress: Metadata are missing");
```
mwu-tow: missing trailing dot. Might be good to log node ID.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+            warning!(self.logger, "Cannot update upload progress: {err}");
+        }
+    }
+
+    async fn establish_remote_file_name(&self, original_name:&str) -> FallibleResult<String> {
+        let list_response     = self.project.json_rpc().client.file_list(&self.data_path()).await?;
+        let files_list        = list_response.paths.into_iter().map(|f| f.take_name());
+        let files_in_data_dir = files_list.collect::<HashSet<String>>();
+        let extension_sep     = original_name.rfind('.');
+        let name_core         = extension_sep.map_or(original_name, |i| &original_name[0..i]);
```
mwu-tow: consider name `stem`

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+                data: Box::new(self.chunks.clone().into_iter())
+            }
+        }
+
+        fn file_to_upload_async(&self) -> (FileToUpload<TestAsyncProvider>,TestAsyncProviderSink) {
+            let (sender,receiver) = futures::channel::mpsc::channel(5);
+            let file_to_upload    = FileToUpload {
+                name: self.file_name.clone(),
+                size: self.file_size,
+                data: receiver
```
mwu-tow:
```suggestion
                name : self.file_name.clone(),
                size : self.file_size,
                data : receiver
```

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+                assert_eq!(*p,path);
+                Ok(response::FileChecksum{checksum})
+            }));
+        }
+
+        fn file_to_upload(&self) -> FileToUpload<TestProvider> {
+            FileToUpload {
+                name: self.file_name.clone(),
+                size: self.file_size,
+                data: Box::new(self.chunks.clone().into_iter())
```
mwu-tow:
```suggestion
                name : self.file_name.clone(),
                size : self.file_size,
                data : Box::new(self.chunks.clone().into_iter())
```

src/rust/ide/src/model/module.rs
```diff
@@ -418,6 +424,22 @@ pub struct MethodId {
...
+#[allow(missing_docs)]
+#[derive(Clone,Debug,Deserialize,Eq,Hash,PartialEq,Serialize)]
+pub struct UploadingFile {
+    /// The name of file dropped in IDE.
+    pub name:String,
+    /// The file's destination name. May differ from original name due to conflict with files
+    /// already present on the remote.
+    pub remote_name    : Option<String>,
+    pub size           : usize,
+    pub bytes_uploaded : usize,
```
mwu-tow: Might be worth noting that `bytes_uploaded` might exceed size

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -3174,6 +3180,21 @@ fn new_graph_editor(app:&Application) -> GraphEditor {
...
 
 
+    // =====================
+    // === Dropped Files ===
+    // =====================
+
+    let files_received = model.drop_manager.files_received().clone_ref();
+    frp::extend! { network
+        files_with_positions <- files_received.map2(&cursor_pos_in_scene, |files,cursor_pos| {
+            files.iter().enumerate().map(|(i,f)| (f.clone_ref(), cursor_pos + Vector2(0.0, 50.0 * i as f32))).collect_vec() // TODO
```
mwu-tow: TODO

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+
+
+
+// ==============
+// === Errors ===
+// ==============
+
+#[allow(missing_docs)]
+#[derive(Clone,Debug,Fail)]
+#[fail(display="Wrong checksum of uploaded file: {}, local checksum is {}",remote,local)]
```
mwu-tow: Not a sentence — missing dot.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,608 @@
...
+
+use enso_protocol::binary;
+use enso_protocol::language_server;
+use enso_protocol::language_server::FileSystemObject;
+use enso_protocol::language_server::Path;
+use json_rpc::error::RpcError;
+use sha3::Digest;
+use enso_protocol::types::Sha3_224;
+
+type Logger = DefaultTraceLogger;
```
mwu-tow: intentional?

mwu-tow (approved):
Ok but please apply comments before merge.

s9ferech:
This is how it looks when I drop a file (The GIF is around two minutes long):

upload

There are multiple problems that I noticed:
1. Every step takes very long. The file has only 1 KB, but it took around 1.5 minutes until everything was finished.
2. The GUI freezes during most of that time. While recording the GIF, I was constantly moving the mouse. But that is not visible because the GUI did not updated.
3. The console outputs multiple error messages.
4. A visual error indicator appears during the upload.
5. The new node appears where my cursor left the window, before I dragged the file. But this is not always the case. A few times, it appeared right where I dropped the file.
6. The output type of the created node is always labeled as text, no matter the file type.
7. There is an invisible div at the top of the window. If I drop on that then the file gets opened by the web browser in a new tab.

All of this was in Chrome on Windows 10 with commit dbec5827cdfbba410178b3edf0acd7cb62bcda2e.

How much of this is expected to happen?

farmaazon:
> 1. Every step takes very long. The file has only 1 KB, but it took around 1.5 minutes until everything was finished.
> 2. The GUI freezes during most of that time. While recording the GIF, I was constantly moving the mouse. But that is not visible because the GUI did not updated.

That's strange. Is this reproducible on your side? I assume you tested the package build with `./run dist`?

> 3. The console outputs multiple error messages.

> 4. A visual error indicator appears during the upload.

Yes it is acceptable so far. The error will be hidden in the GUI part of this task (the next one).

> 5. The new node appears where my cursor left the window, before I dragged the file. But this is not always the case. A few times, it appeared right where I dropped the file.
> 6. The output type of the created node is always labeled as text, no matter the file type.

That because I picked the `File.read` method. Perhaps I should use File.new instead.

> 7. There is an invisible div at the top of the window. If I drop on that then the file gets opened by the web browser in a new tab.

I'll try to fix that

s9ferech:
> > 1. Every step takes very long. The file has only 1 KB, but it took around 1.5 minutes until everything was finished.
> > 2. The GUI freezes during most of that time. While recording the GIF, I was constantly moving the mouse. But that is not visible because the GUI did not updated.
> 
> That's strange. Is this reproducible on your side? I assume you tested the package build with `./run dist`?

I tested it with `./run dist` just now, but it doesn't seem to work at all. The node keeps looking like this:
dist upload

Dropping the file produces no output at all in the dev tools or on the console.

s9ferech:
Actually, the `dist` command produced some errors that I overlooked. Maybe they cause this.

farmaazon:
Update:

> 6. The output type of the created node is always labeled as text, no matter the file type.

This will be covered by engine in https://github.com/enso-org/enso/issues/1801

> I tested it with ./run dist just now, but it doesn't seem to work at all. The node keeps looking like this:

The dropping will work properly with the new engine only. As it is not yet officially released, you need to bump engine_version in your project's package.yaml to `0.2.12-SNAPSHOT.2021-06-15`.

wdanilo:
Regarding the progress bar design, we want something along these lines (shape only, no text):

https://user-images.githubusercontent.com/1623053/122346937-cafbe480-cf49-11eb-9ec7-878636dade78.mov

However, we do not want it to be circular - we want it to be a double circle capsule. The sliders will be around self ports - marked RED here:

image


In order to create such a progress bar, the following steps should be done:
1. There should be 4 shapes - 2 circles and 2 rectangles:
    image
2. All the shapes need to be parametrized in a good way - a percentage beginning and percentage end should be possible to be set. We should think about the percentage this way (sorry, there is an error - there should be 75 instead of 100 - I noticed it too late):
    image
3. Let's draw a shape starting at 25% and finishing at 50%:
    image
4. Or another shape starting at 5% and finishing at 75%:
     image
5. This way we can create the mask. Now, we can also create just filled up shape, make it smaller and subtract them together:
6. image
7. So it aligns with the mask:
image
8. Now we apply the mask:
    image
9. The last part is to place 2 circles at the ending points:
    image

And we should be done! :)

For now, lets put the slider just on the left side of the node. We will integrate them with self ports once the new ports will be here.

src/rust/ide/lib/enso-protocol/src/language_server/types.rs
```diff
@@ -94,6 +94,12 @@ impl Path {
             segments : segments.into_iter().map(|s| s.as_ref().into()).collect()
         }
     }
+
+    /// Constructs a new path from given root ID and segments.
```
s9ferech: The documentation does not seem to fit.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,625 @@
...
+        let bytes_uploaded = 0;
+        let checksum       = sha3::Sha3_224::new();
+        Self {logger,bin_connection,json_connection,file,remote_path,bytes_uploaded,checksum}
+    }
+
+    /// Upload next chunk. Returns true if all data has been uploaded.
+    ///
+    /// After uploading, the checksum of the uploaded file is compared with the file content digest,
+    /// and an error is returned if they do not match.
+    pub async fn upload_chunk(&mut self) -> FallibleResult<bool> {
```
s9ferech: It might be better to return a more descriptive enum.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,625 @@
...
+                self.check_checksum().await?;
+                Ok(true)
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    async fn check_checksum(&mut self) -> FallibleResult {
+        let remote = self.json_connection.file_checksum(&self.remote_path).await?.checksum;
+        let local  = std::mem::take(&mut self.checksum).into();
```
s9ferech: It is a bit difficult to understand when `into` is used before the target type is clear from the context.

s9ferech: Why is it necessary to take `checksum`? I assume repeated calls to `check_checksum` will fail then. That seems like it could lead to problems.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,625 @@
...
+        self.update_expression(node,Self::uploaded_node_expression(&remote_name))?;
+        if let Err(err) = self.graph.module.with_node_metadata(node, Box::new(|md| md.uploading_file = None)) {
+            warning!(self.logger, "Cannot remove uploading metadata: {err}");
+        }
+        Ok(())
+    }
+
+    fn update_metadata(&self, node:ast::Id, f:impl FnOnce(&mut UploadingFile)) {
+        //TODO[ao] see the TODO comment in update_expression.
+        let _tr = self.undo_redo_repository().open_ignored_transaction("Upload Metadata Update");
```
s9ferech, line 246: `open_ignored_transaction` returns a `Result`. Is it fine to ignore the error case?

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,625 @@
...
+            warning!(self.logger, "Cannot update upload progress: {err}");
+        }
+    }
+
+    fn update_expression(&self, node:ast::Id, expression:String) -> FallibleResult {
+        //TODO[ao]: Despite ignoring transaction, this will still not play well with undo-redo.
+        //    Because UR keeps the whole AST in its stack, the transactions created during uploading
+        //    process will keep trail of old metadata and expressions.
+        //    Tracked in https://github.com/enso-org/ide/issues/1623
+        let _tr = self.undo_redo_repository().open_transaction("Update Uploading Node's Expression");
```
s9ferech, line 265: As above, it might be good to check the result.

src/rust/ide/src/controller/upload.rs
```diff
@@ -0,0 +1,625 @@
...
+        self.graph.set_expression(node,expression)
+    }
+
+    async fn establish_remote_file_name(&self, original_name:&str) -> FallibleResult<String> {
+        let list_response     = self.project.json_rpc().client.file_list(&self.data_path()).await?;
+        let files_list        = list_response.paths.into_iter().map(|f| f.take_name());
+        let files_in_data_dir = files_list.collect::<HashSet<String>>();
+        let extension_sep     = original_name.rfind('.');
```
s9ferech: We should make an exception for files that start with a dot, for example `.gitignore`.

s9ferech: Using `rsplit_once` or converting into `Path` could maybe be more elegant.

s9ferech (commented):
Files with non-ASCII characters in their name do not get handled correctly.

There is some stuttering when I drop a file. It is much better when compiled with `node run dist` compared to `node run watch`, but still quite obvious.

src/rust/ide/src/controller/upload.rs
farmaazon: So, you propose to move into() to the usage place?

farmaazon: The checksum is taken, because the `result` method on sha3::Sha3_224 takes `self`. I could only call result_reset instead, but still the errors will be there. And I opt to avoid cloning this structure.

The checksum should be finalized only once per upload; I added a note in `upload_chunk` method about subsequent calls after finishing uploading.

src/rust/ide/src/controller/upload.rs
farmaazon: Yes. The open_ignored_transaction returns Err if there is already opened transaction: but this is normal.

src/rust/ide/src/controller/upload.rs
farmaazon: as above.

src/rust/ide/src/controller/upload.rs
s9ferech: Yes, or use from or add a type annotation to the definition. But it's just a detail.

src/rust/ide/src/controller/upload.rs
farmaazon: Good catch. I fixed it and add a test case for it.

farmaazon: Well, I tried those, but in fact the current solution looks clear and more efficient in handling "dot or not dot" scenarios.

farmaazon:
> Files with non-ASCII characters in their name do not get handled correctly.

One known issue is https://github.com/enso-org/ide/issues/1148. Additionally, as I discovered, the main problem was handling empty files, but that is fixed now.

> There is some stuttering when I drop a file. It is much better when compiled with `node run dist` compared to `node run watch`, but still quite obvious.

After a quick run of profiling: the setting of node expression takes much time, c.a. ~200ms. @wdanilo should it be fixed in this PR, or in a separate issue?

wdanilo (approved):

Issue #1065, opened 2020-12-14
File Viewer


farmaazon:
### Summary
Create a File Viewer EnsoGL compoenent. It should look analogously to the MacOS file dialog: 
![image.png](https://images.zenhubusercontent.com/5dc012d07b4f63000158fc1e/26790318-4cc4-45d8-b904-dd699dea4127)

#### Appearance

* Initially one panel is visible listing content of some specific directory.
* Once user selects directory, the new panel should appear on the right, showing the content of that directory. If there were already some panels on the right for different directory, they should collapse first.
* For simplicity this task does not include control boundaries nor scrolling.

#### API

The main API are the FRP endpoints defined by `define_endpoints` macro.

Inputs:
* Setting size of single panel.
* Setting entry provider. The entry provider should be a dynamic object implementing a (new) Provider trait. The trait should contain a method for getting list of file entries for given location, assuming that the first panel displays a root directory (The implementor of Provider then can decide what exactly the "root" is, for example one could pass provider treating user home directory as root). See List View implementation, where similar Provider pattern is used. 

Output
* File chosen - signal emitted each time clicks file (not directory) in Viewer. 

#### Implementation Details

* In general, implementation should follow the same pattern as other components we have: List View, Drop-down Menu and Text Area. They are defined in crates ensogl-gui-components and ensogl-text. The main structure is View which should consist of 
  * Model (containing all data, sub-components, shapes etc.) 
  * and Frp network implementing whole logic. 
Each of those should be CloneRef-able (see CloneRef trait). The Frp nodes will keep reference to the model, thus the model _must not_ keep any reference of Frp network.

* Each panel should be a re-used ListView component. See src/rust/ensogl/example/src/list_view.rs for usage example.
* All other visual elements and entry icons should be defined using EnsoGL Shape system. 

### Acceptance Criteria & Test Cases
An example scene showing FileViewer for some mocked Provider reflecting the following content:
```
.
├── DirA/
│   ├── SubdirA/
│   │   └── SubSubDirA
│   └── SubdirB/
│       ├── FileA
│       └── FileB
├── DirB/
│   └── Subdir/
├── FileA
└── FilB
```

wdanilo:
![image](https://user-images.githubusercontent.com/1623053/120501135-a6d4cb00-c3c1-11eb-88bd-17fd2e0decf2.png)

farmaazon:
the component should have two FRP outputs, emitted when node should be copied/moved, providing the source and destination path.

[...]

PR #1677, opened 2021-07-07
File browser

s9ferech:
### Pull Request Description

Add a file browser, roughly as described in issue #1065:
file browser

The following video demonstrates the basic interactions:

https://user-images.githubusercontent.com/8134934/124800283-60890380-df4d-11eb-9d75-fc22bd9365f6.mp4

This video shows the current keyboard navigation in action:

https://user-images.githubusercontent.com/8134934/124800038-199b0e00-df4d-11eb-8cb1-8e6599fc3c49.mp4

### Important Notes

The current implementation sets each column width to the amount needed to display the longest file name in the folder. To this end, it creates display objects for all directory entries and measures how wide each one is. If the folder contains many entries then this can be expensive. It could be possible to achieve better performance by computing the width of the text without creating the display object. Then we would only need to create display objects for the entries that are currently in view. An alternative solution would be to leave the column width constant.

Another problem is that columns can get too wide when file names are long. At some point, we should implement the ability to abbreviate file names to a given number of pixels, such that we can keep the column width below some maximum value. An alternative easier solution might be to clip the file names at a maximum width.

I am not sure if this makes a difference, but not all edges in icons are aligned to pixel boundaries.

As demonstrated in the first video, it is possible to scroll beyond the last column. This is intentional. The main benefit is that the view does not have to jump back to the left when closing a column.

The content is not clipped to the boundaries of the file browser because EnsoGL does not support this yet. When clipping is implemented then it should be added to the `ScrollArea` component, which the file browser uses internally. Maybe the file browser should additionally clip the content to fit to the rounded corners.

The PR also changes the appearance of the searcher and visualization chooser, which now look as on the screenshot below and include a scrollbar.
searcher and drop box

### Checklist


src/rust/ensogl/example/src/rendering_error.rs
```diff
@@ -0,0 +1,121 @@
+//! Example scene showing simple usage of a shape system.
```
farmaazon: The docs does not explain the file name.

src/rust/ensogl/lib/components/src/card.rs
```diff
@@ -0,0 +1,71 @@
...
+const SHADOW_PX:f32 = 10.0;
+
+ensogl_core::define_shape_system! {
+    (style:Style,corner_radius:f32) {
+        let sprite_width  : Var<Pixels> = "input_size.x".into();
+        let sprite_height : Var<Pixels> = "input_size.y".into();
+
+        let width  = sprite_width - SHADOW_PX.px() * 2.0;
+        let height = sprite_height - SHADOW_PX.px() * 2.0;
+        let color  = style.get_color(theme::graph_editor::node::background);
```
farmaazon: I'm not sure if card should always take background from the node style. Maybe we should add a new style dedicated for cards background?

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
...
+const CONTENT_OFFSET_Y    : f32 = TOOLBAR_HEIGHT + TOOLBAR_BORDER_SIZE;
+const PADDING             : f32 = 16.0;
+const WIDTH               : f32 = 814.0;
+const HEIGHT              : f32 = 421.0;
+const CONTENT_HEIGHT      : f32 = HEIGHT - CONTENT_OFFSET_Y;
+const SCROLL_SPACING      : f32 = 30.0;
+
+fn title_color() -> color::Rgba {
+    color::Rgba(0.439, 0.439, 0.439, 1.0)
+}
```
farmaazon: Many (if not all) of those properties could be taken from styles.

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
...
+    scroll_area    : ScrollArea,
+    background     : background::View,
+    title          : text::Area,
+    // Invariant: The column with this index (if it exists) has `focused` set, all others have it
+    //            unset. But the index does not have to refer to an existing column.
+    focused_column : Cell<usize>,
+}
+
+impl Model {
+    /// This requires that `index` points to an existing column.
```
farmaazon: What happens if it does not point? A panic? Put it explicitly in docs.

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
...
+    fn move_focus_by(&self, amount:isize) {
+        let old_index  = self.focused_column.get();
+        let max_index  = self.columns.borrow().len() as isize - 1;
+        let new_index  = old_index as isize + amount;
+        let new_index  = new_index.max(0).min(max_index);
+        let new_index  = new_index as usize;
+        let new_column = self.columns.borrow()[new_index].clone();
+        let entries    = new_column.model.entries.borrow().as_ref().cloned();
+        if let Some(entries) = entries {
+            if entries.len() > 0 {
```
farmaazon: what about `is_empty`?

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+use crate::file_browser;
+
+use enso_frp as frp;
+use ensogl_core::application::Application;
+use ensogl_core::{display, Animation};
+use ensogl_core::display::shape::*;
+use std::path::PathBuf;
+use ensogl_text as text;
+use ensogl_core::data::color;
+use ensogl_theme as theme;
```
farmaazon: Imports.

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+// =================
+// === Constants ===
+// =================
+
+const ENTRY_PADDING_LEFT   : f32 = 6.0;
+const ICON_TO_LABEL        : f32 = 5.0;
+const LABEL_TO_ARROW       : f32 = 6.0;
+const ENTRY_PADDING_RIGHT  : f32 = 2.0;
+const STATE_LABEL_PADDING  : f32 = 20.0;
+const STATE_LABEL_OFFSET_Y : f32 = 20.0;
```
farmaazon: Again: I wonder if those could not be taken from styles.

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+    color::Rgba(0.341,0.341,0.341,1.0)
+}
+
+fn entry_color_focused() -> color::Rgba {
+    color::Rgba::white()
+}
+
+fn state_label_color() -> color::Rgba {
+    color::Rgba(0.5,0.5,0.5,1.0)
+}
```
farmaazon: And those too.

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+
+impl ListEntry {
+    fn new(app:&Application, entry:Entry) -> Self {
+        let logger         = Logger::new("ListEntry");
+        let layers         = &app.display.scene().layers;
+        let display_object = display::object::Instance::new(&logger);
+
+        let label = text::Area::new(app);
+        display_object.add_child(&label);
+        label.set_position_x(ENTRY_PADDING_LEFT+icons::ICON_SIZE +ICON_TO_LABEL);
```
farmaazon: strange spacing between operands. 
```suggestion
        label.set_position_x(ENTRY_PADDING_LEFT + icons::ICON_SIZE + ICON_TO_LABEL);
```

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+        label.set_position_y(6.0);
+        label.set_default_color(entry_color_normal());
+        label.set_content(entry.name);
+        label.add_to_scene_layer(&app.display.scene().layers.panel_text);
```
farmaazon: Maybe it is not needed, but on other places the text area is removed from the main layer first.

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+                arrow = None;
+            }
+            EntryType::Folder {type_,..} => {
+                match type_ {
+                    FolderType::Standard => {
+                        icon = Box::new(icons::Folder::new());
+                        layers.add_shapes_order_dependency
+                            ::<list_view::selection::View,icons::folder::View>();
+                        layers.add_shapes_order_dependency
+                            ::<icons::folder::View,list_view::io_rect::View>();
```
farmaazon: This repeated pattern should be a generic function instead.

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+                layers.add_shapes_order_dependency
+                    ::<icons::arrow::View,list_view::io_rect::View>();
+                display_object.add_child(&arrow_icon);
+                app.display.scene().layers.panel.add_exclusive(&arrow_icon);
+                arrow = Some(arrow_icon);
+            }
+        }
+        display_object.add_child(icon.as_ref());
+        app.display.scene().layers.panel.add_exclusive(icon.as_ref());
+        icon.deref().set_position_x(ENTRY_PADDING_LEFT+icons::ICON_SIZE /2.0);
```
farmaazon:
```suggestion
        icon.deref().set_position_x(ENTRY_PADDING_LEFT + icons::ICON_SIZE / 2.0);
```

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+// =============
+// === Model ===
+// =============
+
+#[derive(Clone,Debug)]
+pub struct Model {
+    app            : Application,
+    pub entries    : RefCell<Option<Rc<Vec<Entry>>>>,
+    pub list_view  : ListView,
+    state_label    : text::Area,
```
farmaazon, line 235: What is state_label? Could you add docs?

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+
+#[derive(Debug,Clone,CloneRef)]
+pub struct Column {
+    pub model : Rc<Model>,
+    frp       : Frp,
+}
+
+impl Column {
+    pub fn new(browser:Rc<ModelWithFrp>, index:usize) -> Column {
+        let weak_browser   = Rc::downgrade(&browser);
```
farmaazon, line 274: Because we are unwrapping upgraded pointer, in the future some innocent changes could lead to unexpected panic. We do an assumption that Column won't outlive the file bowser, but this is hard to guarantee given that columns are clone-refable.

Instead, the column should inform the browser (or any other parent which fancies to keep Column component) by its frp outputs. Browser, when creating column, should connect to those outputs accordingly (see, for example, how GraphEditor does it with nodes in `GraphEditorModelWithNetwork::new_node` method).

I know it would require a lot of refactoring we have no time right now, but please create an issue and put todo here.

src/rust/ensogl/lib/components/src/file_browser/icons.rs
```diff
@@ -0,0 +1,487 @@
...
+// =================
+
+/// The width of all icons.
+pub const ICON_SIZE: f32 = 16.0;
+
+// Due to a rendering error, shapes appear too big when zoomed in very closely. To compensate for
+// this, we apply `.shrink(SHRINK_AMOUNT)` to all icons. When the icons are used without zoom,
+// the amount should be 0.0. When zoomed in, for example to examine then 0.4 is more appropriate.
+const SHRINK_AMOUNT: f32 = 0.0;
+// const SHRINK_AMOUNT : f32 = 0.4;
```
farmaazon, line 25: Commented code? Do we still shrink shapes?

src/rust/ensogl/lib/components/src/file_browser/icons.rs
```diff
@@ -0,0 +1,487 @@
...
+
+fn standard_color() -> color::Rgba {
+    color::Rgba(0.5, 0.5, 0.5, 1.0)
+}
+
+fn focused_color() -> color::Rgba {
+    color::Rgba(1.0, 1.0, 1.0, 1.0)
+}
```
farmaazon: Maybe styles?

src/rust/ensogl/lib/components/src/file_browser/model.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn from(dir: D) -> Self {
+        AnyFolderContent(Rc::new(dir))
+    }
+}
+
+
+// === EmptyFolder ===
+
+#[derive(Debug)]
+struct EmptyFolderContent;
```
farmaazon: I don't see any reason this should be private. Please also make it derive copy and clone.

src/rust/ensogl/lib/components/src/list_view.rs
```diff
@@ -6,17 +6,34 @@
...
 use enso_frp as frp;
 use ensogl_core::application;
 use ensogl_core::application::Application;
 use ensogl_core::application::shortcut;
 use ensogl_core::display;
-use ensogl_core::display::scene::layer::LayerId;
 use ensogl_core::display::shape::*;
 use ensogl_core::DEPRECATED_Animation;
-use ensogl_theme as theme;
+use ensogl_core::data::color;
```
farmaazon, line 18: sort alphabetically.

src/rust/ensogl/lib/components/src/list_view.rs
```diff
@@ -6,17 +6,34 @@
...
+
+/// Describes whether list entries should be selected by hovering or clicking. This also affects
+/// how entries can be chosen.
```
farmaazon: Could you add a note how does it affect?

src/rust/ensogl/lib/components/src/list_view.rs
```diff
@@ -26,56 +43,52 @@ use ensogl_theme as theme;
...
             let sprite_height : Var<Pixels> = "input_size.y".into();
-            let padding_inner_x = style.get_number(ensogl_theme::application::searcher::selection::padding::horizontal);
-            let padding_inner_y = style.get_number(ensogl_theme::application::searcher::selection::padding::vertical);
-            let width         = sprite_width  - 2.0.px() * SHAPE_PADDING + 2.0.px() * padding_inner_x;
-            let height        = sprite_height - 2.0.px() * SHAPE_PADDING + 2.0.px() * padding_inner_y;
-            let color         = style.get_color(ensogl_theme::widget::list_view::highlight);
-            let rect          = Rect((&width,&height)).corners_radius(CORNER_RADIUS_PX.px());
-            let shape         = rect.fill(color);
+
+            let width  = sprite_width - PADDING_HORIZONTAL.px() * 2.0;
```
farmaazon, line 70: Why we resigned from taking padding from styles?

src/rust/ensogl/lib/components/src/list_view.rs
```diff
@@ -26,56 +43,52 @@ use ensogl_theme as theme;
...
             let sprite_width  : Var<Pixels> = "input_size.x".into();
             let sprite_height : Var<Pixels> = "input_size.y".into();
-            let width         = sprite_width - SHADOW_PX.px() * 2.0 - SHAPE_PADDING.px() * 2.0;
-            let height        = sprite_height - SHADOW_PX.px() * 2.0 - SHAPE_PADDING.px() * 2.0;
-            let color         = style.get_color(theme::widget::list_view::background);
-            let rect          = Rect((&width,&height)).corners_radius(CORNER_RADIUS_PX.px());
-            let shape         = rect.fill(color);
 
-            let shadow  = shadow::from_shape(rect.into(),style);
+            let rect  = Rect((&sprite_width,&sprite_height));
```
farmaazon: We could just use `Plane` instead.

src/rust/ensogl/lib/components/src/list_view/entry.rs
```diff
@@ -350,156 +265,112 @@ impl List {
...
+            if needs_to_be_created {
+                if let Some(new_entry) = self.provider.get().get(&self.app,id) {
+                    self.display_object.add_child(&new_entry);
+                    self.visible_entries.borrow_mut().insert(id, new_entry);
+                    self.config_entry(id);
                 }
-            });
-            self.entries_range.set(range);
+            }
+        }
```
farmaazon: The previous solution was faster: in case when one entry was hidden and another became visible, we didn't destroy and create entries, but instead update the position and content of the hidden one. Maybe we could remake it, but it would require special handling in entry provider (separate "new" and "update" entry). But maybe it is a story for another issue. Anyway, the searcher performance on scrolling should be checked.

For the future, I think such rewriting demands some discussion with the team.

src/rust/ensogl/lib/components/src/list_view/entry.rs
```diff
@@ -20,266 +19,175 @@ pub const PADDING:f32 = 14.0;
...
+impl display::Object for EmptyProviderEntry {
+    fn display_object(&self) -> &display::object::Instance {
+        match *self {}
     }
 }
 
+impl Entry for EmptyProviderEntry {
+    fn set_focused(&self, _selected: bool) {}
 
-// === Masked Model Provider ===
```
s9ferech, line 139: The `SingleMaskedProvider` was used in drop down menus to hide the currently selected entry from the list. But that was completely broken. For this reason, I removed the `SingleMaskedProvider`. If we still want that feature then I would add a new issue for it, referring to the old implementation.

src/rust/ensogl/lib/components/src/list_view/entry.rs
```diff
@@ -350,156 +265,112 @@ impl List {
...
+                    self.visible_entries.borrow_mut().insert(id, new_entry);
+                    self.config_entry(id);
                 }
-            });
-            self.entries_range.set(range);
+            }
+        }
+    }
+
+    fn config_entry(&self, id:Id) {
```
farmaazon: The name doesn't say me what the function does.

src/rust/ensogl/lib/components/src/scroll_area.rs
```diff
@@ -148,15 +165,26 @@ impl ScrollArea {
...
+        // It is important to take a weak reference to the model here, because we cannot control,
+        // when exactly the closure gets dropped. Otherwise, the model would stay alive for too
+        // long.
+        let scroll_handler = f!([weak_model](event:&mouse::OnWheel)
+            if let Some(model) = weak_model.upgrade() {
+                if hovering.value() {
+                    model.h_scrollbar.scroll_by(event.delta_x() as f32);
+                    model.v_scrollbar.scroll_by(event.delta_y() as f32);
+                }
             }
```
farmaazon, line 177: The issues with lifetime will be fixed when we use the `wheel` endpoint in `mouse.frp`. However, as I see, the endpoint is not connected in `Mouse::new` (in scene.rs), but that should be fixed.

src/rust/ide/view/src/searcher.rs
```diff
@@ -13,25 +13,25 @@ use ensogl::application;
 use ensogl::display::shape::*;
 use ensogl::display;
 use ensogl_gui_components::list_view::ListView;
-use ensogl_gui_components::list_view;
+use ensogl_gui_components::{list_view, card};
```
farmaazon: imports.

src/rust/ensogl/lib/components/src/file_browser/icons.rs
```diff
@@ -0,0 +1,487 @@
...
+        self.0.stroke_width.set(if focused {1.5} else {1.0});
+        self.0.color_rgba.set(if focused {focused_color().into()} else {standard_color().into()})
+    }
+}
+
+impl Default for Project {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
farmaazon, line 492: This is very repetitive pattern. Do you think it is worth deduplicating? Perhaps through macro, or maybe even just the generic structure?

src/rust/ensogl/lib/components/src/list_view/entry.rs
```diff
@@ -308,26 +216,31 @@ impl IdAtYPosition {
 /// Not all entries are displayed at once, only those visible.
 #[derive(Clone,CloneRef,Debug)]
 pub struct List {
```
farmaazon, line 224: Updae docs of this structure. It is no longer handler for the entries, because it also handles selection.

farmaazon (changes requested):

src/rust/ensogl/lib/components/src/file_browser/column.rs
s9ferech: I don't see good way to implement this without any weak pointers at all. The core issue is that events in the column's network can lead to the creation of new columns, which requires a reference to the file browser's network which has references to the browser's `Model` which has a reference to the column. This forms a cycle.

I think the difference to nodes is that they do not fire any events that cause the creation of new nodes.

src/rust/ensogl/lib/components/src/list_view/entry.rs
farmaazon: Was it confirmed with @wdanilo ?

src/rust/ensogl/lib/components/src/list_view/entry.rs
s9ferech: I compared the performance when I made this change and did not see a significant difference. But the running time is dominated by the time that it takes to set the text on text areas, which takes much longer than it should.

src/rust/ensogl/lib/components/src/file_browser/icons.rs
s9ferech: I thought about it too, but I do not think that it is worth it to use a macro for this.

What do you mean by "the generic structure"?

src/rust/ensogl/lib/components/src/list_view/entry.rs
s9ferech: I am not sure what you mean.

src/rust/ensogl/lib/components/src/list_view.rs
s9ferech: After I had to chase yet another bug with styles initializations, I decided not to use styles anymore for now. We should probably discuss this, but I think they bring too many problems and offer us nothing at the current stage.

src/rust/ensogl/lib/components/src/file_browser/icons.rs
s9ferech: We do not currently shrink shapes, since the amount is set to 0. I left the other line there, so it can easily be switched back when it is needed to work on the icons. Do you have a suggestion to make the comment clearer?

src/rust/ensogl/lib/components/src/card.rs
s9ferech: To me, both seems okay.

src/rust/ensogl/lib/components/src/file_browser.rs
s9ferech: I think it panics. I will document that.

src/rust/ensogl/lib/components/src/file_browser.rs
s9ferech: I think I chose this because I instinctively avoid negation and "not empty" almost feels like a double negation. But I am okay with that way to formulate it.

src/rust/ensogl/lib/components/src/file_browser/column.rs
s9ferech: How can we find out if it is needed?

src/rust/ensogl/lib/components/src/file_browser/column.rs
s9ferech: It displays when the folder is loading or empty or could not be loaded. I will document it.

farmaazon (changes requested):
Additional testing notes:
1. When quick jumping through several directories by arrows makes visible lagging:

https://user-images.githubusercontent.com/3919101/124907042-025d2e80-dfe8-11eb-8c1a-966d24203e51.mp4

2. A minor one, however UX: I find it hard to scroll with mouse; I need to be very precise. Perhaps I as an user expect that I can scroll once the scrollbar is visible.

https://user-images.githubusercontent.com/3919101/124913434-67685280-dfef-11eb-95a8-8d850c315bc3.mp4

3. I don't see highlighting in searcher anymore
image

s9ferech:
@farmaazon I cant play your second video. Can you play it?

src/rust/ensogl/lib/components/src/list_view/entry.rs
s9ferech: Not yet

src/rust/ensogl/example/src/file_browser.rs
```diff
@@ -0,0 +1,270 @@
...
+
+impl FolderContent for MockFolderContent {
+    fn request_entries
+    (&self, entries_loaded: frp::Any<Rc<Vec<Entry>>>, _error_occurred: frp::Any<ImString>) {
+        entries_loaded.emit(self.entries.clone());
+    }
+}
+
+
+#[derive(Debug)]
```
wdanilo: 2-enter spacing - either remove it, or create a new section or sub-section here (please apply the same to other places too).

src/rust/ensogl/example/src/file_browser.rs
```diff
@@ -0,0 +1,270 @@
...
+            type_: EntryType::Folder {
+                type_: FolderType::Home,
+                content: MockFolderContent::new(vec![
+                    Entry {
+                        name: "Applications".to_string(),
+                        path: "Applications".into(),
+                        type_: EntryType::Folder {
+                            type_: FolderType::Standard,
+                            content: Default::default(),
+                        },
```
wdanilo: Could you do some `Entry::new` smart constructor that would make this code much shorter, please? The same applies to `EntryType::Folder::new`.

src/rust/ensogl/lib/components/src/list_view/entry.rs
farmaazon: The docs says
```
/// A view containing an entry list, arranged in column.
///
/// Not all entries are displayed at once, only those visible.
``` 
There is nothing about selection. Once selection is added, the docs should mention it I believe.

src/rust/ensogl/lib/components/src/file_browser.rs
farmaazon: Hmm, well, actually there's no need, `> 0` means the same.

src/rust/ensogl/example/src/list_view.rs
```diff
@@ -40,48 +40,91 @@ pub fn entry_point_list_view() {
 // === Mock Entries ===
 // ====================
 
+const PADDING : f32 = 10.0;
```
wdanilo, line 44: We can leave that for the next PR but in every such case, please write a TODO comment about it referencing an existing ticket about it.

farmaazon:
> @farmaazon I cant play your second video. Can you play it?

It was broken, updated.

src/rust/ensogl/example/src/rendering_error.rs
```diff
@@ -0,0 +1,121 @@
...
+
```
wdanilo: Why the above comment of Adam was marked as resolved when it is not resolved? This does not make sense to me. This docs do not explain the file name at all.

src/rust/ensogl/lib/components/src/card.rs
```diff
@@ -0,0 +1,71 @@
+//! This module provides the [`Card`] component.
```
wdanilo: I do not understand what "components/src/card" is, so I look in the docs which tell me that a "card" is a "[`Card`]". Docs should explain what this file is about. I'm skipping review of this file until the docs tell me what it is about.

src/rust/ensogl/lib/components/src/card.rs
```diff
@@ -0,0 +1,71 @@
...
+use ensogl_core::display::object::{ObjectOps, Instance};
+use ensogl_theme as theme;
+
+
+
+// ============
+// === Card ===
+// ============
+
+/// The estimated margin that we need on all sides to include the shadow.
```
wdanilo: I dont understand the sentence. What does it mean "to include the shadow"? Docs should be written in an understandable way to the reader - here is no information about where the shadows will be included, what does "include shadow" mean, etc. Also, the var name is wrong - SHADOW_PX - what is this? this is shadow width, height, spread, or something else?

src/rust/ensogl/lib/components/src/card.rs
wdanilo: > To me, both seems okay.

Could you explain it? Adam's argument is 100% correct, I agree with it. However, if you disagree, could you provide some arguments of why you disagree, please? Telling that it "seems ok" would only make the discussion longer (like it did already), as this does not provide any logical explanation of why. I Also believe that hardcoding the relation between a card (whatever card is) and background is a bad idea and this is why we have relations in the stylesheet instead.

src/rust/ensogl/lib/components/src/card.rs
```diff
@@ -0,0 +1,71 @@
...
+        let sprite_height : Var<Pixels> = "input_size.y".into();
+
+        let width  = sprite_width - SHADOW_PX.px() * 2.0;
+        let height = sprite_height - SHADOW_PX.px() * 2.0;
+        let color  = style.get_color(theme::graph_editor::node::background);
+        let rect   = Rect((&width,&height)).corners_radius(corner_radius);
+        let shape  = rect.fill(color);
+
+        let shadow = shadow::from_shape(rect.into(),style);
+
```
wdanilo, line 37: What is this spacing for?

src/rust/ensogl/lib/components/src/card.rs
wdanilo: Plase fix tha naming of variables in this PR (check all vars created in the whole PR please) - names should tell semantically what they are about. Here, I need to read the code to understand that this is background color, the var should be `bg_color` or something like that. If I see var "color" I need to go to the sources to understand what it is about. There are many places with similar var names, so please, look trough all var names and fix them – I'll not copy paste this comment to all the places, so in order to make the next review easier, please think of all new vars and check if they have semantically correct names (semantic naming = descriibng their purpose)

src/rust/ensogl/lib/components/src/card.rs
```diff
@@ -0,0 +1,71 @@
...
+        (shadow + shape).into()
+    }
+}
+
+/// A simple display object consisting of a white rectangle with rounded corners and a shadow. This
+/// can be used as a background for other components.
```
wdanilo, line 43: Oh, I see now what a card is. Right now I'm even more concerned about this discussion tbh:

image

Are you sure that every component in IDE should have the same background as nodes?

src/rust/ensogl/lib/components/src/card.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// Create a new card.
+    pub fn new() -> Self {
+        Card(View::new(Logger::new("Card")))
+    }
+
+    /// Set the size of the card's base shape. The shadow can reach beyound this.
+    pub fn resize(&self,size:Vector2) {
+        let shadow_margin = Vector2(SHADOW_PX,SHADOW_PX);
+        self.0.size.set(size+2.0*shadow_margin);
+    }
```
wdanilo: 1. some functions are named `resize`, some are named `set_corner_radius` - this is inconsistent. It should be `set_size` instead.
2. We normally allow component properties to be set with FRP. Either this should not be placed in `lib/components` or it should have these properties set with FRP. I'd definitely go with the latter option.

s9ferech:
@farmaazon 
> 1. When quick jumping through several directories by arrows makes visible lagging:

The core problem is the bad performance of creating or destroying text areas. On every move from one folder to another, we create the whole list of files and destroy the old one. I hope that we can fix the performance of text areas at some point. Otherwise, we would have to construct the columns asynchronously, which seems like a big project to me.


> 2. A minor one, however UX: I find it hard to scroll with mouse; I need to be very precise. Perhaps I as an user expect that I can scroll once the scrollbar is visible.

Do you have a suggestion how to improve it?

src/rust/ensogl/lib/components/src/drop_down_menu.rs
```diff
@@ -74,7 +76,7 @@ pub mod chooser_hover_area {
 
 ensogl_core::define_endpoints! {
     Input {
-        set_entries         (list_view::entry::AnyModelProvider),
+        set_entries         (Rc<Vec<String>>),
```
wdanilo, line 79: This change is very bad. The previous type had some meaning `AnyModelProvider` - tells something, has some documentation etc. Allowing barebone `Rc<Vec<....>>` to be passed via FRP is a bad idea – it is semantically much worse, limits what we can pass here, does not have right documentation and requires (this is the worst part) users of this API to use such types instead of nice, well named types for this thing.

src/rust/ensogl/lib/components/src/drop_down_menu.rs
```diff
@@ -98,18 +100,18 @@ ensogl_core::define_endpoints! {
...
 
-    label           : text::Area,
-    selection_menu  : list_view::ListView,
+    label          : text::Area,
+    selection_menu : ListView,
+    background     : Card,
 
-    // `SingleMaskedProvider` allows us to hide the selected element.
-    content         : RefCell<Option<list_view::entry::SingleMaskedProvider>>,
+    content        : RefCell<Option<Rc<Vec<String>>>>,
```
wdanilo, line 114: Plase revert all of such changes. As explained above, types should be named in a nice way, as they were before this PR! Passing barebone Rc , RefCells etc in FRP makes it a much worse design.

src/rust/ensogl/lib/components/src/drop_down_menu.rs
```diff
@@ -119,17 +121,25 @@ impl Model {
...
     }
 
     fn init(self) -> Self {
         self.add_child(&self.icon);
         self.add_child(&self.icon_overlay);
         self.add_child(&self.label);
+        self.selection_menu.add_child(&self.background);
+        self.app.display.scene().layers.add_shapes_order_dependency::<card::View,list_view::selection::View>();
+
+        self.background.set_corner_radius(list_view::CORNER_RADIUS_PX);
```
wdanilo, line 142: this should be part of FRP.

src/rust/ensogl/lib/components/src/drop_down_menu.rs
```diff
@@ -149,23 +159,20 @@ impl Model {
...
+        if let Some(id) = id {
+            if self.content.borrow().is_some() {
+                Some(Ref::map(self.content.borrow(), |content|
+                    content.as_ref().unwrap().get(id).unwrap().as_str()))
+            } else {
+                None
+            }
+        } else {
+            None
+        }
```
wdanilo: use .as_some(...) instead - all `} else { None }` can be written shorter always.

src/rust/ensogl/lib/components/src/drop_down_menu.rs
```diff
@@ -298,26 +305,16 @@ impl DropDownMenu {
...
-                        };
-                        // Remove selected item from menu list
-                        content.set_mask(*entry_id);
-                        // Update menu content.
-                        let entries:list_view::entry::AnyModelProvider = content.clone().into();
-                        model.selection_menu.frp.set_entries.emit(entries);
+                    if let Some(_content) = model.content.borrow().as_ref() {
+                        if let Some(item) = model.get_item_label(Some(*entry_id)) {
+                            model.set_label(item.as_ref())
+                        }
```
wdanilo, line 310: Why this code was removed ? Is it not needed anymore? What happened to the logic it is not needed? I was unable to track it down but I'd love to understand it better.

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
+//! This module provides the [`FileBrowser`] component. See type documentation for more details.
```
wdanilo: `See type documentation for more details.` THIS is the documentation that should explain what the file `file_browser` implements. It should have at least a few sentences describing it. File docs are for users to understand what this file is about - it they tell the same thing as file name, there is something wrong with docs.

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
...
+// =================
+
+const TOOLBAR_HEIGHT      : f32 = 45.0;
+const TOOLBAR_BORDER_SIZE : f32 = 1.0;
+const CONTENT_OFFSET_Y    : f32 = TOOLBAR_HEIGHT + TOOLBAR_BORDER_SIZE;
+const PADDING             : f32 = 16.0;
+const WIDTH               : f32 = 814.0;
+const HEIGHT              : f32 = 421.0;
+const CONTENT_HEIGHT      : f32 = HEIGHT - CONTENT_OFFSET_Y;
+const SCROLL_SPACING      : f32 = 30.0;
```
wdanilo: Everything is hardcoded here.

src/rust/ensogl/lib/components/src/file_browser.rs
wdanilo: It can be as a seaprate PR, but please add fixme docs with issue references to EVERY VARIABLE that needs to be fixed - not in one place in the file, as later they may not be seen. Add it to every hardcoded value that is written in non proper way now.

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
...
+
+mod background {
+    use super::*;
+
+    pub const SHADOW_PX        : f32 = 10.0;
+    pub const CORNER_RADIUS_PX : f32 = 16.0;
+
+    // This defines a rounded rectangle, filling the whole sprite, up to a padding of `SHADOW_PX` on
+    // all sides. The rectangle has a shadow and there is a gray line separating the upper part that
+    // covers the toolbar of the file browser.
```
wdanilo: Based on this docs I cannot understand what it does. Could you do a small ascii-art showing it or create a more precise description? I know it is a rectangle, filling the whole spirte, up to shadow_px-  (whatever "filling whole sprite up to shadow px means". This recangle has a gray line - but where - what is this line about, its horizontal, vertical - ho it is placed - what this whole shape represents? After this docs is fixed, please send it to Sylwia and ask here if based only on the docs she is able to tell what visual component it is described here. If she approves  the new version. elts then leave it. Please also check other docs in this PR – many of them have similar issues. Let's ask Sylwia for now only about this doc, so you will see the reaction of someone reading the code. Lets fix other docs without her help, but if we would not be able to fix them, lets do the same with other docs too.

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
...
+#[derive(Clone,Debug)]
+struct Model {
+    app            : Application,
+    display_object : display::object::Instance,
+    columns        : RefCell<Vec<Column>>,
+    scroll_area    : ScrollArea,
+    background     : background::View,
+    title          : text::Area,
+    // Invariant: The column with this index (if it exists) has `focused` set, all others have it
+    //            unset. But the index does not have to refer to an existing column.
```
wdanilo: I dont understand these docs at all. They do not explain what this field is aobut.

src/rust/ensogl/lib/components/src/file_browser.rs
```diff
@@ -0,0 +1,328 @@
...
+    focused_column : Cell<usize>,
+}
+
+impl Model {
+    /// This requires that `index` points to an existing column.
+    fn focus_column(&self, index:usize) {
```
wdanilo, line 112: Regarding the above conversation - Adam asked "What happens if it does not point? A panic? Put it explicitly in docs.". Felix, you have replied "I think it panics. I will document that." ... and you marked it as resolved. Please do not mark things as resolved until the person that asked the question thinks it is resolved. Our public API should not panic if it is possible not to panic. Focus column should focus the closest column instead. Panicking on every bad occasion is bad. Visual components should be as panic-proof as possible. 

Also, you marked this as resolved, while the docs do not still mention the panic behavior. I do not understand what happened here.

src/rust/ensogl/lib/components/src/file_browser/column.rs
wdanilo: Yes, this si not extensible at all. We will have many more icons and hardcoding their support in one place seems like a very bad idea.

src/rust/ensogl/lib/components/src/file_browser/column.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+}
+
+
+// =========================
+// === ListEntryProvider ===
+// =========================
+
+#[derive(Debug)]
+struct ListEntry {
```
wdanilo, line 87: IS this a ListEntryProvider? It is in this section

src/rust/ensogl/lib/components/src/file_browser/column.rs
wdanilo: I agree a lot with Adam's comment. Felix please talk with Adam how to do it correctly here.

src/rust/ensogl/lib/components/src/file_browser/icons.rs
wdanilo: I do not understand this comment at all - this is about zooming with camera, not sclaing the icons right? So what does it mean to set this value ever? Making some camera zoom look correctly and the default zoom not? Instead of this there should be just issue about zoomed things looking incorrectly AND it should be linked here. Why would anyone ever want to change this variable now? (again, this is not explained in the docs).

src/rust/ensogl/lib/components/src/file_browser/icons.rs
```diff
@@ -0,0 +1,487 @@
...
+// === DynamicIcon ===
+// ===================
+
+/// Icons to be used in the file browser. They have a normal state and a focused state, in which
+/// they may change their color and weight.
+pub trait DynamicIcon: display::Object+Debug {
+
+    /// Sets whether this icon is focused, changing it's appearance.
+    fn set_focused(&self, focused:bool);
+}
```
wdanilo, line 49: I do not understand why normal state + focused state is named "dynamic" icon. This name does not reflect the purpose of this triat.

src/rust/ensogl/lib/components/src/file_browser/icons.rs
wdanilo: Generic structure = traits.

Why not macro? You have written " I do not think that it is worth it to use a macro for this." - but you have not explained why you think its not worth.

src/rust/ensogl/lib/components/src/file_browser/model.rs
wdanilo: why clippy didnt catch the derives errors? Is this code linted?

src/rust/ensogl/lib/components/src/list_view/entry.rs
wdanilo: It was used for visualization chooser. The currently selected visualization should not be on the list - and it was not, it was working correctly. So it looks like a regression. Also, Felix, before removing such things, talk about what they are used / are meant to be used for with people. Talking about that on the PR day when there is a deadline of merging that is a bad idea.

src/rust/ensogl/lib/components/src/list_view/entry.rs
```diff
@@ -20,266 +19,175 @@ pub const PADDING:f32 = 14.0;
...
-    fn from(content:AnyModelProvider) -> Self {
-        let mask = default();
-        SingleMaskedProvider{content,mask}
+impl display::Object for StringEntry {
+    fn display_object(&self) -> &display::object::Instance {
+        &self.display_object
     }
 }
 
+impl Entry for StringEntry {
```
wdanilo, line 165: there are no secitons in the file for these entires / providers

src/rust/ensogl/lib/components/src/list_view/entry.rs
wdanilo: "I compared the performance when I made this change and did not see a significant difference" - Imo this is not a correct explanation. Until you made extensive benchmarks this does not explain anything. 

"But the running time is dominated by the time that it takes to set the text on text areas" – this doesn't matter at all. When text areas will be improved, this PR introduces slowdown, it should be reverted. I agree with Adam, before making such changes, a design call / discussion should be made. For now, please revert to previous, faster solution.

wdanilo (changes requested):
> The current implementation sets each column width to the amount needed to display the longest file name in the folder. To this end, it creates display objects for all directory entries and measures how wide each one is. If the folder contains many entries then this can be expensive. It could be possible to achieve better performance by computing the width of the text without creating the display object. Then we would only need to create display objects for the entries that are currently in view. An alternative solution would be to leave the column width constant.

I haven't seen this explanation anywhere in the codebase. It should be there with a link to an issue. 

> Another problem is that columns can get too wide when file names are long. At some point, we should implement the ability to abbreviate file names to a given number of pixels, such that we can keep the column width below some maximum value. An alternative easier solution might be to clip the file names at a maximum width.

I haven't seen this explanation anywhere in the codebase. It should be there with a link to an issue. 

> I am not sure if this makes a difference, but not all edges in icons are aligned to pixel boundaries.

It is bad – they should be aligned. In the figma I provided they are not. Please list all icons where it applies and please compare it to the provided figma file.

> As demonstrated in the first video, it is possible to scroll beyond the last column. This is intentional. The main benefit is that the view does not have to jump back to the left when closing a column.

Whioch scrolling are you referring to - left / right? Which video / which time in it? Please be precise, otherwise it takes a lot of time during review to get info of wat to look at.

 
> The content is not clipped to the boundaries of the file browser because EnsoGL does not support this yet. When clipping is implemented then it should be added to the ScrollArea component, which the file browser uses internally. Maybe the file browser should additionally clip the content to fit to the rounded corners.

Clipping will be able to clip with rounded corners.

> The PR also changes the appearance of the searcher and visualization chooser, which will like the lists in the file browser, just without icons. They will also include a scrollbar.

There are no screenshots / videos provided. If you are changing a visual element, always provide extensive visual materials so this can be seen, please

src/rust/ensogl/lib/components/src/scroll_area.rs
s9ferech: The `wheel` node even has type `()` at the moment. It does not provide a delta. Should I fix that as part of this PR?

src/rust/ensogl/lib/components/src/card.rs
farmaazon: But it explains, doesn't it? This is the module with `Card` component. This is the point of the module.

Personally:
1. I would avoid duplicate descriptions of `Card` in `Card`'s docs and module's docs; otherwise there's high chance of them being desynchronized.
2. I prefer having main explanation of `Ca in structure, not in module. This is how I usually check documentation in rust: rather go directly to structure I use, or want to use, or see used. Rarely I jump to the module.

src/rust/ensogl/lib/components/src/file_browser/column.rs
farmaazon: > The core issue is that events in the column's network can lead to the creation of new columns,

but it is not responsibility of columns to manage columns; The column should have an output saying "A directory is chosen; a new column should be created". Then the browser, after creating column, should connect with endpoint mentioned and create new column.

src/rust/ensogl/lib/components/src/list_view/entry.rs
farmaazon: The simple reverting is not possible: we can keep it or rewrite as I described.

The thing is, that before the providers provided _models_ only with icon and text. But now we want support for anything (e.g. icon + text + icon).

src/rust/ensogl/lib/components/src/file_browser/icons.rs
farmaazon: Actually, by generic structure I meant exactly a structure with generic parameters, like `struct Icon<View>` perhaps making some helper traits is not avoidable.

src/rust/ensogl/lib/components/src/file_browser/column.rs
s9ferech: But this would not eliminate the need for weak pointers, would it?

s9ferech: Are you suggesting something like this in the file browsers network to create the cycle?
```
push_column <- any_mut::<AnyFolderContent>();
eval push_column([push_column](content) {
    let new_column = Column::new(&content,&push_column);
    ...
});

**Comment by wdanilo** [comment]
@s9ferech @farmaazon 

`A minor one, however UX: I find it hard to scroll with mouse; I need to be very precise. Perhaps I as an user expect that I can scroll once the scrollbar is visible.` - this is not minor one. This is extremely broken UX. Because of such things people are constantly reporting us they have trouble using Enso – I agree with 100% with Adam that this should be easy to grab and scroll.

File: src/rust/ensogl/lib/components/src/card.rs (hunk 33 above)
Comment (wdanilo) [reply src/rust/ensogl/lib/components/src/card.rs]: @farmaazon there sohuld be at least 1 sentence description of what a card is. Card is a background with rounded corners shared between different components. Detailed doc can be on the struct, but the file docs should explain it already in short way.

**Comment by s9ferech** [comment]
I just added a screenshot to the PR description, showing the searcher and visualization chooser.

File: src/rust/ensogl/lib/components/src/list_view/entry.rs (hunk 23 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/list_view/entry.rs:139]: Adam confirmed that it was not working correctly. But he suggested to include the implementation again and fix it either as part of this PR or later.

@wdanilo Is it okay for you if we include the code, as it was (with some adoptions to the new list views) but do not fix it right now?

File: src/rust/ensogl/lib/components/src/card.rs (hunk 2 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs]: I don't disagree. I wanted to express that I see no need for discussion and can live with anything that you or Adam suggest.

File: src/rust/ensogl/lib/components/src/file_browser/icons.rs (hunk 15 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/file_browser/icons.rs:25]: Yes, this is about camera zooming. Setting this value applies `.shrink(SHRINK_AMOUNT)` to all icons defined in that file. If `ZOOM_AMOUNT` is 0.0 then the icons are rendered correctly with default zoom. If you zoom in very closely, to examine the pixel alignment of icons for example, then a `SHRINK_AMOUNT` of 0.4 makes them more accurate. So, you would temporarily set it to this value whenever you work on the icons. I will try to clarify the comment inside the code and link to a GitHub issue.

File: src/rust/ensogl/lib/components/src/list_view/entry.rs (hunk 22 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/list_view/entry.rs]: What were the reasons to choose the previous implementation? Where there any benchmarks? Or is there a reason to assume that the new one is significantly slower?

File: src/rust/ensogl/lib/components/src/file_browser/icons.rs (hunk 27 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/file_browser/icons.rs:492]: > Why not macro? You have written " I do not think that it is worth it to use a macro for this." - but you have not explained why you think its not worth.

I have not explained it because it sounded like Adam is aware of some of the downsides and his comment was more a suggestion than a request for a discussion.

Some downsides of using macros are: Currently, it is obvious what types are defined and what traits are implemented in this file (apart from the `define_shape_system!` definitions). If we use macros then you have to understand the macro first, before you can make any sense of this file. Macro definitions provide few static correctness guarantees. If a macro fails then the error messages are typically quite difficult to understand. IntelliJ IDEA and VS Code with the popular Rust extension provide bad support for things that you pass to a macro. And finally, not every developer is familiar with macro definitions.

**Comment by s9ferech** [comment]
@wdanilo 
> > I am not sure if this makes a difference, but not all edges in icons are aligned to pixel boundaries.
> 
> It is bad – they should be aligned. In the figma I provided they are not. Please list all icons where it applies and please compare it to the provided figma file.

Here are the icons from the Figma file with the pixel grid. The icons that I implemented look the same, up to tiny differences in round parts of the project icon and the roof of the home icon.

icons

To clarify: The parts that I am talking about are the door and chimney of the home icon, the vertical lines of the folder icon, the horizontal lines of the file icon and, to some degree, the round parts of the project icon and the outer circle of the root icon.

**Comment by s9ferech** [comment]
> > As demonstrated in the first video, it is possible to scroll beyond the last column. This is intentional. The main benefit is that the view does not have to jump back to the left when closing a column.
> 
> Whioch scrolling are you referring to - left / right? Which video / which time in it? Please be precise, otherwise it takes a lot of time during review to get info of wat to look at.

It is at 0:18 in the first video in the PR description. When you scroll to the right, such that you can see a large empty space to the right of the rightmost column.

The alternative would be that there is no such space and you can scroll just far enough to the right to make the  rightmost column visible. But if you closed the rightmost column in that solution then the view would have to scroll to the left again to bring the new rightmost column back to the right edge.

File: src/rust/ensogl/example/src/file_browser.rs (hunk 30 above)
Comment (s9ferech) [reply src/rust/ensogl/example/src/file_browser.rs]: How should the signature of that constructor look in you opinion?

File: src/rust/ensogl/example/src/rendering_error.rs (hunk 32 above)
Comment (s9ferech) [reply src/rust/ensogl/example/src/rendering_error.rs]: I usually mark comments as resolved when I am sure that I have fixed them localy. Otherwise, it is difficult to keep an overview. But in the future, I will leave it to the reviewer then.

File: src/rust/ensogl/lib/components/src/card.rs (hunk 34 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs]: It means to include the shadow inside the shape boundaries, such that it will not be clipped. I will try to clarify the comment.

Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs]: `SHADOW_PX` is the margin used on all sides. I will also try to clarify that.

File: src/rust/ensogl/lib/components/src/card.rs (hunk 35 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs:37]: I don't know. I did not write the code, I just moved it here with some modifications. I would guess that the author wanted so separate the definition of the two shapes that are combined here, to make the code clearer.

But if it annoys you then I will remove the spacing for you.

File: src/rust/ensogl/lib/components/src/card.rs (hunk 2 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs]: In this case, we are defining a card shape and this is the color of the card. `bg_color` would be confusing, in my opinion. Maybe we should call it `card_color`. But that seems a bit redundant. Do you have a better suggestion?

Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs]: By the way, such "fix it everywhere" comments take a very long time for me, especially since I am often unaware why a name is not clear to the reader. It would safe me a lot of time if you left just a short comment whenever a name is unclear to you. Suggested improvements or other thoughts would also be helpful, of course.

File: src/rust/ensogl/lib/components/src/card.rs (hunk 36 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs:43]: I am not sure. My philosophy is to generalize things when need to. But I can generalize it now if you prefer.

File: src/rust/ensogl/lib/components/src/card.rs (hunk 37 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/card.rs]: I object to 1. All other components use `resize`. If we rename it here then we should rename it everywhere consistently. But I also disagree that `set_size` is a better name. `resize` makes it clearer that this has an actual effect on screen, rather than just setting some internal attribute, in my opinion.

File: src/rust/ensogl/lib/components/src/drop_down_menu.rs (hunk 40 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/drop_down_menu.rs:142]: I do not understand what you mean by this comment.

File: src/rust/ensogl/lib/components/src/drop_down_menu.rs (hunk 41 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/drop_down_menu.rs]: I will do that, but I think that it makes the code harder to understand.

File: src/rust/ensogl/lib/components/src/drop_down_menu.rs (hunk 42 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/drop_down_menu.rs:310]: This code is only needed if we use the `MaskedModelProvider`. I removed it because it was not working and it did not seem worth it to adapt broken code to the new list view interface.

File: src/rust/ensogl/lib/components/src/file_browser.rs (hunk 45 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/file_browser.rs]: I would suggest to just include links to pictures on GitHub in such shape descriptions. Does that sound good to you?

File: src/rust/ensogl/lib/components/src/file_browser/column.rs (hunk 48 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/file_browser/column.rs:87]: It is here because it is used for the `ListEntryProver` definition.

File: src/rust/ensogl/lib/components/src/file_browser/icons.rs (hunk 49 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/file_browser/icons.rs:49]: What name would be better for you? Maybe `FocusableIcon` or `FileBrowserIcon`?

File: src/rust/ensogl/example/src/list_view.rs (hunk 31 above)
Comment (s9ferech) [reply src/rust/ensogl/example/src/list_view.rs:44]: This file is just an example. Are you sure that we should take this value from styles?

File: src/rust/ensogl/lib/components/src/file_browser.rs (hunk 3 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/file_browser.rs]: I opened an issue [here](https://github.com/enso-org/ide/issues/1694).

File: src/rust/ensogl/lib/components/src/file_browser/column.rs (hunk 14 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/file_browser/column.rs:274]: I opened an issue for it (#1695).

File: src/rust/ensogl/lib/components/src/drop_down_menu.rs (hunk 38 above)
Comment (s9ferech) [reply src/rust/ensogl/lib/components/src/drop_down_menu.rs:79]: Would it suffice to use a type alias for `RC<Vec<String>>`?

File: src/rust/ensogl/lib/components/src/list_view/entry.rs (hunk 23 above)
Comment (farmaazon) [reply src/rust/ensogl/lib/components/src/list_view/entry.rs:139]: I would postpone even the restoring old code, as we are really short of time.

File: src/rust/ensogl/lib/components/src/file_browser/column.rs (hunk 10 above)
Comment (farmaazon) [reply src/rust/ensogl/lib/components/src/file_browser/column.rs]: Looking at the code, I see we use `add_symbol_exclusive` so it is not.

File: src/rust/ensogl/lib/components/src/list_view/entry.rs (hunk 22 above)
Comment (farmaazon) [reply src/rust/ensogl/lib/components/src/list_view/entry.rs]: Never said it is _significantly_ slower for sure. I only asked to make sure it isn't. 

There were no benchmarks for the previous solution. Actually, it was picked, because I just implemented it that way (it wasn't so hard after all). I checked the scrolling performance on list_view debug scene, however I think that was in times when setting text in text_area was not so slow.

File: src/rust/ensogl/lib/components/src/file_browser/column.rs (hunk 14 above)
Comment (farmaazon) [reply src/rust/ensogl/lib/components/src/file_browser/column.rs:274]: > Are you suggesting something like this in the file browsers network to create the cycle?
> 
> ```
> push_column <- any_mut::<AnyFolderContent>();
> eval push_column([push_column](content) {
>     let new_column = Column::new(&content,&push_column);
>     ...
> });
> ```

No, rather something like 
```rust
push_column <- any_mut::<AnyFolderContent>();
eval push_column([push_column](content) {
    let new_column = Column::new(&content);
    push_column <+ new_column.column_should_be_pushed.
});
```
But of course this is a simplification: in fact, the column should not care if another one should be pushed, or maybe columns on the right should be hidden first: it should just have an endpoint "user wants to open directory", and the parent "FileBrowser" should decide if that means pushing new column, or whatever else.

src/rust/ensogl/lib/components/src/list_view/entry.rs
s9ferech: I created an issue for it (#1701).

src/rust/ensogl/lib/components/src/scroll_area.rs
s9ferech: I created an issue for this (#1702).

src/rust/ensogl/lib/components/src/card.rs
s9ferech: I created an issue for it (#1706).

src/rust/ensogl/lib/components/src/file_browser/column.rs
s9ferech: I fixed the original concern by using `if let ...` in place of `unwrap`. But I am currently hesitant to refactor the code as proposed, since I cannot discuss it with Adam.

src/rust/ensogl/lib/components/src/file_browser/column.rs
s9ferech: This is fixed.

src/rust/ensogl/example/src/file_browser.rs
s9ferech: This is fixed.

src/rust/ensogl/lib/components/src/card.rs
s9ferech: The description is included now.

src/rust/ensogl/lib/components/src/file_browser.rs
s9ferech: The documentation is updated.

src/rust/ensogl/lib/components/src/list_view/entry.rs
s9ferech: There are sections now.

src/rust/ensogl/example/src/file_browser.rs
wdanilo: It should take generic things that can be converted to string and to other type.s Then you'd be able to provide it with .e.g &str.

src/rust/ensogl/example/src/list_view.rs
wdanilo: If this is example, then no, you are right!

src/rust/ensogl/example/src/rendering_error.rs
wdanilo: Please leave them to the reviewer :)

src/rust/ensogl/lib/components/src/card.rs
wdanilo: Clarification doesnt help here - the name of the variable is not self-explaining, so it needs to be changed to be more semantically descriptive.

src/rust/ensogl/lib/components/src/card.rs
wdanilo: The names "color" are not semantically descriptive. It is like naming a variable "var". Sure it is a color, but what this color is used for? `bg_color` is million times better, nad yes, I'd use `bg_color`. Please remember that this should be fixed as the second PR here. It means this should probably be applied by @MichaelMauderer now then

src/rust/ensogl/lib/components/src/card.rs
wdanilo: This file is already a generalization - you are creating here a generic component that is a background of other components. Either remove this file or keep this component useful in various component implementations. Now, this is defined as a "component" but in fact it is a hardcoded background of another component. IMO this file should either not exist or this should be a real component defined that could be (as documentation states) be used as background in other components (not necessarily with exactly this background).

src/rust/ensogl/lib/components/src/card.rs
wdanilo: I don't agree with that TBH. Naming consistency should be the most basic guideline for us. If all such "setters" start with `set_` prefix, then not following this rule with one function seems like a dirty naming design to me. And yes, if other components use it the same way, it should be changed there as well. Especially when included in FRP API, as we use `set_` prefix there for all setters.

src/rust/ensogl/lib/components/src/drop_down_menu.rs
wdanilo: No, because the type checker when infering types would expand that and will hint you with a messy type instead of the right one. Such things should be defined with new types, the previous implementation used a very nice design. This needs to be reverted.

src/rust/ensogl/lib/components/src/drop_down_menu.rs
wdanilo: Hmm, this might be invalid actually! Sorry for that!

src/rust/ensogl/lib/components/src/drop_down_menu.rs
wdanilo: Let's see the new version then and let's discuss looking at it :)

src/rust/ensogl/lib/components/src/drop_down_menu.rs
wdanilo: If I understand correctly, MaskedModelProvider was working (at least partially) as it was hiding the currently selected visualization in the visualization chooser. I understand that this PR introduces regression there, is this correct? I mean, in visualization chooser we need to hide the currently selected entry and it was working, so it had to work somehow. Anyway, removing such things should be done by first talking about it with people who implemented it. Have you conducted such talks with Adam and did he agree for it to be removed? If not, this should be reverted / fixed and discussed properly.

src/rust/ensogl/lib/components/src/file_browser.rs
wdanilo: nope, because then you need to open browser to understand docs. Docs should be understandable by reading the code. Creating a very simple description / ascii art is rather simple task and we did it already in multiple places in the code.

src/rust/ensogl/lib/components/src/file_browser/column.rs
wdanilo: But it is in section "ListEntryProvider" - so the question is if ListEntry IS a ListEntryProvider. If it just uses it, it should be in different section.

src/rust/ensogl/lib/components/src/file_browser/icons.rs
wdanilo: Just `Icon`. it should be used qualified as `file_browser::Icon`.

MichaelMauderer:
Closed, as this will be further developed as part of a different future PR.

PR #1700, opened 2021-07-12
File Open Dialog

farmaazon:
### Pull Request Description
This PR contains the ctrl/cmd + o dialog where user can pick a project to open, and in future the file browser will be displayed.

This PR contains the API for not-yet-implemented file browser component, and the API is already integrated with controllers, and the integration was tested on the old file browser, which, alas, has been reverted.

image

### Checklist


wdanilo (approved):

s9ferech:
When I use the scroll wheel on the file manager then it also scrolls the graph in the background.

s9ferech (commented):
If I open the dialogue immediately after starting the IDE then the "Press tab key to search for components" label is in front of it:
press tab

s9ferech:
When a folder cannot be opened, the browser displays a message that was likely not written for users:
error message

s9ferech:
I would suggest to put folders at the top, by default, since that is how I know it from most systems.

I would also propose to use [natural sort order](https://en.wikipedia.org/wiki/Natural_sort_order) among files and folders in the future.

s9ferech:
I would expect that the dialogue closes when I click on the background.

s9ferech:
Since the browser allows to paste files, I think it we should soon add the feature to delete files. I assume this will not be difficult to implement.

s9ferech:
We should probably also soon add an icon for the `Standard.Base` library. At least, it should use the "Project" icon rather than the root icon, in my opinion.

s9ferech:
Dropping files onto the file browser inserts them into the graph behind the browser, which is unexpected for me.

s9ferech:
This is not a surprise, but the file browser does not adopt to narrow windows. Neither does it make good use of large windows.

farmaazon:
Additionally, after navigating in graph, the file browser events are misplaced. I'll tried to apply a workaround, but it does not work.

s9ferech:
The ordering of content roots changes between restarts. A consistent ordering would make it more predictable for the user. This might also affect the ordering of files and folders on some systems, but they were always sorted alphabetically for me.

PR #1715, opened 2021-07-16
Searcher icons 1

s9ferech:
### Pull Request Description

This PR provides the following icons that will be used in the searcher.

searcher icons

### Checklist

[ci no changelog needed]


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@
 
 - [File browser component][1677]. A basic file browser component has been
   implemented, but is not integrated into the user interface yet.
+- [Searcher icons][1715]. Some icons have been created that will be used in the
+  next version of the searcher.
```
wdanilo: This are meant to be user product docs. Telling user "some icons have been created" doesn't really make any sense.

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
```diff
@@ -203,6 +203,28 @@ define_sdf_shapes! {
     }
 
 
+    // === RoundedLineSegment ===
+
+    /// A line segment from the origin to `target` with rounded endpoints.
+    RoundedLineSegment (target:Vector2<Pixels>, width:Pixels) {
```
wdanilo: Why "LineSegment" and not just "Line" ?

wdanilo: Also, I asked you to provide info where these equation come from. Please next to each such function link to sources.

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
```diff
@@ -203,6 +203,28 @@ define_sdf_shapes! {
...
+        float half_width = width / 2.0;
+
+        // A value between 0.0 and 1.0 indicating the position of the point on the line segment that
+        // is closest to `position`. 0.0 stands for the origin and 1.0 for `target`.
+        float projection = clamp(dot(position,target)/dot(target,target),0.0,1.0);
+        vec2 closest_point = projection * target;
```
wdanilo: alignment

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
```diff
@@ -203,6 +203,28 @@ define_sdf_shapes! {
...
+
+        float left         = min(target.x,0.0) - half_width;;
```
wdanilo: what are these double semicolons?

src/rust/ensogl/lib/theme/src/lib.rs
```diff
@@ -166,6 +166,18 @@ define_themes! { [light:0, dark:1]
         }
         searcher {
             action_list_gap = 10.0, 10.0;
+            icons {
+                favorites = Rgba(0.98,0.584,0.122,1.0)  , Rgba(0.98,0.584,0.122,1.0);
+                io        = Rgba(0.475,0.494,0.145,1.0) , Rgba(0.475,0.494,0.145,1.0);
```
wdanilo, line 178: what are favorites and io?

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,387 @@
...
+const ICON_SIZE : f32 = 16.0;
+
+// Due to a rendering error, shapes appear too big when the camera is zoomed in very closely.
+// (Documented here: https://github.com/enso-org/ide/issues/1698)
+// To compensate for this, we apply `.shrink(SHRINK_AMOUNT)` to all icons in this file. By default,
+// `SHRINK_AMOUNT` should be set to 0.0 to make the icons appear right on the default zoom level.
+// When examining the icons closely, with a strong zoom, `SHRINK_AMOUNT` should be set to 0.4 to
+// make the icons apear right on that zoom level.
+const SHRINK_AMOUNT : f32 = 0.0;
+// const SHRINK_AMOUNT : f32 = 0.38;
```
wdanilo: I was already writing comment about it somewhere. I do not understand it. First of all it is set to 0.0. What does it mean that you want to watch icons with close zoom? With a camera zoom? What is the idea this variable should serve for? Do you want it to be ever changed? I do not understand the purpose of this tbh. And also commented code.

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,387 @@
...
+            let fragment3 = fragment1.rotate((rotation_step * 2.0).radians());
+            let fragment4 = fragment1.rotate((rotation_step * 3.0).radians());
+            let fragment5 = fragment1.rotate((rotation_step * 4.0).radians());
+
+            let shape = fragment1 + fragment2 + fragment3 + fragment4 + fragment5;
+            let shape = shape.fill(style.get_color(theme::favorites));
+            let shape = shape.translate_y((-0.5).px());
+            shape.shrink(SHRINK_AMOUNT.px()).into()
+        }
+    }
```
wdanilo: There is a highly optimised version of start here: https://iquilezles.org/www/articles/distfunctions2d/distfunctions2d.htm 

please use it instead

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+
+
+// ==========
+// === IO ===
+// ==========
+
+// A rounded rectangle with an arrow pointing in from the left.
```
wdanilo: Could you use in docs a small ascii art? It will take 5 mins per comment and would increase the readability sooooo much, like:

```
╭──╮
│ <─
╰──╯
```

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,387 @@
...
+
+
+            // === Shape ===
+
+            let shape = rect + arrow;
+            let shape = shape.fill(style.get_color(theme::io));
+            shape.shrink(SHRINK_AMOUNT.px()).into()
+        }
+    }
+}
```
wdanilo: This has many common things with data input. Are we able to extract them?

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,387 @@
...
+            let rect = rect - gap;
+
+
+            // === Cursor ===
+
+            let cursor_middle = Rect((1.0.px(),15.0.px()));
+            let cursor_top    = Rect((5.0.px(),1.0.px())).translate_y(7.5.px());
+            let cursor_bottom = Rect((5.0.px(),1.0.px())).translate_y((-7.5).px());
+            let cursor        = (cursor_middle + cursor_top + cursor_bottom).translate_x(3.5.px());
+
```
wdanilo: Please refactor such things out. Lets build shapes library here and build icons from them instead.

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,387 @@
...
+            let aperture = aperture.translate_y((-aperture_height / 2.0).px());
+            // Make the triangle's left side point downward.
+            let aperture = aperture.rotate((-aperture_angle / 2.0).radians());
+            // Make the triangle's left side point right.
+            let aperture = aperture.rotate(-PI.radians() / 2.0);
+            // Make the triangle's left side point to the connection between vertical bar and arc.
+            // We have to negate the angle because positive results of `atan2` stand for counter-
+            // clockwise rotations.
+            let connection_direction = -connection_offset.y.atan2(connection_offset.x);
+            let aperture             = aperture.rotate(connection_direction.radians());
```
wdanilo, line 255: Absolutely stunning work with the comments. Absolutely ❤️

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,387 @@
...
+            shape.shrink(SHRINK_AMOUNT.px()).into()
+        }
+    }
+}
+
+
+
+// ===================
+// === Debug Scene ===
+// ===================
```
wdanilo, line 275: I love that the debug scene is here. It makes the file so easy to read!

wdanilo (commented):
Just minor comments really! Amazing work with these!

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
s9ferech: The word "line" commonly refers to lines that run to infinity, while the line segment has two ends.

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
s9ferech: Just a typo.

src/rust/ensogl/lib/theme/src/lib.rs
s9ferech: `favorites` is used just for the "Favorites" icon. `io` is used for all icons in the input/output category.

src/rust/ide/view/src/searcher/icons.rs
s9ferech: I reformulated it again. Is this clearer?

> Due to a rendering error, shapes appear too big when the camera is zoomed in very closely.
(Documented here: https://github.com/enso-org/ide/issues/1698)
In the user interface, this is not a big problem, since icons are usually shown at lower zoom levels. But it is a problem during development of icons when it becomes necessary to inspect them closely. In those situations, one can apply `.shrink(0.38)` to shapes to compensate for the bug make them appear at the correct size while the camera is zoomed in. But that work-around will make them appear too thin on the default zoom level.

> To make it easy to turn this shrinking on and off before and after working on icons, we define the constant `SHRINK_AMOUNT` and apply `.shrink(SHRINK_AMOUNT)` to all icons. In every commit, `SHRINK_AMOUNT` should be set to 0.0 to make icons look best in the user interface. But during work on the icons, it can temporarily be set to 0.38.

src/rust/ide/view/src/searcher/icons.rs
s9ferech: I am a bit reluctant to do this. I would rather refer to the debug scene in the module documentation and maybe add labels to the scene to make it easy to find the corresponding icon for each shape definition. Would you be happy with that?

src/rust/ide/view/src/searcher/icons.rs
s9ferech: Arrows could be extracted. I think the rest isn't worth it.

s9ferech: Can you tell me, what should be the return type of such a function?

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
wdanilo: Hmm, in graphics we nomaly use line for line segment, but as in SDF there is something like infinite line, then yeah, lets leave this name then :)

src/rust/ensogl/lib/theme/src/lib.rs
wdanilo: Ah, ok, makes sense. Hmm, I dont know why but I did not get it from the code. Maybe no change is needed.

src/rust/ide/view/src/searcher/icons.rs
wdanilo: Yes, now I understand that it is for development purposes only. I would start the whole docs with "For development purposes only." to make it even more clear!

src/rust/ide/view/src/searcher/icons.rs
wdanilo: Ok, lets skip it altogether. It would make the docs much nicer but yeah, lets just skip it.

src/rust/ide/view/src/searcher/icons.rs
wdanilo: You can always click RMB in intellij and expand the macro - then you'll get the types. This is something like `AnyShape` or something like that.

s9ferech: Which macro should I expand? I can take the lines from the shape system definition and move them into a function. Then the return type will be something like `Union<Translate<Triangle>,...>`. But this type contains so much irrelevant information. I would rather use `impl ...` as return type, or any other solution that doesn't require me to have that full type in the code.

src/rust/ensogl/lib/core/src/display/shape/primitive/def/primitive.rs
s9ferech: After working with it for a while, I think that the line segment should take a start point and an end point as arguments. That would more symmetric and would make chains of line segments much clearer. Would it be fine for you @wdanilo?

wdanilo: Yes, it makes sesnee to me as well after longer thinking about it!

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,376 @@
...
+// the constant `SHRINK_AMOUNT` and apply `.shrink(SHRINK_AMOUNT.px())` to all icons. In every
+// commit, `SHRINK_AMOUNT` should be set to 0.0 to make icons look best in the user interface. But
+// during work on the icons, it can temporarily be set to 0.35.
+const SHRINK_AMOUNT : f32 = 0.35;
+
+
+
+// =============
+// === Arrow ===
+// =============
```
wdanilo: I'd rename it to "Shape Components" and I'd put here all functions used to build other components, like arrow or cursor. This way it will be named the same way as the next section, which is "IO" and contains all icons belonging to i/o

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+/// A rounded rectangle with the letter "A" and a text cursor.
+mod text_input {
+    use super::*;
+
+    ensogl::define_shape_system! {
+        (style:Style) {
+
+            // === Rectangle ===
```
wdanilo: This is a border, right? If so, rename it to border pls (the same in other icons)

src/rust/ide/view/src/searcher/icons.rs
```diff
@@ -0,0 +1,376 @@
...
+
+            // === Number ===
+
+            // The number "5" consists of a short horizontal bar at the top, a vertical bar
+            // connected to it on the left and a big arc below, connected to the vertical bar.
+            let top  = Rect((3.0.px(),1.0.px()));
+            let left = Rect((1.0.px(),3.0.px())).translate_x((-1.0).px()).translate_y((-1.0).px());
+
+
+            // == Arc ==
```
wdanilo: Please call it "Number 5 Arc" - it would be more understandable this way

wdanilo (approved):

PR #1738, opened 2021-07-28
Fix zooming with a scroll wheel

s9ferech:
### Pull Request Description

This fixes issue #1613 by changing the equations used to calculate the zoom amount. Zooming with a scroll wheel, which scrolls in big individual steps, will behave more appropriately and consistently with zooming on a touch pad or a stepless scroll wheel.

### Checklist

### DIFF
--- a/src/rust/ensogl/lib/core/src/display/navigation/navigator.rs
+++ b/src/rust/ensogl/lib/core/src/display/navigation/navigator.rs
@@ -97,7 +97,7 @@ impl NavigatorModel {
             // The direction in which we zoom. We negate x and y, because this vector points behind
             // the camera: When we get a positive scroll input, we want to move the camera
             // backwards, away from the mouse cursor.
-            let direction = Vector3(-focus_at_dist_1.x,-focus_at_dist_1.y,1.0);
+            let direction = Vector3(focus_at_dist_1.x,focus_at_dist_1.y,-1.0);
 
             let current_position = simulator.target_value();
             // We need to take the exponent because we do multiply `zoom_factor` onto the z
@@ -107,7 +107,7 @@ impl NavigatorModel {
             let zoom_factor      = zoom.amount.exp2();
             let new_z            = current_position.z * zoom_factor;
             let new_z            = new_z.max(camera.clipping().near + min_zoom).min(max_zoom);
-            let zoom_distance    = new_z - current_position.z;
+            let zoom_distance    = current_position.z - new_z;
             let zoom_delta       = direction * zoom_distance;
             let new_position     = current_position + zoom_delta;
 
--- a/src/rust/ensogl/lib/core/src/display/navigation/navigator.rs
+++ b/src/rust/ensogl/lib/core/src/display/navigation/navigator.rs
@@ -85,29 +85,33 @@ impl NavigatorModel {
         );
 
         let zoom_callback = enclose!((scene,camera,simulator) move |zoom:ZoomEvent| {
-            let point       = zoom.focus;
-            let normalized  = normalize_point2(point,scene.shape().value().into());
-            let normalized  = normalized_to_range2(normalized, -1.0, 1.0);
-            let half_height = 1.0;
+            let half_screen_size = Vector2::from(scene.shape().value()) / 2.0;
+            // We consider the focus point relative to the center of the screen.
+            let focus_from_center = zoom.focus - half_screen_size;
+            // We scale the focus point, such that y=1 stands for the upper edge and y=-1 for the
+            // lower edge of the screen.
+            let normalized_focus = focus_from_center / half_screen_size.y;
+            // The focus point projected to an imagined plane that is one unit in front of the
+            // camera.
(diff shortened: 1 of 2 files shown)


src/rust/ensogl/lib/core/src/display/navigation/navigator.rs
```diff
@@ -85,29 +85,33 @@ impl NavigatorModel {
...
+            // consistently for large and for small scroll steps. More infos can be found here:
+            // https://github.com/enso-org/ide/issues/1613
+            let zoom_factor      = zoom.amount.exp2();
+            let new_z            = current_position.z * zoom_factor;
+            let new_z            = new_z.max(camera.clipping().near + min_zoom).min(max_zoom);
+            let zoom_distance    = new_z - current_position.z;
+            let zoom_delta       = direction * zoom_distance;
+            let new_position     = current_position + zoom_delta;
+
+            simulator.set_target_value(new_position);
```
wdanilo: On order to understand that and not discuss it for hours, I need a better drawing/explanation there. Then I just want to analyze the drawing from the geometrical perspective and check if all assumptions regarding physical properties are correct (like the same camera acceleration/velocity with different angles). At the first glance it seems to me that this doesn't work well for camera movements near the limits (it shifts camera to the sides with higher velocity of zoom) and this line `current_position.z * zoom_factor;` suggests to me that the velocity/acceleration si different depending on the angle we are zooming, which is also bad. I might be wrong, but I need a good description of the algorithm with a drawing here first.

s9ferech:
In this comment, I will try to explain the implementation with some drawings. When we get a zoom event, we do not actually zoom the camera (by changing the field of view), we rather move it closer to the scene, which is another way to make objects appear bigger. We want to move the camera in the direction of the mouse cursor, so the first section of our event handler computes a vector in 3D space pointing in that direction. We start out with the `zoom.focus` vector that indicates the position of the cursor on screen:

We translate `zoom.focus`, such that we have the position relative to the center of the screen instead of the corner:

With the next two operations, we normalize the cursor position, such that the y coordinate is between `-camera.half_fovy_slope()` and `camera.half_fovy_slope()`:
 
Now, we can add the z coordinate -1 (which makes it point in the viewing direction) to get a vector in 3D space that points from the camera in the direction of the cursor:

From the side, and including the plane that contains our virtual scene, it looks like this:



As soon as we know `new_z`, we can scale `direction` by the appropriate amount to get the new camera position:



The remaining question is only how we determine `new_z`. If `zoom.amount` is -1, that means the user is zooming in. In this case, we want double the size of objects on screen, which is equivalent to halving the distance between the camera and the scene. If the user zooms twice with `zoom.amount` = -1 each time then we want to halve the distance twice, which together divides it by four. If we get just a single zoom event, but with `zoom.amount` = -2 then we want to produce exactly the same behavior. And if `zoom.amount` = -3 then we want to divide the distance by 8, as we would if we got three events with `zoom.amount` = -1 each. This pattern is the exponential function. To be precise, we compute `new_z = current_position.z * zoom.amount.exp2()`. In the last step, we modify `new_z` to make sure that it is bounded between a constant minimum and maximum zoom level to prevent the user from zooming arbitrarily far and to keep things at a reasonable size.

wdanilo:
Felix, thanks for all the drawings, this helps a lot. TBH the last few days I'm too tired to analyse both old a new way and compare them deeply – I'm not sure whether this new way provides a physically correct results that we always wanted here. Give me additional 1-2 days to get back to that, please. Also, please ping me in case this get lost for more than 2 other days.

s9ferech:
> Felix, thanks for all the drawings, this helps a lot. TBH the last few days I'm too tired to analyse both old a new way and compare them deeply – I'm not sure whether this new way provides a physically correct results that we always wanted here. Give me additional 1-2 days to get back to that, please. Also, please ping me in case this get lost for more than 2 other days.

The important difference with the new method is that, if I turn my scroll wheel by one step then this will double the size of objects on the screen, no matter in which direction I am zooming. (Actually, it is currently not exactly factor 2, but close to that.) And when I turn it back then I return to the exact zoom level at which I started. That is what I would expect from such a user interface and how other apps like Figma or Google Maps do it. It also feels right with pinch gestures, where the movement on screen will fit with the movement of your fingers.

The previous solution kept the speed of the camera in 3D space independent from the from the zoom direction. If you zoomed towards the center of the screen, a single scroll step would still double the size of everything (if there wasn't the bug that I fixed in this PR), but if you zoomed towards the corners of the screen then the zoom would be weaker, because the camera was moving sideways. There are many reasons why this is undesirable. A simple reason is that the zoom speed in the corners varies between full speed and almost 0, depending on the field of view. But since the user interface is currently strictly 2D, the users will not even see what the field of view is, they will just feel that zooming in the corners is slowed down by some magical amount.

MichaelMauderer:
@s9ferech What is the status of this?

s9ferech:
> @s9ferech What is the status of this?

Wojciech needs to check if the behavior of the new implementation is desirable.

MichaelMauderer:
Closed, as there was no clear way to bring this to a conclusion at the moment.

Issue #1550, opened 2021-05-06
Auto-layout when creating nodes


MichaelMauderer:
### Summary
New nodes should be placed, so they don't overlap existing nodes. 

### Value
Better overview over the graph while editing it, and less manual moving nodes around required. 

### Specification
When creating a new node, the new node should be placed at an empty spot in the graph that is 
* close to the source of the creation (i.e., where the user initiated the node creation action, which is where we currently place the node) and 
* in a spot that has enough space to contain the node.

The new size of the node should be estimated by some basic assumption, if they are known: standard node size (just come constant that we guesstimate), known variable names of the new node(if there is some input already defined through the previously selected node). 
We do not relocate a node that was newly created but gets too large, e.g., due to default arguments. 

### Acceptance Criteria & Test Cases
Scenario showing the node created in the described way.
Test cases that assert a new created node does not overlap other nodes (under the given size assumptions).

s9ferech:
At what exact moment should the node be positioned? When it is initially created with an empty content, or when the user has entered some initial text and hits enter for the first time? @wdanilo

wdanilo:
@s9ferech When we press tab the searcher should already be positioned in the "right" position. The "right" position means the Y-position as it is now and the X position on the right in the first place that is free for a hardcoded width. So the algorithm after pressing tab should be:
1. Move down A pixels (A is a value, like current offset in the product)
2. Assume the node width required is B, move right as far as needed to find a spot free for a node of B width
3. Allow editing of the node. Even if the node width will be bigger than B, keep it at that position.

s9ferech:
The acceptance criteria contain "A scenario showing the node created in the described way." What exactly is meant by that?

MichaelMauderer:
> The acceptance criteria contain "A scenario showing the node created in the described way." What exactly is meant by that?

Show a video of the working feature, and create a description for our test scenario to verify it is working correctly.

s9ferech:
I don't think it will be possible to create automated tests for this without larger refactorings. What do you think @MichaelMauderer?

MichaelMauderer:
Agree that automated tests at the moment are beyond scope. But can you try to set up a demo scene that initializes nodes that would usually overlap but now get auto-layouted? That should be feasible, and we can check that demo scene as part of the test scenario.

s9ferech:
[...]

PR #1755, opened 2021-08-04
Auto-layout when creating nodes

s9ferech:
### Pull Request Description

This PR implements the auto-layout described in issue #1550.

New nodes are pushed to the right into the first gap that is large enough:

If there isn't enough spacing then it will be pushed further:

The "interface" debug scene was updated to contain a "foo", "bar" and "baz" node that demonstrate the new auto-layout:

### Checklist


src/rust/ide/view/src/project.rs
```diff
@@ -236,6 +230,48 @@ impl Model {
...
+                }
+            }
+        }
+        let pos = Vector2(x,y);
+        self.graph_editor.add_node.emit(());
+        let node_id = self.graph_editor.frp.output.node_added.value();
+        self.graph_editor.set_node_position((node_id,pos));
+        node_id
+    }
+
```
wdanilo: Why this is not a function of graph editor? Graph editor is responsible for creating new nodes, it should be its responsibility to find the place as well IMO

src/rust/ide/view/src/project.rs
```diff
@@ -325,8 +361,8 @@ mod js {
 #[allow(missing_docs)]
 #[derive(Clone,CloneRef,Debug)]
 pub struct View {
-    model   : Model,
-    pub frp : Frp,
+    pub model : Model,
```
wdanilo: making model pub is rather a bad practice. We should rather not do it. Why is it needed?

src/rust/ide/view/src/project.rs
```diff
@@ -236,6 +230,48 @@ impl Model {
         node_id
     }
 
+    /// Ads a new node below `parent` and returns its ID. If there is not enough space right below
+    /// `parent` then the new node is moved to the right to first gap that is large enough.
+    pub fn add_node_below(&self, parent:NodeId) -> NodeId {
+        let parent_pos = self.graph_editor.model.get_node_position(parent).unwrap_or_default();
+        let styles     = StyleWatch::new(&self.app.display.scene().style_sheet);
+        use ensogl_theme::project as theme;
```
wdanilo: This line inbetween var defs looks strange

src/rust/ide/view/src/project.rs
```diff
@@ -236,6 +230,48 @@ impl Model {
...
+        let x_gap    = styles.get_number(theme::default_x_gap_between_nodes);
+        let y_gap    = styles.get_number(theme::default_y_gap_between_nodes);
+        let y_offset = y_gap + node::HEIGHT;
+        let mut x    = parent_pos.x;
+        let y        = parent_pos.y - y_offset;
+        // Push x to the right until we find a position where we have enough space for the new
+        // node, including a margin of size `x_gap`/`y_gap` on all sides.
+        {
+            let nodes             = self.graph_editor.model.nodes.all.raw.borrow();
+            // We subtract `f32::EPSILON` to be sure that we do not include the selected node.
```
wdanilo: I do not understand neither this comment nor this line of code. What would happen if we do not subtract it? Please describe it more clear in the docs :)

src/rust/ide/view/src/project.rs
```diff
@@ -236,6 +230,48 @@ impl Model {
...
+            let maybe_overlapping = nodes.values().filter(|node|
+                (node.position().y - y).abs() < y_offset - f32::EPSILON);
+            let maybe_overlapping = maybe_overlapping.sorted_by_key(|n|
+                OrderedFloat(n.position().x));
+            // This is how much horizontal space we are looking for.
+            let assumed_width = styles.get_number(theme::assumed_node_width);
```
wdanilo: I think that `theme:: assumed_node_width` is confusing. The name should be more explicit, like "minimal_x_spacing_for_a_new_node" or something like that

src/rust/ide/view/src/project.rs
```diff
@@ -236,6 +230,48 @@ impl Model {
...
+                let node_right = node.position().x + node.view.model.width() + x_gap;
+                if x + assumed_width > node_left {
+                    x = x.max(node_right);
+                } else {
+                    // Since `maybe_overlapping` is sorted, we know that the if-condition will
+                    // be false for all following `node`s as well. Therefore, we can skip the
+                    // remaining iterations.
+                    break;
+                }
+            }
```
wdanilo: can we rework the loop to use e.g. while loop and not use `break`? We try to avoid `break` always when possible to make the function/block exit in one place.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ these updates be shipped in a stable release before the end of the year.
 - [New look of open project dialog.][1700]. Now it has "Open project" title on
   the top.
 - [Documentation cooments are displayed next to the nodes.][1744].
+- [Auto-layout for new nodes.][1755]. When a node is selected and a new node
+  gets created below using <kbd>Tab</kbd> then the new node is automatically
+  positioned far enough to the right to find sufficient space and avoid
+  overlapping with existing nodes.
```
wdanilo: Ask Ara for help in rewording that to something sounding slightly nicer, please :)

src/rust/ide/view/src/project.rs
s9ferech: I just left it here because the position was also computed here before (It was much simpler at that time). I suppose it could easily be defined as a method on `GraphEditor`. I will try that out.

src/rust/ide/view/src/project.rs
s9ferech: `add_node_below` needs to be defined as a method on `Model`, such that we can easily use it from the FRP network. But we also need the method to demonstrate the auto-layout in the demo scene. The model needs to be public for that.

src/rust/ide/view/src/project.rs
s9ferech: What could happen is that rounding errors lead us to consider the parent root as overlapping, because lies exactly on the threshold where we should not count it.

CHANGELOG.md
s9ferech: @iamrecursion, would you help here, please?

src/rust/ide/view/src/project.rs
s9ferech: Then I would prefer to just remove the `break` if that is fine. It has a minimal performance impact, but keeps the code more readable.

src/rust/ide/view/src/project.rs
s9ferech: I just noticed that `default_y_gap_between_nodes` is defined in the `project` part of the theme as well. Should that also be moved to the graph-editor part then?

wdanilo: yes, it seems so

src/rust/ide/view/src/project.rs
wdanilo: nah, keeping performance optimal is alwasy better. So its better to keep it, but its even more nicer to use while

s9ferech: Unless you have a concrete suggestion to implement it with a while loop, I will leave it like this, because I don't see how I could do it better.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2039,6 +2042,50 @@ impl GraphEditor {
         self.frp.add_node.emit(());
         self.frp.output.node_added.value()
     }
+
+    /// Ads a new node below `parent` and returns its ID. If there is not enough space right below
+    /// `parent` then the new node is moved to the right to first gap that is large enough.
+    pub fn add_node_below(&self, parent:NodeId) -> NodeId {
```
farmaazon: I would rename `parent` to something else, maybe `above`? Because parent in this place may suggest rather some display objects hierarchy.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2039,6 +2042,50 @@ impl GraphEditor {
...
+        use ensogl_theme::graph_editor as theme;
+        let parent_pos = self.model.get_node_position(parent).unwrap_or_default();
+        let styles     = StyleWatch::new(&self.model.app.display.scene().style_sheet);
```
wdanilo: This is slow. Style watch should be created once and remembered instead.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2039,6 +2042,50 @@ impl GraphEditor {
...
+        let x_gap      = styles.get_number(theme::default_x_gap_between_nodes);
+        let y_gap      = styles.get_number(theme::default_y_gap_between_nodes);
```
wdanilo: This is slow as well - this should be read in FRP by using StyleWatchFRP instead. Now, we are using a lot of string queries in hashmaps everytime we create a new node. When using FRP, this will be done only when theme changes.

src/rust/ide/view/graph-editor/src/lib.rs
```diff
@@ -2039,6 +2042,50 @@ impl GraphEditor {
...
+                (node.position().y - y).abs() < y_offset - f32::EPSILON);
+            let maybe_overlapping = maybe_overlapping.sorted_by_key(|n|
+                OrderedFloat(n.position().x));
+            // This is how much horizontal space we are looking for.
+            let min_spacing = styles.get_number(theme::minimal_x_spacing_for_new_nodes);
+            for node in maybe_overlapping {
+                let node_left  = node.position().x - x_gap;
+                let node_right = node.position().x + node.view.model.width() + x_gap;
+                if x + min_spacing > node_left {
+                    x = x.max(node_right);
```
wdanilo: amazing job with the comments!!! ❤️

wdanilo (commented):
We are almost there! Let's fix these small perf things, and lets merge it :)

farmaazon (approved):
I'm ok with the code if you apply the @wdanilo comments.

MichaelMauderer:
@s9ferech What is the status of this?

s9ferech:
> @s9ferech What is the status of this?

Waiting for re-review by Wojciech.

MichaelMauderer (approved):

PR #1776, opened 2021-08-10
Fix freezing after inactivity

s9ferech:
### Pull Request Description

This PR fixes issue #1748.

It solves multiple problems:
1. Animation state is handled relative to the target value. For animations of floats or similar types, this means that the numbers get smaller as the animation approaches its target value. This leads to higher precision during computations and avoids a bug where the termination condition was never satisfied due to rounding errors with large numbers.
2. We can properly handle animations where the distance between the start and target value is NaN. Those will just instantly skip to the target value.
3. We initialize the graph editors collection of profiling statuses with the correct minimum and maximum.
4. Profiling labels will have an actual color, even when there is no profiling information and the label is invisible. Before, they they assumed a color with hue NaN in that situation. But this might lead to problems, since the color types were likely not intended to be used with such values.

### Checklist

- [x] All code has been manually tested by the PR owner against our [test scenarios](https://airtable.com/shr7KPRypRpanF7TO).


src/rust/ensogl/lib/core/src/animation/physics/inertia.rs
```diff
@@ -220,14 +220,19 @@ impl Thresholds {
...
-    mass         : Mass,
-    spring       : Spring,
-    drag         : Drag,
-    thresholds   : Thresholds,
-    active       : bool,
+    // We store the current value as an offset from the target rather than an absolute value,
+    // because this provides higher precision when the value is getting close to the target and `T`
+    // is a type of floating-point numbers, vectors of floating point numbers, or similar. The main
+    // benefit is that we avoid non-termination that could otherwise happen due to rounding errors
+    // when the target and current value are large.
```
wdanilo: I do not understand this description. "this provides higher precision when the value is getting close to the target" - why? Could you describe it with some examples and better explanation, please?

src/rust/ensogl/lib/core/src/animation/physics/inertia.rs
```diff
@@ -723,3 +727,28 @@ impl Default for EndStatus {
...
+    }
+
+    #[test]
+    fn animation_from_nan() {
+        let mut data = SimulationData::<f32>::new();
+        data.set_value(f32::NAN);
+        data.set_target_value(0.0);
+        data.step(1.0);
+        assert_eq!(data.value(),0.0);
+        assert!(!data.active);
```
wdanilo: please add comments what these tests are for and how they should behave. They are testing a corner cases and this needs additional explanation

wdanilo (changes requested):

src/rust/ensogl/lib/core/src/animation/physics/inertia.rs
s9ferech: I have tried to improve the explanation. Let me know if it is better.

s9ferech:
As mentioned [here](https://github.com/enso-org/ide/issues/758#issuecomment-903709906), I propose to remove the warning for large list views as part of this PR. What do you think @wdanilo?

CHANGELOG.md
```diff
@@ -48,9 +48,14 @@ these updates be shipped in a stable release before the end of the year.
 
 #### Visual Environment
 
-- [Fixed a bug where edited node expression was sometimes altered.][1743]. When
+- [Fixed a bug where edited node expression was sometimes altered.][1743] When
   editing node expression, the changes were occasionally reverted, or the
   grayed-out parameter names were added to the actual expression.
+- [Fixed freezing after inactivity.][1776] When the IDE window was minimized or
```
MichaelMauderer: This should be in the "Next Release" section.

src/rust/ensogl/lib/core/src/animation/physics/inertia.rs
```diff
@@ -240,29 +256,26 @@ impl<T:Value> SimulationData<T> {
...
+            let distance      = self.offset_from_target.magnitude();
             let snap_velocity = velocity < self.thresholds.speed;
             let snap_distance = distance < self.thresholds.distance;
             let should_snap   = snap_velocity && snap_distance;
-            if should_snap {
-                self.value    = self.target_value;
-                self.velocity = default();
-                self.active   = false;
+            if should_snap || distance.is_nan() {
+                self.offset_from_target = default();
```
MichaelMauderer, line 264: Do we have a `::zero()` method available for `Value`? That might be clearer.

src/rust/ensogl/lib/core/src/animation/physics/inertia.rs
```diff
@@ -240,29 +256,26 @@ impl<T:Value> SimulationData<T> {
...
         }
     }
 
     /// Compute spring force.
     fn spring_force(&self) -> T {
-        let value_delta    = self.target_value + self.value * -1.0;
-        let spring_stretch = value_delta.magnitude();
-        let coefficient    = spring_stretch * self.spring.value;
-        value_delta.normalize() * coefficient
+        self.offset_from_target * -self.spring.value
```
MichaelMauderer, line 278: Is there a reason this subtraction is written as `-value`, while other ones are `value * -1.0`?

MichaelMauderer (changes requested):

src/rust/ensogl/lib/core/src/animation/physics/inertia.rs
s9ferech: I don't think we do. But I agree with you. Maybe the `Zero` trait should be included in the definition of `Value`. But the code used `default()` for zero before. It might have been a conscious decision.

src/rust/ensogl/lib/core/src/animation/physics/inertia.rs
s9ferech: The first one is more readable. But the second one is necessary for values satisfying only the trait `Value`, because we do not have subtraction on that trait.

MichaelMauderer (approved):

PR #1814, opened 2021-08-25
Update build script to separate builds for different targets and engine versions.

s9ferech:
### Pull Request Description

This implements task #1132.

[ci build]

### Checklist

- [ ] All code has been manually tested in the IDE.
- [ ] All code has been manually tested by the PR owner against our [test scenarios](https://airtable.com/shr7KPRypRpanF7TO).

### DIFF
--- a/build/paths.js
+++ b/build/paths.js
@@ -28,6 +28,7 @@ paths.dist.bin            = path.join(paths.dist.root, 'bin')
 paths.dist.init           = path.join(paths.dist.root,'init')
 paths.dist.buildInit      = path.join(paths.dist.root,'build-init')
 paths.dist.buildInfo      = path.join(paths.dist.root,'build.json')
+paths.dist.tmp            = path.join(paths.dist.root,'tmp')
 
 paths.dist.wasm           = {}
 paths.dist.wasm.root      = path.join(paths.dist.root,'wasm')
--- a/build/run.js
+++ b/build/run.js
@@ -282,6 +282,9 @@ commands.watch.common   = async function(argv) {
         return commands.build.rust(argv);
     })
     await cmd.with_cwd(paths.js.root, async () => {
+        // FIXME: Among other things, this will call the build script of the project-manager
+        // package. But this is unnecessary because that script is already called by
+        // `build_project_manager` above. 
         return  commands.build.js(argv)
     })
 
--- a/build/run.js
+++ b/build/run.js
@@ -282,9 +282,9 @@ commands.watch.common   = async function(argv) {
         return commands.build.rust(argv);
     })
     await cmd.with_cwd(paths.js.root, async () => {
-        // FIXME: Among other things, this will call the build script of the project-manager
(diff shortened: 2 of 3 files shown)


s9ferech:
Should this update have an entry in the CHANGELOG?

build/paths.js
```diff
@@ -28,6 +28,7 @@ paths.dist.bin            = path.join(paths.dist.root, 'bin')
 paths.dist.init           = path.join(paths.dist.root,'init')
 paths.dist.buildInit      = path.join(paths.dist.root,'build-init')
 paths.dist.buildInfo      = path.join(paths.dist.root,'build.json')
+paths.dist.tmp            = path.join(paths.dist.root,'tmp')
```
MichaelMauderer, line 31: I would suggest something more specific like `tmpEngines`

build/run.js
```diff
@@ -282,6 +282,9 @@ commands.watch.common   = async function(argv) {
         return commands.build.rust(argv);
     })
     await cmd.with_cwd(paths.js.root, async () => {
+        // FIXME: Among other things, this will call the build script of the project-manager
```
MichaelMauderer: Actually, since we won't fix it, let's just leave the comment without the "FIXME" prefix.

src/js/lib/project-manager/src/build.ts
```diff
@@ -124,45 +122,93 @@ async function download_project_manager(file_url: string, overwrite: boolean): P
...
+    // The file at path `buildInfoFile` should always contain the build info of the project manager
+    // that is currently installed in the dist directory. We read the file if it exists and compare
+    // it with the version and target platform that we need. If they already agree then the right
+    // project manager is already installed and there is nothing to do.
+    const build_info_file = path.join(distPath, 'installed-enso-version')
+    let existing_build_info: BuildInfo | null
+    try {
+        let build_info_file_content = await fs.readFile(build_info_file)
+        existing_build_info = JSON.parse(build_info_file_content.toString())
+    } catch {
```
MichaelMauderer: If there is an exception here, should we at least log it?

MichaelMauderer (changes requested):
I just did a code review without testing yet as I'm on a weak connection at the moment. It might be worth to let this PR build packages to check that CI will have a working full build.

build/paths.js
s9ferech: The engines are actually downloaded into a subdirectory of `tmp` that is named `project-manager`. Do you still thing that `tmp` should be renamed?

src/js/lib/project-manager/src/build.ts
s9ferech: I will check which kind of error it is. If the file does not exist then this expected and no logging necessary. Otherwise, It will throw the error again.

build/paths.js
MichaelMauderer: No then this should be fine.

